PR #13, opened 2024-10-25
Cicd self hosted arm32

bduffy-spideroak:
Testing self hosted runner targeting by tags


gknopf-aranya:
This PR is no longer needed

PR #15, opened 2024-11-05
Update README.md

mattsinspace:
Updated to have a more direct link with SpiderOak and call to action at the top of the README while a more cultivated link is prepared.


cmarsh-spideroak (approved):

ycarmy:
This was added in a commit on the latest release.

PR #31, opened 2025-01-03
rust example app

gknopf-aranya:
Refer to this PR for original review:
https://git.spideroak-inc.com/spideroak-inc/aranya/pull/100


deny.toml
```diff
@@ -100,7 +100,6 @@ allow = [
     "BSD-3-Clause",
     "MIT",
     "Unicode-3.0",
-    "Unicode-DFS-2016",
```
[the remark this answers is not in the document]
gknopf-aranya: It seems visible enough to see that the `deny.toml` changed on the PR

README.md
```diff
@@ -73,6 +75,32 @@ configuring and running the daemon.
 
 After the daemon has started up, start the application.
 
+### Rust Example Application
+
+An example Rust program for using Aranya is located here:
+[Aranya Rust Example](templates/aranya-example)
```
nikki-aranya, line 81: Should shorten to just this first little chunk now that we have a fleshed out readme in templates/aranya-example/README.md

templates/aranya-example/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Running the example
+
+```
+target/release/aranya-example
+```
+
+Optionally, you can set the tracing log level with:
+```
+ARANYA_EXAMPLE=info
```
nikki-aranya, line 36: I'd prefer to keep the invocation as part of this line; coming from Windows we don't even have this syntax to set environment variables (that's a bash thing, cmd never bothered), so it's a little easier to see `ARANYA_EXAMPLE=info target/release/aranya-example` than just the raw prefix.

nikki-aranya (changes requested):
The separate README is probably a better move, just need to shorten the section in the root README to not clutter it and we should be good.

gknopf-aranya:
@jdygert-spok, @nikki-aranya could you take another look at this?

nikki-aranya (approved):

jdygert-spok (approved):

PR #50, opened 2025-01-08
release: keep all crate versions in sync and release 0.4.0

gknopf-aranya:
Keep all crate versions in sync by using workspace version


jdygert-spok:
Need to updated used versions in workspace cargo.toml

jdygert-spok (approved):

PR #51, opened 2025-01-08
ci: configure release-plz

jdygert-spok:
We are keeping the crate versions in sync so we only need one tag. We are just putting the changelog in the github releases.


gknopf-aranya (approved):

jdygert-spok:
Not the tag though, I don't think we want to tie all the crate versions together.

elagergren-spideroak (approved):

PR #52, opened 2025-01-09
ci: use `cargo-release` for publishing

jdygert-spok:
Rather than contorting `release-plz` to work better for a single version and our workflow, use `cargo-release` for publishing.

Process:
1. Obtain the tag version from `Cargo.toml` with `cargo release tag`.
2. Create a release tag with `cargo release tag`.
3. Push the tag with `cargo release push`.
4. Create a release for the new tag in GitHub.
5. Build and upload release artifacts to the GitHub release.

If there is an existing tag for the current version in `Cargo.toml`, the release process will be skipped.

Workflows tested on this fork:
https://github.com/gknopf-aranya/aranya/pull/2


.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+          cargo release publish -v --execute
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.ARANYA_BOT_CRATESIO_CARGO_LOGIN_KEY }}
+
+      - name: Push tag to github
+        run: |
+          cargo release tag -v --execute
+          cargo release push -v --execute
+        env:
+          # TODO: Use PAT so it will trigger publish.yml
```
gknopf-aranya, line 46: Do we need a new github release every time there is a new crate published? Or can we continue to manually create releases with desired feature sets?

gknopf-aranya:
We can remove the `release-plz` action and dep forks after this PR is merged

.github/workflows/release-plz.yml
```diff
@@ -15,25 +15,6 @@ on:
       - main
 
 jobs:
-
-  # Release unpublished packages.
```
gknopf-aranya, line 19: Do we need to change `release-plz.toml` after removing this?

.github/workflows/release.yml
jdygert-spok: Why would we not line up the github releases with the crate publishes? Keep in mind this is entirely detached from anything happening in aranya-core.

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+    concurrency:
+      group: release-${{ github.ref }}
+      cancel-in-progress: false
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/actions/setup
```
gknopf-aranya, line 28: Is this setup really needed?

.github/workflows/release.yml
gknopf-aranya: Are we only going to change the workspace version when there's a new release? If so, this seems fine. Publishing a new release each time a PR is merged seems like too much.

jdygert-spok: Yes, it'll still be the same pattern of having a PR that bumps it to trigger the release and publish.

jdygert-spok: Will need a change to not mark the action as failed though.

.github/workflows/release.yml
jdygert-spok: Currently only for sccache for the build verification that cargo publish does, but that isn't really needed :shrug:

.github/workflows/release-plz.yml
jdygert-spok: I think the config is still right for the PRs

.github/workflows/release.yml
gknopf-aranya: Rather than triggering `publish.yml` from here, we could invoke the workflow directly or combine the workflows into one where `cargo-release` is one job and publishing the release artifacts are other jobs in the workflow.

gknopf-aranya: https://docs.github.com/en/actions/sharing-automations/reusing-workflows

jdygert-spok: Yeah that'd be good, just need to set it up not to trigger when no new release

gknopf-aranya: There are ways to conditionally run jobs, get output from one job as input into another job, and make certain jobs required before running other jobs. So I think we could get that working somehow.

gknopf-aranya:
@jdygert-spok is this PR still needed after our `release-plz` discussions?

jdygert-spok:
Yes, we still want something to do the publishing automatically.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+        if: success() && steps.tag.outcome == 'success'
+        run: cargo release push -v --execute
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+
+      - name: Call publish workflow
+        if: success() && steps.tag.outcome == 'success'
+        uses: ./.github/workflows/publish.yml
+        with:
+          workflow: publish
```
jdygert-spok, line 58: @gknopf-aranya @nikki-aranya This condition hopefully makes it work right, but would be nice to test...

gknopf-aranya: @jdygert-spok should we test it on a fork?

jdygert-spok: Yeah I guess we can test both paths in another repo if we just take out the `cargo release publish` step.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Push tags to github
+        if: success() && steps.tag.outcome == 'success'
+        run: cargo release push -v --execute
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+
+      - name: Call publish workflow
+        if: success() && steps.tag.outcome == 'success'
+        uses: ./.github/workflows/publish.yml
```
gknopf-aranya, line 56: Rather than invoking the `publish.yml` workflow directly, we can allow it to run once it is triggered by the new tag.

This is the error I see when attempting to invoke the workflow directly on my fork:
https://github.com/gknopf-aranya/aranya/actions/runs/13687767626/job/38274928519?pr=2
```
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/aranya/aranya/.github/workflows/publish.yml'. Did you forget to run actions/checkout before running your local action?
```

Rather than troubleshooting why that is happening, I think it's simpler to continue using the tag trigger for the `publish.yml` workflow since we already know that works.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Create tags
+        id: tag
+        run: cargo release tag -v --execute
+        # Continue on error so we don't get a CI failure (red X) when no release is needed.
+        # Each following step must check `steps.tag.outcome` (NOT `steps.tag.conclusion`).
+        continue-on-error: true
+
+      - name: Publish to crates.io
+        if: success() && steps.tag.outcome == 'success'
+        run: cargo release publish -v --execute
```
gknopf-aranya, line 44:
```suggestion
        run: cargo release publish -v --execute  --no-confirm
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+      - uses: ./.github/actions/setup
+
+      - name: Install cargo-release
+        uses: aranya-project/install-action@main
+        with:
+          tool: cargo-release@0.25.15
+
+      - name: Create tags
+        id: tag
+        run: cargo release tag -v --execute
```
gknopf-aranya, line 37:
```suggestion
        run: cargo release tag -v --execute  --no-confirm
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.ARANYA_BOT_CRATESIO_CARGO_LOGIN_KEY }}
+
+      - name: Push tags to github
+        if: success() && steps.tag.outcome == 'success'
+        run: cargo release push -v --execute
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+
+      - name: Call publish workflow
```
gknopf-aranya, line 54: The new tag is not being passed into the `publish.yml` workflow.

.github/workflows/release.yml
jdygert-spok: Oh, since publish.yml is a workflow we need to call it as a job instead of a step, or we could change it to an action instead.

The tag trigger should work, we'll just need to use a PAT for this workflow.

.github/workflows/release.yml
gknopf-aranya: Tested on a fork here:
https://github.com/gknopf-aranya/aranya/pull/2

.github/workflows/release.yml
gknopf-aranya: Combined `publish.yml` and `release.yml` since they will share the same trigger. It simplifies the implementation to keep them in a single workflow.

.github/workflows/release.yml
djohnson-spideroak, line 1: is this creating the semver tag in github? or does this take in the desired semver tag?

gknopf-aranya: It is using the version that's in `Cargo.toml`. I've confirmed that changing the version in that file results in a different tag.

djohnson-spideroak: what happens if we use a tag we have already released?

gknopf-aranya: Added some more details to the PR description to describe the high-level process.

.github/workflows/release.yml
```diff
@@ -1,38 +1,112 @@
...
+        # Each following step must check `steps.push.outcome` (NOT `steps.push.conclusion`).
+        continue-on-error: true
+
+      - name: Version
+        if: success()
+        run: |
+          echo "VERSION=$(cargo metadata --format-version=1 --no-deps | jq --raw-output '.packages[0].version')" >> $GITHUB_OUTPUT
+        id: version
+
+      - name: Show Outputs
```
gknopf-aranya, line 63: Forgot to remove this after debugging

.github/workflows/release.yml
```diff
@@ -1,38 +1,112 @@
...
+
   release:
+    needs:
+      - tag
+    if: needs.tag.outputs.outcome == 'success'
     permissions:
       contents: write
     runs-on: ubuntu-latest
     steps:
+      - name: Show Outputs
```
gknopf-aranya, line 81: Remove debug

djohnson-spideroak (dismissed):

gknopf-aranya:
For the case where tag creation and crate publishing succeeds, but creating the github release fails, we may want to be able to continue triggering the publish.yml workflow to create a github release separately. A github tag/release can be deleted and recreated, but a Rust crate cannot be republished with the same version.

I originally combined release.yml and publish.yml for simplicity. But I think keeping them separate will give us more flexibility to handle the event when things don't run perfectly during a release.

gknopf-aranya:
Separated the workflows out again so we can manually trigger `publish.yml`.

smckeon-so (dismissed):
✅

nikki-aranya (dismissed):

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+        id: push
+        if: success() && steps.tag.outcome == 'success'
+        run: |
+          cargo release push -v --execute --no-confirm
+        # Continue on error so we don't get a CI failure (red X) when no release is needed.
+        # Each following step must check `steps.push.outcome` (NOT `steps.push.conclusion`).
+        continue-on-error: true
+
+      - name: Version
+        if: success()
```
djohnson-spideroak, line 58: do we need to check `steps.tag.outcome == 'success'` here as well?

.github/workflows/publish.yml
```diff
@@ -1,15 +1,26 @@
...
       workflow:
         required: true
         type: string
+      tag:
+        required: true
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}-${{ inputs.tag }}
+  cancel-in-progress: true
```
djohnson-spideroak, line 23: is this correct? I thought we needed to restrict publish to happen after release.

gknopf-aranya: True. If there is a release in-progress we want that to complete without being interrupted.

djohnson-spideroak (approved):

nikki-aranya (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
+# Workflow for automating Rust crate release.
+
+name: Release
```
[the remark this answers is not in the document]
gknopf-aranya: True. Not sure if we want to hold up the next release on that though.

PR #59, opened 2025-01-28
Update Aranya-Example

nikki-aranya:
Updates the example to use the latest release of aranya, removes the loose dependency to AFC, and polishes the READMEs. Package versions are based off what's already in supply-chain/config.toml.

### DIFF
--- a/crates/aranya-daemon/src/sync.rs
+++ b/crates/aranya-daemon/src/sync.rs
@@ -15,7 +15,7 @@ use aranya_util::Addr;
 use futures_util::StreamExt;
 use tokio::sync::mpsc;
 use tokio_util::time::{delay_queue::Key, DelayQueue};
-use tracing::{debug, error, instrument};
+use tracing::{error, info, instrument};
 
 use crate::{
     daemon::{Client, EF},
@@ -170,7 +170,7 @@ impl Syncer {
 
     #[instrument(skip_all, fields(peer = %peer, graph_id = %id))]
     async fn sync(&mut self, id: &GraphId, peer: &Addr) -> Result<()> {
-        debug!("syncing with peer");
+        info!("syncing with peer");
 
         let effects: Vec<EF> = {
             let mut sink = VecSink::new();
@@ -186,7 +186,7 @@ impl Syncer {
             .send(effects)
             .await
             .context("unable to send effects")?;
-        debug!(?n, "completed sync");
+        info!(?n, "completed sync");
         Ok(())
     }
 }
--- a/crates/aranya-daemon/src/sync.rs
+++ b/crates/aranya-daemon/src/sync.rs
@@ -15,7 +15,7 @@ use aranya_util::Addr;
 use futures_util::StreamExt;
 use tokio::sync::mpsc;
 use tokio_util::time::{delay_queue::Key, DelayQueue};
-use tracing::{error, info, instrument};
+use tracing::{debug, error, instrument};
 
 use crate::{
     daemon::{Client, EF},
@@ -170,7 +170,7 @@ impl Syncer {
 
     #[instrument(skip_all, fields(peer = %peer, graph_id = %id))]
     async fn sync(&mut self, id: &GraphId, peer: &Addr) -> Result<()> {
-        info!("syncing with peer");
+        debug!("syncing with peer");
 
         let effects: Vec<EF> = {
             let mut sink = VecSink::new();
@@ -186,7 +186,7 @@ impl Syncer {
             .send(effects)
             .await
             .context("unable to send effects")?;
-        info!(?n, "completed sync");
+        debug!(?n, "completed sync");
         Ok(())
     }
 }
--- a/templates/README.md
+++ b/templates/README.md
@@ -1,3 +1,3 @@
 # cargo-generate templates
 
-This folder contains [`cargo-generate`](https://github.com/cargo-generate/cargo-generate) templates.
+This folder contains `cargo-generate` templates.
--- a/templates/README.md
+++ b/templates/README.md
@@ -1,3 +1,3 @@
 # cargo-generate templates
 
-This folder contains `cargo-generate` templates.
+This folder contains [`cargo-generate`](https://github.com/cargo-generate/cargo-generate) templates.
--- a/templates/aranya-example/Cargo.toml
+++ b/templates/aranya-example/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "aranya-example"
+name = "{{project-name}}"
 description = "Example Rust program using the Aranya client library"
 version = "0.1.0"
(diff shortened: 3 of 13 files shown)


nikki-aranya:
cc: @gknopf-aranya I'm having trouble actually getting the example to run on my machine. This new version gives me the error:
```
2025-01-28T03:46:48.506129Z ERROR error responding to sync request err=sync resp poll failed

Caused by:
    0: storage error: no such storage
    1: no such storage
2025-01-28T03:46:48.506951Z ERROR sync_peer error

Caused by:
    sync error: sync resp poll failed
    
    Caused by:
        0: storage error: no such storage
        1: no such storage
2025-01-28T03:46:48.506996Z ERROR unable to sync with peer err=sync_peer error

Caused by:
    sync error: sync resp poll failed
    
    Caused by:
        0: storage error: no such storage
        1: no such storage
```
while the old one gives me:
```
2025-01-28T03:48:16.658177Z ERROR unable to sync with peer err=sync error: sync resp poll failed peer=Addr { host: 127.0.0.1, port: 65250 }
2025-01-28T03:48:16.658479Z ERROR unable to sync with peer err=unable to sync with peer

Caused by:
    sync error: sync resp poll failed
```

nikki-aranya:
New errors, seems like a client doesn't have a graph generated yet:
```
2025-02-03T18:32:49.184381Z ERROR error responding to sync request err=sync resp poll failed

Caused by:
    0: storage error: no such storage
    1: no such storage
2025-02-03T18:32:49.184784Z ERROR sync_peer error

Caused by:
    sync error: sync resp poll failed
    
    Caused by:
        0: storage error: no such storage
        1: no such storage
2025-02-03T18:32:49.184871Z ERROR unable to sync with peer err=sync_peer error

Caused by:
    sync error: sync resp poll failed
    
    Caused by:
        0: storage error: no such storage
        1: no such storage
```

gknopf-aranya:
> New errors, seems like a client doesn't have a graph generated yet:

@nikki-aranya sometimes syncs timeout and show a sync error in the logs. These errors are not fatal.

Try increasing the verbosity of the tracing to see that the example runs to completion:
`ARANYA_EXAMPLE=info target/release/aranya-example`

I see this in my logs indicating the example completed successfully:
`2025-02-03T19:33:50.052124Z  INFO completed example Aranya application`

templates/README.md
```diff
@@ -1,3 +1,3 @@
 # cargo-generate templates
 
-This folder contains `cargo-generate` templates.
+This folder contains [`cargo-generate`](https://github.com/cargo-generate/cargo-generate) templates.
```
gknopf-aranya, line 3: I think this link is a little more user friendly. Plus it links out to the repo and docs.rs site:
https://crates.io/crates/cargo-generate

gknopf-aranya:
We may want to change the log level of the message that prints out when the example app completes successfully so it's easier to tell when the application has completed. Or change the sync timeout log level so it shows up as a warning rather than an error.

nikki-aranya:
> We may want to change the log level of the message that prints out when the example app completes successfully so it's easier to tell when the application has completed. Or change the sync timeout log level so it shows up as a warning rather than an error.

no such storage seems to be a race condition, and we should be investigating the underlying cause to actually fix it because otherwise this will come back to bite us in the future. This also happens consistently with every run. 

> > New errors, seems like a client doesn't have a graph generated yet:
> 
> @nikki-aranya sometimes syncs timeout and show a sync error in the logs. These errors are not fatal.
> 
> Try increasing the verbosity of the tracing to see that the example runs to completion: `ARANYA_EXAMPLE=info target/release/aranya-example`
> 
> I see this in my logs indicating the example completed successfully: `2025-02-03T19:33:50.052124Z INFO completed example Aranya application`

It does complete successfully, I needed the poll fix for it to actually resolve though.

gknopf-aranya (dismissed):

nikki-aranya:
#70 is the counterpart to this, I've enabled `INFO` logging so even with the error, it'll show that the example finished running successfully. #77 contains future work to make the logging on this example more readable, and then we can revisit making it more polished in the future.

gknopf-aranya (approved):

smckeon-so (changes requested):
The approach we should take here isn't to replace tarpc in #77 or even demoting the info log level in syncs. A simpler solution and what I'd like to see is just updating the demo output to use a different logging mechanism. A good solution is the approach we take in our integration demos. We have a macro for `dout` (demo output), that we use in TACFI. We can use this to show "starting example Aranya application", "Creating team", "Adding sync peers", "completed example Aranya application", etc. 

``` rust
#[macro_export]
/// dout is used to output a log line that should be viewed in the demo.
/// It uses the 'demo_out' target with tracing so that the output can
/// be filtered.
///
/// # Example
/// ```ignore
/// dout!("Adding the TokenIssuer")
/// ```
macro_rules! dout {
    ($($tt:tt)+) => {
        ::tracing::info!(target: "demo_out", $($tt)*);
    };
}
```

If the user then wants to see all the nuts and bolts of the under the hood logging they can opt for the verbose output of `info`.

nikki-aranya:
Here's all logging the example now does:
image

smckeon-so (approved):
Looks great!

crates/aranya-daemon/src/sync.rs
```diff
@@ -186,7 +186,7 @@ impl Syncer {
             .send(effects)
             .await
             .context("unable to send effects")?;
-        info!(?n, "completed sync");
+        debug!(?n, "completed sync");
```
djohnson-spideroak, line 189: we might want to leave this as info, it is very useful when debugging integrations.

djohnson-spideroak (dismissed):

templates/aranya-example/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "{{project-name}}"
+name = "aranya-example"
```
jdygert-spok, line 2: I'm guessing this is currently the only change needed to make this directly runnable, but I don't necessarily think we want that. I think we may want to  but this is really meant to be a template and not just an example, which I think is easy to conflate. The template is currently very minimal but we might end up adding more templating, so I think we may want to have a separate example that is directly runnable.

nikki-aranya: This is invalid toml so I can't even get LSP hints is the problem, and none of the cargo generate example templates do this either

crates/aranya-daemon/src/sync.rs
jdygert-spok: Let's do this then:
```rust
debug!("starting sync");
info!(effects.len = n, "completed sync");
```

djohnson-spideroak (approved):
LGTM once CI passes and others rereview

jdygert-spok (approved):

PR #66, opened 2025-01-29
gh-pages: lanyon theme, directory listing

gknopf-aranya:
Add lanyon theme to GH pages site.
Manually list C API links until we find a way to autogenerate them.


nikki-aranya:
...is this deployment supposed to be broken? https://aranya-project.github.io/aranya/
I also assume the CI isn't supposed to fail

gknopf-aranya:
> Looks good, I'm not sure we necessarily want just a raw directory listing on the landing page though.

I'm working on adding a plugin that can list files/directories so we can automatically link to them from the landing page.

gknopf-aranya:
Spent time trying to figure out why the custom liquid filter wasn't working, only to discover that GitHub pages doesn't allow non-whitelisted plugins:
https://talk.jekyllrb.com/t/jekyll-github-pages-cannot-load-custom-liquid-tags/802

gknopf-aranya:
I'm giving up on trying to get the automatic directory tree working on this PR. I've entered the links manually for now. We'll want to come up with a solution eventually so the links don't need to change for every release.

Jekyll doesn't have a built-in way to list files/directories (which they claim is for security purposes).
There aren't any whitelisted plugins I could find that do this already.
GitHub pages doesn't allow us to run custom plugins in the `_plugins` folder (which I realized after adding a new plugin in that was supposed to work).
JavaScript doesn't have access to the files on the server.

Best solutions I can think of right now are:
- Writing our own script to generate the markdown/html content and automating pushing that to the `gh-pages` branch.
- Using this action to generate the directory tree for us:
https://github.com/marketplace/actions/github-pages-directory-listing
- Using the GitHub content API to dynamically fetch directory structure of the repo (but this isn't ideal since the API is rate-limited and the code would run every time someone renders the site)

_config.yml
```diff
@@ -0,0 +1,18 @@
+# Setup
+title:               Aranya API Documentation
```
smckeon-so, line 2: This should call out that these are C API docs.

_config.yml
```diff
@@ -0,0 +1,18 @@
...
+tagline:             'An overview of the Aranya project APIs'
```
smckeon-so, line 3: Here too

gknopf-aranya:
@smckeon-so, @jdygert-spok let's get this merged in. Then we can try to figure out how to automate the directory list on another PR.

gknopf-aranya:
Created a follow-on issue here to automate the directory listings:
https://github.com/aranya-project/aranya/issues/67

README.md
```diff
@@ -0,0 +1,20 @@
...
+- [v0.3.0](capi/v0.3.0/)
+
+## Deploying Docs For New Releases
+
+When a new version of Aranya is released, the docs are automatically pushed to a subfolder corresponding to the release tag on the `gh-pages` branch.
+
+This is done by the `publish.yml` workflow on the `main` branch of this repo.
+
+## Deploying the site
+Deploying is easy, just merge a PR to the `gh-pages` branch. The documentation repo is configured to use Github's built-in branch push actions to trigger builds and deploy to GH pages. The target branch and directory can be configured in the GH Pages settings section of the repo, if you need to test a deployment. Just note however, we don't have a proper staging environment, so these deployments will go live to the production github.io site.
```
jdygert-spok, line 17: Bad copy?

jdygert-spok (approved):

README.md
gknopf-aranya: I copied it from the `aranya-docs` repo. It's true that merging into the `gh-pages` branch will deploy the website. Which part looks incorrect?

nikki-aranya: Yeah, we have it set up to push to multiple sub-pages (see https://aranya-project.github.io/aranya/ vs https://aranya-project.github.io/aranya-docs/) based on repo name

smckeon-so (commented):
This could be okay, if a user only views the C API site. But if the user's path is from the doc site, having this look identical to the main doc site with no way to navigate back is going to be very confusing and jarring. I think the better approach is to incorporate the version menu into the Doxygen docs and just open this in a new window. If that's not an option we can figure something else out.

_config.yml
gknopf-aranya: What if we want to add other APIs in the future?

smckeon-so: Like Python?

jdygert-spok:
> This could be okay, if a user only views the C API site. But if the user's path is from the doc site, having this look identical to the main doc site with no way to navigate back is going to be very confusing and jarring. I think the better approach is to incorporate the version menu into the Doxygen docs and just open this in a new window. If that's not an option we can figure something else out.

Yeah I was thinking this too, a header with dropdown for version like docs.rs would nice.

_config.yml
gknopf-aranya: Yeah, that's one some customers have mentioned. We could change the wording to be more general later, after those APIs are added.

smckeon-so: My biggest concern is https://aranya-project.github.io/aranya/ and https://aranya-project.github.io/aranya-docs/ now look identical with the exact same messaging, but different content. How do you know which is which, if you don't notice the url, you'd think you're on the same site with suddenly different content.

gknopf-aranya:
> Yeah I was thinking this too, a header with dropdown for version like docs.rs would nice.

Does Doxygen support this? How would we have older versions of the docs link to newer versions if the older versions were published before the newer ones? Would we import a JavaScript library into all the Doxygen sites in order to have the same menu on each one?

gknopf-aranya:
> This could be okay, if a user only views the C API site. But if the user's path is from the doc site, having this look identical to the main doc site with no way to navigate back is going to be very confusing and jarring. I think the better approach is to incorporate the version menu into the Doxygen docs and just open this in a new window. If that's not an option we can figure something else out.

We could add a link back to the main Aranya docs site from here.
Or add the docs links to the navigation menu on the main docs site instead of in a separate landing page here.

_config.yml
gknopf-aranya: Let's put this landing page into https://aranya-project.github.io/aranya-docs/ with a dropdown on the navigation bar for each version of the C API docs.

smckeon-so:
> > Yeah I was thinking this too, a header with dropdown for version like docs.rs would nice.
> 
> Does Doxygen support this? How would we have older versions of the docs link to newer versions if the older versions were published before the newer ones? Would we import a JavaScript library into all the Doxygen sites in order to have the same menu on each one?

I sent a blog post of someone who did it, but I can't find it now. I believe they were rebuilding all the docs each time from release tags and modifying the doxygen header to dynamically populate the versions.

smckeon-so:
> > > Yeah I was thinking this too, a header with dropdown for version like docs.rs would nice.
> > 
> > 
> > Does Doxygen support this? How would we have older versions of the docs link to newer versions if the older versions were published before the newer ones? Would we import a JavaScript library into all the Doxygen sites in order to have the same menu on each one?
> 
> I sent a blog post of someone who did it, but I can't find it now. I believe they were rebuilding all the docs each time from release tags and modifying the doxygen header to dynamically populate the versions.

This one https://oleksandrkvl.github.io/2024/06/05/multi-version-doxygen.html

_layouts/default.html
```diff
@@ -0,0 +1,32 @@
...
+        {% include sidebar.html %}
+        <!--
+            Wrap is the content to shift when toggling the sidebar. We wrap the
+            content to avoid any CSS collisions with our real content.
+        -->
+        <div class="wrap">
+            <div class="masthead">
+                <div class="container">
+                    <h3 class="masthead-title">
+                        <a href="{{ site.url }}/" title="Home">{{ site.title }}</a>
```
smckeon-so, line 16: I would have this link back to the aranya-docs site. The menu logo too.

_includes/sidebar.html
```diff
@@ -0,0 +1,57 @@
...
+        <a href="{{ site.url }}">
+            <img class="logo" src="{{ site.url }}/assets/images/aranya-logo-white.png" alt="Aranya logo">
+        </a>
+    </div>
+    <div class="sidebar-item">
+        <p>{{ site.description }}</p>
+    </div>
+
+    <nav class="sidebar-nav">
+        <a class="sidebar-nav-item{% if page.title == 'Home' %} active{% endif %}" href="{{ '/' | absolute_url }}">Home</a>
```
smckeon-so, line 20: This home menu item isn't really usable at all. In fact we may want to just lose the whole sidebar menu and just have a branded landing page for someone who lands on this repo. But the main navigation will be done in te aranya-docs site.

smckeon-so (changes requested):

gknopf-aranya:
Closing in favor of:
https://github.com/aranya-project/aranya-docs/pull/47

PR #91, opened 2025-03-04
Update core deps

jdygert-spok:


jdygert-spok:
@gknopf-aranya wanted to get this started to see how far we have diverged and any potential issues before we do a release of the core deps.

There is a lot stubbed out for the syncer since that has had significant change.

One thing for core deps I realized is that it might be awkward to do a release of `aranya-policy-*` in the middle of the change to version 2.

gknopf-aranya:
Added this to the next sprint

Cargo.lock
```diff
@@ -119,41 +119,41 @@ checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
 [[package]]
 name = "aranya-afc-util"
 version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3813c9abf60b265f70c300887908595da2185b45ebe3770f7ca142b409508519"
+source = "git+https://github.com/aranya-project/aranya-core.git#644427dc67ddb2c045c88432cce693d4193b7b56"
```
gknopf-aranya, line 122: This is causing aranya-core deps to be pulled in from a github commit instead of the crates.io release

Cargo.toml
```diff
@@ -35,23 +35,25 @@ unwrap_used = "warn"
...
-aranya-idam-ffi = { version = "0.3.0" }
-aranya-libc = { version = "0.1.0" }
-aranya-perspective-ffi = { version = "0.3.0" }
-aranya-policy-compiler = { version = "0.3.0" }
-aranya-policy-ifgen = { version = "0.3.1", features = ["serde"] }
-aranya-policy-ifgen-build = { version = "0.1.0" }
-aranya-policy-lang = { version = "0.1.0" }
-aranya-policy-vm = { version = "0.3.0", features = ["std"] }
-aranya-runtime = { version = "0.3.0", features = ["std", "libc"] }
+aranya-afc-util = { git = "https://github.com/aranya-project/aranya-core.git", features = ["alloc"] }
```
gknopf-aranya, line 38: I believe we'll want to upgrade the versions of the aranya-core crates rather than getting the latest commit from github.

jdygert-spok: Yes, once they have been released.

crates/aranya-daemon/src/policy.md
```diff
@@ -1,5 +1,5 @@
 ---
-policy-version: 1
+policy-version: 2
```
gknopf-aranya, line 2: Will this change require any changes to the policy in the `aranya` repo?

gknopf-aranya: I'm guessing the answer is no if this still compiles. Seems odd to bump the policy version without making any changes to the policy though.
cc: @ycarmy or @chip-so

jdygert-spok: See discussion on https://github.com/aranya-project/aranya-core/pull/98. The actual version 2 changes are in progress. I'm still not sure exactly how we want to publish this.

Cargo.toml
gknopf-aranya: @jdygert-spok when are they going to be released?

crates/aranya-daemon/src/aranya.rs
```diff
@@ -115,9 +119,15 @@ where
                 let mut trx = client.transaction(id);
                 // TODO: save PeerCache somewhere.
                 client
-                    .add_commands(&mut trx, sink, &cmds, &mut PeerCache::new())
+                    .add_commands(&mut trx, sink, &cmds)
                     .context("unable to add received commands")?;
                 client.commit(&mut trx, sink).context("commit failed")?;
+                // TODO: Update heads
```
djohnson-spideroak, line 125: do we have an issue for this?

djohnson-spideroak (approved):

gknopf-aranya (approved):
lgtm other than pulling aranya-core crates from github instead of crates.io. Since we're preparing for an incremental release I think we should release the aranya-core crates then use the released crates here.

PR #103, opened 2025-03-06
naming: user -> device

gknopf-aranya:
Change references of user to device in the codebase.

Depends on `aranya_crypto` crate changing `UserId` to `DeviceId`:
https://github.com/aranya-project/aranya-core/pull/122

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,9 +119,9 @@ checksum = "dcfed56ad506cb2c684a14971b8861fdc3baaaae314b9e5f9bb532cbe3ba7a4f"
 
 [[package]]
 name = "aranya-afc-util"
-version = "0.4.0"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a6bbb43a35f8680475795e51ce64d124e603402a160f2d5f1ce60629d59cd41"
+checksum = "4facc29a9d9559909391589a8c7222171486204661c936ceb624ecbbf2cbbc0a"
 dependencies = [
  "aranya-crypto",
  "aranya-fast-channels",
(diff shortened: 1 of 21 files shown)


smckeon-so (dismissed):
Looks good!

smckeon-so (approved):

supply-chain/audits.toml
jdygert-spok, line 1: Why are there unrelated dependency updates on this PR?

templates/aranya-example/src/main.rs
jdygert-spok, line 1: Making changes early again, #130 :shrug:

jdygert-spok:
Generally looks good at a glance but haven't poured over for any find/replace errors

templates/aranya-example/src/main.rs
gknopf-aranya: I will remove template changes from this PR. Nice to know what to change it to after the next release.

supply-chain/audits.toml
gknopf-aranya: Because I needed to update the aranya-core crates and figured why not run `cargo update` and update everything else while I'm at it.

templates/aranya-example/src/main.rs
gknopf-aranya: It doesn't appear that there were breaking changes when updating the deps. Would like to stay on top of `cargo vet` audits rather than auditing everything right before the MVP release.

supply-chain/audits.toml
gknopf-aranya: It doesn't appear that there were breaking changes when updating the deps. Would like to stay on top of cargo vet audits rather than auditing everything right before the MVP release.

elagergren-spideroak: then let's do the updates on a separate PR

elagergren-spideroak: except for those that are required for the renaming to work, of course

gknopf-aranya: Moving to another PR

gknopf-aranya:
@elagergren-spideroak, @jdygert-spok ready for another look

djohnson-spideroak (approved):

jdygert-spok (approved):

PR #105, opened 2025-03-10
supply chain updates

gknopf-aranya:
Shortens a couple commonly used `cargo make` commands to make them easier to type.
Updates supply chain audits for `cargo vet` checks.


gknopf-aranya:
Out of the diffs I looked at, this one seemed the most questionable:
https://diff.rs/winapi-i686-pc-windows-gnu/0.4.0/0.4.0/lib
Because the library binaries are checked in.

scripts/cargo-vet/diff.bash
```diff
@@ -0,0 +1,37 @@
...
+if command -v shellcheck; then
+    shellcheck "${0}"
+fi
+
+# cargo vet diff
+diff=${DIFF:-1}
+# cargo vet certify
+certify=${CERTIFY:-1}
+
+pushd ../../
```
jdygert-spok, line 23: If you want to set the directory, it should be relative to the script, but IMO it is unnecessary. `cargo vet` will work as long as you're anywhere within the repo.

gknopf-aranya:
Removed `diff.bash` to unblock the incremental product release until @jdygert-spok is able to complete a similar script that does not require `eval`.

djohnson-spideroak (approved):

jdygert-spok (approved):

PR #107, opened 2025-03-10
afc: add afc prefix to afc apis

gknopf-aranya:
Add `afc` prefix the AFC API (Rust, C, and daemon unix domain socket APIs).
Makes relevant changes to the daemon API and Aranya policy in preparation for adding the AQC channels feature which will use its own `aqc` prefix.


djohnson-spideroak (approved):

nikki-aranya (approved):

djohnson-spideroak (dismissed):

smckeon-so (dismissed):

crates/aranya-client/src/client.rs
```diff
@@ -144,7 +144,7 @@ impl Client {
     /// It is NOT safe to cancel the resulting future. Doing so
     /// might lose data.
     #[instrument(skip_all, fields(self = self.debug(), %team_id, %peer, %label))]
-    pub async fn create_bidi_channel(
+    pub async fn afc_create_bidi_channel(
```
elagergren-spideroak, line 147: nit: for *Rust* it might be nice to reorder the name as `create_afc_bidi_channel` b/c it reads more naturally and Rust doesn't have the same namespacing concerns as C.

elagergren-spideroak (dismissed):

crates/aranya-client/src/client.rs
nikki-aranya: That was my thought too, a lot of them read better when they're not prefixed, I have them combined into a single impl block in the rustdoc API anyways

elagergren-spideroak: > I wonder if they should be under another type, used like `client.afc().create_bidi_channel()`

That works too. It would also allow us to (eventually) split out functionality. The initial impl could just be

```rust
pub struct Afc<'a> {
    client: &'a Client,
}
impl Afc<'_> {
    pub async fn create_bidi_channel(...) { ... }
}
```

or similar.

nikki-aranya: The problem is that we don't currently expose Afc like we do Team, we'd have to add an accessor. Regardless, this PR majorly overlaps with my rustdoc/API work, rebasing yesterday was hell 😅

gknopf-aranya:
@nikki-aranya I found some more prefixes we'll want to add:
assign_net_identifier -> assign_afc_net_identifier
remove_net_identifier -> remove_afc_net_identifier

gknopf-aranya:
Not sure about the `*_label` API methods yet since I don't know if we're reusing AFC labels for ACQ or using different labels.

@djohnson-spideroak could you confirm which approach we're using for MVP?

nikki-aranya:
wrt the conversation: we decided we're gonna go with "natural names" for any Rust functions, C API gets "aranya_afc_" for everything, and we can address moving those Client methods to the Afc/FastChannel type in a future PR, probably alongside #116

gknopf-aranya:
We also confirmed that AFC and AQC are both using the same label methods.
So `assign_label` and `create_label` methods should not have either an afc or aqc prefix associated with them.

gknopf-aranya (commented):
lgtm! I can't approve since I opened the PR. Would like @elagergren-spideroak or @jdygert-spok to take another look at it.

gknopf-aranya:
Did you retest the Rust example app with these changes?

djohnson-spideroak (approved):

jdygert-spok (approved):

gknopf-aranya:
Let's wait to merge this until after the 0.5.0 release.

PR #108, opened 2025-03-10
101/create rust example app in example dir

smckeon-so:
There was some confusion from a client about running the Rust example. This PR adds a Rest example alongside the C example in the example directory. This will give the user several ways to run a Rust example. 

This will need to be merged in after https://github.com/aranya-project/aranya/pull/107

### DIFF
--- a/examples/rust/README.md
+++ b/examples/rust/README.md
@@ -1,0 +1,66 @@
+# Example Rust Application
+
+This crate contains example code showing how to use Aranya in a Rust application.
+
+This example shows how to use the [`aranya-client`](../../crates/aranya-client) library to:
+- [Setup a team](https://aranya-project.github.io/aranya-docs/getting-started/walkthrough/#create-team)
+- [Sync Aranya graphs](https://aranya-project.github.io/aranya-docs/getting-started/walkthrough/#syncer)
+- [Create an Aranya Fast Channel](https://aranya-project.github.io/aranya-docs/getting-started/walkthrough/#off-graph-messaging)
+- [Send encrypted data between peers](https://aranya-project.github.io/aranya-docs/getting-started/walkthrough/#send-messages)
+
+During setup, the example application starts an instance of the [`aranya-daemon`](../../crates/aranya-daemon) for each Aranya device in the background. [The daemon](https://aranya-project.github.io/aranya-docs/technical-apis/rust-api/#aranya-daemon) handles low-level operations such as automatically syncing graph states between different devices so that [the client](https://aranya-project.github.io/aranya-docs/technical-apis/rust-api/#aranya-client) can focus on the operations it wants to perform on the team.
+
+## Running the example
+
+First, download the source code from this repository:
+
+`$ git clone git@github.com:aranya-project/aranya.git`
+
+### Running the example script:
+
+A `run.bash` bash script is provided to help with running the example.
+Note that the script automatically starts the `aranya-daemon` executable before
+running the example Rust application.
+
+```bash
+$ cd aranya
+$ cargo make run-capi-example
+```
+
+This starts the example and automatically runs through the aforementioned
+steps.
+
+### Building and running the example manually
+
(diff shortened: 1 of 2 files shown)


examples/rust/README.md
```diff
@@ -0,0 +1,32 @@
...
+- [Create an Aranya Fast Channel](https://aranya-project.github.io/aranya-docs/getting-started/walkthrough/#off-graph-messaging)
+- [Send encrypted data between peers](https://aranya-project.github.io/aranya-docs/getting-started/walkthrough/#send-messages)
+
+During setup, the example application starts an instance of the [`aranya-daemon`](../../crates/aranya-daemon) for each Aranya device in the background. [The daemon](https://aranya-project.github.io/aranya-docs/technical-apis/rust-api/#aranya-daemon) handles low-level operations such as automatically syncing graph states between different devices so that [the client](https://aranya-project.github.io/aranya-docs/technical-apis/rust-api/#aranya-client) can focus on the operations it wants to perform on the team.
+
+# Building the Example
+
+You can then build your new project (release mode will provide faster binaries):
+```
+$ cd examples/rust && cargo build --release
```
gknopf-aranya, line 17:
```suggestion
$ cd <repo>/examples/rust && cargo build --release
```

examples/rust/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+# Running the Example
+
+Finally, you can run your newly created project[^1]:
+```
+$ target/release/aranya-example
```
gknopf-aranya, line 24:
```suggestion
$ <repo>/target/release/aranya-example
```

gknopf-aranya:
This PR is copying the `templates/aranya-example` into `examples/rust`.
I think what we should do instead is move the existing template to the `examples/rust` folder and simplify instructions/scripts so it's easier for a user to run the example. That way there is only a single Rust example app to maintain.

cc: @jdygert-spok

examples/rust/run.bash
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# Copyright (c) SpiderOak, Inc. All rights reserved.
+
+cargo install --locked cargo-generate
+cd examples/rust
+cargo generate aranya-project/aranya templates/aranya-example
```
gknopf-aranya, line 7: Let's add the following to the script to make the user's life even easier:
```
cargo build ...
ARANYA_EXAMPLE=info target/release/aranya-example
```

gknopf-aranya:
Like this approach even better. This will let use keep the `templates` folder while adding a script to run it in `examples/rust`.

examples/rust/README.md
gknopf-aranya, line 1: There is a lot of overlap between the READMEs in `examples/rust` and `templates/aranya-example`.
We should pick a spot to keep the primary readme and possibly reference the primary one from the other one.
I vote to keep the primary README for the Rust example in `examples/rust/README.md`. The README in `templates/aranya-example` can link to the primary one.

smckeon-so:
> This PR is copying the `templates/aranya-example` folder to `examples/rust`. I think what we should do instead is move the existing template to the `examples/rust` folder and simplify instructions/scripts so it's easier for a user to run the example. That way there is only a single Rust example app to maintain.
> 
> cc: @jdygert-spok

Yeah, I'm going to bring it together into one thing.

gknopf-aranya:
Once you have the `run.bash` script working, it should be trivial to call that script from `Makfile.toml` and get it running in ci.

examples/rust/run.bash
gknopf-aranya, line 7: This `cargo generate` command is generating a project based on the template in the `main` branch of the repo.
We'll want to use the current commit of the repo that's checked out instead:
```suggestion
cd <repo>
cargo generate --path templates/aranya-example --name examples/rust/aranya-example-template
```

nikki-aranya: If we do this, then we need to pin the template's dependencies to a path/git because the example will be updated but the crates are still from crates.io, if we use the public release then we're guaranteed it'll compile

gknopf-aranya: Eventually, we'll want to be able to test on the current branch during CI tests.

PR #111, opened 2025-03-11
sync now

### LINKED ISSUE #80: sync: AddSyncPeer should take a config
A config parameter should be added to the AddSyncPeer calls. This config will be used to configure the interval between syncs, and other configuration values that might be useful. This issue can be considered resolved once a new config type is added and the AddSyncPeer function is updated.

### LINKED ISSUE #81: sync: AddSyncPeer config should have option to sync immediately
blocked on #80 

When adding a sync peer, the config should have an option to initiate a sync immediately. Determine if this behavior should be on-by-default or off-by-default.

### LINKED ISSUE #82: sync: implement SyncNow
SyncNow should immediately initiate a sync with the given user. 

- `SyncNow(address, team_id, Option<config>) -> bool` - Trigger an immediate sync for the peer. If a sync config is provided, use that. If no config arg is provided, fallback to the config used when the peer was added via AddSyncPeer. If the peer was not added, use a default config or error. 

from the MVP spec

gknopf-aranya:
~~Add boolean flag to `add_sync_peer()` to sync immediately when adding a new peer.~~

Resolves #80 - Adds a new `SyncPeerConfig` type. `AddSyncPeer` client API function takes  ``SyncPeerConfig` type as an arg

Resolves #81 - `SyncPeerConfig` type has a field that determines when the first sync should occur.
Resolves #82 - Add `sync_now()` method to API to allow syncing with peer immediately.


gknopf-aranya:
Currently getting an error compiling the C API. Doesn't like the `sync_now: bool` that was added:
```
gknopf@geoffreys-macbook-pro aranya % cargo make run-capi-example
[cargo-make] INFO - cargo make 0.37.24
[cargo-make] INFO - 
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: run-capi-example
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: install-nightly
[cargo-make] INFO - Execute Command: "rustup" "run" "nightly-2024-10-13" "cbindgen" "--config=cbindgen.toml" "--output=output/aranya-client.h"
...
build command for `aranya-client-capi v0.0.0 (/Users/gknopf/scm/aranya/crates/aranya-client-capi)`\n\nCaused by:\n  process didn't exit successfully: `/Users/gknopf/scm/aranya/target/debug/build/aranya-client-capi-5c50c3915377d98c/build-script-build` (exit status: 101)\n  --- stdout\n  cargo::rustc-check-cfg=cfg(cbindgen)\n  cargo::rustc-cfg=cbindgen\n\n  --- stderr\n  thread 'main' panicked at /Users/gknopf/.cargo/registry/src/index.crates.io-6f17d22bba15001f/aranya-capi-codegen-0.1.0/src/ast/expand.rs:1230:62:\n  unknown type\n  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace\nwarning: build failed, waiting for other jobs to finish...\n")
ERROR: Couldn't generate bindings for /Users/gknopf/scm/aranya/crates/aranya-client-capi.
Error while executing command, exit code: 1
```

crates/aranya-client/src/client.rs
```diff
@@ -335,11 +337,20 @@ pub struct Team<'a> {
...
+    pub async fn add_sync_peer(&mut self, addr: Addr, config: SyncPeerConfig) -> Result<()> {
+        Ok(self
+            .client
+            .daemon
+            .add_sync_peer(context::current(), addr, self.id, config)
+            .await??)
+    }
+
+    /// Sync with peer immediately.
+    pub async fn sync_now(&mut self, addr: Addr, maybe_cfg: Option<SyncPeerConfig>) -> Result<()> {
```
gknopf-aranya, line 349: @ssweetney-spok, @djohnson-spideroak, @elagergren-spideroak what happens if `sync_now` is called without a config and the peer does not have an existing peer config from `add_sync_peer`?
I think this should error because a config is needed to use AQC.
A config isn't needed to use AFC though, so the failure could check for a specific error case.

crates/aranya-client/tests/tests.rs
```diff
@@ -290,8 +290,8 @@ impl Drop for UserCtx {
 /// `RUST_LOG="aranya_client=debug,aranya_daemon=info" cargo test integration_test -- --show-output --nocapture`
 #[test(tokio::test(flavor = "multi_thread"))]
 async fn integration_test() -> Result<()> {
-    let sync_interval = Duration::from_millis(100);
-    let sleep_interval = sync_interval * 6;
+    let sync_config: SyncPeerConfig = Default::default();
```
gknopf-aranya, line 293: Can use shorter variables names:
```suggestion
    let cfg = SyncPeerCfg::default();
```

crates/aranya-client/tests/tests.rs
```diff
@@ -325,44 +325,46 @@ async fn integration_test() -> Result<()> {
     let mut operator_team = team.operator.client.team(team_id);
     let mut member_team = team.membera.client.team(team_id);
 
+    owner_team.sync_now(admin_addr.into(), None).await?;
+
     owner_team
-        .add_sync_peer(admin_addr.into(), sync_interval)
+        .add_sync_peer(admin_addr.into(), sync_config)
```
gknopf-aranya, line 331:
```suggestion
        .add_sync_peer(admin_addr.into(), sync_cfg)
```
or
```suggestion
        .add_sync_peer(admin_addr.into(), cfg)
```

crates/aranya-daemon-api/src/service.rs
```diff
@@ -132,6 +132,22 @@ impl From<Id> for AfcId {
 // serialized command which must be passed over AFC.
 pub type AfcCtrl = Vec<Box<[u8]>>;
 
+/// Configuration values for syncing with a peer
+#[derive(Serialize, Deserialize, Clone, Copy, Debug)]
+pub struct SyncPeerConfig {
```
gknopf-aranya, line 137: We may want to add a todo comment here for the AQC config.

gknopf-aranya: Nevermind, that's not needed for sync. Only for AQC.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -132,6 +132,22 @@ impl From<Id> for AfcId {
...
+    pub interval: Duration,
+    pub sync_now: bool,
+}
+
+impl Default for SyncPeerConfig {
+    fn default() -> Self {
```
gknopf-aranya, line 143: We'll want to have a `new(interval, sync_now)` method so the config can be set to something other than the default without manually constructing the object struct.

crates/aranya-client/src/client.rs
djohnson-spideroak: MVP spec says we should either use a default or error, I think we should use a default config. I see a usecase where you dont want to add a user as a sync peer, but instead just want to sync once with them.

crates/aranya-daemon-api/src/service.rs
gknopf-aranya: I'd prefer not to provide users with a `default()` constructor for `SyncPeerConfig`.
The chance of someone misconfiguring the sync interval and not knowing why seems too high.
It's okay for testing.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -144,7 +160,10 @@ pub trait DaemonApi {
     async fn get_device_id() -> Result<DeviceId>;
 
     /// Adds the peer for automatic periodic syncing.
-    async fn add_sync_peer(addr: Addr, team: TeamId, interval: Duration) -> Result<()>;
+    async fn add_sync_peer(addr: Addr, team: TeamId, config: SyncPeerConfig) -> Result<()>;
+
+    /// Sync with peer immediately.
+    async fn sync_now(addr: Addr, team: TeamId, maybe_cfg: Option<SyncPeerConfig>) -> Result<()>;
```
gknopf-aranya, line 166: `Option<SyncPeerConfig>` is enough to know the parameter is optional without including `maybe` in the parameter name.
```suggestion
    async fn sync_now(addr: Addr, team: TeamId, cfg: Option<SyncPeerConfig>) -> Result<()>;
```

crates/aranya-daemon/src/sync.rs
```diff
@@ -78,6 +96,40 @@ impl SyncPeers {
...
+
+        Ok(())
+    }
+
+    /// Sync with a peer immediately.
+    pub async fn sync_now(
+        &self,
+        addr: Addr,
+        graph_id: GraphId,
+        maybe_cfg: Option<SyncPeerConfig>,
```
gknopf-aranya, line 110:
```suggestion
        cfg: Option<SyncPeerConfig>,
```

crates/aranya-daemon/src/sync.rs
```diff
@@ -78,6 +96,40 @@ impl SyncPeers {
...
+    ) -> Result<()> {
+        let _cfg = match maybe_cfg {
+            Some(c) => c,
+            None => self
+                .cfgs
+                .get(&(addr, graph_id))
+                .cloned()
+                .unwrap_or_default(),
+        };
```
gknopf-aranya, line 119: Or use a default then set to something else if the config is set.
```suggestion
let mut cfg = SyncPeerCfg::default()
if let Some(sync_cfg) = self.cfgs.get(...)... {
    cfg = sync_cfg
}
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -349,6 +349,36 @@ pub struct AfcConfig {
     pub addr: *const c_char,
 }
 
+/// A type that represents when the first sync with a peer should occur.
```
gknopf-aranya, line 352: We should add a todo to change this when booleans are supported that includes a link to the relevant capi codegen issue.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -478,7 +534,7 @@ pub unsafe fn add_sync_peer(
         client
             .inner
             .team(team.0)
-            .add_sync_peer(addr, interval.into()),
+            .sync_now(addr, Some(config.into())),
```
gknopf-aranya, line 537: The way this is currently implemented, the sync peer config is optional for the Rust API but must always be specified for the C API. We should allow a user of the C API to provide a `NULL` value for the `SyncPeerConfig` which translates to an optional value of `None` in Rust.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -132,6 +132,22 @@ impl From<Id> for AfcId {
 // serialized command which must be passed over AFC.
 pub type AfcCtrl = Vec<Box<[u8]>>;
 
+/// Configuration values for syncing with a peer
+#[derive(Serialize, Deserialize, Clone, Copy, Debug)]
+pub struct SyncPeerConfig {
+    pub interval: Duration,
```
gknopf-aranya, line 138: Rust docs for public fields

gknopf-aranya:
@ssweetney-spok this job appears to be failing due to formatting in the `defs.rs` file:
https://github.com/aranya-project/aranya/actions/runs/13847263451/job/38748157519?pr=111
@nikki-aranya saw something similar on the afc prefix PR and could help if needed.

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: @elagergren-spideroak @jdygert-spok Is using NULL for optional params like this acceptable?

ssweetney-spok: Yea, null pointers should work once https://github.com/aranya-project/aranya-core/pull/159 is done

djohnson-spideroak (commented):
pretty much LGTM pending resolution of comments

crates/aranya-daemon-api/src/service.rs
elagergren-spideroak, line 137: A couple of things:

1. `aranya-client` shouldn't expose the `aranya-daemon` RPC type. It should have its own config and convert into an internal RPC type.
2. Given (1), let's make `aranya-client`'s config type use the builder pattern: We're going to need to support a number of different arguments in the future and the builder pattern lets us make API changes _much_ more easily.

```rust 
pub struct PeerSyncConfig {}
impl PeerSyncConfig {
    pub fn builder() -> PeerSyncConfigBuilder { }
}
pub struct PeerSyncConfigBuilder { }
impl PeerSyncConfigBuilder {
    pub fn build() -> PeerSyncConfig { }
}
```

elagergren-spideroak (changes requested):

crates/aranya-client/src/client.rs
elagergren-spideroak, line 349: In the docs, please describe the difference between `Some` and `None` for `maybe_cfg`.

gknopf-aranya:
As discussed today, we are going to unblock this PR by not requiring the C API for adding a sync peer to use a boolean arg in the config. When adding a sync peer, it will always sync right away when the C API method is invoked. The Rust API can continue to allow `sync_now` to be configurable.

ssweetney-spok:
> As discussed today, we are going to unblock this PR by not requiring the C API for adding a sync peer to use a boolean arg in the config. When adding a sync peer, it will always sync right away when the C API method is invoked. The Rust API can continue to allow `sync_now` to be configurable.

I believe @jdygert-spok suggested to mark the config as "opaque" since it wouldn't require all the fields to implement `aranya_capi_core::types::ByConstPtr`. So if we do that and use the builder pattern that should work fine

crates/aranya-client/src/client.rs
ssweetney-spok: Going with the default in this case

crates/aranya-daemon-api/src/service.rs
ssweetney-spok: > I'd prefer not to provide users

This is different from the type that users will use. They'll use the the type defined in "crates/aranya-client/src/client.rs"

Also, it looks like we'll want to keep a default implementation if a config isn't provided when a user calls "sync now"

crates/aranya-daemon/src/sync.rs
ssweetney-spok: Generally, it's better to generate values until you need them. With this suggestion, we'd generate the default config and may not even use it.

jdygert-spok: 🤓 
```suggestion
        let _cfg = maybe_cfg
            .or_else(|| self.cfgs.get(&(addr, graph_id)).cloned())
            .unwrap_or_default();
```

ssweetney-spok: Nice

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: I removed this type so it's no longer used

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+#[aranya_capi_core::opaque(size = 16, align = 8)]
+pub struct SyncPeerConfigBuilder {
+    interval: Duration,
+    sync_now: bool,
+}
```
djohnson-spideroak, line 38: side note: do we have any protections in place in the event that an underlying type (like Duration) has it's size changed?

crates/aranya-client/src/client.rs
```diff
@@ -263,3 +274,61 @@ impl Team<'_> {
...
+    pub fn interval(mut self, duration: Duration) -> Self {
+        self.interval = duration;
+        self
+    }
+
+    pub fn sync_now(mut self, sync_now: bool) -> Self {
+        self.sync_now = sync_now;
+        self
+    }
+}
```
djohnson-spideroak, line 325: docs

crates/aranya-daemon/src/sync.rs
```diff
@@ -44,29 +46,45 @@ struct SyncPeer {
...
+        &mut self,
+        addr: Addr,
+        graph_id: GraphId,
+        cfg: SyncPeerConfig,
+    ) -> Result<()> {
         let peer = Msg::AddPeer {
             peer: SyncPeer { addr, graph_id },
-            interval,
+            interval: cfg.interval,
         };
```
djohnson-spideroak, line 72: should this be sending the full SyncPeerConfig? Could a future syncer implementation benefit from having the additional information?

crates/aranya-client-capi/src/imp/config.rs
ssweetney-spok: Good question. There is [this check](https://github.com/aranya-project/aranya-core/blob/1f2befa16d071a317e2eecb61157d1b8e99ca9a4/crates/aranya-capi-macro/src/opaque.rs#L50-L57) which happens at compile time.

So it looks like if the size of `Duration` grows so that the size of `SyncPeerConfigBuilder` exceeds 16 bytes it'll be caught by that check but I'm not sure what happens if `Duration` gets smaller

djohnson-spideroak (commented):
What happens if I create a config with an interval of 0? or 1 nanosecond?

crates/aranya-client-capi/src/imp/config.rs
elagergren-spideroak: yeah, steve's correct. And smaller is fine.

crates/aranya-daemon/src/sync.rs
ssweetney-spok: Yea, that makes sense

gknopf-aranya:
@ssweetney-spok it looks like there are formatting issues with the header file checked in. Could you run `cargo make fmt` and check in the reformatted aranya header file?
https://github.com/aranya-project/aranya/actions/runs/14070301050/job/39402699495?pr=111

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -478,7 +487,37 @@ pub unsafe fn add_sync_peer(
...
+/// will appear in the tracing logs and
+/// Aranya will be unable to sync state with that peer.
+///
+/// @param client the Aranya Client [`Client`].
+/// @param team the team's ID [`TeamId`].
+/// @param addr the peer's Aranya network address [`Addr`].
+/// @param config configuration values for syncing with a peer.
+/// Default values for a sync config will be used if `config` is `NULL`
+/// @relates AranyaClient.
+pub unsafe fn sync_now(
```
gknopf-aranya, line 507: Missing a test for this method in `example.c`

templates/aranya-example/src/main.rs
gknopf-aranya, line 1: Should test `sync_now` here too.

templates/aranya-example/src/main.rs
```diff
@@ -187,6 +187,7 @@ async fn main() -> Result<()> {
 
     let sync_interval = Duration::from_millis(100);
     let sleep_interval = sync_interval * 6;
+    let sync_config = SyncPeerConfig::builder().interval(interval).build();
```
gknopf-aranya, line 190:
```suggestion
    let sync_cfg = SyncPeerConfig::builder().interval(interval).build();
```

examples/c/example.c
```diff
@@ -229,7 +228,43 @@ AranyaError run(Team *t) {
...
+    // Allocate memory for the sync peer config struct
+    struct AranyaSyncPeerConfig *cfg = (struct AranyaSyncPeerConfig *)malloc(sizeof(struct AranyaSyncPeerConfig));
+
+    if (cfg == NULL) {
+        // Handle allocation failure
+        fprintf(stderr, "Memory allocation failed\n");
+        return 1;
+    }
+
+    err = aranya_sync_peer_config_builder_build(builder, cfg);
```
gknopf-aranya, line 261: `builder` is built then cleaned up without being used.
If an object is built it should be used somewhere in the test.

examples/c/example.c
```diff
@@ -229,7 +228,43 @@ AranyaError run(Team *t) {
     EXPECT("error initializing team", err);
 
     // add sync peers.
-    err = add_sync_peers(t);
+
+    // Allocate memory for the config builder struct
+    struct AranyaSyncPeerConfigBuilder *builder = (struct AranyaSyncPeerConfigBuilder *)malloc(sizeof(struct AranyaSyncPeerConfigBuilder));
```
gknopf-aranya, line 233: We should let the C API provide a method for allocating/deallocating types defined in the API rather than manually invoking `malloc`. E.g. `new()` and `cleanup()` methods.

gknopf-aranya: Or define the type without using malloc like this:
```suggestion
    AranyaSyncPeerConfigBuilder builder;
```

examples/c/example.c
```diff
@@ -229,7 +228,43 @@ AranyaError run(Team *t) {
...
+    AranyaDuration interval = ARANYA_DURATION_MILLISECONDS * 100;
+    err = aranya_sync_peer_config_builder_set_interval(builder, interval);
+    EXPECT("error setting duration on config builder", err);
+
+    // Set syncing to happen immediately on the config builder
+    err = aranya_sync_peer_config_builder_set_sync_now(builder);
+    EXPECT("error setting `sync_now` on config builder", err);
+
+    // Allocate memory for the sync peer config struct
+    struct AranyaSyncPeerConfig *cfg = (struct AranyaSyncPeerConfig *)malloc(sizeof(struct AranyaSyncPeerConfig));
```
gknopf-aranya, line 253: Ditto on not using `malloc` here.

gknopf-aranya:
We should add a test to confirm `sync_now` works as expected.
For example, add a command to the graph with no sync peers configured.
Manually sync the command between peers via `sync_now` then confirm that a subsequent operation relying on peers being in sync works.
This will prove `sync_now` caused a graph sync to occur.
The negative test would be to confirm the operation fails before invoking `sync_now`.

examples/c/example.c
jdygert-spok: Yeah these should be stack allocated.

jdygert-spok:
> @ssweetney-spok it looks like there are formatting issues with the header file checked in. Could you run `cargo make fmt` and check in the reformatted aranya header file? https://github.com/aranya-project/aranya/actions/runs/14070301050/job/39402699495?pr=111

It is not formatting but the header needs to be regenerated and checked in. `cargo make build-capi-header`

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: Good call

examples/c/example.c
ssweetney-spok: Good call

examples/c/example.c
gknopf-aranya: Ah okay, so `builder` and `cfg` are both used after they are constructed. I thought this was constructing `builder` but it is constructing `cfg`. `builder_build` is a confusing name for this function because it makes it sound like it's building the builder when it's really building the cfg.

elagergren-spideroak: @gknopf-aranya in Rust terms, `aranya_sync_peer_config_builder_build()` is `aranya::SyncPeerConfigBuilder::build()`

elagergren-spideroak: I don't think the naming can be changed without changing aranya-capi-codegen. I also can't think of a better naming scheme.

examples/c/example.c
elagergren-spideroak: > We should let the C API provide a method for allocating/deallocating types defined in the API rather than manually invoking `malloc`. E.g. `new()` and `cleanup()` methods.

All types in our APIs are either

a. Completely defined and initialized via `Foo_init(Foo* foo)` methods
b. Forward declared and constructed via `Foo* Foo_new()` methods.

Providing `Foo* Foo_new()` constructors for completely defined types is a mistake. It means users can accidentally write code like this:

```c
Foo* foo = malloc(sizeof(Foo)); // oops, should have used `Foo_new`
[...]
Foo_free(foo); // oops, undefined behavior
```

All or almost all of our types use option (a) so that users can allocate our types however they please—stack, heap, or static memory. So, @ssweetney-spok's original code was correct (although I agree allocating it on the stack is simpler and better for this example).

ssweetney-spok:
> What happens if I create a config with an interval of 0? or 1 nanosecond?

Bad things will happen 🙃. But on a serious note, we should probably have a lower bound on sync interval that we enforce

gknopf-aranya:
> Bad things will happen 🙃. But on a serious note, we should probably have a lower bound on sync interval that we enforce

It won't actually sync that quickly. It will turn the delay queue into a regular queue that syncs as quickly as possible with the next peer in the queue in a round-robin fashion.

examples/c/example.c
ssweetney-spok: Well the naming can be changed since they're just defined as free functions in "crates/aranya-client-capi/src/api/defs.rs" but I agree with you on keeping this as the naming scheme

gknopf-aranya:
If a peer has a sync interval of 0 could it starve syncing with other peers or will it fairly rotate between peers because eventually other sync intervals will also have a zero delay in the delay queue after waiting a while?

elagergren-spideroak:
> > Bad things will happen 🙃. But on a serious note, we should probably have a lower bound on sync interval that we enforce
> 
> It won't actually sync that quickly. It will turn the delay queue into a regular queue that syncs as quickly as possible with the next peer in the queue in a round-robin fashion.

Note that ~most non-real-time systems have a granularity in microseconds or milliseconds depending on the system configuration. I think it's probably fine to not enforce a lower bound and just say that it'll do whatever your system is set up to do.

djohnson-spideroak (dismissed):

crates/aranya-client/tests/tests.rs
```diff
@@ -1094,3 +1103,62 @@ async fn test_afc_monotonic_seq() -> Result<()> {
 
     Ok(())
 }
+
+#[test(tokio::test(flavor = "multi_thread"))]
```
gknopf-aranya, line 1107:
```suggestion
/// Tests sync_now() by demonstrating that an admin cannot assign a role to a device until it syncs with the owner.
#[test(tokio::test(flavor = "multi_thread"))]
```

crates/aranya-client/tests/tests.rs
```diff
@@ -1094,3 +1103,62 @@ async fn test_afc_monotonic_seq() -> Result<()> {
...
+    let owner_addr = team.owner.aranya_local_addr().await?;
+
+    // setup team handles.
+    let mut owner_team = team.owner.client.team(team_id);
+    let mut admin_team = team.admin.client.team(team_id);
+
+    // add admin to team.
+    info!("adding admin to team");
+    owner_team.add_device_to_team(team.admin.pk.clone()).await?;
+    // owner_team.assign_role(team.admin.id, Role::Admin).await?;
```
gknopf-aranya, line 1133: Remove or uncomment this code

crates/aranya-client/tests/tests.rs
```diff
@@ -1094,3 +1103,62 @@ async fn test_afc_monotonic_seq() -> Result<()> {
...
+
+    // add operator to team.
+    info!("adding operator to team");
+    owner_team
+        .add_device_to_team(team.operator.pk.clone())
+        .await?;
+
+    // Assign role to Admin
+    owner_team.assign_role(team.admin.id, Role::Admin).await?;
+    sleep(Duration::from_secs(1)).await;
```
gknopf-aranya, line 1143: This sleep was needed when there was a periodic sync in the background. If syncing is manual, no delay is needed.

crates/aranya-client/tests/tests.rs
```diff
@@ -1094,3 +1103,62 @@ async fn test_afc_monotonic_seq() -> Result<()> {
...
+
+    // Admin tries to assign a role
+    match admin_team
+        .assign_role(team.operator.id, Role::Operator)
+        .await
+    {
+        Ok(_) => bail!("Expected role assignment to fail"),
+        Err(aranya_client::Error::Daemon(_)) => {}
```
gknopf-aranya, line 1151: This should verify that the expected error occurs in the daemon.
But given that detailed error reporting in the daemon API isn't implemented that could be added on a different PR.

djohnson-spideroak (dismissed):

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -993,6 +1032,39 @@ pub unsafe fn afc_recv_data(
...
+    cfg.deref_mut().sync_now(true);
+}
+
+/// Set the sync_now field on the config builder to false
+/// When this field is set to false, syncing occurs after some specifed duration from when the peer is added
+///
+/// @param cfg a pointer to the builder for a sync config
+pub fn sync_peer_config_builder_set_sync_later(cfg: &mut SyncPeerConfigBuilder) {
+    cfg.deref_mut().sync_now(false);
+}
```
ssweetney-spok, line 1057: Note that this is split into 2 functions because https://github.com/aranya-project/aranya-core/issues/129 has not been implemented yet

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -993,6 +1032,39 @@ pub unsafe fn afc_recv_data(
...
+/// Set the interval field on the config builder
+///
+/// @param cfg a pointer to the builder for a sync config
+/// @param interval Set the interval at which syncing occurs
+pub fn sync_peer_config_builder_set_interval(cfg: &mut SyncPeerConfigBuilder, interval: Duration) {
+    cfg.deref_mut().interval(interval);
+}
+
+/// Set the sync_now field on the config builder to true.
+/// When this field is set to true, syncing happens immediately after a peer is added
```
nikki-aranya, line 1044: Extraneous since nothing outside this function implies true/false and especially not in the C header

crates/aranya-client/tests/tests.rs
```diff
@@ -818,59 +824,59 @@ async fn test_afc_monotonic_seq() -> Result<()> {
...
-        .add_sync_peer(owner_addr.into(), sync_interval)
+        .add_sync_peer(owner_addr.into(), sync_config.clone())
         .await?;
     memberb_team
-        .add_sync_peer(admin_addr.into(), sync_interval)
+        .add_sync_peer(admin_addr.into(), sync_config.clone())
         .await?;
     memberb_team
-        .add_sync_peer(operator_addr.into(), sync_interval)
+        .add_sync_peer(operator_addr.into(), sync_config.clone())
```
nikki-aranya, line 876: We'll have to make sure sync_config stays cheaply cloneable (right now it's a u64+u32+bool) but it seems fine for now, will probably optimize to a double u64 move

examples/c/example.c
```diff
@@ -228,36 +227,75 @@ AranyaError run(Team *t) {
...
     // add operator to team.
     err = aranya_add_device_to_team(&t->clients.owner.client, &t->id,
                                     &t->clients.operator.pk);
     EXPECT("error adding operator to team", err);
 
-    // upgrade role to operator.
+    // upgrade role to admin.
     err = aranya_assign_role(&t->clients.owner.client, &t->id,
-                             &t->clients.operator.id, ARANYA_ROLE_OPERATOR);
+        &t->clients.admin.id, ARANYA_ROLE_ADMIN);
```
nikki-aranya, line 242: nit: these should stay aligned with the &t on the previous line but I don't know that we have an automatic formatter we use for C so it's eh

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -17,25 +46,6 @@ impl Typed for ClientConfig {
...
-    pub max_channels: usize,
-    /// Address to bind AFC server to.
-    pub addr: *const c_char,
-}
-
-#[cfg(feature = "afc")]
-impl Typed for AfcConfig {
-    const TYPE_ID: TypeId = TypeId::new(0x227DFC9F);
-}
-
```
nikki-aranya, line 38: Cannot remove

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -354,6 +373,11 @@ typedef struct ARANYA_ALIGNED(1) AranyaChannelId {
 } AranyaChannelId;
 #endif
 
+/**
+ * A type to represent a span of time.
+ */
+typedef uint64_t AranyaDuration;
```
elagergren-spideroak, line 379: This needs to specify units.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1087,6 +1126,39 @@ pub unsafe fn afc_recv_data(
...
 
+/// Set the interval field on the config builder
+///
+/// @param cfg a pointer to the builder for a sync config
+/// @param interval Set the interval at which syncing occurs
+pub fn sync_peer_config_builder_set_interval(cfg: &mut SyncPeerConfigBuilder, interval: Duration) {
+    cfg.deref_mut().interval(interval);
+}
+
+/// Set the sync_now field on the config builder to true.
```
elagergren-spideroak, line 1137:
```rust
/// Updates the config to immediately sync with the peer.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1087,6 +1126,39 @@ pub unsafe fn afc_recv_data(
     Ok(true)
 }
 
+/// Set the interval field on the config builder
```
elagergren-spideroak, line 1129:
```rust
/// Configures how often the peer will be synced with.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1087,6 +1126,39 @@ pub unsafe fn afc_recv_data(
...
+
+/// Set the sync_now field on the config builder to true.
+///
+/// @param cfg a pointer to the builder for a sync config
+// TODO: aranya-core#129
+pub fn sync_peer_config_builder_set_sync_now(cfg: &mut SyncPeerConfigBuilder) {
+    cfg.deref_mut().sync_now(true);
+}
+
+/// Set the sync_now field on the config builder to false
```
elagergren-spideroak, line 1145:
```rust
/// Updates the config to *not* immediately sync with the peer.
///
/// Overrides [`sync_peer_config_builder_set_sync_now`] if invoked afterward.
```

ssweetney-spok (approved):
LGTM! Thanks @gknopf-aranya

crates/aranya-daemon-api/src/service.rs
```diff
@@ -132,6 +132,24 @@ impl From<Id> for AfcId {
...
+    /// The interval at which syncing occurs
+    pub interval: Duration,
+    /// Determines if a peer should be synced with immediately after they're added
+    pub sync_now: bool,
+}
+
+impl Default for SyncPeerConfig {
+    fn default() -> Self {
+        Self {
+            interval: Duration::from_millis(100),
```
elagergren-spideroak, line 147: Given our target audience, 10x per second seems a bit much. Maybe 1x per second?

crates/aranya-client/src/client.rs
```diff
@@ -446,3 +457,68 @@ impl Queries<'_> {
...
+
+impl Default for SyncPeerConfigBuilder {
+    fn default() -> Self {
+        Self {
+            interval: Duration::from_millis(100),
```
elagergren-spideroak, line 520: ditto for duration

crates/aranya-daemon-api/src/service.rs
```diff
@@ -132,6 +132,24 @@ impl From<Id> for AfcId {
 // serialized command which must be passed over AFC.
 pub type AfcCtrl = Vec<Box<[u8]>>;
 
+/// Configuration values for syncing with a peer
+#[derive(Serialize, Deserialize, Clone, Copy, Debug)]
```
elagergren-spideroak, line 136: nit: fix the ordering `Clone, Debug, Serialize, Deserializes`

elagergren-spideroak, line 136: Get rid of the `Copy` bound.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -132,6 +132,24 @@ impl From<Id> for AfcId {
...
+pub struct SyncPeerConfig {
+    /// The interval at which syncing occurs
+    pub interval: Duration,
+    /// Determines if a peer should be synced with immediately after they're added
+    pub sync_now: bool,
+}
+
+impl Default for SyncPeerConfig {
```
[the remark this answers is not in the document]
ssweetney-spok: Nah, this was left over from before I implemented the builder pattern. It can be removed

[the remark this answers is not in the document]
ssweetney-spok: Wait, I forgot it is used when the `sync_now` API is called with no config

```
pub async fn sync_now(
        &self,
        addr: Addr,
        graph_id: GraphId,
        cfg: Option<SyncPeerConfig>,
    ) -> Result<()> {
        // TODO: Use this value later
        let _cfg = cfg
            .or_else(|| self.cfgs.get(&(addr, graph_id)).cloned())
            .unwrap_or_default();
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1087,6 +1126,40 @@ pub unsafe fn afc_recv_data(
     Ok(true)
 }
 
+/// Configures how often the peer will be synced with.
```
elagergren-spideroak, line 1129: Also:

```rust
/// Configures [...]
///
/// By default, the peer will be synced with every XXX, including
/// once immediately if [`sync_peer_config_builder_set_sync_now`]
/// is specified.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1087,6 +1126,40 @@ pub unsafe fn afc_recv_data(
...
+///
+/// @param cfg a pointer to the builder for a sync config
+/// @param interval Set the interval at which syncing occurs
+pub fn sync_peer_config_builder_set_interval(cfg: &mut SyncPeerConfigBuilder, interval: Duration) {
+    cfg.deref_mut().interval(interval);
+}
+
+/// Updates the config to immediately sync with the peer.
```
elagergren-spideroak, line 1137: Also:

```rust
/// Updates [...]
///
/// By default, the peer is (not ?) synced with immediately.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1087,6 +1126,40 @@ pub unsafe fn afc_recv_data(
...
+///
+/// @param cfg a pointer to the builder for a sync config
+// TODO: aranya-core#129
+pub fn sync_peer_config_builder_set_sync_now(cfg: &mut SyncPeerConfigBuilder) {
+    cfg.deref_mut().sync_now(true);
+}
+
+/// Updates the config to *not* immediately sync with the peer.
```
elagergren-spideroak, line 1145: Also:

```rust
/// Updates [...]
///
/// By default, the peer is (not ?) synced with immediately.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -566,7 +575,37 @@ pub unsafe fn add_sync_peer(
...
+    )?;
+    Ok(())
+}
+
+/// Sync with peer immediately.
+///
+/// If a peer is not reachable on the network, sync errors
+/// will appear in the tracing logs and
+/// Aranya will be unable to sync state with that peer.
+///
```
elagergren-spideroak, line 588:
```rust
/// This function ignores [`sync_peer_config_builder_set_sync_later`], if set.
```

crates/aranya-client/src/client.rs
```diff
@@ -446,3 +457,68 @@ impl Queries<'_> {
...
+
+    /// Build a [`SyncPeerConfig`]
+    pub fn build(self) -> SyncPeerConfig {
+        SyncPeerConfig {
+            interval: self.interval,
+            sync_now: self.sync_now,
+        }
+    }
+
+    /// Set the interval at which syncing occurs
```
elagergren-spideroak, line 503:
```rust
/// Sets the interval [...]
///
/// By default, the peer is synced with every XXX and once immediately
/// if [`sync_now`][Self::sync_now] is enabled.
```

crates/aranya-client/src/client.rs
```diff
@@ -208,10 +208,21 @@ pub struct Team<'a> {
...
             .daemon
-            .add_sync_peer(context::current(), addr, self.id, interval)
+            .add_sync_peer(context::current(), addr, self.id, config.into())
+            .await?
+            .map_err(Into::into)
+    }
+
+    /// Sync with peer immediately.
+    /// Default values for a sync config will be used if `config` is `None`.
+    /// Otherwise the provided config will be used for syncing.
```
elagergren-spideroak, line 221: Let's hedge a little:

```rust
/// Immediately syncs with the peer.
///
/// If `config` is `None`, default values (including those from the daemon) will
/// be used.
```

elagergren-spideroak (changes requested):
Mostly doc changes.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -115,7 +116,7 @@ impl Default for SyncPeerConfigBuilder {
     fn default() -> Self {
         Self {
             interval: Duration {
-                nanos: 100 * ARANYA_DURATION_MILLISECONDS,
+                nanos: 5 * ARANYA_DURATION_SECONDS,
```
djohnson-spideroak, line 119: nit: imo 5 seconds is too long, once per second is probably better.

djohnson-spideroak (approved):

crates/aranya-client-capi/src/imp/config.rs
gknopf-aranya: Agree with 1s. Network latency is generally less than 100ms.

crates/aranya-client/src/client.rs
```diff
@@ -446,3 +458,73 @@ impl Queries<'_> {
...
+    pub fn build(self) -> SyncPeerConfig {
+        SyncPeerConfig {
+            interval: self.interval,
+            sync_now: self.sync_now,
+        }
+    }
+
+    /// Set the interval at which syncing occurs
+    ///
+    /// /// By default, the peer is synced with every 5 seconds and once immediately
```
elagergren-spideroak, line 507: double `///`

crates/aranya-client-capi/src/imp/config.rs
elagergren-spideroak: > Agree with 1s. Network latency is generally less than 100ms.

It is not just a function of latency. It's a function of how often the graph is updated, how fast we can process graph updates, and a number of other things.

Also, keep in mind that this is supposed to be a "not awful" default. I'd even be in favor of getting rid of the default (making this field `Option<Duration>`) and `add_sync_peer` return an error if the duration is unspecified.

elagergren-spideroak: Actually, the more I think about it the more I think we _should_ do what I wrote above ^

elagergren-spideroak: Pretty much any duration value is going to be awful for somebody, so it's better to force users to always specify it.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -69,6 +80,71 @@ impl ClientConfigBuilder {
...
+    ///
+    /// By default, the peer is immediately synced with.
+    pub fn sync_now(&mut self, sync_now: bool) {
+        self.sync_now = sync_now;
+    }
+
+    /// Build a [`SyncPeerConfig`]
+    pub fn build(&self) -> Result<SyncPeerConfig, super::Error> {
+        let Some(interval) = self.interval else {
+            bug!("Tried to create a `SyncPeerConfig` without setting the interval!");
```
ssweetney-spok, line 106: I used `bug` here but this doesn't really seem like an internal bug. Should I just create a new variant in `aranya-client-capi::imp::Error`?

elagergren-spideroak: InvalidArg https://github.com/aranya-project/aranya/blob/98906c899657a51498ef07e39638212047321f0c/crates/aranya-client-capi/src/imp/error.rs#L22-L24

crates/aranya-daemon-api/src/service.rs
```diff
@@ -132,6 +132,24 @@ impl From<Id> for AfcId {
...
+    fn default() -> Self {
+        Self {
+            interval: Duration::from_secs(1),
```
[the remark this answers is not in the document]
elagergren-spideroak: then maybe let's not use it inside `sync_now`? We're not even using the `_cfg` anyway.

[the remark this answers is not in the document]
ssweetney-spok: Yea, it's not used currently but I think we may in the future. Here's what the spec says

> Trigger an immediate sync for the peer. If a sync config is provided, use that. If no config arg is provided, fallback to the config used when the peer was added via AddSyncPeer. If the peer was not added, use a default config or error.

Should we just error instead?

cc: @djohnson-spideroak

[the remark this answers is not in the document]
elagergren-spideroak: "...or error" is there in case the config has some value that we absolutely need. E.g., the peer's public key for a future "encrypted syncing" extension. The current impl of `sync_now` doesn't need anything from the config, so we can just ignore it for now.

crates/aranya-client/src/client.rs
```diff
@@ -495,19 +496,24 @@ impl SyncPeerConfigBuilder {
     }
 
     /// Build a [`SyncPeerConfig`]
-    pub fn build(self) -> SyncPeerConfig {
-        SyncPeerConfig {
-            interval: self.interval,
+    pub fn build(self) -> Result<SyncPeerConfig> {
+        let Some(interval) = self.interval else {
+            bug!("Tried to create a `SyncPeerConfig` without setting the interval!");
```
djohnson-spideroak, line 501: Is this a bug or just an invalid config?

ssweetney-spok: Not a bug, just an invalid config. The `InvalidArg` type I got suggested to use for C API error only exists in `aranya-capi-core`. I'll add a new error variant  and do something similar for the Rust API

djohnson-spideroak (approved):

gknopf-aranya (commented):
lgtm! I can't approve since I originally opened the PR.

elagergren-spideroak (approved):

PR #113, opened 2025-03-11
client: add config to create_team, add_team

djohnson-spideroak:
Adds a new `TeamConfig` to aranya-client and aranya-daemon-api for use with create_team and add_team. Currently only contains a single field with the current version of the `TeamConfig`, which can later be expanded.


crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -276,6 +276,15 @@ typedef struct ARANYA_ALIGNED(1) AranyaDeviceId {
     uint8_t __for_size_only[64];
 } AranyaDeviceId;
 
+typedef struct ARANYA_ALIGNED(1) AranyaTeamOperationConfig {
```
gknopf-aranya, line 279: Could shorten to:
```suggestion
typedef struct ARANYA_ALIGNED(1) AranyaTeamConfig {
```

crates/aranya-client/tests/tests.rs
```diff
@@ -302,7 +302,7 @@ async fn integration_test() -> Result<()> {
     let team_id = team
         .owner
         .client
-        .create_team()
+        .create_team(TeamOperationConfig {})
```
gknopf-aranya, line 305: May want to include a builder method for the `TeamConfig` type to make it easier to construct in the future.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -73,6 +73,10 @@ pub enum Role {
     Member,
 }
 
+/// A configuration for creating or adding a team to a daemon.
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
```
gknopf-aranya, line 77: `Eq`, `PartialEq` in case we want to compare configs to each other. I'm guessing we'll want to compare version fields when they are added.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -435,9 +440,13 @@ pub fn create_team(client: &mut Client) -> Result<TeamId, imp::Error> {
 /// @param team the team's ID [`TeamId`].
 ///
 /// @relates AranyaClient.
-pub fn add_team(client: &mut Client, team: &TeamId) -> Result<(), imp::Error> {
+pub fn add_team(
```
gknopf-aranya, line 443: Note that `add_team` is not used anywhere in the C example code and is unimplemented in `api.rs`.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -73,6 +73,10 @@ pub enum Role {
...
+pub struct TeamOperationConfig {}
```
gknopf-aranya, line 78: TODO to add version field. We'll want to add a method to check if the version is valid.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -435,9 +440,13 @@ pub fn create_team(client: &mut Client) -> Result<TeamId, imp::Error> {
...
+    client: &mut Client,
+    team: &TeamId,
+    cfg: &TeamOperationConfig,
+) -> Result<(), imp::Error> {
```
gknopf-aranya, line 447: Possibly want a new error type if the version check fails

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok, @elagergren-spideroak completing this PR is blocked on a discussion opened in the Product Engineering channel regarding what fields will be in the Team Config and what value will be used for the version.

crates/aranya-daemon-api/src/service.rs
elagergren-spideroak, line 78: `TeamConfig` is probably a nicer name. I don't know what "Operation" means.

crates/aranya-client/src/client.rs
```diff
@@ -298,13 +300,16 @@ impl Client {
     }
 
     /// Create a new graph/team with the current device as the owner.
-    pub async fn create_team(&mut self) -> Result<TeamId> {
-        Ok(self.daemon.create_team(context::current()).await??)
+    pub async fn create_team(&mut self, cfg: TeamOperationConfig) -> Result<TeamId> {
```
elagergren-spideroak, line 303: Please add a `TeamConfig` to `aranya-client` and then have a conversion from `aranya_client::TeamConfig` to `aranya_daemon_api::TeamConfig`.

crates/aranya-daemon-api/src/service.rs
gknopf-aranya: Similar thoughts

crates/aranya-daemon-api/src/service.rs
elagergren-spideroak, line 77: Remove the `Copy` impl. It's likely that we need to add a non-copy type, like `Vec`.

crates/aranya-client/src/client.rs
elagergren-spideroak: You'll also want to make `aranya_client::TeamConfig` a builder. `aranya_daemon_api::TeamConfig` doesn't really matter.

crates/aranya-daemon-api/src/service.rs
gknopf-aranya: @elagergren-spideroak, @djohnson-spideroak I posted a question on the product engineering channel a while back on what version we want to use in the TeamConfig version field.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -177,6 +177,11 @@ pub struct TeamId(aranya_daemon_api::TeamId);
 #[aranya_capi_core::opaque(size = 64, align = 1)]
 pub struct DeviceId(aranya_daemon_api::DeviceId);
 
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug)]
+#[aranya_capi_core::opaque(size = 0, align = 1)]
+pub struct TeamOperationConfig(aranya_daemon_api::TeamOperationConfig);
```
gknopf-aranya, line 183: Implement a builder pattern for `TeamConfig` with `new()` and `build()` methods. More methods can be added later when other fields are added to the struct.

crates/aranya-daemon-api/src/service.rs
nikki-aranya: If we want to use this in the C API, it has to implement Copy in the current approach. Also if we want a Vec, then we probably want a [u8; LEN] anyways
```
error[E0277]: the trait bound `aranya_daemon_api::TeamConfig: std::marker::Copy` is not satisfied
    --> /aranya/target/debug/build/aranya-client-capi-fc71209ee8f6b9a5/out/generated.rs:2178:62
     |
2178 |             __capi::internal::util::check_valid_input_ty_val(cfg),
     |             ------------------------------------------------ ^^^ the trait `std::marker::Copy` is not implemented for `aranya_daemon_api::TeamConfig`
     |             |
     |             required by a bound introduced by this call
     |
     = help: the trait `ByValue` is implemented for `__AranyaTeamConfigFfiWrapper<Inner>`
note: required for `__AranyaTeamConfigFfiWrapper<aranya_daemon_api::TeamConfig>` to implement `ByValue`
    --> /aranya/target/debug/build/aranya-client-capi-fc71209ee8f6b9a5/out/generated.rs:7082:24
     |
7082 |     unsafe impl<Inner> __capi::types::ByValue for __AranyaTeamConfigFfiWrapper<Inner>
     |                        ^^^^^^^^^^^^^^^^^^^^^^     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
7083 |     where
7084 |         Inner: ::core::marker::Copy,
     |                -------------------- unsatisfied trait bound introduced here
note: required by a bound in `check_valid_input_ty_val`
    --> /.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aranya-capi-core-0.2.1/src/internal/util.rs:64:50
     |
64   | pub const fn check_valid_input_ty_val<T: Input + ByValue>(v: T) -> T {
     |                                                  ^^^^^^^ required by this bound in `check_valid_input_ty_val`

For more information about this error, try `rustc --explain E0277`.
```

crates/aranya-client/src/client.rs
nikki-aranya: What's wrong with reusing TeamConfig? We already expose all of these using aranya_daemon_api
```rs
#[cfg(feature = "afc")]
use aranya_daemon_api::CS;
use aranya_daemon_api::{
    DaemonApiClient, DeviceId, KeyBundle, NetIdentifier, Role, TeamConfig, TeamId,
};
```

crates/aranya-daemon-api/src/service.rs
elagergren-spideroak: (It shouldn't be used in the C API because `aranya-client` should have its own type.)

crates/aranya-client/src/client.rs
elagergren-spideroak: `aranya-daemon-api` is the RPC between `aranya-client` and `aranya-daemon`. It's an implementation detail and it's wrong for us to expose it to users (other than what we're technically forced to expose b/c it's a "glue" crate for `aranya-client` and `aranya-daemon`).

If we re-export `TeamConfig` then we have to support its behavior ~forever, which could prevent us from making RPC changes. Dumb example: maybe we stop using `serde` b/c a new RPC library handles it for us. But then we can't b/c that would be a breaking change.

crates/aranya-daemon-api/src/service.rs
nikki-aranya: Alright, I'll change it, but I'm dissatisfied with us now having to maintain another two types in parallel since either way the RPC has to expose a common TeamConfig that both client and daemon share, I can see the appeal of "masking" the public client one but it's not enough of a justification for me

elagergren-spideroak: The choice is essentially between maintaining two types in parallel or exposing internal implementation details. From an engineering perspective, the former is going to cause much less pain than the latter.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -85,6 +85,11 @@
 
 #define ARANYA_DURATION_NANOSECONDS 1
 
+/**
+ * The current version of `TeamConfig`.
+ */
+#define TEAM_CONFIG_VERSION 1
```
gknopf-aranya, line 91: I have this from my notes on one of the MVP planning or Rust calls:
"Team config - no version, will implement builder APIs new(), build(). Update spec to not have version in config."

gknopf-aranya: Rather than using a version to check whether the configs are compatible, we will use a builder pattern to ensure the config is backward compatible.

nikki-aranya: This is a builder pattern, we don't need an explicit build() right now, I mostly added version due to the above comments

nikki-aranya: https://github.com/aranya-project/aranya/pull/113/files/f10ae3c27f30eba2faccbf1ce951a2fea835a815#diff-76db0923c419899dc4cdf3b1b4310f6298fd8824824cb78b4d5d31642a6799b9

nikki-aranya: On second thought, we probably should isolate it behind a proper Builder type, I'll do that tomorrow

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -102,3 +102,37 @@ impl AfcConfigBuilder {
...
+pub struct TeamConfigBuilder {
+    pub version: u32,
+}
+
+impl TeamConfigBuilder {
+    /// Attempts to construct a [`TeamConfig`], returning an [`Error::Bug`](super::Error::Bug) if
+    /// there are invalid parameters.
+    pub fn build(self) -> Result<TeamConfig, super::Error> {
+        if self.version < aranya_daemon_api::TeamConfig::MINIMUM_VERSION {
+            bug!("Tried to create a TeamConfig with an unsupported version!");
```
ssweetney-spok, line 131: We should decide on a consistent way to handle these misconfiguration errors. I got [some feedback](https://github.com/aranya-project/aranya/pull/111#discussion_r2021781228) on another PR to use `InvalidArg`

nikki-aranya: InvalidArg works

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -515,27 +521,54 @@ pub fn get_device_id(client: &mut Client) -> Result<DeviceId, imp::Error> {
...
     let client = client.deref_mut();
-    let id = client.rt.block_on(client.inner.create_team())?;
+    let cfg = aranya_client::TeamConfigBuilder::new().build()?;
+    let id = client.rt.block_on(client.inner.create_team(cfg))?;
     Ok(TeamId(id))
 }
 
 /// Add a team to the local device store.
 ///
+/// NOTE: this function is unfinished and will panic if called.
```
gknopf-aranya, line 560: Should we return an `Unimplemented` error for now instead of panicking?

nikki-aranya: That's a problem for a different PR, ideally we'd just implement it, it panics because daemon side is a todo!() macro

gknopf-aranya: Seems risky if we don't have time to get to that before MVP.
And we can't remove the API method because we're aiming for API stability in MVP.
I'd rather return an error than panic, until there is an implementation.
Unless the implementation is trivial, then we can just implement it soon.

gknopf-aranya: Here's an issue for implementing `add_team`:
https://github.com/aranya-project/aranya/issues/167

gknopf-aranya (approved):
lgtm! We shouldn't be panicking in methods provided as part of the stable MVP API but that can be solved on a different PR.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -515,27 +521,54 @@ pub fn get_device_id(client: &mut Client) -> Result<DeviceId, imp::Error> {
     Ok(DeviceId(id))
 }
 
+#[aranya_capi_core::opaque(size = 24, align = 8)]
+pub type TeamConfig = Safe<imp::TeamConfig>;
+
+#[aranya_capi_core::opaque(size = 4, align = 4)]
+pub type TeamConfigBuilder = imp::TeamConfigBuilder;
+
+/// Attempts to construct a [`TeamConfig`], returning an `Error::Bug`
```
elagergren-spideroak, line 530: don't you mean `Error::InvalidArgment`?

crates/aranya-client-capi/src/api/defs.rs
elagergren-spideroak: yeah, either remove it or implement it

crates/aranya-client/src/client.rs
```diff
@@ -154,19 +157,19 @@ impl Client {
     }
 
     /// Create a new graph/team with the current device as the owner.
-    pub async fn create_team(&mut self) -> Result<TeamId> {
-        self.daemon
-            .create_team(context::current())
-            .await?
-            .map_err(Into::into)
+    pub async fn create_team(&mut self, cfg: TeamConfig) -> Result<TeamId> {
+        Ok(self
```
elagergren-spideroak, line 161: minor nit: we've gone back and forth on the preferred style for calling the daemon. Please let's just choose `Ok(<big expression>)` or `.map_err(Into::into)`, but not both, thanks.

crates/aranya-client/src/error.rs
```diff
@@ -35,9 +40,14 @@ pub enum Error {
     },
 }
 
+#[derive(Debug, thiserror::Error)]
```
elagergren-spideroak, line 43: needs docs

crates/aranya-client/src/config.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct TeamConfigBuilder {}
+
+impl TeamConfigBuilder {
+    /// Creates a new builder for [`TeamConfig`]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Creates a [`TeamConfig`] using the provided parameters
+    pub fn build(self) -> Result<TeamConfig, ConfigError> {
```
elagergren-spideroak, line 14: unless we flesh out `ConfigError` before MVP, we should probably just make this return `Error`, since it's a superset of `ConfigError`.

crates/aranya-client/src/error.rs
nikki-aranya: I knew I was forgetting something

crates/aranya-client/src/client.rs
nikki-aranya: That was a leftover from Declan's attempt at this but you're right. I personally prefer the latter tbh, just need to enforce it

gknopf-aranya (dismissed):

djohnson-spideroak (commented):
LGTM, I cant approve because I opened the PR

nikki-aranya (approved):

gknopf-aranya:
@elagergren-spideroak waiting for your approval to confirm feedback was incorporated correctly

gknopf-aranya:
@elagergren-spideroak, @jdygert-spok this is ready for review

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok, @elagergren-spideroak this PR is ready for a final review

gknopf-aranya (dismissed):

crates/aranya-client/src/config.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn default() -> Self {
+        Self {
+            interval: None,
+            sync_now: true,
+        }
+    }
+}
+
+/// Configuration info for adding and creating teams.
+pub struct TeamConfig {}
```
jdygert-spok, line 84: add an empty field `_priv: ()` so it is not publicly constructable.

crates/aranya-client/src/config.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl From<TeamConfig> for aranya_daemon_api::TeamConfig {
+    fn from(_value: TeamConfig) -> Self {
+        Self {}
+    }
+}
+
+/// Builder for a [`TeamConfig`]
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct TeamConfigBuilder {}
```
jdygert-spok, line 101: ditto private field

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -68,6 +68,9 @@ pub enum Error {
 
     #[capi(msg = "invalid index")]
     InvalidIndex,
+
+    #[capi(msg = "invalid config")]
+    Config,
```
jdygert-spok, line 73: Needs doc comment. Please add to the above two as well while we're here.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -85,13 +88,16 @@ impl From<&imp::Error> for Error {
                 aranya_client::Error::Connecting(_) => Self::Connecting,
                 aranya_client::Error::Rpc(_) => Self::Rpc,
                 aranya_client::Error::Daemon(_) => Self::Daemon,
+                aranya_client::Error::Config(_) => Self::Config,
                 #[cfg(feature = "afc")]
                 aranya_client::Error::Afc(_) => Self::Afc,
                 aranya_client::Error::Bug(_) => Self::Bug,
                 aranya_client::Error::InvalidArg { .. } => Self::InvalidArgument,
+                _ => todo!(),
```
jdygert-spok, line 96: debug panic and/or log, and return `Self::Bug` ?

crates/aranya-client-capi/src/imp/config.rs
jdygert-spok, line 1: Is there a reason we don't wrap the rust config and builder, for all of these?

crates/aranya-daemon-api/src/service.rs
```diff
@@ -73,6 +73,12 @@ pub enum Role {
     Member,
 }
 
+/// A configuration for creating or adding a team to a daemon.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TeamConfig {
+    // TODO(nikki): any fields added to this should be public
```
jdygert-spok, line 79: I'm not really sure how we will want to handle this. Adding public fields will be a breaking change. This isn't _really_ a public crate, but we're currently keeping the versions in lockstep, so we can't properly signify breakage here when doing a non-breaking update to aranya-client.

elagergren-spideroak: > so we can't properly signify breakage here when doing a non-breaking update to aranya-client.

I'm not exactly sure what you mean by this.

jdygert-spok: Scenario: We add an optional field to the config. This is a public field in this config, so aranya-daemon-api needs to bump to 0.7.0. The builder and default value means we don't have a breaking change in aranya-client so we can just update to 0.6.1. These versions are currently updated in lockstep so we can't do this.

Maybe we'll stop updating the support crates in lockstep?

elagergren-spideroak: Oh. Yeah, I don't see why `aranya-client` and `aranya-daemon-api` need to have the same version. Only our "public" crates do.

crates/aranya-client-capi/src/imp/config.rs
nikki-aranya: I dunno, just didn't really make sense, it'll be roughly the same amount of lines either way but I guess we could

nikki-aranya: Also the Rust-style builder pattern doesn't work in CAPI land, we're unable to just take `mut self`, so we have to work around that

examples/rust/src/main.rs
```diff
@@ -164,7 +164,7 @@ async fn main() -> Result<()> {
     let team_id = team
         .owner
         .client
-        .create_team()
+        .create_team(TeamConfigBuilder::new().build()?)
```
gknopf-aranya, line 167: @nikki-aranya, @jdygert-spok there were some hoops we needed to jump through to get the C API builder pattern working, but we should be using an idiomatic builder pattern in Rust if possible. E.g.:
```suggestion
        .create_team(TeamConfigBuilder::new()?)
```

nikki-aranya: if/when that actually has parameters, it'll be split out into a let block

gknopf-aranya (dismissed):

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -579,27 +585,56 @@ pub fn get_device_id(client: &mut Client) -> Result<DeviceId, imp::Error> {
...
+/// @param cfg the Team Configuration [`TeamConfig`].
 /// @param __output the team's ID [`TeamId`].
 ///
 /// @relates AranyaClient.
-pub fn create_team(client: &mut Client) -> Result<TeamId, imp::Error> {
+#[allow(unused_variables)] // TODO(nikki): once we have fields on TeamConfig, remove this for cfg
+pub fn create_team(client: &mut Client, cfg: &TeamConfig) -> Result<TeamId, imp::Error> {
     let client = client.deref_mut();
-    let id = client.rt.block_on(client.inner.create_team())?;
+    let cfg = aranya_client::TeamConfig::builder().build()?;
```
djohnson-spideroak, line 617: why are we taking the cfg as an argument, and building a blank one here?

gknopf-aranya:
@djohnson-spideroak, @nikki-aranya I haven't seen much activity on this PR recently. Would be nice to get this across the finish line soon.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -732,27 +738,56 @@ pub fn get_device_id(client: &mut Client) -> Result<DeviceId, imp::Error> {
     Ok(id.into())
 }
 
+#[aranya_capi_core::opaque(size = 24, align = 8)]
+pub type TeamConfig = Safe<imp::TeamConfig>;
+
+#[aranya_capi_core::opaque(size = 4, align = 4)]
+pub type TeamConfigBuilder = imp::TeamConfigBuilder;
```
ssweetney-spok, line 745: Needs to be wrapped with `Safe`. I know there's no fields to be set on this type yet but we may as well handle it now since #188 is complete

nikki-aranya: Why does it need to be wrapped with Safe? If it's concerns about public fields, opaque should already be taking care of that, it doesn't *need* Safe.

gknopf-aranya:
@nikki-aranya, @djohnson-spideroak let's prioritize getting this team configs PR updated and landed today/tomorrow

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: Yea, I don't think there's a technical reason for it so I should've said "we want".

I'm assuming we want to keep things consistent among the builders in the C API.

That issue was created because of the convo in this [thread](https://github.com/aranya-project/aranya/pull/145#discussion_r2042813030)

ssweetney-spok: From what I can tell from reading the code, `Safe` is about "guarding" against reading/writing potentially uninitialized data.

cc: @jdygert-spok @elagergren-spideroak

jdygert-spok: Yes, it protects against reading uninitialized data and double frees. Not currently needed but we will want it once we put any complex types in the config/builder, so might as well add it now.

gknopf-aranya: This supports long-term API stability which is a goal for MVP.

gknopf-aranya (approved):

gknopf-aranya (approved):

jdygert-spok (approved):

PR #114, opened 2025-03-11
`aranya-client` API refactoring

nikki-aranya:
This is focused on refactoring our public API for aranya-client.


nikki-aranya:
Note to self: refactor to create a new FastChannel wrapper like we do Team and hide the current FastChannel as a `FastChannelImpl` so we don't leak a way to swap channels between multiple Client instances

elagergren-spideroak (changes requested):

nikki-aranya:
Alright, here's an overview of everything that's changed:

- We now have an afc() function on Client that returns a FastChannels client "wrapper" which exposes the same functions as we originally did on Client.
- We now publicly share afc.rs as a module for better organization/documentation (need a better module doc in a future PR).
- I had to re-format some comments (shoutout Rewrap Revived VSCode plugin and Alt+Q) to better match our line length.
- I also had to recreate all our use blocks in afc/client/tests due to tests failing from weird serialization issues and shortened paths on a handful of things, as well as enforcing `use Trait as _` for any traits we don't actively use in the file and are only there to provide functionality. 
- I moved around the error cases to show up in a better order on https://docs.rs/
- I rearranged the files to a better logical order of types (e.g. StreamMsg below all the wire format stuff that precedes it in our packet format)
- `channel` and `address` shouldn't be shortened imo, especially on user-facing stuff like `Message`
- I've structured this so we can expose `FastChannel` in a future PR (see add_channel/remove_channel).
- A couple clippy fixes (io::Error::other).
- I had to change do_poll due to new lifetime annoyances. It works, it just could probably be cleaner if we use an alternative to tokio::select!{}

nikki-aranya:
Here's what the APIs look like:
root:
image
afc:
image
client:
image
error:
image

All rustdoc stuff subject to change, and AfcError/afc will eventually be gated behind a feature flag/experimental

crates/aranya-client/src/error.rs
```diff
@@ -1,25 +1,146 @@
-/// Errors that could occur in the Aranya client.
+use aranya_daemon_api::AfcId;
+use aranya_fast_channels::Version;
```
gknopf-aranya, line 2: This would be more clear:
```suggestion
use aranya_fast_channels::AfcVersion;
```

crates/aranya-client/src/lib.rs
```diff
@@ -11,12 +11,13 @@
 //! [README]: https://github.com/aranya-project/aranya/tree/main/crates/aranya-client/README.md
 //! [walkthrough]: https://github.com/aranya-project/aranya/tree/main/docs/walkthrough.md
 
-mod afc;
-mod client;
-mod error;
+pub mod afc;
```
gknopf-aranya, line 14: Is this `pub` because specific types are used by the application?
If so, we could reexport just those types instead of making the entire afc module pub.

crates/aranya-client/tests/tests.rs
```diff
@@ -264,11 +268,11 @@ impl UserCtx {
     }
 
     async fn aranya_local_addr(&self) -> Result<SocketAddr> {
-        Ok(self.client.aranya_local_addr().await?)
+        Ok(self.client.local_addr().await?)
```
gknopf-aranya, line 271: Since we're going to use `client.afc` and `client.aqc` to differentiate AFC from AQC, we might as well use `client.sync` to separate out the sync related methods.
```suggestion
        Ok(self.client.sync().local_addr().await?)
```

gknopf-aranya:
This PR is changing addr -> address for some APIs while keeping `local_addr()`.
Seems like we should pick one and stay consistent.

crates/aranya-client/src/error.rs
nikki-aranya: Version is only use'd because otherwise it would make VersionMismatch take up 4 lines formatted

crates/aranya-client/src/lib.rs
nikki-aranya: afc is public so we don't have to append "Afc" to all types like you suggest above, as well as giving another "area" for rustdoc documentation about AFC-specific features, it gives the option of afc::Message. I'm specifically trying to avoid the pitfalls we've already hit

nikki-aranya:
> This PR is changing addr -> address for some APIs while keeping `local_addr()`. Seems like we should pick one and stay consistent.

I only changed "address" specifically, but you're right that I should probably change local_addr too

nikki-aranya:
Settled on "address if it's public, otherwise it's fine to abbreviate"

jdygert-spok:
I was curious, `std` and `tokio` entirely use `addr`. `quinn` and `s2n-quic` have a `mix` of `addr` and `address` in the api.

crates/aranya-client/src/lib.rs
elagergren-spideroak: I prefer using `pub mod {afc,aqc}`. In addition to what @nikki-aranya said, re-exporting all of our submodules makes the rustdocs a Big Wall of Text.

nikki-aranya:
Alright, alright, I changed back to chan internally. Should be good again.

crates/aranya-client/tests/tests.rs
```diff
@@ -613,13 +621,14 @@ async fn test_afc_one_way_two_chans() -> Result<()> {
...
+        .afc()
+        .try_recv_data()
         .expect("should have a message");
-    let want = AfcMsg {
+    let want = Message {
         data: msgs[0].as_bytes().to_vec(),
         // We don't know the address of outgoing connections, so
         // assume `got.addr` is correct here.
-        addr: got.addr,
+        address: got.address,
```
gknopf-aranya, line 631: Switch back to using `addr` instead of `address` for public APIs and internal code:
```suggestion
        addr: got.addr,
```

nikki-aranya: This is part of Message which is a public type as we talked about

gknopf-aranya (dismissed):

gknopf-aranya:
lgtm!

crates/aranya-client/src/client.rs
```diff
@@ -1,292 +1,83 @@
...
+    ///   on.
+    #[instrument(skip_all, fields(?daemon_socket, ?afc_shm_path, max_channels))]
     pub async fn connect<A>(
-        daemon_sock: &Path,
+        daemon_socket: &Path,
         afc_shm_path: &Path,
-        max_chans: usize,
-        afc_listen_addr: A,
+        max_channels: usize,
+        afc_address: A,
```
djohnson-spideroak, line 53: very nit: are we using address or addr here? I am fine with either, just want to make sure this is consistent.

crates/aranya-client/src/afc.rs
```diff
@@ -341,37 +259,34 @@ impl<S: AfcState> Afc<S> {
         stream.flush().await.map_err(AfcError::StreamWrite)?;
         debug!("sent control message");
 
-        // TODO(eric): This throws away `stream` if we already
-        // have a stream with this address.
+        // TODO(eric): This throws away `stream` if we already have a stream
```
djohnson-spideroak, line 262: lets make sure this is covered by an issue.

djohnson-spideroak (approved):
LGTM, I think we should require a review from Eric and/or Dygert before merging.

crates/aranya-client/src/client.rs
nikki-aranya: address, Client::connect is public

nikki-aranya:
> LGTM, I think we should require a review from Eric and/or Dygert before merging.

I'm blocked until @elagergren-spideroak reviews it anyways because they requested changes

crates/aranya-client/src/afc.rs
```diff
@@ -6,241 +6,155 @@
...
     streams: TcpStreams,
     /// All open channels.
-    chans: BTreeMap<AfcId, Chan>,
+    // TODO: expose FastChannel to the user and add more methods to it
+    chans: BTreeMap<AfcId, FastChannel>,
     /// Incrementing counter for unique [`NodeId`]s.
     // TODO: move this counter into the daemon.
     next_node_id: u32,
+    /// Messages from `handle_data`.
+    msgs: VecDeque<Message>,
```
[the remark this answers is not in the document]
nikki-aranya: dynamic currently, afaik it's the same allocation behavior as a normal Vec. We could switch to https://docs.rs/heapless/latest/heapless/struct.Deque.html but right now I'm just copying the old behavior

crates/aranya-client/src/afc.rs
nikki-aranya: This should be addressed once we expose some form of add_channel and remove_channel to allow FastChannel::read/write directly

nikki-aranya: https://github.com/aranya-project/aranya/issues/134

crates/aranya-client/src/afc.rs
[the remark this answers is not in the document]
djohnson-spideroak: Yeah I dont see a reason to change it this moment. AFC is normally where we put special consideration on performance and memory, so I thought it was good to bring up.

crates/aranya-client/src/afc.rs
```diff
@@ -6,241 +6,155 @@
...
 //! ```
 //!
-//! - `magic` is a 32-bit little-endian integer with the magic
-//!   value `"AFC\0"`.
-//! - `len` is a 32-bit little endian integer that contains the
-//!   size in bytes of `msg`.
-//! - `msg`: A postcard-encoded [`Msg`].
+//! - `magic` is a 32-bit little-endian integer with the magic value `"AFC\0"`.
+//! - `len` is a 32-bit little endian integer that contains the size in bytes of
+//!   `msg`.
```
elagergren-spideroak, line 11: nit: reformatting these two lines is just diff noise

crates/aranya-client/src/afc.rs
```diff
@@ -6,241 +6,155 @@
...
     /// The underlying AFC client.
     afc: Client<S>,
     /// Listens for incoming connections from peers.
     listener: TcpListener,
     /// Open TCP connections.
     // TODO(eric): prune unused/idle streams.
-    // TODO(eric): use different maps for streams we opened vs
-    // streams that peers opened.
+    // TODO(eric): use different maps for streams we opened vs streams that
+    // peers opened.
```
elagergren-spideroak, line 128: diff noise

crates/aranya-client/src/afc.rs
```diff
@@ -6,241 +6,155 @@
...
     streams: TcpStreams,
     /// All open channels.
-    chans: BTreeMap<AfcId, Chan>,
+    // TODO: expose FastChannel to the user and add more methods to it
```
elagergren-spideroak, line 131: nit: conventionally we use `TODO(name)`. It helps b/c it's much quicker than stepping through git history to see who let the TODO (like when comments get moved around, etc.).

crates/aranya-client/tests/tests.rs
```diff
@@ -132,19 +133,22 @@ fn trim(mut d: u128, mut width: usize) -> (u128, usize) {
...
             "start `do_poll`",
         );
         loop {
+            let mut afcs = [ $($client.afc()),* ];
+            let mut afcs = afcs.iter_mut();
             tokio::select! {
                 biased;
-                $(data = $client.poll_afc_data() => {
-                    $client.handle_afc_data(data?).await?
+                $(data = afcs.next().unwrap().poll_data() => {
```
elagergren-spideroak, line 150: Please add a comment about why you're using `unwrap` and why it doesn't panic. Or, better, use `next().assume("...")?` from `buggy::BugExt`.

templates/aranya-example/src/main.rs
```diff
@@ -124,19 +124,22 @@ impl UserCtx {
...
             "start `do_poll`",
         );
         loop {
+            let mut afcs = [ $($client.afc()),* ];
+            let mut afcs = afcs.iter_mut();
             tokio::select! {
                 biased;
-                $(data = $client.poll_afc_data() => {
-                    $client.handle_afc_data(data?).await?
+                $(data = afcs.next().unwrap().poll_data() => {
```
elagergren-spideroak, line 141: ditto about unwrap

crates/aranya-client/src/afc.rs
```diff
@@ -6,241 +6,155 @@
...
 use std::{
-    collections::btree_map::{self, BTreeMap},
+    collections::{btree_map, BTreeMap, VecDeque},
     ffi::c_int,
     fmt,
     future::Future,
     io::{self, IoSlice},
     net::SocketAddr,
-    os::fd::AsRawFd,
+    os::fd::AsRawFd as _,
```
elagergren-spideroak, line 21: is this triggering a lint or something? if possible, i'd prefer to leave it as-is to minimize the diff noise (so we can see the important changes).

crates/aranya-client/src/afc.rs
```diff
@@ -341,37 +259,34 @@ impl<S: AfcState> Afc<S> {
         stream.flush().await.map_err(AfcError::StreamWrite)?;
         debug!("sent control message");
 
-        // TODO(eric): This throws away `stream` if we already
-        // have a stream with this address.
+        // TODO(eric): This throws away `stream` if we already have a stream
+        // with this address.
```
elagergren-spideroak, line 263: nit: diff noise

elagergren-spideroak:
> > LGTM, I think we should require a review from Eric and/or Dygert before merging.
> 
> I'm blocked until @elagergren-spideroak reviews it anyways because they requested changes

It lgtm other than my comments about `unwrap`.

I would like it if we could minimize some of the diff noise though—like reformatting the comments. It obscures the important changes in the PR and 80 vs 120 chars for a comment is ~irrelevant.

nikki-aranya:
The only time you're going to care about diff noise is in the review process which we're already mostly through and potentially in git blame. We should instead just have a discussion around enabling the rustfmt rule that reflows comments and calling it a day, I just had to do it manually here.

crates/aranya-client/src/client.rs
```diff
@@ -1,292 +1,83 @@
...
-    fn debug(&self) -> tracing::field::Empty {
-        tracing::field::Empty
+        Ok(Self { daemon, afc })
     }
 
-    /// Returns the address that the Aranya sync server is bound
-    /// to.
-    pub async fn aranya_local_addr(&self) -> Result<SocketAddr> {
+    /// Returns the address that the Aranya sync server is bound to.
+    pub async fn local_addr(&self) -> Result<SocketAddr> {
```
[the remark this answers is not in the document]
nikki-aranya: This will be addressed in #133

crates/aranya-client/src/afc.rs
nikki-aranya: As mentioned in my list above, I'm introducing this change specifically to not "pollute" our namespace, however much credit you want to give that, it's good practice

nikki-aranya:
> git blame, yeah, or when looking through the git history, which we try to keep relatively orderly. I also don't really see the point in reflowing comments that are unrelated to the PR (i.e., that aren't being rewritten for other reasons). Especially when they're already <= 120 chars and we don't have a minimum comment length. But like I said, it's a nit, not a blocker ¯_(ツ)_/¯

Normally I'd agree with you, in my mind this and the rustdoc PR that's gonna follow this are one big change, which is why I'm letting a lot of this slide since a lot of it's going to get immediately rewritten anyways.

djohnson-spideroak (approved):

gknopf-aranya:
If formatting changes, I'd update it on whatever PR the `rustfmt` config is changed on.

gknopf-aranya (dismissed):

nikki-aranya:
> > Normally I'd agree with you; in my mind this and the rustdoc PR that's gonna follow this are one big change, which is why I'm letting a lot of this slide since a lot of it's going to get immediately rewritten anyways.
> 
> If the comments are going to be immediately rewritten, why change any of it here? Were those mangled through some automation you didn't have good control over?

Because I didn't think y'all would take such issue. As mentioned above, I manually reflowed them as I dug through the file making my changes using the Rewrap Revived plugin and it's a lot harder to un-flow once it's changed but it's fine.

gknopf-aranya (approved):

djohnson-spideroak (approved):

gknopf-aranya:
Would like to get a final approval from @elagergren-spideroak to verify feedback was incorporated

gknopf-aranya (approved):

Issue #79, opened 2025-02-24
afc: make AFC experimental


djohnson-spideroak:
This issue is blocked on the implementation of AQC. We decided that AFC should be kept behind an "experimental" flag. I believe the original intent was to supply this flag at compile time.

This issue will also likely result in follow-on issues to update the docs.

gknopf-aranya:
@djohnson-spideroak we can complete adding the afc prefixes before making the AFC API experimental.
Adding the API prefixes is a simpler change and will take afc out of the namespace that acq needs:
https://github.com/aranya-project/aranya/issues/83

gknopf-aranya:
Discussion related to feature flags for AFC and AQC:
https://github.com/aranya-project/aranya/pull/145#discussion_r2011079306

PR #124, opened 2025-03-12
mvp: factdb queries

gknopf-aranya:
Implement factdb queries by having ephemeral commands query the factdb and return a result via an effect.
Add the factdb query methods from the MVP spec to the Rust and C APIs.

MVP spec containing factdb queries:
https://github.com/aranya-project/aranya-docs/blob/main/docs/aranya-mvp.md

PR for updating the factdb queries in the MVP spec:
https://github.com/aranya-project/aranya-docs/pull/81

Rust API:
- [x] `QueryDevicesOnTeam(team_id) -> Vec<device_id>`
- [x] `QueryDeviceRole(team_id, device_id) -> Role`
- [x] `QueryDeviceKeybundle(team_id, device_id) -> Keybundle`
- [x] `QueryAfcNetIdentifier(team_id, device_id) -> network_str`
- [x] `QueryAqcNetIdentifier(team_id, device_id) -> network_str`
- [x] `QueryLabelExists(team_id, label) -> bool`
- [x] `QueryDeviceLabelAssignments(team_id, device_id) -> Vec<Label>` 

C API (future work https://github.com/aranya-project/aranya/issues/131):
- [x] `QueryDevicesOnTeam(team_id) -> Vec<device_id>`
- [ ] `QueryDeviceRole(team_id, device_id) -> Role` (blocked by https://github.com/aranya-project/aranya-core/issues/160)
- [x] `QueryDeviceKeybundle(team_id, device_id) -> Keybundle`
- [x] `QueryAfcNetIdentifier(team_id, device_id) -> network_str`
- [x] `QueryAqcNetIdentifier(team_id, device_id) -> network_str`
- [x] `QueryLabelExists(team_id, label) -> bool`
- [x] `QueryDeviceLabelAssignments(team_id, device_id) -> Vec<Label>`

The C API portion of factdb queries will be implemented in future PRs since it is blocked by capi generator features:
https://github.com/aranya-project/aranya-core/issues/160


crates/aranya-daemon/src/policy.md
```diff
@@ -1499,6 +1499,38 @@ command CreateUniChannel {
...
+command QueryDevices {
+    fields {}
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        let author = get_valid_user(envelope::author_id(envelope))
+
+        // TODO: needs vec support.
```
gknopf-aranya, line 1524: @ycarmy, @djohnson-spideroak does the policy language have a way to return a vector of device IDs?

crates/aranya-daemon/src/policy.md
```diff
@@ -1499,6 +1499,91 @@ command CreateUniChannel {
...
+
+```policy
+
+// Returns the user's key bundle.
+function get_user_keybundle(user_id id) struct KeyBundle {
+    let ident_key = check_unwrap query UserIdentKey[user_id: user_id]
+    let sign_key = check_unwrap query UserSignKey[user_id: user_id]
+    let enc_key = check_unwrap query UserEncKey[user_id: user_id]
+
+    return KeyBundle {
```
gknopf-aranya, line 1548: @ycarmy do you know why the queried keys are not `bytes`? As far as I can tell they were bytes when they were added to the factdb.
```
Caused by:
    invalid type: `Struct KeyBundle` field `ident_key` is not bytes at line 1548 col 5:
      let key_bundle = KeyBundle {
            ident_key: ident_key,
            sign_key: sign_key,
            enc_key: enc_key,
        }
```

crates/aranya-daemon/src/api.rs
```diff
@@ -570,6 +572,88 @@ impl DaemonApi for DaemonApiHandler {
...
+        device: DeviceId,
+    ) -> ApiResult<ApiKeyBundle> {
+        let mut session = self.client.session_new(&team.into_id().into()).await?;
+        let (ctrl, _effects) = self
+            .client
+            .actions(&team.into_id().into())
+            .query_device_keybundle_off_graph(device.into_id().into())
+            .await?;
+        for cmd in ctrl {
+            let effects = self.client.session_receive(&mut session, &cmd).await?;
```
ycarmy, line 616: Why are you re-processing the command through session_receive? I think you should be able to use the effects returned from your call to the action.

crates/aranya-daemon/src/policy.md
ycarmy: Unwrapping a fact query returns a struct with the fact fields, including its key fields. You need to access the value field for the fact that holds the public key bytes:
```
let key_bundle = KeyBundle {
  ident_key: ident_key.key,
  sign_key: sign_key.key,
  enc_key: enc_key.key,
}
```

gknopf-aranya:
Because there are some limitations to the type conversions in the capi for return values, this PR is going to focus on completing the Rust API first, then implement the C API once the changes are supported by the C API generation tool.

@jdygert-spok is there an issue for the C API fixes to support returning a device Role?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -922,3 +933,23 @@ pub unsafe fn recv_data(
...
+    client: &mut Client,
+    team: &TeamId,
+    device_id: &DeviceId,
+) -> Result<Role, imp::Error> {
+    let client = client.deref_mut();
+
+    let role = client
+        .rt
+        .block_on(client.inner.team(team.0).query_device_role(device_id.0))?;
+    Ok(role.into())
```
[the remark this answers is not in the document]
gknopf-aranya: @jdygert-spok has an issue been created for this?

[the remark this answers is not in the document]
jdygert-spok: https://github.com/aranya-project/aranya-core/issues/160

crates/aranya-daemon/src/api.rs
gknopf-aranya: No reason. I copy-pasted this code from what was used to process ctrl messages for AFC because I knew queries would use ephemeral sessions.
It works without re-processing with `session_receive`.

gknopf-aranya:
Setting work on this aside until the next sprint. Completing this PR is blocked by:
https://github.com/aranya-project/aranya-core/issues/160

@jdygert-spok will the capi generator be able to handle `Vec` return types?

gknopf-aranya:
This PR has been tested by running the Rust template/example.
Wouldn't hurt to add some unit tests but that could be done on another PR.

crates/aranya-daemon/src/api.rs
```diff
@@ -567,6 +608,153 @@ impl DaemonApi for DaemonApiHandler {
...
+    ) -> ApiResult<ApiKeyBundle> {
+        let (_ctrl, effects) = self
+            .client
+            .actions(&team.into_id().into())
+            .query_device_keybundle_off_graph(device.into_id().into())
+            .await?;
+        if let Some(Effect::QueryDeviceKeyBundleResult(e)) =
+            find_effect!(&effects, Effect::QueryDeviceKeyBundleResult(_e))
+        {
+            return Ok(ApiKeyBundle::from(e.user_keys.clone()));
```
ssweetney-spok, line 667: Nit: The key bundle doesn't need to be cloned

```suggestion
            find_effect!(effects, Effect::QueryDeviceKeyBundleResult(_e))
        {
            return Ok(ApiKeyBundle::from(e.user_keys));
```

crates/aranya-daemon/src/api.rs
```diff
@@ -567,6 +608,153 @@ impl DaemonApi for DaemonApiHandler {
...
+    ) -> ApiResult<NetIdentifier> {
+        let (_ctrl, effects) = self
+            .client
+            .actions(&team.into_id().into())
+            .query_afc_net_identifier_off_graph(device.into_id().into())
+            .await?;
+        if let Some(Effect::QueryAfcNetIdentifierResult(e)) =
+            find_effect!(&effects, Effect::QueryAfcNetIdentifierResult(_e))
+        {
+            return Ok(NetIdentifier(e.net_identifier.clone()));
```
ssweetney-spok, line 714:
```suggestion
        if let Some(Effect::QueryAfcNetIdentifierResult(e)) =
            find_effect!(effects, Effect::QueryAfcNetIdentifierResult(_e))
        {
            return Ok(NetIdentifier(e.net_identifier));
```

crates/aranya-daemon/src/api.rs
```diff
@@ -567,6 +608,153 @@ impl DaemonApi for DaemonApiHandler {
...
+    ) -> ApiResult<NetIdentifier> {
+        let (_ctrl, effects) = self
+            .client
+            .actions(&team.into_id().into())
+            .query_aqc_net_identifier_off_graph(device.into_id().into())
+            .await?;
+        if let Some(Effect::QueryAqcNetIdentifierResult(e)) =
+            find_effect!(&effects, Effect::QueryAqcNetIdentifierResult(_e))
+        {
+            return Ok(NetIdentifier(e.net_identifier.clone()));
```
ssweetney-spok, line 734: Can also remove the clone call here

ssweetney-spok (commented):
Looks good, just a few small comments

crates/aranya-daemon/src/api.rs
gknopf-aranya: If I remove `clone()`:
```
error[E0507]: cannot move out of `e.user_keys` which is behind a shared reference
   --> crates/aranya-daemon/src/api.rs:667:42
    |
667 |             return Ok(ApiKeyBundle::from(e.user_keys));
    |                                          ^^^^^^^^^^^ move occurs because `e.user_keys` has type `policy::KeyBundle`, which does not implement the `Copy` trait
```

crates/aranya-daemon/src/api.rs
gknopf-aranya:
```
error[E0507]: cannot move out of `e.net_identifier` which is behind a shared reference
   --> crates/aranya-daemon/src/api.rs:714:37
    |
714 |             return Ok(NetIdentifier(e.net_identifier));
    |                                     ^^^^^^^^^^^^^^^^ move occurs because `e.net_identifier` has type `std::string::String`, which does not implement the `Copy` trait

```

crates/aranya-daemon/src/api.rs
gknopf-aranya: Under the hood, the net identifier is a string which does need to be cloned.
```
error[E0507]: cannot move out of `e.net_identifier` which is behind a shared reference
   --> crates/aranya-daemon/src/api.rs:714:37
    |
714 |             return Ok(NetIdentifier(e.net_identifier));
    |                                     ^^^^^^^^^^^^^^^^ move occurs because `e.net_identifier` has type `std::string::String`, which does not implement the `Copy` trait

```

crates/aranya-daemon/src/api.rs
ssweetney-spok: You can pass the effects variable by value rather than using a reference. It doesn't look like it's used after line 665

crates/aranya-daemon/src/api.rs
ssweetney-spok: Pass effects by value instead

ssweetney-spok:
> > Looks good, just a few small comments
> 
> I tried these suggestions but the compiler said `clone` was needed.

The clone is needed if the effect is borrowed.
The effect returned by `find_effect!` will borrowed if you pass it a slice instead of a Vec of owned effects

crates/aranya-client/src/client.rs
```diff
@@ -224,6 +224,36 @@ impl Team<'_> {
...
+        device: DeviceId,
+        net_identifier: NetIdentifier,
+    ) -> Result<()> {
+        Ok(self
+            .client
+            .daemon
+            .remove_aqc_net_identifier(context::current(), self.id, device, net_identifier)
+            .await??)
+    }
+
```
djohnson-spideroak, line 256: are we saving these in memory like AFC? or is it only the policy API?

crates/aranya-daemon/src/api.rs
```diff
@@ -435,7 +444,39 @@ impl DaemonApi for DaemonApiHandler {
...
+            .client
+            .actions(&team.into_id().into())
+            .set_aqc_network_name(device.into_id().into(), name.0)
+            .await?;
+        self.handle_effects(&effects, None).await?;
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    async fn remove_aqc_net_identifier(
```
djohnson-spideroak, line 470: does this match the verbiage of AFC net ident operations?

crates/aranya-daemon/src/api.rs
```diff
@@ -567,6 +608,153 @@ impl DaemonApi for DaemonApiHandler {
...
+            .client
+            .actions(&team.into_id().into())
+            .query_afc_net_identifier_off_graph(device.into_id().into())
+            .await?;
+        if let Some(Effect::QueryAfcNetIdentifierResult(e)) =
+            find_effect!(effects, Effect::QueryAfcNetIdentifierResult(_e))
+        {
+            return Ok(NetIdentifier(e.net_identifier));
+        };
+        Err(anyhow!("unable to query afc network identifier").into())
```
djohnson-spideroak, line 716: nit, could do this without the return by having the if let evaluate to Ok/Err

crates/aranya-daemon/src/policy.md
```diff
@@ -127,10 +127,13 @@ fact TeamEnd[]=>{}
 fact Label[label int]=>{}
 
 // Records that a user is allowed to use an AFC label.
-fact AssignedLabel[label int, user_id id]=>{op enum ChanOp}
+fact AssignedLabel[user_id id, label int]=>{op enum ChanOp}
```
djohnson-spideroak, line 130: why is this being reversed? also, will conflict with the device rename PR.

crates/aranya-daemon/src/policy.md
```diff
@@ -1499,6 +1612,287 @@ command AfcCreateUniChannel {
...
+
+    policy {
+        finish {
+            emit QueryDeviceLabelAssignmentsResult {
+                label: this.label
+            }
+        }
+    }
+}
+```
```
djohnson-spideroak, line 1762: should we restrict the users that can query labels?

crates/aranya-client/src/client.rs
gknopf-aranya: Only in the policy. I wasn't originally planning to add these APIs on this PR, but they are needed in order to test the query. The AQC implementation need to handle storing the AQC net identifiers in memory if needed.

crates/aranya-daemon/src/policy.md
gknopf-aranya: The keys are reversed to support the `map` operation for the assigned labels query. This was a suggestion from @jdygert-spok.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Possibly? That is not in the MVP spec.

jdygert-spok: Worth noting it'd be somewhat bypassable since it is just a local operation.

crates/aranya-daemon/src/api.rs
gknopf-aranya: I copied the AFC net identifier operations and changed afc -> aqc in the function names, types, docs, etc.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Here is the original discussion:
https://github.com/aranya-project/aranya-docs/pull/81#discussion_r1992315584

crates/aranya-daemon/src/api.rs
gknopf-aranya: Switching over to if let else

crates/aranya-daemon/src/policy.md
gknopf-aranya: Restricting who can query labels from the factdb is too late in the process to place that restriction since the information already exists locally in the factdb. If we are concerned about who has read access to that information, they shouldn't be allowed to decrypt the command and update their factdb with the information in the first place.

djohnson-spideroak (dismissed):

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1226,6 +1304,134 @@ AranyaError aranya_afc_recv_data_ext(struct AranyaClient *client,
...
+ * @param device the device's ID [`AranyaDeviceId`](@ref AranyaDeviceId).
+ * @param network identifier string [`AranyaNetIdentifier`](@ref AranyaNetIdentifier).
+ *
+ * @relates AranyaClient.
+ */
+AranyaError aranya_query_afc_net_identifier(struct AranyaClient *client,
+                                            const struct AranyaTeamId *team,
+                                            const struct AranyaDeviceId *device,
+                                            char *ident,
+                                            size_t ident_len);
```
djohnson-spideroak, line 1352: What do you think about a function for querying all AFC/AQC net identifiers?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -928,3 +988,95 @@ pub unsafe fn afc_recv_data(
...
+    client: &mut Client,
+    team: &TeamId,
+    device: &DeviceId,
+    ident: &mut [MaybeUninit<c_char>],
+) -> Result<(), imp::Error> {
+    let client = client.deref_mut();
+    let net_identifier = client
+        .rt
+        .block_on(client.inner.team(team.0).query_afc_net_identifier(device.0))?;
+    let mut ident_len = net_identifier.0.len();
```
elagergren-spideroak, line 1036: This is throwing away then length instead of returning it to the caller. You want this:

```rust
pub unsafe fn query_afc_net_identifier(..., ident: &mut MaybeUninit<c_char>], ident_len: &mut usize) -> Result<(), imp::Error> {
    [...]
    let net_identifier = ...;
    let ident = aranya_capi_core::try_as_mut_slice!(ident, ident_len);
    aranya_capi_core::write_c_str(ident, &net_identifier, ident_len)?;
}
```

examples/c/example.c
```diff
@@ -300,17 +300,61 @@ AranyaError run(Team *t) {
...
+    printf("%s aqc net identifier: %s \r\n", t->clients_arr[MEMBERB].name,
+           memberb_aqc_net_identifier);
+    free(memberb_aqc_net_identifier);
+    bool exists = false;
+    err = aranya_query_label_exists(&t->clients.membera.client, &t->id, &label,
+                                    &exists);
+    EXPECT("error querying label exists", err);
+    printf("%s label exists: %s \r\n", t->clients_arr[MEMBERB].name,
+           exists ? "true" : "false");
+
```
djohnson-spideroak, line 357: I dont see a test for removal of idents

crates/aranya-client-capi/output/aranya-client.h
elagergren-spideroak: Like returning a `Vec<NetIdentifier>`? What's the use case?

djohnson-spideroak: Given that all the net identifiers live on the graph, it might be useful to find the identifiers of peers the current node is not currently in contact with. Clearly it isnt meant as peer discovery but I think it is worth considering. It would need to return the net identifier and device ID to be useful though.

examples/c/example.c
djohnson-spideroak: We also want a test where we try and query a nonexistent net identifier.

crates/aranya-client-capi/output/aranya-client.h
elagergren-spideroak: I _think_ policy isn't capable of performing this query right now.

jdygert-spok: If you mean returning many results, it's doable with an action producing many commands and thus many effects: https://github.com/aranya-project/aranya/pull/124/files#diff-50b0db5827a2e66fcb36996671bcdb2241f59d65b4bfe0ac96aa8f32840aec07R1619-R1645

elagergren-spideroak: 👀

gknopf-aranya: I don't think we should add that on this PR. There is enough here already. Combining those together could make hiding the AFC functionality behind a feature flag more difficult on https://github.com/aranya-project/aranya/issues/79

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Thank you, pushed a commit for this

examples/c/example.c
gknopf-aranya: Added tests for this. Required some policy fixes.

gknopf-aranya:
@djohnson-spideroak I've added a test that attempts to query the net ident after it's removed. It works but seems like the error could be improved:
```
2025-03-24T19:45:52.817127Z  INFO RPC{rpc.deadline=2025-03-24T19:46:02.816832416Z otel.kind="client" otel.name=DaemonApi.query_afc_net_identifier rpc.trace_id=00}: tarpc::client: SendRequest
2025-03-24T19:45:52.820329Z ERROR err=not authorized
2025-03-24T19:45:52.820500Z ERROR aranya_capi_core::internal::error: err=client error: Daemon reported error: not authorized
2025-03-24T19:45:52.820506Z DEBUG aranya_client_capi::api::defs: err=Client(Daemon(Error("not authorized")))
error querying memberb afc net identifier: daemon reported error
```

gknopf-aranya:
Added more ctx to the errors in the daemon API

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: I'd prefer to make any changes to the query API to the spec first, before implementing things differently on this PR.
There's an open PR here with a few MVP spec suggestions:
https://github.com/aranya-project/aranya-docs/pull/81

examples/c/example.c
```diff
@@ -26,6 +26,16 @@
...
 
+// Macro for printing AranyaError to stderr and returning the error.
+// Does nothing if error value is ARANYA_SUCCESS.
+#define EXPECT_ERR(M, E)                                                       \
+    do {                                                                       \
+        AranyaError error = (E);                                               \
+        if (err != ARANYA_ERROR_SUCCESS) {                                     \
+            fprintf(stderr, "%s: %s\r\n", (M), aranya_error_to_str(error));    \
+        }                                                                      \
+    } while (0)
```
djohnson-spideroak, line 37: shouldnt this terminate/fail if it is ARANYA_SUCCESS?

crates/aranya-client/src/client.rs
```diff
@@ -224,6 +224,36 @@ impl Team<'_> {
...
+    ///
+    /// If the address already exists for this device, it is replaced with the new address. Capable
+    /// of resolving addresses via DNS, required to be statically mapped to IPV4. For use with
+    /// OpenChannel and receiving messages. Can take either DNS name or IPV4.
+    pub async fn assign_aqc_net_identifier(
+        &mut self,
+        device: DeviceId,
+        net_identifier: NetIdentifier,
+    ) -> Result<()> {
+        Ok(self
```
elagergren-spideroak, line 237: nit: 

```rust
self.client.daemon.assign_aqc_net_identifier(...).await??;
Ok(())
```

elagergren-spideroak: or `.await?.map_err(Into::into)`

crates/aranya-client/src/client.rs
```diff
@@ -262,4 +292,67 @@ impl Team<'_> {
             .revoke_label(context::current(), self.id, device, label)
             .await??)
     }
+
+    /// Query devices on team.
```
elagergren-spideroak, line 296: The docs could be a bit less terse:

```rust
/// Returns the list of current team members.
```

or something like that.

crates/aranya-client/src/client.rs
```diff
@@ -262,4 +292,67 @@ impl Team<'_> {
...
+    pub async fn query_devices_on_team(&mut self) -> Result<Vec<DeviceId>> {
```
elagergren-spideroak, line 297: For future proofing purposes, we /may/ want to consider returning `Result<impl Iterator<Item = DeviceId>>`.

crates/aranya-client/src/client.rs
```diff
@@ -262,4 +292,67 @@ impl Team<'_> {
...
+    pub async fn query_aqc_net_identifier(&mut self, device: DeviceId) -> Result<NetIdentifier> {
+        Ok(self
+            .client
+            .daemon
+            .query_aqc_net_identifier(context::current(), self.id, device)
+            .await??)
+    }
+
+    /// Query label exists.
+    pub async fn query_label_exists(&mut self, label: Label) -> Result<bool> {
```
elagergren-spideroak, line 351: All of the queries are named `query_foo` or `query_bar`. They're a little unnatural. E.g.:

```rust
if client.query_label_exists(FOO).await? { ... }
```

We could do what we've done elsewhere and add

```rust
pub struct Queries<'a> {
    client: &'a mut Client,
}

impl Queries {
    pub async fn aqc_net_identifier(&mut self, device: DeviceId) -> Result<NetIdentifier> { ... }
    pub async fn label_exists(&mut self, label: Label) -> Result<bool> { ... }
```

crates/aranya-client/src/client.rs
```diff
@@ -262,4 +292,67 @@ impl Team<'_> {
...
+    pub async fn query_afc_net_identifier(&mut self, device: DeviceId) -> Result<NetIdentifier> {
+        Ok(self
+            .client
+            .daemon
+            .query_afc_net_identifier(context::current(), self.id, device)
+            .await??)
+    }
+
+    /// Query AQC network ID.
+    pub async fn query_aqc_net_identifier(&mut self, device: DeviceId) -> Result<NetIdentifier> {
```
elagergren-spideroak, line 342: Not all devices have net identifiers, so this should return `Result<Option<NetIdentifier>>`

examples/c/example.c
gknopf-aranya: Yes

crates/aranya-client/src/client.rs
gknopf-aranya: Are you suggesting implementing an error type for `client.rs` to map the errors to? We should do that but I don't want to block completing factdb queries on that.

crates/aranya-client/src/client.rs
gknopf-aranya: Making this change locally. Marking as resolved to track what's left.

gknopf-aranya: Are do we return a bool/err indicating that the string wasn't set?

elagergren-spideroak: Yeah, return a bool.

crates/aranya-client/src/client.rs
elagergren-spideroak: No, I'm saying that returning `Ok(<expression that evaluates to ()>)` is a /little/ misleading and would be more clear using one of the two styles I listed above.

gknopf-aranya:
Addressed most of the feedback except for the future proofing suggestion so far.
Need to figure out how many of the unimplemented C API methods we want to incorporate into this PR vs. another PR.
Enum return is still blocked by a capi generator issue.
Will take some time to figure out the Vec<> return methods.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -928,3 +988,103 @@ pub unsafe fn afc_recv_data(
...
+/// @relates AranyaClient.
+pub unsafe fn query_aqc_net_identifier(
+    client: &mut Client,
+    team: &TeamId,
+    device: &DeviceId,
+    ident: &mut MaybeUninit<c_char>,
+    ident_len: &mut usize,
+) -> Result<bool, imp::Error> {
+    let client = client.deref_mut();
+    if let Some(net_identifier) = client
```
elagergren-spideroak, line 1061: nit

```rust
let Some(net_identifier) = client else { return Ok(false); };
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -928,3 +988,103 @@ pub unsafe fn afc_recv_data(
...
+    device: &DeviceId,
+) -> Result<KeyBundle, imp::Error> {
+    let client = client.deref_mut();
+    let keys = client
+        .rt
+        .block_on(client.inner.queries(team.0).device_keybundle(device.0))?;
+    Ok(KeyBundle::from_underlying(keys))
+}
+
+// TODO: query_device_label_assignments
```
elagergren-spideroak, line 1016: Something like this:

```rust
fn query_device_label_assignments(...,
    labels: &mut MaybeUninit<u32>,
    labels_len: &mut usize,
) -> Result<(), imp::Error> {
    let data = client.device_label_assignments()?;
    let out = aranya_capi_core::try_as_mut_slice!(labels, *labels_len);
    for (dst, src) in out.iter_mut().zip(&data) { *dst = src }
    Ok(())
}
```

but: you _should_ be able to do this:

```c
size_t len =0;
AranyaError err = query_device_label_assignments(client, team, device, NULL, &len);
if (err != SUCCESS) { ... }
uint32_t* labels = malloc(len * sizeof(uint32_t));
if (labels == NULL) { ... }
err = query_device_label_assignments(client, team, device, labels, &len);
```

so the `labels` parameter should probably be `Option<&mut MaybeUninit<u32>>`.

/cc: @jdygert-spok

crates/aranya-client/src/client.rs
gknopf-aranya: @jdygert-spok, @djohnson-spideroak what are your thoughts on this?
If we decide to take this future proofing approach, should that be implemented on this PR or should we tackle that in the next sprint as we focus on API stability more broadly?

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I've implemented this differently on this PR:
https://github.com/aranya-project/aranya/pull/149

gknopf-aranya: Marking this as resolved since I've addressed this feedback on the other PR.

crates/aranya-client/src/client.rs
gknopf-aranya: Pushed some changes for this

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -928,3 +992,197 @@ pub unsafe fn afc_recv_data(
...
+    team: &TeamId,
+    devices: Option<&mut MaybeUninit<DeviceId>>,
+    devices_len: &mut usize,
+) -> Result<(), imp::Error> {
+    let client = client.deref_mut();
+    let data = client
+        .rt
+        .block_on(client.inner.queries(team.0).devices_on_team())?;
+    let data = data.__data();
+    if let Some(devices) = devices {
```
elagergren-spideroak, line 1015: nit: switch the conditions:

```rust
let Some(devices) = devices else {
    *devices_len = data.len();
    return Err(BufferTooShort);
};
let out = aranya_capi_core::try_as_mut_slice!(...);
[...]
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -928,3 +992,197 @@ pub unsafe fn afc_recv_data(
...
+        let out = aranya_capi_core::try_as_mut_slice!(devices, *devices_len);
+        for (dst, src) in out.iter_mut().zip(data) {
+            dst.write(DeviceId(*src));
+        }
+        *devices_len = data.len();
+        return Ok(());
```
elagergren-spideroak, line 1021: You should be able to simplify this:

```rust
let n = out.len().min(data.len());
out[..n].copy_from_slice(&data[..n])
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -928,3 +992,197 @@ pub unsafe fn afc_recv_data(
...
+    }
+    *devices_len = data.len();
```
elagergren-spideroak, line 1023: We should return an error (BufferTooShort, or whatever it's current called) when `devices_len` is too short.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -928,3 +992,197 @@ pub unsafe fn afc_recv_data(
...
+}
+
+/// The formula for computing the amount of characters in a base64 string from the original number of bytes is:
+/// base64_str_len = (bytes*1375)/1000
+///
+/// A DeviceId is 64 bytes.
+///  
+/// ARANYA_DEVICE_ID_STR_LEN is the number of characters required to hold the base64 string plus the null terminator:
+/// (sizeof(AranyaDeviceId)*1375)/1000+1 = 89
+// TODO: compute size from size_of::<DeviceId>()
```
elagergren-spideroak, line 1034: We don't need all of this context in the public API docs. How about just:

```rust
/// The size in bytes of a [`DeviceId`] converted to a human-readable string.
// <your explanation here>
```

NB: `///` vs `//`

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -928,3 +992,197 @@ pub unsafe fn afc_recv_data(
...
+pub const ARANYA_DEVICE_ID_STR_LEN: u64 = (64 * 1375) / 1000 + 1;
+
+/// Returns a human-readable message for a [`DeviceId`].
+///
+/// This method converts the DeviceId to a base64 encoded string.
+///
+/// Before calling this method, allocate a string of size ARANYA_DEVICE_ID_STR_LEN.
```
elagergren-spideroak, line 1041: These docs are still wrong (it doesn't return anything, it's not a message, it doesn't use base64, etc.).

How about:

```rust
/// Writes the human-readable encoding of `device` to `str`.
///
/// To always succeed, `str` must be at least `ARANYA_DEVICE_ID_STR_LEN` bytes long.
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya:
```
    --> crates/aranya-client-capi/src/api/defs.rs:1115:30
     |
1115 |     out[..n].copy_from_slice(&data[..n]);
     |              --------------- ^^^^^^^^^^ expected `&[MaybeUninit<u32>]`, found `&[Label]`
     |              |
     |              arguments to this method are incorrect
     |
     = note: expected reference `&[std::mem::MaybeUninit<u32>]`
                found reference `&[aranya_client::Label]`
```

elagergren-spideroak: oh, right, yeah, what you're doing is fine then.

elagergren-spideroak: modulo the "buffer too short" stuff i mentioned

djohnson-spideroak (approved):

djohnson-spideroak (approved):

djohnson-spideroak (approved):

ssweetney-spok (approved):

PR #127, opened 2025-03-13
chore: release 0.5.1

jdygert-spok:
Fixes a publishing issue of `aranya-client@0.5.0` using `aranya-daemon@0.5.0` as a dev-dependency. You generally don't use both path and version for dev-dependencies. In this case it should technically be possible but we hit https://github.com/crate-ci/cargo-release/issues/829

### DIFF
--- a/crates/aranya-client/Cargo.toml
+++ b/crates/aranya-client/Cargo.toml
@@ -38,7 +38,7 @@ tokio = { workspace = true, features = ["io-util", "macros", "net", "sync"] }
 tracing = { workspace = true }
 
 [dev-dependencies]
-aranya-daemon = { workspace = true }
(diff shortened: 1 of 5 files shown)


crates/aranya-client/Cargo.toml
```diff
@@ -38,7 +38,7 @@ tokio = { workspace = true, features = ["io-util", "macros", "net", "sync"] }
 tracing = { workspace = true }
 
 [dev-dependencies]
-aranya-daemon = { workspace = true }
+aranya-daemon = { path = "../aranya-daemon" }
```
jdygert-spok, line 41: This is the fix

gknopf-aranya (approved):

djohnson-spideroak (approved):

PR #129, opened 2025-03-15
101/create rust example app in example dir

### LINKED ISSUE #101: Create a rust example app in the examples directory
Even though we have the rust template, we should also have a rust example app alongside the c example in the examples directory. The user should be able to run this example app from the dir.

### LINKED ISSUE #106: rust example: add cargo make task to test the example app
Add a task to `cargo make` for testing the Rust example application.
Ensure the Rust example app is tested as part of cicd.

cc: @djohnson-spideroak, @jdygert-spok, @ycarmy

smckeon-so:
There was some confusion from a client about running the Rust example. This PR adds a Rest example alongside the C example in the examples directory. This will give the user several ways to run a Rust example. 

- Question regarding https://github.com/aranya-project/aranya/pull/107. The way we had the old template dependencies configured, we're pulling from the crate.io releases. Which does not include the afc prefix changes. Do we want to continue with this? Should we instead switch to pulling in the latest `main` changes (use workspace dependencies)? I had to back out the prefix changes.

- This recreates a previous PR. https://github.com/aranya-project/aranya/pull/108

To test cargo generate on this PR branch run `cargo generate aranya-project/aranya --branch 101/Create-rust-example-app-in-example-dir examples/rust --name test-rust-example`

Closes #101, Closes #106


nikki-aranya:
As discussed previously, we should probably just pin the aranya crate dependencies to git, so eventually we can run the example in CI and see if any breakage occurred.

examples/rust/README.md
```diff
@@ -0,0 +1,80 @@
...
+Install [cargo-generate](https://crates.io/crates/cargo-generate) using cargo:
+```bash
+#  Install cargo-generate
+$ cargo install --locked cargo-generate
+```
+
+Navigate to your desired destination parent directory. Then, generate a new project from the template, which will prompt you for the name of the new project you wish to create:
+```bash
+# Create Aranya example project
+$ cargo generate aranya-project/aranya examples/rust
```
gknopf-aranya, line 70: By using `--name` the user would no longer need to manually enter the name of the project:
```suggestion
$ cargo generate aranya-project/aranya examples/rust --name rust-example
```

examples/rust/run.bash
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Copyright (c) SpiderOak, Inc. All rights reserved.
+
+cargo build --release
```
gknopf-aranya, line 5: What if `run.bash` is executed in a different directory than `examples/rust`? There is a way to get the directory where the script lives so commands can be executed relative to `examples/rust`, making `cd examples/rust` optional.
https://stackoverflow.com/questions/59895/how-do-i-get-the-directory-where-a-bash-script-is-located-from-within-the-script

examples/rust/src/main.rs
```diff
@@ -124,7 +124,7 @@ impl UserCtx {
     }
 }
 
-/// Repeatedly calls `poll_afc_data`, followed by `handle_afc_data`, until all
+/// Repeatedly calls `poll_data`, followed by `handle_data`, until all
```
gknopf-aranya, line 127: Changes removing `afc` prefix should be reverted.

gknopf-aranya:
I'm leaning towards keeping the template at `templates/aranya-example` while the `examples/rust/run.bash` script uses that template to execute the example.
Rather than moving the template to `examples/rust`.

Makefile.toml
```diff
@@ -250,3 +250,10 @@ description = "Run C example application"
 cwd = "examples/c"
 script = "./run.bash"
 dependencies = ["build-capi-header", "build-capi-lib"]
+
+# Rust Example
+[tasks.run-rust-example]
```
gknopf-aranya, line 255: Let's invoke this from `tests.yml` to add it to ci

examples/rust/Cargo.toml
```diff
@@ -29,20 +29,14 @@ unwrap_used = "warn"
 wildcard_imports = "warn"
 
 [dependencies]
-# To allow `cargo-generate` to work, explicitly define dependency versions rather than importing the repo's workspace `Cargo.toml`.
-aranya-client = { version = "0.5.1" }
-aranya-daemon = { version = "0.5.1" }
-aranya-daemon-api = { version = "0.5.1" }
-aranya-util = { version = "0.5.1" }
+aranya-client = { git = "https://github.com/aranya-project/aranya.git" }
```
gknopf-aranya, line 32: If we use a path here, we can ensure the example/ci job is running on the current checked out branch of the repo rather than `main`.
```suggestion
aranya-client = { path = "../../crates/aranya-client" }
```
@jdygert-spok any better ideas for handling this?

examples/rust/README.md
```diff
@@ -0,0 +1,83 @@
...
+Install [cargo-generate](https://crates.io/crates/cargo-generate) using cargo:
+```bash
+#  Install cargo-generate
+$ cargo install --locked cargo-generate
+```
+
+Navigate to your desired destination parent directory. Then, generate a new project from the template, which will prompt you for the name of the new project you wish to create:
+```bash
+# Create Aranya example project
+$ cargo generate aranya-project/aranya examples/rust --name rust-example
```
gknopf-aranya, line 70: It's possible to specify a specific branch or tag to generate the template from. `cargo generate --help`:
```
   -b, --branch <BRANCH>
          Branch to use when installing from git
  -t, --tag <TAG>
          Tag to use when installing from git
```

examples/rust/Cargo.toml
gknopf-aranya: My concern is this may not work when a user is trying to generate a template from the repo without checking out the repo first.

smckeon-so: The path approach won't work with the cargo-generate command. It will create the new project, but still look for a relative path which wouldn't exist. 

What we may want to do is loose the cargo generate approach all together and just focus on using the aranya-example. Then, we can use relative paths and be able to test the current state of the codebase in CI.

examples/rust/README.md
smckeon-so: Yeah, that's not an issue `cargo generate aranya-project/aranya --branch 101/Create-rust-example-app-in-example-dir examples/rust --name aranya-example `

examples/rust/Cargo.toml
gknopf-aranya: I agree that the Rust example is more important than cargo generate working.
If we have a running example, it should be simple enough for a user to copy the example to another project.

Another idea: copy a different `Cargo.toml` file into the template directory when running the example than what's used for the template by default. This would allow `cargo generate` to work as expected and give us more control over what branch is used by `examples/rust` and ci.
If we do this, I'd want to copy the entire template to a temporary directory first that is `.gitignore`'ed so we don't end up checking in the alternate `Cargo.toml` to the template.

gknopf-aranya: We should check with @mooreoak and @jdygert-spok first before removing the template since `cargo generate` is intended to make it easier for a user to setup an example repo.

examples/rust/README.md
```diff
@@ -0,0 +1,54 @@
...
+$ git clone git@github.com:aranya-project/aranya.git
+
+# Change into the directory
+$ cd aranya/examples/rust
+
+# Build the binary
+$ cargo build --release
+
+# Run the binary:
+$ target/release/aranya-example
```
gknopf-aranya, line 47:
```suggestion
# Run the example
$ ./run.bash
```

examples/rust/src/main.rs
gknopf-aranya, line 1: Since this example has been added to ci, we could remove the `integration_test` unit tests as long as they have the same test coverage.

templates/aranya-example/Cargo.toml
```diff
@@ -30,10 +30,10 @@ wildcard_imports = "warn"
...
 [dependencies]
 # To allow `cargo-generate` to work, explicitly define dependency versions rather than importing the repo's workspace `Cargo.toml`.
-aranya-client = { version = "0.5.1" }
-aranya-daemon = { version = "0.5.1" }
-aranya-daemon-api = { version = "0.5.1" }
-aranya-util = { version = "0.5.1" }
+aranya-client = { git = "https://github.com/aranya-project/aranya.git" }
+aranya-daemon = { git = "https://github.com/aranya-project/aranya.git" }
+aranya-daemon-api = { git = "https://github.com/aranya-project/aranya.git" }
+aranya-util = { git = "https://github.com/aranya-project/aranya.git" }
```
gknopf-aranya, line 36: This change should be reverted. The example will run on the current branch. Template will continue to use released crate versions.

examples/rust/src/main.rs
gknopf-aranya, line 1: Some changes to the `main.rs` for this example are needed to get things to compile:
https://github.com/aranya-project/aranya/actions/runs/13981036203/job/39146150253?pr=129

Looks like afc prefix differences and the user -> device rename.

smckeon-so: It was the client API changes

examples/rust/README.md
smckeon-so: I wouldn't add the bash script to this section, we could add
```
# or
$ ./examples/rust/run.bash
```
to the "1. Running the example from a script" section though.

nikki-aranya (approved):

nikki-aranya (approved):

crates/aranya-client/tests/tests.rs
```diff
@@ -1,1096 +0,0 @@
...
-    owner_team.remove_device_from_team(team.operator.id).await?;
-    info!("removing admin");
-    owner_team.revoke_role(team.admin.id, Role::Admin).await?;
-    owner_team.remove_device_from_team(team.admin.id).await?;
-
-    Ok(())
-}
-
-#[test(tokio::test(flavor = "multi_thread"))]
-async fn test_afc_one_way_two_chans() -> Result<()> {
```
gknopf-aranya, line 422: Removing `integration_test` is okay because that is covered by the Rust example.
However, the other tests in this file are testing cases that the example does not so we should leave them in.

examples/rust/run.bash
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# Copyright (c) SpiderOak, Inc. All rights reserved.
```
gknopf-aranya, line 3: We usually add this to the beginning of bash scripts to make them easier to debug:
```suggestion
#!/usr/bin/env bash

# Copyright (c) SpiderOak, Inc. All rights reserved.

set -xeuo pipefail
```

examples/rust/run.bash
```diff
@@ -0,0 +1,11 @@
...
+
+# Change into the example directory if needed
+SCRIPT_PATH=$(dirname "$0")
+cd $SCRIPT_PATH
+
+# Build and run the example
+cargo build --release
+./target/release/aranya-example
```
gknopf-aranya, line 11: The output in the ci run is not very verbose. Ideally, we want to print out enough info to troubleshoot ci failures.
```suggestion
ARANYA_EXAMPLE=info ./target/release/aranya-example
```

gknopf-aranya (commented):
lgtm pending open feedback

gknopf-aranya:
This could use review from @djohnson-spideroak, @jdygert-spok, or @elagergren-spideroak

examples/rust/run.bash
smckeon-so: Yes, but one of the goals with this example is to make things easier on the new user. This verbose output ran in that context can be a bit overwhelming. We could just run the info logging level in CI though.

crates/aranya-client/tests/tests.rs
```diff
@@ -283,141 +283,6 @@ impl Drop for DeviceCtx {
...
-/// More extensive integration testing is conducted inside the daemon crate.
-/// The goal of this integration test is to validate the user library's end-to-end functionality.
-/// This includes exercising the user library's idiomatic Rust API as well as the daemon's `tarpc` API.
-///
-/// Example of debugging test with tracing:
-/// `RUST_LOG="debug" cargo test integration_test -- --show-output --nocapture`
-/// `RUST_LOG="aranya_client::afc=debug" cargo test integration_test -- --show-output --nocapture`
-/// `RUST_LOG="aranya_client=debug,aranya_daemon=info" cargo test integration_test -- --show-output --nocapture`
-#[test(tokio::test(flavor = "multi_thread"))]
-async fn integration_test() -> Result<()> {
```
djohnson-spideroak, line 297: do we know why this was being removed?

djohnson-spideroak (commented):
LGTM if the integration test isnt removed. I dont see a reason not to keep it.

crates/aranya-client/tests/tests.rs
djohnson-spideroak: If the example is run during CI/tests, then its fine with me.

djohnson-spideroak (approved):

gknopf-aranya (approved):

ssweetney-spok (approved):
LGTM. I like that Rust Analyzer works for the example now that it's contained in a proper Rust project

jdygert-spok:
There's some stale links in the main readme

gknopf-aranya:
> There's some stale links in the main readme

Opened a PR to fix that:
https://github.com/aranya-project/aranya/pull/161

PR #136, opened 2025-03-20
capi: factdb queries

gknopf-aranya:
C API progress:
- [ ] `QueryDevicesOnTeam(team_id) -> Vec<device_id>`
- [ ] `QueryDeviceRole(team_id, device_id) -> Role`
- [x] `QueryDeviceKeybundle(team_id, device_id) -> Keybundle`
- [x] `QueryAfcNetIdentifier(team_id, device_id) -> network_str`
- [x] `QueryAqcNetIdentifier(team_id, device_id) -> network_str`
- [x] `QueryLabelExists(team_id, label) -> bool`
- [ ] `QueryDeviceLabelAssignments(team_id, device_id) -> Vec<Label>`

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1,6 +1,6 @@
 use core::{ffi::c_char, ops::DerefMut, ptr, slice};
 use std::{
-    ffi::{CString, OsStr},
+    ffi::OsStr,
     os::unix::ffi::OsStrExt,
 };
 
@@ -240,11 +240,6 @@ impl NetIdentifier {
         Ok(aranya_daemon_api::NetIdentifier(String::from(
             cstr.to_str()?,
         )))
-    }
-    // TODO: not sure if this is right.
-    fn from_underlying(net_identifier: aranya_daemon_api::NetIdentifier) -> Self {
-        let cstr = CString::new(net_identifier.0.as_str()).expect("expected valid C str");
-        NetIdentifier(cstr.as_c_str().as_ptr())
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -238,6 +241,11 @@ impl NetIdentifier {
             cstr.to_str()?,
         )))
     }
+    // TODO: not sure if this is right.
```
gknopf-aranya, line 244: @jdygert-spok, @elagergren-spideroak I'm fairly certain this is not right. Could use some help.

elagergren-spideroak: Yeah, it's not. You'll need something like

```rust
fn try_from_underlying(ident: aranya_daemon_api::NetIdentifier) -> Result<Self, imp::Error> {
    let Ok(cstr) = CString::new(ident) else {
        return Err(imp::Error::something_idk_what());
    };
    let ptr = cstr.into_raw();
    Ok(Self(ptr))
}
```

However, that's still not really correct because now we're leaking the `CString`. We _could_ add a function like this:

```rust
/// Frees `ident`.
///
/// # Safety
///
/// - `ident` must have been allocated by Aranya.
/// - `ident` must not have been changed in any way.
pub unsafe "C" fn net_identifier_free(ident: NetIdentifier) {
    if ident.0.is_null() || !ident.0.is_aligned() { return; }
    // SAFETY: See the function's safety docs.
    let cstr = unsafe { CString::from_raw(ident.0) };
    drop(cstr); // for clarity
}
```

But then the user has to manage memory, which is annoying and error prone. Instead, we should write the query APIs like this:

```rust
pub unsafe fn query_afc_net_identifier(
    client: &mut Client,
    team: &TeamId,
    device: &DeviceId,
    ident: &mut MaybeUninit<c_char>,
    ident_len: &mut usize,
) -> Result<(), imp::Error> {
    let net_identifier = client
        .rt
        .block_on(client.inner.team(team.0).query_afc_net_identifier(device.0))?;
    aranya_capi_core::write_c_str(ident, net_identifier, len)?;
    Ok(())
}
```

See [`ext_err_msg`][0] as an example.

(We can probably do better than `write_c_str` since that goes through `String`'s `fmt::Display` impl, which is probably less efficient than just doing a memcpy into the provided buffer, but that's an optimization we can do later.)

[0]: https://github.com/aranya-project/aranya/blob/d145066954a0eace6d105abae8220d608e7e3887/crates/aranya-client-capi/src/api/defs.rs#L127-L151

gknopf-aranya: Trying this out now:
```
    --> crates/aranya-client-capi/src/api/defs.rs:1041:35
     |
1041 |     aranya_capi_core::write_c_str(ident, &mut net_identifier, ident_len)?;
     |     ----------------------------- ^^^^^ expected `&mut [MaybeUninit<i8>]`, found `&mut MaybeUninit<i8>`
     |     |
     |     arguments to this function are incorrect
```

gknopf-aranya:
The query APIs that haven't been implemented on this PR are blocked by features needed in the capi tooling.
We should merge what we have working now then add anything else that's blocked on another PR.

This is one of the capi tooling blockers:
https://github.com/aranya-project/aranya-core/issues/160

I believe we're also blocked on returning vecs but I don't know if we have an issue created for that yet.

elagergren-spideroak:
I'm sorry, this contains code that isn't correct and needs to be reverted.

elagergren-spideroak:
Ah, sorry, never mind. I thought this was merged into `main`.

PR #145, opened 2025-03-24
aqc: policy interface and scaffolding

gknopf-aranya:
This PR updates Aranya policy in preparation for the AQC implementation.
While we're at it, makes corresponding updates to Aranya such as:
- AQC key management (encap passed via IPC, secret passed via key store)
- Effect processing
- Daemon unix domain socket API
- Testing
- Rust/C application APIs
- Rust/C examples (label management and channel creation, but not data transmission)

TODOs:
- [x] AQC Policy Interface. More policy improvements can be made on another PR.
- [x] AQC Policy new label types
- [x] TARPC API (returning AQC encap/meta-data)
- [x] Loading PSKs from key store
- [ ] Make `aqc_receive_ctrl()` private once ctrl messages are passed via the network on Ben's AQC implementation PR. Must be `pub` for testing currently.
- [x] Rust AQC example application and Rust API
- [x] C AQC example application and C API

Some parts of this PR are blocked by:
- AQC crypto https://github.com/aranya-project/aranya-core/pull/184
- AQC FFIs
- AQC implementation

Future work:
- Implementing the rest of the policy permission checks.
- Implementing AQC networking in `aqc.rs`
- Adding AQC data transmission to C/Rust APIs and example applications.
- Updating AFC to use new LabelId instead of integer labels.


gknopf-aranya:
Started working on a PR for the AQC FFIs, but that's blocked on implementations inside other `aranya-core` crates:
https://github.com/aranya-project/aranya-core/pull/182

crates/aranya-client/tests/tests.rs
```diff
@@ -575,6 +591,13 @@ async fn test_afc_one_way_two_chans() -> Result<()> {
     operator_team
         .assign_afc_net_identifier(team.memberb.id, NetIdentifier(memberb_afc_addr.to_string()))
         .await?;
+    // TODO: use aqc addr
+    operator_team
+        .assign_aqc_net_identifier(team.membera.id, NetIdentifier(membera_afc_addr.to_string()))
+        .await?;
+    operator_team
+        .assign_aqc_net_identifier(team.memberb.id, NetIdentifier(memberb_afc_addr.to_string()))
+        .await?;
```
nikki-aranya, line 600: This is inside a test for AFC, we should have separate tests for AQC

crates/aranya-client/src/client.rs
```diff
@@ -42,14 +43,18 @@ impl Client {
...
     ///   on.
-    #[instrument(skip_all, fields(?daemon_socket, ?afc_shm_path, max_channels))]
+    // TODO: aqc_address
+    #[instrument(skip_all, fields(?daemon_socket, ?afc_shm_path, ?aqc_shm_path, max_channels))]
     pub async fn connect<A>(
         daemon_socket: &Path,
         afc_shm_path: &Path,
+        aqc_shm_path: &Path,
         max_channels: usize,
         afc_address: A,
```
nikki-aranya, line 59: ideally we should instead have a enum ChannelInfo { AFC(shm_path, address), AQC(shm_path, address) } so we don't clutter this, on the assumption you'll only want one or the other since they provide identical(?) functionality

crates/aranya-client/src/client.rs
```diff
@@ -64,12 +69,21 @@ impl Client {
...
 
-        let read = setup_afc_shm(afc_shm_path, max_channels)?;
-        let afc = FastChannelsImpl::new(afc::Client::new(read), afc_address).await?;
+        let afc_read = setup_afc_shm(afc_shm_path, max_channels)?;
+        let afc = FastChannelsImpl::new(afc::Client::new(afc_read), afc_address).await?;
         debug!(
             addr = ?afc.local_addr().map_err(Error::Afc)?,
             "bound AFC router",
         );
+        let _aqc_read = setup_afc_shm(aqc_shm_path, max_channels)?;
```
nikki-aranya, line 78: we should instead have setup_aqc_shim inside aqc.rs. It might duplicate code but it'll be cleaner in the long run

gknopf-aranya: Yes, that should be added in @benz-spideroak's AQC implementation PR

crates/aranya-client/tests/tests.rs
```diff
@@ -201,13 +201,21 @@ impl DeviceCtx {
...
+            afc_shm_path = format!("/{}", t.to_base58())
+        };
+        let mut aqc_shm_path = format!("/aqc_{team_name}_{name}");
+        if cfg!(target_os = "macos") && aqc_shm_path.len() > 31 {
+            // Shrink the size of the team name down to 22 bytes
+            // to work within macOS's limits.
+            let d = Sha256::hash(aqc_shm_path.as_bytes());
+            let t: [u8; 16] = d[..16].try_into().unwrap();
+            aqc_shm_path = format!("/{}", t.to_base58())
         };
```
nikki-aranya, line 219: I feel like this should be split out into a function now that we're using it twice and its relative complexity

crates/aranya-client/src/client.rs
gknopf-aranya: This PR is setting up some scaffolding for AQC that's related to the policy changes, but not implementing AQC

crates/aranya-client/tests/tests.rs
gknopf-aranya: Agreed, I had similar thoughts but wanted to get this up quickly to identify any blockers.

crates/aranya-client/tests/tests.rs
nikki-aranya: I'd take a TODO

crates/aranya-client/src/client.rs
gknopf-aranya: We'll want to be able to hide them behind feature flags on your PR

gknopf-aranya: I was under the impression that both aqc and afc features could be enabled at the same time.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Added a todo for this

crates/aranya-client/tests/tests.rs
gknopf-aranya: Created a function for this

crates/aranya-client/src/client.rs
gknopf-aranya: Created an `aqc.rs` file that only has `setup_aqc_shm` inside. More will be added later on a different PR.

crates/aranya-client/tests/tests.rs
```diff
@@ -161,6 +161,17 @@ macro_rules! do_poll {
     };
 }
 
+fn shm_path(path: String) -> String {
```
nikki-aranya, line 164:
```suggestion
fn get_shm_path(path: String) -> String {
```

crates/aranya-client/tests/tests.rs
```diff
@@ -201,22 +212,8 @@ impl DeviceCtx {
...
-        let mut aqc_shm_path = format!("/aqc_{team_name}_{name}");
-        if cfg!(target_os = "macos") && aqc_shm_path.len() > 31 {
-            // Shrink the size of the team name down to 22 bytes
-            // to work within macOS's limits.
-            let d = Sha256::hash(aqc_shm_path.as_bytes());
-            let t: [u8; 16] = d[..16].try_into().unwrap();
-            aqc_shm_path = format!("/{}", t.to_base58())
-        };
+        let afc_shm_path = shm_path(format!("/afc_{team_name}_{name}"));
+        let aqc_shm_path = shm_path(format!("/aqc_{team_name}_{name}"));
```
nikki-aranya, line 216:
```suggestion
        let afc_shm_path = get_shm_path(format!("/afc_{team_name}_{name}"));
        let aqc_shm_path = get_shm_path(format!("/aqc_{team_name}_{name}"));
```

crates/aranya-client/src/client.rs
benz-spideroak: I think they can both be enabled.

gknopf-aranya: I can see a use-case where AFC would be used over one interface while AQC is used over another depending on the network properties. Having the ability to enable each one individually gives us the most flexibility.

gknopf-aranya: @nikki-aranya this decision should be handled by the AFC experimental flag issue that you're working on here: https://github.com/aranya-project/aranya/issues/79

gknopf-aranya: Resolving this because it is not in scope for this PR. Moved discussion to relevant issue.

nikki-aranya (dismissed):

gknopf-aranya:
Next this to do on this PR is integrate with the aqc ffi here:
https://github.com/aranya-project/aranya-core/pull/184

gknopf-aranya:
After attempting to use the new aqc ffis, I'm getting a whole bunch of compiler errors.

Cargo.toml
```diff
@@ -35,10 +35,13 @@ unwrap_used = "warn"
 wildcard_imports = "warn"
 
 [workspace.dependencies]
-aranya-afc-util = { version = "0.5.0", features = ["alloc"] }
+aranya-afc-util = { git = "https://github.com/aranya-project/aranya-core.git", branch = "issue/183", features = ["alloc"] }
+# TODO: use released crate
+aranya-aqc-util = { git = "https://github.com/aranya-project/aranya-core.git", branch = "issue/183", features = ["alloc"] }
```
gknopf-aranya, line 40: These are some of the compiler errors I'm seeing after attempting to use the new aqc ffis:
```
error[E0277]: the trait bound `aranya_crypto::default::DefaultCipherSuite: aranya_fast_channels::crypto::aranya_crypto::CipherSuite` is not satisfied
   --> crates/aranya-daemon/src/daemon.rs:213:13
    |
213 | /             WriteState::open(
214 | |                 &path,
215 | |                 Flag::Create,
216 | |                 Mode::ReadWrite,
217 | |                 self.cfg.afc.max_chans,
218 | |                 Rng,
219 | |             )
    | |_____________^ the trait `aranya_fast_channels::crypto::aranya_crypto::CipherSuite` is not implemented for `aranya_crypto::default::DefaultCipherSuite`
    |
    = help: the trait `aranya_fast_channels::crypto::aranya_crypto::CipherSuite` is implemented for `aranya_fast_channels::crypto::aranya_crypto::default::DefaultCipherSuite`
note: required by a bound in `WriteState::<CS, R>::open`
   --> /Users/gknopf/.cargo/registry/src/index.crates.io-6f17d22bba15001f/aranya-fast-channels-0.5.0/src/shm/write.rs:34:9
    |
34  |     CS: CipherSuite,
    |         ^^^^^^^^^^^ required by this bound in `WriteState::<CS, R>::open`
```
```
error[E0277]: the trait bound `E: aranya_fast_channels::crypto::Engine` is not satisfied
  --> crates/aranya-daemon/src/vm_policy.rs:86:13
   |
86 |             Box::from(PerspectiveFfi),
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `aranya_fast_channels::crypto::Engine` is not implemented for `E`, which is required by `FfiPerspective: FfiCallable<E>`
   |
   = note: required for `FfiPerspective` to implement `FfiCallable<E>`
   = note: required for the cast from `Box<FfiPerspective>` to `Box<dyn FfiCallable<E> + std::marker::Send>`
help: consider further restricting this bound
   |
59 |     E: aranya_crypto::Engine + aranya_fast_channels::crypto::Engine,
   |                              ++++++++++++++++++++++++++++++++++++++=
```
```
error[E0277]: the trait bound `Store: aranya_fast_channels::crypto::aranya_crypto::KeyStore` is not satisfied
  --> crates/aranya-daemon/src/vm_policy.rs:82:13
   |
82 |             Box::from(CryptoFfi::new(store.try_clone()?)),
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `aranya_fast_channels::crypto::aranya_crypto::KeyStore` is not implemented for `Store`, which is required by `aranya_crypto_ffi::Ffi<Store>: FfiCallable<E>`
   |
   = help: the trait `aranya_fast_channels::crypto::aranya_crypto::KeyStore` is implemented for `aranya_fast_channels::crypto::aranya_crypto::keystore::memstore::MemStore`
   = note: required for `aranya_crypto_ffi::Ffi<Store>` to implement `FfiModule`
   = note: required for `aranya_crypto_ffi::Ffi<Store>` to implement `FfiCallable<E>`
   = note: required for the cast from `Box<aranya_crypto_ffi::Ffi<Store>>` to `Box<dyn FfiCallable<E> + std::marker::Send>`

   --> crates/aranya-daemon/src/vm_policy.rs:83:38
    |
83  |             Box::from(DeviceFfi::new(device_id)),
    |                       -------------- ^^^^^^^^^ expected `DeviceId`, found `aranya_crypto::DeviceId`
    |                       |
    |                       arguments to this function are incorrect
    |
    = note: `aranya_crypto::DeviceId` and `DeviceId` have similar names, but are actually distinct types
```

gknopf-aranya: Resolved this by setting the rest of the aranya-core deps to the aqc ffi feature branch

crates/aranya-client/src/error.rs
```diff
@@ -154,4 +160,119 @@ pub enum AfcError {
...
+    #[error("stream not found: {0}")]
+    StreamNotFound(std::net::SocketAddr),
+
+    // Protocol-related errors
+    // TODO: replace with QUIC header error.
+    /// Invalid AC header.
+    //#[error("invalid AQC header: {0}")]
+    //InvalidHeader(#[from] aranya_fast_channels::HeaderError),
+
+    /// Invalid AFC magic.
```
gknopf-aranya, line 224: AQC

crates/aranya-client/src/error.rs
```diff
@@ -154,4 +160,119 @@ pub enum AfcError {
     Bug(#[from] buggy::Bug),
 }
 
+/// Possible errors that could happen when using Aranya QUIC Channels.
+#[derive(Debug, thiserror::Error)]
+pub enum AqcError {
```
gknopf-aranya, line 165: AFC -> AQC for all the error variants/docs

Cargo.toml
```diff
@@ -35,23 +35,25 @@ unwrap_used = "warn"
...
-aranya-libc = { version = "0.1.1" }
-aranya-perspective-ffi = { version = "0.5.0" }
-aranya-policy-compiler = { version = "0.5.0" }
-aranya-policy-ifgen = { version = "0.5.0", features = ["serde"] }
-aranya-policy-ifgen-build = { version = "0.2.0" }
-aranya-policy-lang = { version = "0.2.0" }
-aranya-policy-vm = { version = "0.5.0", features = ["std"] }
-aranya-runtime = { version = "0.5.0", features = ["std", "libc"] }
+aranya-afc-util = { version = "0.6.0", features = ["alloc"] }
+# TODO: use released crate
```
gknopf-aranya, line 39: rm todos

examples/c/example.c
```diff
@@ -363,23 +380,23 @@ AranyaError run(Team *t) {
...
-    err = aranya_assign_label(&t->clients.operator.client, &t->id,
-                              &t->clients.membera.id, label);
+    err = aranya_assign_afc_label(&t->clients.operator.client, &t->id,
+                                  &t->clients.membera.id, label);
     EXPECT("error assigning afc label to membera", err);
 
-    err = aranya_assign_label(&t->clients.operator.client, &t->id,
-                              &t->clients.memberb.id, label);
+    err = aranya_assign_afc_label(&t->clients.operator.client, &t->id,
+                                  &t->clients.memberb.id, label);
```
djohnson-spideroak, line 399: probably dont want the AFC stuff in here anymore.

examples/c/example.c
```diff
@@ -605,9 +636,59 @@ AranyaError run(Team *t) {
...
+    AranyaChanOp op = ARANYA_CHAN_OP_SEND_RECV;
+    err             = aranya_assign_label(&t->clients.operator.client, &t->id,
+                                          &t->clients.membera.id, &label_id, op);
+    EXPECT("error assigning label to membera", err);
+    err = aranya_assign_label(&t->clients.operator.client, &t->id,
+                              &t->clients.memberb.id, &label_id, op);
+    EXPECT("error assigning label to memberb", err);
+    sleep(1);
+
+    // Create channel
```
djohnson-spideroak, line 668:
```suggestion
    // Create channel using Member A's client
```

examples/c/example.c
```diff
@@ -605,9 +636,59 @@ AranyaError run(Team *t) {
...
+    printf("creating AQC channel \r\n");
+    AranyaAqcChannelId chan_id;
+    err =
+        aranya_aqc_create_bidi_channel(&t->clients.membera.client, &t->id,
+                                       afc_addrs[MEMBERB], &label_id, &chan_id);
+    EXPECT("error creating aqc bidi channel", err);
+
+    // TODO: send AQC data
+
+    // Revoke/delete label
```
djohnson-spideroak, line 678:
```suggestion
    // Revoke/delete label using the Operator
```

examples/c/example.c
gknopf-aranya: This is hidden behind `#if defined(ENABLE_AFC)` so will get compiled out if AFC isn't being used.

gknopf-aranya: The reason we haven't taken this out of the example is so it's easier to add back when adding AFC back (after the label ID fix). It's possible we may never add AFC back and end up deleting all the AFC related code. That would happen on a different PR though.

crates/aranya-client/tests/tests.rs
```diff
@@ -517,15 +526,21 @@ async fn test_afc_one_way_two_chans() -> Result<()> {
...
-            .await?;
-        operator_team
-            .assign_aqc_net_identifier(team.memberb.id, NetIdentifier(memberb_afc_addr.to_string()))
-            .await?;
     }
 
+    // get aqc addresses.
+    // TODO: use aqc_local_addr()
+    let membera_aqc_addr = SocketAddr::new(std::net::IpAddr::V4(Ipv4Addr::LOCALHOST), 0);
+    let memberb_aqc_addr = SocketAddr::new(std::net::IpAddr::V4(Ipv4Addr::LOCALHOST), 1);
```
djohnson-spideroak, line 534: isnt this assigning port 1?

gknopf-aranya: This will be fixed by the TODO which will not be implemented until Ben's PR binds the QUIC server to a socket.

I needed to use different addresses because if they are the same, this map `aqc_peers: Arc<Mutex<BiBTreeMap<NetIdentifier, DeviceId>>>,` will overwrite the first net identifier with the next one.

crates/aranya-daemon/src/policy.md
```diff
@@ -941,6 +1028,785 @@ command UnsetAqcNetworkName {
...
+    }
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        let author = get_valid_device(envelope::author_id(envelope))
+        let target = get_valid_device(this.device_id)
+
+        // NB: Check roles, other ACLs here.
```
djohnson-spideroak, line 1723: need permission check

crates/aranya-daemon/src/policy.md
```diff
@@ -941,6 +1028,785 @@ command UnsetAqcNetworkName {
...
```
djohnson-spideroak, line 1657: permission check

crates/aranya-daemon/src/policy.md
```diff
@@ -241,6 +272,62 @@ function select_peer_id(device_id id, id_a id, id_b id) id {
 }
 ```
 
+##### AQC Channel Functions
+
+```policy
+// Reports whether the devices have permission to create
+// a bidirectional AQC channel with each other.
+function can_create_aqc_bidi_channel(device1 id, device2 id, label_id id) bool {
```
jdygert-spok, line 280: The mix of returning bool but doing e.g. `check_unwrap` in `get_allowed_op` is kinda awkward. I think I'd make it entirely checks and name it `check_can_create...` with no return value but I don't think that's currently allowed :/

crates/aranya-daemon/src/policy.md
gknopf-aranya: Adding additional permissions checks will be in a follow-on PR

crates/aranya-daemon/src/policy.md
gknopf-aranya: Yes, the policy is still work-in-progress. The goal is to get an interface for AQC policy added in this PR so Ben can start integrating with it in `aqc.rs`. A follow-on PR will clean up the AQC policy and add additional permissions checks.

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok, @elagergren-spideroak, @ycarmy here's a follow-on issue for completing the AQC policy:
https://github.com/aranya-project/aranya/issues/186

gknopf-aranya:
Planning to merge this PR today to unblock:
- @benz-spideroak's integration with AQC networking code (e.g. QUIC client/server)
- @nikki-aranya's AQC net identifier work
- Other AQC integrations

crates/aranya-daemon/src/policy.md
gknopf-aranya: Adding additional permissions checks will be in a follow-on PR

crates/aranya-client/tests/tests.rs
gknopf-aranya: @benz-spideroak fyi, that I'm anticipating your AQC PR will resolve this TODO

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -185,3 +192,36 @@ impl AfcConfigBuilder {
...
     }
 }
+
+#[repr(C)]
+#[derive(Copy, Clone, Debug)]
+#[aranya_capi_core::opaque(size = 24, align = 8)]
+/// Configuration info for Aranya Fast Channels
+pub struct AqcConfig {
+    /// Address to bind AQC server to.
+    pub addr: *const c_char,
```
elagergren-spideroak, line 202: doesn't need to be `pub`

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -185,3 +192,36 @@ impl AfcConfigBuilder {
...
+impl Typed for AqcConfig {
+    const TYPE_ID: TypeId = TypeId::new(0x227DFC9F);
+}
+
+#[derive(Copy, Clone, Debug)]
+#[aranya_capi_core::opaque(size = 24, align = 8)]
+/// Builder for an [`AqcConfig`]
+pub struct AqcConfigBuilder {
+    /// Address to bind AQC server to.
+    pub addr: *const c_char,
```
elagergren-spideroak, line 214: ditto

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -457,12 +531,53 @@ pub fn afc_config_builder_build(
...
+
 /// Configuration info for Aranya.
-#[aranya_capi_core::opaque(size = 48, align = 8)]
+#[aranya_capi_core::opaque(size = 56, align = 8)]
 pub type ClientConfig = Safe<imp::ClientConfig>;
 
 /// Configuration info builder for Aranya.
-#[aranya_capi_core::opaque(size = 40, align = 8)]
+#[aranya_capi_core::opaque(size = 56, align = 8)]
 pub type ClientConfigBuilder = imp::ClientConfigBuilder;
```
elagergren-spideroak, line 581: This should probably be `Safe<imp::ClientConfigBuilder>`

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -38,12 +38,13 @@ impl From<&SyncPeerConfig> for aranya_client::client::SyncPeerConfig {
 
 #[repr(C)]
 #[derive(Copy, Clone, Debug)]
-#[aranya_capi_core::opaque(size = 32, align = 8)]
+#[aranya_capi_core::opaque(size = 40, align = 8)]
```
elagergren-spideroak, line 41: You shouldn't need to make this struct `opaque`. Or even `repr(c)`.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -52,12 +53,13 @@ impl Typed for ClientConfig {
 
 #[repr(C)]
 #[derive(Copy, Clone, Debug)]
-#[aranya_capi_core::opaque(size = 40, align = 8)]
+#[aranya_capi_core::opaque(size = 56, align = 8)]
```
elagergren-spideroak, line 56: ditto

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -52,12 +53,13 @@ impl Typed for ClientConfig {
...
 /// Builder for a [`ClientConfig`]
 pub struct ClientConfigBuilder {
     pub daemon_addr: *const c_char,
     #[cfg(feature = "afc")]
     pub afc: Option<AfcConfig>,
+    pub aqc: Option<AqcConfig>,
```
elagergren-spideroak, line 62: I don't think these need to be `pub`.

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,319 @@
...
+                    &mut self.client.aqc.eng.clone(),
+                    &BidiChannelCreated {
+                        parent_cmd_id: v.parent_cmd_id,
+                        author_id: v.author_id.into_id().into(),
+                        author_enc_key_id: v.author_enc_key_id,
+                        peer_id: v.peer_id.into_id().into(),
+                        peer_enc_pk: &v.peer_enc_pk,
+                        label_id: v.label_id.into_id().into(),
+                        channel_id: v.channel_id,
+                        psk_length_in_bytes: 32, // TODO: don't hard-code
```
elagergren-spideroak, line 128: you can get this from the effect now

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,319 @@
...
+                    &UniChannelCreated {
+                        parent_cmd_id: v.parent_cmd_id,
+                        author_id: v.author_id.into_id().into(),
+                        author_enc_key_id: v.author_enc_key_id,
+                        send_id: v.send_id.into_id().into(),
+                        recv_id: v.recv_id.into_id().into(),
+                        peer_enc_pk: &v.peer_enc_pk,
+                        label_id: v.label_id.into_id().into(),
+                        channel_id: v.channel_id,
+                        psk_length_in_bytes: 32, // TODO: don't hard-code
```
elagergren-spideroak, line 186: ditto

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,319 @@
...
+
+            // TODO: send ctrl msg via network.
+        }
+
+        // TODO: for testing only. Send ctrl via network instead of returning.
+        Ok((aqc_id, aqc_ctrl))
+    }
+
+    /// Deletes an AQC channel.
+    // TODO(eric): Is it an error if the channel does not exist?
```
elagergren-spideroak, line 201: yes :)

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,319 @@
...
+use crate::error::AqcError;
+
+// TODO: use same generics as daemon.
+/// CE = Crypto Engine
+pub(crate) type CE = DefaultEngine;
+/// KS = Key Store
+pub(crate) type KS = Store;
+
+/// Length of a PSK key.
+const PSK_KEY_LEN: u16 = 32;
```
elagergren-spideroak, line 37: no longer needed since the length is returned in an effect

crates/aranya-client/src/client.rs
```diff
@@ -137,6 +182,11 @@ impl Client {
             .map_err(Into::into)
     }
 
+    /// Returns the address that the AQC server is bound to.
+    pub async fn aqc_local_addr(&mut self) -> Result<SocketAddr> {
+        Ok(self.aqc().local_addr()?)
+    }
+
```
elagergren-spideroak, line 189: Let's remove this, users can just write `client.aqc().local_addr()`.

elagergren-spideroak: But _if_ we want this, then it should be `&self`, not `&mut self`.

crates/aranya-daemon/src/policy.md
gknopf-aranya: As mentioned previously in a group chat, I plan to open a follow on PR where we review the entire AQC policy in that PR. I'll remove the entire `policy.md` file on a `rm-policy` branch (that will not be merged), then rebase the AQC policy PR on that branch so it shows the entire policy as a delta. Once we're good with the policy, we can merge it into `main`.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: True, but that's not in scope for this PR.
@ssweetney-spok and @nikki-aranya have worked on the builder configs recently and would be more familiar with making these updates.

crates/aranya-client-capi/src/imp/config.rs
gknopf-aranya: This is only changing the `size` because of the added `AqcConfig`.
`opaque` and `repr(c)` were already present.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: IIRC, I didn't mention it during code review since I didn't think it was worth adding the `Safe` wrapper since all the fields implemented `ByMutPtr` and the like. Also, @nikki-aranya wrote included validation checks in the builder methods so there shouldn't be a way to construct an invalid client config

crates/aranya-client/src/aqc.rs
gknopf-aranya: Thanks, I'll update this

crates/aranya-client/src/client.rs
gknopf-aranya: Yeah, it was kindof clunky anyways and really only used for temporary testing until there is a QUIC server that binds to an interface to make the `local_addr()` available.

crates/aranya-client/src/aqc.rs
gknopf-aranya: @elagergren-spideroak I noticed the handler methods are expecting a `u16` while the effects are returning a `i64` for the key length.

elagergren-spideroak: Yes, this is expected b/c policy only supports `i64` integers. Just use `TryFrom` + `BugExt`.

crates/aranya-client-capi/src/imp/config.rs
gknopf-aranya: Tried removing `pub` from this field and got this:
```
error[E0616]: field `addr` of struct `config::AqcConfigBuilder` is private
   --> crates/aranya-client-capi/src/api/defs.rs:559:9
    |
559 |     cfg.addr = address;
    |         ^^^^ private field

error[E0616]: field `aqc` of struct `config::ClientConfigBuilder` is private
   --> crates/aranya-client-capi/src/api/defs.rs:615:9
    |
615 |     cfg.aqc = Some(**aqc_config);
    |         ^^^ private field

```

elagergren-spideroak: yes, b/c it should be `Safe<...>` and use getters/setters.

Or just be `pub(crate)` for now.

gknopf-aranya: I'll use `pub(crate)` for now. Let's clean this up on a different PR while we're auditing the APIs.

gknopf-aranya: Here's an issue for cleaning up `pub` fields in the C API as part of the MVP API audit:
https://github.com/aranya-project/aranya/issues/188

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Created an issue to handle this here:
https://github.com/aranya-project/aranya/issues/188

crates/aranya-client-capi/src/imp/config.rs
gknopf-aranya: Created an issue to clean this up on another PR:
https://github.com/aranya-project/aranya/issues/189

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1368,6 +1625,46 @@ pub unsafe fn query_device_keybundle(
...
 
+/// Query device label assignments.
+///
+/// @param client the Aranya Client [`Client`].
+/// @param team the team's ID [`TeamId`].
+/// @param device the device's ID [`DeviceId`].
+/// @param labels returns a list of labels assigned to the device [`LabelId`].
+/// @param labels_len returns the length of the labels list [`LabelId`].
+///
+/// @relates AranyaClient.
```
djohnson-spideroak, line 1636: nit: I think even though "returns" is part of the output params, it is still a bit unclear on first glance which params are output.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1470,12 +1767,69 @@ pub unsafe fn query_aqc_net_identifier(
...
+    for (dst, src) in out.iter_mut().zip(data) {
+        dst.write((*src).into());
+    }
+    if *labels_len < data.len() {
+        *labels_len = data.len();
+        return Err(imp::Error::BufferTooSmall);
+    }
+    *labels_len = data.len();
+    Ok(())
+}
```
djohnson-spideroak, line 1824: looks like this only returns the label IDs, what about the label names?

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: How could we make that more clear?

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Yeah, it would be nice to be able to print out the name of each label

gknopf-aranya: The current method of returning an array of labels would no longer work because instead of returning an array of
```
LabelId {
  id: Id,
} 
```
We would want to return a struct with strings:
```
LabelId {
  id: Id,
  str: String,
} 
```
So far, we've been returning individual strings to a preallocated buffer like this:
```
    str: &mut MaybeUninit<c_char>,
    str_len: &mut usize,
```
But that method wouldn't work for an array of strings.
I suggest that we add a `label_to_str()` method to return the string of an individual label to avoid the complexity of returning an array of strings.

gknopf-aranya: We'll want to come up with a standard approach for returning an array of strings to the application.
So far, we've only been returning individual strings.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1368,6 +1625,46 @@ pub unsafe fn query_device_keybundle(
...
+        return Err(imp::Error::BufferTooSmall);
+    };
+    let out = aranya_capi_core::try_as_mut_slice!(labels, *labels_len);
+    for (dst, src) in out.iter_mut().zip(data) {
+        dst.write((*src).into());
+    }
+    if *labels_len < data.len() {
+        *labels_len = data.len();
+        return Err(imp::Error::BufferTooSmall);
+    }
```
ssweetney-spok, line 1663: Should probably document that a partial list can be returned when there's an error or move the conditional check before the data is written

examples/c/example.c
```diff
@@ -633,9 +665,112 @@ AranyaError run(Team *t) {
...
+    size_t device_str_len = ARANYA_ID_STR_LEN;
+    char *device_str      = malloc(ARANYA_ID_STR_LEN);
+    aranya_id_to_str(&t->clients.memberb.id.id, device_str, &device_str_len);
+    printf("query labels assigned to device: %s\r\n", device_str);
+    size_t labels_len     = BUF_LEN;
+    AranyaLabelId *labels = malloc(labels_len * sizeof(AranyaLabelId));
+    err = aranya_query_device_label_assignments(&t->clients.operator.client,
+                                                &t->id, &t->clients.memberb.id,
+                                                labels, &labels_len);
+    EXPECT("error querying labels assigned to device", err);
```
ssweetney-spok, line 715: We should show how to handle `ARANYA_ERROR_BUFFER_TOO_SMALL` errors because it's not obvious without looking at the source code that the actual length of the data is written to the `labels_len` even when this error occurs

examples/c/example.c
```diff
@@ -633,9 +665,112 @@ AranyaError run(Team *t) {
...
+    }
+    free(device_str);
+
+    size_t team_str_len = ARANYA_ID_STR_LEN;
+    char *team_str      = malloc(ARANYA_ID_STR_LEN);
+    aranya_id_to_str(&t->id.id, team_str, &team_str_len);
+    printf("query labels on team: %s\r\n", team_str);
+    err = aranya_query_labels(&t->clients.operator.client, &t->id, labels,
+                              &labels_len);
+    EXPECT("error querying labels on team", err);
```
ssweetney-spok, line 735: Same as above. Handle `ARANYA_ERROR_BUFFER_TOO_SMALL`

examples/c/example.c
ssweetney-spok: Should make this more clear in the docs too

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -245,6 +250,35 @@ impl From<&DeviceId> for aranya_daemon_api::DeviceId {
...
+    fn from(value: &LabelId) -> Self {
+        value.id.bytes.into()
+    }
+}
+
+/// Channel ID for AQC channel.
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug)]
+#[aranya_capi_core::opaque(size = 16, align = 1)]
+pub struct AqcChannelId(aranya_daemon_api::AqcId);
```
ssweetney-spok, line 280: I don't think we want `opaque` for IDs anymore

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1470,12 +1767,69 @@ pub unsafe fn query_aqc_net_identifier(
...
+        return Err(imp::Error::BufferTooSmall);
+    };
+    let out = aranya_capi_core::try_as_mut_slice!(labels, *labels_len);
+    for (dst, src) in out.iter_mut().zip(data) {
+        dst.write((*src).into());
+    }
+    if *labels_len < data.len() {
+        *labels_len = data.len();
+        return Err(imp::Error::BufferTooSmall);
+    }
```
ssweetney-spok, line 1821: [Same](https://github.com/aranya-project/aranya/pull/145/files#r2043432007)

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok, line 280: This type as it is won't work with the existing `id_to_str` and `id_from_str` functions. Should there be new ones just for this type or should this ID be extended to 64 bytes?

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: > or move the conditional check before the data is written

Yeah, I think if the buffer is too small, we shouldnt write anything.

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: We might have to, or modify the `id_*_str` functions to handle the AQC IDs as a different case. I think adding a special case is less ideal.

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: Maybe some spacing between input/output params, with a label?

```suggestion
/// Query device label assignments.
///
/// @param client the Aranya Client [`Client`].
/// @param team the team's ID [`TeamId`].
/// @param device the device's ID [`DeviceId`].
///
/// Output params:
/// @param labels returns a list of labels assigned to the device [`LabelId`].
/// @param labels_len returns the length of the labels list [`LabelId`].
///
/// @relates AranyaClient.
```

jdygert-spok: Doxygen has an in/out modifier, e.g. `@param[out]` Not sure how it actually looks rendered.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Yep, making this change

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: > We might have to, or modify the `id_*_str` functions to handle the AQC IDs as a different case. I think adding a special case is less ideal.

Oh yea, we could modify the existing functions. I don't think it'll be too hard to handle if we create a new Enum.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: I'd rather use the attributes which assign semantic meaning. Especially since `labels_len` is both, so `@param[in,out]`. But it can be done on a cleanup PR to update existing as well.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: I don't see any reason not to use a normal aranya ID. It was copied from afc, which was truncated for perf reasons.

examples/c/example.c
gknopf-aranya: Updated docs and showed an example of handling the error in `example.c`

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Changing the Aqc channel ID type would require a change in `aranya-core` to the AQC crypto/ffi code.
@elagergren-spideroak should we make that change?

elagergren-spideroak: Sorry, why does it require changes to `aranya-core`?

crates/aranya-client/src/client.rs
```diff
@@ -53,6 +59,24 @@ impl Labels {
     }
 }
 
+/// List of AFC labels.
+#[cfg(feature = "afc")]
+pub struct AfcLabels {
+    pub data: Vec<AfcLabel>,
```
jdygert-spok, line 65:
```suggestion
   data: Vec<AfcLabel>,
```

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,316 @@
...
+        debug!("keystore wrapped key path: {:?}", keystore_info.wrapped_key);
+        let store = KS::open(keystore_info.path).context("unable to open keystore")?;
+        let handler = Handler::new(
+            device_id.into_id().into(),
+            store.try_clone().context("unable to clone keystore")?,
+        );
+        let eng = {
+            let key = load_or_gen_key_wrap_key(keystore_info.wrapped_key).await?;
+            CE::new(&key, Rng)
+        };
```
jdygert-spok, line 60: Have we considered implications of needing a second engine in the client for this?

gknopf-aranya: This is needed to derive the psk from the encap. We're unable to pass the psk directly from the daemon to the client because the unix domain socket API is not encrypted.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Updating to not use `opaque` for this type.

examples/c/example.c
```diff
@@ -633,9 +665,142 @@ AranyaError run(Team *t) {
...
+    // `ARANYA_ERROR_BUFFER_TOO_SMALL` error handling.
+    size_t labels_len     = 1;
+    AranyaLabelId *labels = malloc(labels_len * sizeof(AranyaLabelId));
+    err = aranya_query_device_label_assignments(&t->clients.operator.client,
+                                                &t->id, &t->clients.memberb.id,
+                                                labels, &labels_len);
+    if (err == ARANYA_ERROR_BUFFER_TOO_SMALL) {
+        printf("handling buffer too small error\r\n");
+        free(labels);
+        labels = malloc(labels_len * sizeof(AranyaLabelId));
```
gknopf-aranya, line 731: nit: should use `realloc` here

crates/aranya-daemon-api/src/service.rs
```diff
@@ -67,8 +64,13 @@ custom_id! {
...
 custom_id! {
-    /// An AQC channel ID.
-    pub struct ChannelId;
+    /// An AQC bidi channel ID.
+    pub struct AqcBidiChannelId;
+}
+
+custom_id! {
+    /// An AQC uni channel ID.
+    pub struct AqcUniChannelId;
```
djohnson-spideroak, line 73: do we need different IDs for bidi/uni?

gknopf-aranya: Fundamentally they are both Aranya IDs.
However, they are defined as different types in `aranya_crypto`:
```
use aranya_crypto::{
    aead::Aead,
    aqc::{BidiChannelId, UniChannelId},
```
We'll have better type safety if we keep them as different types.

crates/aranya-client/src/aqc.rs
gknopf-spideroak: @jdygert-spok, @elagergren-spideroak I've confirmed there is 1 instance of the engine in the Aranya client, and 1 instance in the Aranya daemon.

djohnson-spideroak (dismissed):

gknopf-aranya:
@elagergren-spideroak, @jdygert-spok I need approval from you on this PR.

crates/aranya-client/src/client.rs
gknopf-aranya: Resolved

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,336 @@
...
+
+/// Sends and receives AQC messages.
+pub(crate) struct AqcChannelsImpl {
+    // TODO: add Aqc fields.
+    handler: Handler<Store>,
+    eng: CE,
+}
+
+impl AqcChannelsImpl {
+    /// Creates a new `FastChannelsImpl` listening for connections on `address`.
```
nikki-aranya, line 44: VSCode Rename Symbol to QuicChannelsImpl, and fix the comment?

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,336 @@
...
+    pub(crate) async fn new(
+        device_id: DeviceId,
+        keystore_info: KeyStoreInfo,
+    ) -> Result<Self, AqcError> {
+        debug!("device ID: {:?}", device_id);
+        debug!("keystore path: {:?}", keystore_info.path);
+        debug!("keystore wrapped key path: {:?}", keystore_info.wrapped_key);
```
nikki-aranya, line 51: How is this handled elsewhere? Might want to make an #[instrument()] span

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,336 @@
...
+            CE::new(&key, Rng)
+        };
+
+        Ok(Self { handler, eng })
+    }
+}
+
+/// Aranya Fast Channels client that allows for opening and closing channels and
+/// sending data between peers.
+pub struct AqcChannels<'a> {
```
nikki-aranya, line 68: ditto, Rename Symbol to QuicChannels and fix the comment

crates/aranya-client/src/client.rs
```diff
@@ -102,14 +128,27 @@ impl Client {
...
+        let afc = FastChannelsImpl::new(afc::Client::new(afc_read), afc_address).await?;
         debug!(
             addr = ?afc.local_addr().map_err(Error::Afc)?,
             "bound AFC router",
         );
+        debug!("getting key store info");
+        let keystore_info = daemon.get_keystore_info(context::current()).await??;
+        debug!("getting device id");
+        let device_id = daemon.get_device_id(context::current()).await??;
+        let aqc = AqcChannelsImpl::new(device_id, keystore_info).await?;
```
jdygert-spok, line 141: We are effectively sharing the keystore and engine, is this OK? I guess I'm not entirely sure what our conversation ended up at. If we don't share it, does that work for aqc? Are there reasons we _can't_ share it in the future, e.g. hardware modules or elevated permissions, etc?

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,336 @@
...
+
+            // TODO: send ctrl msg via network.
+
+            // TODO: for testing only. Send ctrl via network instead of returning.
+            return Ok((v.channel_id.into_id().into(), aqc_ctrl));
+        }
+        Err(crate::Error::Aqc(AqcError::Other(anyhow!(
+            "unable to create bidi channel"
+        ))))
+    }
```
nikki-aranya, line 140: nit: This ending pattern is a bit ugly with return vs Err, could refactor to if let { Ok } else { Err }

crates/aranya-client/src/client.rs
gknopf-aranya: We are using the keystore as the IPC mechanism for passing the AQC PSK from the daemon to the client.
This was an alternative @elagergren-spideroak suggested so we're not passing the AQC PSK via an unsecured IPC mechanism such as UDS.

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,336 @@
...
+                        author_id: v.author_id.into_id().into(),
+                        author_enc_key_id: v.author_enc_key_id,
+                        send_id: v.send_id.into_id().into(),
+                        recv_id: v.recv_id.into_id().into(),
+                        peer_enc_pk: &v.peer_enc_pk,
+                        label_id: v.label_id.into_id().into(),
+                        channel_id: v.channel_id,
+                        psk_length_in_bytes: v.psk_length_in_bytes,
+                        author_secrets_id: v.author_secrets_id,
+                    },
```
nikki-aranya, line 188: nit: with how large the object is, I'd prefer it to be a separate let; above and passed in as a variable, `&effect`

crates/aranya-client/src/aqc.rs
gknopf-aranya: We decided to call it AQC everywhere. So this would change to:
`AqcChannelsImpl`

crates/aranya-client/src/client.rs
```diff
@@ -102,14 +128,27 @@ impl Client {
...
+        let afc_read = setup_afc_shm(afc_shm_path, max_channels)?;
+        let afc = FastChannelsImpl::new(afc::Client::new(afc_read), afc_address).await?;
         debug!(
             addr = ?afc.local_addr().map_err(Error::Afc)?,
             "bound AFC router",
         );
+        debug!("getting key store info");
+        let keystore_info = daemon.get_keystore_info(context::current()).await??;
+        debug!("getting device id");
+        let device_id = daemon.get_device_id(context::current()).await??;
```
nikki-aranya, line 140: These would be more useful if they actually printed out keystore_info/device_id, if the calls fail then we already get a nice error message in return

crates/aranya-daemon-api/src/service.rs
```diff
@@ -231,18 +368,46 @@ pub trait DaemonApi {
...
+    async fn create_aqc_uni_channel(
+        team: TeamId,
+        peer: NetIdentifier,
+        node_id: NodeId,
+        label_id: LabelId,
+    ) -> Result<(AqcCtrl, AqcChannelInfo)>;
+    /// Delete a QUIC bidi channel.
+    async fn delete_aqc_bidi_channel(chan: AqcBidiChannelId) -> Result<AqcCtrl>;
+    /// Delete a QUIC uni channel.
+    async fn delete_aqc_uni_channel(chan: AqcUniChannelId) -> Result<AqcCtrl>;
```
nikki-aranya, line 390: have we had a discussion over whether these should be delete or destroy? destroy would be the natural pairing to create

crates/aranya-daemon/src/api.rs
```diff
@@ -11,37 +11,41 @@ use std::{
...
 use aranya_daemon_api::{
-    AfcCtrl, AfcId, DaemonApi, DeviceId as ApiDeviceId, KeyBundle as ApiKeyBundle, NetIdentifier,
-    Result as ApiResult, Role as ApiRole, SyncPeerConfig, TeamId, CS,
+    AfcCtrl, AfcId, AqcBidiChannelCreatedInfo, AqcBidiChannelId as ApiAqcBidiChannelId,
+    AqcBidiChannelReceivedInfo, AqcChannelInfo, AqcCtrl, AqcUniChannelCreatedInfo,
+    AqcUniChannelId as ApiAqcUniChannelId, AqcUniChannelReceivedInfo, ChanOp as ApiChanOp,
+    DaemonApi, DeviceId as ApiDeviceId, KeyBundle as ApiKeyBundle, KeyStoreInfo, Label as ApiLabel,
+    LabelId as ApiLabelId, NetIdentifier, Result as ApiResult, Role as ApiRole, SyncPeerConfig,
+    TeamId, CS,
 };
```
nikki-aranya, line 22: with how bulky this is, I think you're better off doing what we do in aranya_client and doing `use aranya_daemon_api::{self as api, ...}; api::Result`

crates/aranya-client/src/aqc.rs
gknopf-aranya: This file is the bare minimum needed to get things compiling.
Ben will be completing the implementation of aqc.rs in a future PR.

crates/aranya-daemon/src/api.rs
```diff
@@ -101,18 +107,27 @@ impl DaemonApiServer {
...
+                ))),
+                aqc_peers: Arc::default(),
             },
         })
     }
 
     /// Create new RPC server.
     #[instrument(skip_all)]
     #[cfg(not(feature = "afc"))]
+    #[allow(clippy::too_many_arguments)]
```
nikki-aranya, line 125: Do we have an issue to make this (and any other functions that need the clippy exemption) take a config/builder? That's pretty much the only solution to "way too many args" is stuff them in a struct and optionally add a builder pattern

crates/aranya-daemon-api/src/service.rs
gknopf-aranya: I don't think we've had a discussion about that yet.
I'm open to changing this as we work on finalizing the API next sprint.
Not needed on this PR.

crates/aranya-daemon/src/api.rs
gknopf-aranya: I could use that `KeystoreInfo` type to combine `keystore_path` and `wrapped_key_path`

crates/aranya-daemon/src/api.rs
```diff
@@ -779,6 +825,193 @@ impl DaemonApi for DaemonApiHandler {
...
+            .client
+            .actions(&team.into_id().into())
+            .create_aqc_bidi_channel_off_graph(peer_id, label.into_id().into())
+            .await?;
+        let id = self.pk.ident_pk.id()?;
+
+        let Some(Effect::AqcBidiChannelCreated(e)) =
+            find_effect!(&effects, Effect::AqcBidiChannelCreated(e) if e.author_id == id.into())
+        else {
+            return Err(anyhow::anyhow!("unable to find AqcBidiChannelCreated effect").into());
```
nikki-aranya, line 857: nit: this is what anyhow's bail!() is for (I'd import `use anyhow::bail;`)

crates/aranya-client/src/client.rs
gknopf-aranya: In order to secure the keystore, I believe we could set file permissions in the OS to only allow the aranya client and daemon to use the keystore.

crates/aranya-daemon/src/api.rs
jdygert-spok: `bail!` is strictly `anyhow::Error`, it doesn't convert :(

crates/aranya-client/src/client.rs
gknopf-aranya: @elagergren-spideroak are keys stored in the keystore encrypted to prevent other unauthorized processes from reading them? Or is the only protection based on file system permissions?

elagergren-spideroak: > Are there reasons we _can't_ share it in the future, e.g. hardware modules or elevated permissions, etc?

Basically, yeah. I don't think we can assume that the client and daemon can share a keystore or a engine.

elagergren-spideroak: > Seems unfortunate that this limits how much we can protect the keys/engine.

What do you mean?

> But if you and Eric think this isn't an issue or can probably be worked through without fundamental changes then it can be resolved after merging this.

The correct fix (for MVP) is for the client and daemon to have a shared key in their respective keystores, then to use that to encrypt the PSK.

The correct fix (post MVP) is for the client and daemon to use a better IPC mechanism.

crates/aranya-daemon/src/aranya.rs
```diff
@@ -542,6 +542,56 @@ where
...
 
+    /// Create a label.
+    #[instrument(skip(self), fields(name = %name))]
+    fn create_label(&self, name: String) -> impl Future<Output = Result<Vec<Effect>>> + Send {
+        self.with_actor(move |actor| {
+            actor.create_label(name)?;
+            Ok(())
+        })
+        .in_current_span()
+    }
```
nikki-aranya, line 553: with how many functions there are with this pattern, it would be nice to macro-ize it, that'd have to be a new issue though

crates/aranya-daemon/src/policy.md
```diff
@@ -73,11 +85,17 @@ enum Role {
...
-    ReadWrite,
+    // The device can only receive data in channels with this
+    // label.
+    RecvOnly,
+    // The device can only send data in channels with this
+    // label.
+    SendOnly,
+    // The device can send and receive data in channels with this
+    // label.
+    SendRecv,
```
nikki-aranya, line 98: nit: these comments should be single-line

crates/aranya-daemon/src/api.rs
gknopf-aranya: That doesn't solve the `as` to make sure afc and aqc types don't collide.

crates/aranya-client/src/client.rs
elagergren-spideroak: > @elagergren-spideroak are keys stored in the keystore encrypted to prevent other unauthorized processes from reading them? Or is the only protection based on file system permissions?

So, to be clear: "keystore" means `KeyStore` the trait. Aranya just happens to use the filesystem based keystore. And yes, `KeyStore`s only store wrapped keys [docs]. But that's only as secure as the keys used to wrap those keys. `DefaultEngine` (the `Engine` impl Aranya currently uses) uses a user-provided key. IIRC, Aranya currently writes that key to disk with 0600 permissions, which is bleh.

[docs]: https://docs.rs/aranya-crypto/0.5.0/aranya_crypto/keystore/trait.KeyStore.html

crates/aranya-daemon/src/vm_policy.rs
```diff
@@ -19,11 +20,31 @@ use aranya_runtime::{
...
+    }
+}
+
+/// Display implementation for [`ChanOp`]
+impl fmt::Display for ChanOp {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "ChanOp::{:?}", self)
+    }
+}
+
```
nikki-aranya, line 47: nit: It'd be pulling in another dep but I think we could find something to do this more elegantly

crates/aranya-daemon/src/api.rs
elagergren-spideroak: Yeah, you'd use `api::Result` and `afc::Result` and `aqc::Result`.

crates/aranya-client/src/client.rs
gknopf-aranya: @elagergren-spideroak, @jdygert-spok, @djohnson-spideroak this this fix need to be implemented on this PR?
Or can we patch this later prior to the MVP release?

crates/aranya-daemon-api/src/service.rs
nikki-aranya: destroy would be my choice, unless we want add+delete like we do for create_team+add_team/delete_team but I don't see add_bidi_channel and I'm not sure how much sense it would make

crates/aranya-client/src/client.rs
jdygert-spok: > > Seems unfortunate that this limits how much we can protect the keys/engine.
>
> What do you mean?

Your fix suggestion helps, but just generally thinking that the client (which is a library being used who knows where) now has to be run in a process with sufficient OS permissions to access the keystore and engine, rather than only the daemon process, which are often run under their own user, etc.

gknopf-aranya: Configuring permissions for the keystore file is something we'd need to solve at the linux admin or install level.
I don't see how the Aranya client or daemon could configure the permissions themselves without knowing what user/group each process is running as and admin-level permissions to set the filesystem permissions.

elagergren-spideroak: We need to talk about the engine/keystore in abstract, so please let's not focus on file system permissions specifically here.

elagergren-spideroak: > Configuring permissions for the keystore file is something we'd need to solve at the linux admin or install level. I don't see how the Aranya client or daemon could configure the permissions themselves without knowing what user/group each process is running as and admin-level permissions to set the filesystem permissions.

We need to talk about the engine/keystore in abstract, so please let's not focus on file system permissions specifically here.

elagergren-spideroak: > > > Seems unfortunate that this limits how much we can protect the keys/engine.
> > 
> > 
> > What do you mean?
> 
> Your fix suggestion helps, but just generally thinking that the client (which is a library being used who knows where) now has to be run in a process with sufficient OS permissions to access the keystore and engine, rather than only the daemon process, which are often run under their own user, etc.

If a keystore impl is half decent then each process should only have access to a subset of all the keys stored in the keystore. This is how keystores work on Windows, macOS, iOS, Android, etc. But, yeah—ideally the client process wouldn't have access to anything. But until we switch over to an IPC we're comfortable with, we don't have much of a choice here.

We should make sure the keystore impl that's shared by the client and daemon in order to store the key used to encrypt PSKs sent over IPC is separate from the keystore used for everything else.

crates/aranya-client/src/client.rs
gknopf-aranya: I can print them out with `debug!` tracing

crates/aranya-daemon-api/src/service.rs
gknopf-aranya: Here's an issue for it:
https://github.com/aranya-project/aranya/issues/192

Small enough change it could be pulled into a different PR.

crates/aranya-daemon/src/aranya.rs
gknopf-aranya: The `self.with_actor()` is already shortening things from what it wouldn't been without it.

crates/aranya-daemon/src/vm_policy.rs
gknopf-aranya: If it pulls in another dep, let's handle that on a different PR.

crates/aranya-daemon/src/api.rs
gknopf-aranya: Was able to combine the keystore related info into a single type to remove the `too_many_arguments` from this function. It's still there for the corresponding afc method, but that doesn't matter since afc is a disabled/experimental feature right now.

gknopf-aranya:
@elagergren-spideroak, @jdygert-spok, @djohnson-spideroak looks like the last thing to resolve on this PR is how we're handling passing the PSK from the daemon to the client.

djohnson-spideroak (approved):

crates/aranya-client/src/aqc.rs
```diff
@@ -0,0 +1,336 @@
...
+                debug!("psk id: {:?}", psk.identity());
+            }
+            _ => {}
+        }
+
+        Ok(())
+    }
+}
+
+// TODO: borrowed from daemon.
```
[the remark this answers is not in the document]
gknopf-aranya: Yeah, it should probably be moved to a utility crate where the client and daemon can both use it.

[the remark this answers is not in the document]
gknopf-aranya: It's something I only want to do if we have time to complete it before the MVP release.
Otherwise, it could be refactored later. It doesn't impact the MVP API.

PR #139, opened 2025-03-20
`rustfmt` Comments + Style Edition

nikki-aranya:
This implements the couple changes we talked about on the Rust Call earlier. Namely, we enforce style_edition=2024 which includes a number of logic fixes, we wrap comments to 100 characters (65 for capi crates), and we format any code examples inside comments.

PR #148, opened 2025-03-25
Add Experimental Feature Flag

nikki-aranya:
This adds an "afc" feature flag to aranya, which is off by default. Additionally, this changes the C-API to use a Builder-style pattern for client configuration info. Other changes include moving "_afc_net_identifier" below "_label" due to labels being reused for AQC.


nikki-aranya:
Note: this cannot be merged until https://github.com/aranya-project/aranya-core/pull/180 gets merged and we update aranya-capi-codegen, but the changes are ready to review.

Makefile.toml
```diff
@@ -65,7 +65,7 @@ category = "build"
 description = "Build C API shared library"
 cwd = "crates/aranya-client-capi"
 command = "cargo"
-args = ["build", "--release"]
+args = ["build", "--release", "--features", "experimental"]
```
gknopf-aranya, line 68: @nikki-aranya, @djohnson-spideroak do we really want a single feature flag for all experimental features?
We would have more flexibility if each feature could be enabled independently.
E.g. `afc`, `aqc`, etc. features.

There's nothing stoping us from gating afc behind `experimental` and `afc`, but it's easier to enable it using a single feature rather than 2.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1065,6 +995,93 @@ AranyaError aranya_revoke_label_ext(struct AranyaClient *client,
                                     AranyaLabel label,
                                     struct AranyaExtError *__ext_err);
 
+#if defined(EXPERIMENTAL)
```
gknopf-aranya, line 998: @nikki-aranya, @jdygert-spok, @djohnson-spideroak I'd prefer to move all the AFC methods inside the same block. That way the logic is centralized in one spot and it's less likely to forget to add the gate to new methods that are added. Disadvantage of doing this is that it's harder to tell that methods are inside the same block if it gets large. Indentation will help a bit. Ideally, we could include a separate afc.h header file that's only included if the afc feature is enabled.
```
#if defined(EXPERIMENTAL)
...
#endif
```

Makefile.toml
jdygert-spok: I just worry about testing explosion

gknopf-aranya:
Some code in `example.c` should also be gated behind whether the afc feature is enabled.

gknopf-aranya:
Rust example template also needs `afc` feature so it can be run against Aranya whether afc is enabled/disabled.

Makefile.toml
gknopf-aranya: Executable size is another thing to consider. If we allow users to select the minimum number of features they need, we can optimize size for specific hardware or use-cases.

jdygert-spok:
> @nikki-aranya this security check is failing: https://github.com/aranya-project/aranya/actions/runs/14069268932/job/39399351522?pr=148

That's just from using the unreleased capi tooling changes.

Makefile.toml
gknopf-aranya, line 68: Have you tried building/testing without the flag to confirm there is no unused code or code that is gated behind the flag that's needed somewhere else? We should do this as a sanity check to confirm the correct code is behind the flag and that it exhaustively covers all code related to afc.

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: @jdygert-spok, @elagergren-spideroak would it be possible to have a separate afc.h header file? Not sure if the capi generator would support this since I've only seen a single `defs.rs` file so far.

crates/aranya-daemon/tests/tests.rs
```diff
@@ -137,7 +141,9 @@ async fn test_bidirectional_channel() -> Result<()> {
 
 #[test(tokio::test(flavor = "multi_thread"))]
 #[serial]
-async fn test_revoke_label() -> Result<()> {
+#[cfg(feature = "experimental")]
+//TODO(nikki): we should add a separate test for AQC once that's in
```
gknopf-aranya, line 145: We should mention relevant issue numbers in todos

crates/aranya-client-capi/output/aranya-client.h
nikki-aranya: > @nikki-aranya, @jdygert-spok, @djohnson-spideroak I'd prefer to move all the AFC methods inside the same block.

There's no way to do this unless we fork bindgen. If your Rust code is correct, your header is correct, that's what I'm going for.

crates/aranya-daemon/src/api.rs
```diff
@@ -567,6 +622,56 @@ impl DaemonApi for DaemonApiHandler {
         }
         Err(anyhow!("unable to find AfcBidiChannelReceived effect").into())
     }
+
+    #[cfg(not(feature = "experimental"))]
+    async fn assign_afc_net_identifier(
```
gknopf-aranya, line 627: Rather than returning an error if a method for a disabled feature is invoked, we should completely remove those methods from the API.

gknopf-aranya: Let's verify that the API has a reasonably clear error if a user attempts to invoke a method which does not exist and add it if that's not already there.

crates/aranya-daemon/tests/tests.rs
nikki-aranya: And what issue would that be? After having gone through #139 and co, the couple times we've left comments relating to specific issues they've basically just decayed like any other TODO cuz of how limited our time is. You'd be better off leaving a comment on whatever issue this pertains to than in the actual code.

crates/aranya-daemon/src/api.rs
nikki-aranya: If I do that then there's no way to control what error you get back from the RPC call

nikki-aranya: > Let's verify that the API has a reasonably clear error if a user attempts to invoke a method which does not exist and add it if that's not already there.

That would require entirely replacing tarpc which is on the TODO but isn't gonna get any attention until MVP is done.

crates/aranya-daemon/tests/tests.rs
gknopf-aranya: AQC implementation issue:
https://github.com/aranya-project/aranya-core/issues/101

nikki-aranya:
> Some code in `example.c` should also be gated behind whether the afc feature is enabled.

There's no really good way to do that. With the way things are now, you have to decide one way or the other, which is why I'm just leaving experimental enabled on the example

crates/aranya-daemon/src/api.rs
gknopf-aranya: What error does the RPC call return if the method doesn't exist? That may be good enough.
I believe we're adding our own tracing before invoking the daemon's API anyways so it should be clear which method is failing.

Makefile.toml
nikki-aranya: We basically have to decide whether or not to enable experimental unless we make multiple Makefile options which I'm really not a fan of. I'm trying to contain the configuration complexity

gknopf-aranya:
> There's no really good way to do that. With the way things are now, you have to decide one way or the other, which is why I'm just leaving experimental enabled on the example

My concern is that this PR is not testing the case where the experimental feature is disabled. If we're going to make the feature configurable, we should test that things work whether it's enabled or disabled. Otherwise, I'd prefer to not have the feature flag at all and have better test coverage.

crates/aranya-daemon/src/api.rs
nikki-aranya: It has to exist because it exists as a single trait, you can't skip certain methods on a trait.

jdygert-spok: We _could_ cfg the trait definition in aranya-daemon-api

gknopf-aranya: @nikki-aranya, @jdygert-spok is it possible modify the API trait by hiding methods within the trait behind the feature flag? If not, can we conditionally specify which trait to use based on the feature flag? It seems like there should be a way to do this. We could make traits composable by including the common methods in one trait and adding the extra ones in an afc trait that's only included based on the feature flag.

gknopf-aranya: > We _could_ cfg the trait definition in aranya-daemon-api

Yes, that's my preference. From a security standpoint, it makes sense to reduce the API surface if a feature is disabled.

nikki-aranya:
Note to self: main blockers are replacing config with a builder pattern to work around capi/expand.rs limitations and figuring out how (ta)rpc behaves when an invalid RPC call is made.

nikki-aranya:
Not sure what I'm missing here

```
2025-03-27T18:31:25.264842Z ERROR aranya_capi_core::internal::error: err=bug: Tried to create an AfcConfig without a valid shm_path!
location: crates/aranya-client-capi/src/imp/config.rs:146:13

2025-03-27T18:31:25.264894Z DEBUG aranya_client_capi::api::defs: err=Bug(Bug(BugInner { msg: "Tried to create an AfcConfig without a valid shm_path!", location: Location { file: "crates/aranya-client-capi/src/imp/config.rs", line: 146, col: 13 } }))
2025-03-27T18:31:25.264939Z ERROR aranya_capi_core::internal::error: err=bug: Tried to create a ClientConfig without a valid address!
location: crates/aranya-client-capi/src/imp/config.rs:72:13

2025-03-27T18:31:25.264941Z DEBUG aranya_client_capi::api::defs: err=Bug(Bug(BugInner { msg: "Tried to create a ClientConfig without a valid address!", location: Location { file: "crates/aranya-client-capi/src/imp/config.rs", line: 72, col: 13 } }))
AddressSanitizer:DEADLYSIGNAL
```

gknopf-aranya:
> Not sure what I'm missing here
> 
> 2025-03-27T18:31:25.264842Z ERROR aranya_capi_core::internal::error: err=bug: Tried to create an AfcConfig without a valid shm_path! location: crates/aranya-client-capi/src/imp/config.rs:146:13

What shm_path is it trying to use?

nikki-aranya:
I figured it out, builder pattern doesn't really work the way I was trying to make it work with taking `mut self`.

gknopf-aranya:
@nikki-aranya could you update the PR description to match what's currently implemented?

nikki-aranya:
Alright, done

examples/c/example.c
```diff
@@ -116,30 +118,70 @@ typedef struct {
     };
 } Team;
 
-AranyaError init_client(Client *c, const char *name, const char *daemon_sock,
+#if defined(ENABLE_AFC)
```
gknopf-aranya, line 121: The way this code is checking `ENABLE_AFC` in multiple places and implementing different versions of functions makes the program much more complicated.

I think we should check `ENABLE_AFC` once and use it to set a boolean in the `Team` object which is already passed around everywhere. Then check whether that boolean is set in the corresponding places in the code.

nikki-aranya: that doesn't fix the problem of needing variable amounts of arguments sadly, I need to gate out e.g. the fprintf at the very least because it tries to read from a method that doesn't exist which is a compile error, I can't just if {} else {} it

gknopf-aranya (dismissed):
lgtm

crates/aranya-client/src/client.rs
```diff
@@ -74,6 +80,23 @@ impl Client {
...
+
+        let transport = tarpc::serde_transport::unix::connect(daemon_socket, Json::default)
+            .await
+            .map_err(Error::Connecting)?;
+        let daemon = DaemonApiClient::new(tarpc::client::Config::default(), transport).spawn();
+        debug!("connected to daemon");
+
+        Ok(Self { daemon })
+    }
+
```
djohnson-spideroak, line 99: why is this only enabled when not afc?

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -492,6 +520,163 @@ AranyaError aranya_client_cleanup(struct AranyaClient *ptr);
...
+/**
+ * Sets the maximum number of channels that are stored in shared memory.
+ *
+ * @param cfg a pointer to the afc config builder
+ * @param max_channels the maximum number of channels allowed
+ */
+AranyaError aranya_afc_config_builder_set_max_channels_ext(struct AranyaAfcConfigBuilder *cfg,
+                                                           size_t max_channels,
+                                                           struct AranyaExtError *__ext_err);
+#endif
```
djohnson-spideroak, line 567: is there a reason to not use a single large block of if defined?

nikki-aranya: Because that's not how cbindgen works, I don't control that

crates/aranya-client/src/client.rs
nikki-aranya: Because above it is the function that's only enabled when afc

examples/c/run.bash
```diff
@@ -41,7 +41,7 @@ trap 'trap - SIGTERM && cleanup && kill -- -$$ || true' SIGINT SIGTERM EXIT
 rm -rf "${out}"
 
 # build the daemon.
-cargo build --bin aranya-daemon --release
+cargo build --bin aranya-daemon --release --features afc
```
gknopf-aranya, line 44: We should test this code with the afc feature enabled and disabled.
One possible way to do this would be to run the tests back-to-back like this:
run.bash:
```
AFC=0 ./test.bash
AFC=1 ./test.bash
```
`test.bash` would be similar to what `run.bash` is now but with an environment variable to configure how to build Aranya and the example code so they have compatible feature sets.

gknopf-aranya: @nikki-aranya I've opened an issue to tackle this on another PR:
https://github.com/aranya-project/aranya/issues/156

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: Yes, I had made a similar comment previously. Then again, if the block is too large, it's difficult to tell that you're inside the block. So this isn't that bad.

djohnson-spideroak (dismissed):

examples/rust/Cargo.toml
djohnson-spideroak, line 1: Probably best to use cratesio versions. if not it will not be standalone.

crates/aranya-client/src/client.rs
```diff
@@ -370,4 +331,117 @@ impl Team<'_> {
...
+    /// Returns the AFC network identifier assigned to the current device.
+    pub async fn afc_net_identifier(&mut self, device: DeviceId) -> Result<Option<NetIdentifier>> {
+        self.client
+            .daemon
+            .query_afc_net_identifier(context::current(), self.id, device)
+            .await?
+            .map_err(Into::into)
+    }
```
djohnson-spideroak, line 428: should this be feature gated?

examples/rust/Cargo.toml
gknopf-aranya: We are using the version in the checked out workspace to allow us to test the Rust example in CI on the current branch.
This will also allow us to keep the example up-to-date on each PR rather than breaking it and periodically updating it to the latest version.

crates/aranya-client/src/client.rs
gknopf-aranya: Yes, good catch

nikki-aranya:
Latest commits seem fine at a glance, I'll do a full pass of the diff in the morning.

djohnson-spideroak (approved):

gknopf-aranya (approved):

crates/aranya-client/tests/tests.rs
```diff
@@ -8,36 +8,55 @@
...
+#[cfg(feature = "afc")]
 use tempfile::tempdir;
+#[cfg(feature = "afc")]
 use test_log::test;
+#[cfg(feature = "afc")]
 use tokio::{
     fs,
     task::{self, AbortHandle},
     time::{self, Sleep},
 };
```
nikki-aranya, line 54: We'll have to fix this whole mess once we have some more tests added but I guess for now it's fine, nothing in this file is used outside afc right now

nikki-aranya (commented):
Alright, lgtm

crates/aranya-client/tests/tests.rs
gknopf-aranya: We'll add some tests for AQC here in the next sprint

PR #155, opened 2025-03-27
policy: make labels strings, not integers

### LINKED ISSUE #152: policy: map string labels to integer

elagergren-spideroak:
To keep compatibility with AFC, map labels to 32-bit integers.

Fixes #152

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -311,6 +311,7 @@ dependencies = [
  "aranya-crypto",
  "aranya-fast-channels",
  "aranya-util",
+ "buggy",
  "postcard",
  "serde",
  "spideroak-base58",
--- a/crates/aranya-client/src/afc.rs
+++ b/crates/aranya-client/src/afc.rs
@@ -30,10 +30,9 @@ use anyhow::anyhow;
 use aranya_crypto::{Random as _, Rng};
 pub use aranya_daemon_api::AfcId;
 use aranya_daemon_api::{AfcCtrl, NetIdentifier, TeamId, CS};
-pub use aranya_fast_channels::Label;
 use aranya_fast_channels::{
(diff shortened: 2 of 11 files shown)


elagergren-spideroak:
NB: I need to update `aranya-afc-util` for this to work.

jdygert-spok:
Isn't a counter an issue if two clients make a label then braid?

elagergren-spideroak:
> Isn't a counter an issue if two clients make a label then braid?

ugh, yeah, i guess we could make the caller provide a string _and_ an integer :/

elagergren-spideroak:
> > Isn't a counter an issue if two clients make a label then braid?
> 
> ugh, yeah, i guess we could make the caller provide a string _and_ an integer :/

...and somehow require both be unique :/

elagergren-spideroak:
> > > Isn't a counter an issue if two clients make a label then braid?
> > 
> > 
> > ugh, yeah, i guess we could make the caller provide a string _and_ an integer :/
> 
> ...and somehow require both be unique :/

Well, I mean, it's actually not _that_ bad. If `DefineLabel` asserts that the label ID is unique then it's no different from two branches that choose the same label (string).

elagergren-spideroak:
> > > > Isn't a counter an issue if two clients make a label then braid?
> > > 
> > > 
> > > ugh, yeah, i guess we could make the caller provide a string _and_ an integer :/
> > 
> > 
> > ...and somehow require both be unique :/
> 
> Well, I mean, it's actually not _that_ bad. If `DefineLabel` asserts that the label ID is unique then it's no different from two branches that choose the same label (string).

...except that it's easier to screw up.

gknopf-aranya:
@elagergren-spideroak given that we're planning to implement ID-based labels, should this PR be closed?

PR #158, opened 2025-03-28
docs: keygen rustdocs

djohnson-spideroak:
Initial draft of keygen rustdocs.

Needs: probably ignore the examples so they don't get run in tests, general wording updates.


crates/aranya-keygen/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+- Generation of secure cryptographic key bundles
+- Management of identity, encryption, and signing keys
+- Utilities for loading key bundles from storage
+
+## Overview
+
+The `aranya-keygen` crate simplifies the process of generating and managing cryptographic keys for Aranya applications. It provides a unified interface to create key bundles containing:
+
+- Identity keys (for device identification)
```
gknopf-aranya, line 13:
```suggestion
- Identity keys (for uniquely identifying devices)
```

crates/aranya-keygen/README.md
```diff
@@ -0,0 +1,44 @@
...
+    // Use the public keys as needed
+    // ...
+    
+    Ok(())
+}
+```
+
+## License
+
+This crate is licensed under the same terms as the Aranya project.
```
gknopf-aranya, line 44: Should we link to the license or github repo here so it's easier to look up the license?

crates/aranya-keygen/src/keygen.rs
```diff
@@ -6,32 +6,120 @@ use aranya_crypto::{
 };
 use serde::{Deserialize, Serialize};
 
-/// A key bundle.
+/// A bundle of cryptographic keys for secure communication in Aranya.
+///
+/// A KeyBundle contains identifiers for three types of keys:
+/// - Device identity key: Used to identify a device in the Aranya network
```
gknopf-aranya, line 12:
```suggestion
/// - Device identity key: Used to identify a device in an Aranya team
```

crates/aranya-keygen/src/keygen.rs
```diff
@@ -6,32 +6,120 @@ use aranya_crypto::{
...
+    /// Device identifier derived from the identity key.
+    ///
+    /// This ID uniquely identifies the device in the Aranya network.
+    ///
     /// See [`IdentityKey`].
     pub device_id: DeviceId,
+
+    /// Identifier for the encryption key.
+    ///
+    /// The encryption key is used for secure data encryption and decryption.
```
gknopf-aranya, line 47: Private key would encrypt. Public key would decrypt.

crates/aranya-keygen/src/keygen.rs
```diff
@@ -6,32 +6,120 @@ use aranya_crypto::{
...
 }
 
-/// Public keys from key bundle.
+/// Collection of public keys derived from a [`KeyBundle`].
+///
+/// This structure contains the public portions of the keys referenced in a KeyBundle.
+/// These public keys can be shared with other devices for secure communication.
+///
+/// - `ident_pk`: Public identity key for device identification
+/// - `enc_pk`: Public encryption key for encrypting messages to this device
```
gknopf-aranya, line 67: Public key is for decrypting

crates/aranya-keygen/src/keygen.rs
```diff
@@ -59,7 +147,43 @@ impl KeyBundle {
         })
     }
 
-    /// Loads the public keys from `store`.
+    /// Retrieves the public keys associated with this key bundle.
```
gknopf-aranya, line 150: "Loads" was a better word since it implies loading from storage as opposed to retrieving a value from memory.

crates/aranya-keygen/src/keygen.rs
```diff
@@ -6,32 +6,121 @@ use aranya_crypto::{
...
     pub sign_id: SigningKeyId,
 }
 
-/// Public keys from key bundle.
+/// Collection of public keys derived from a [`KeyBundle`].
+///
+/// This structure contains the public portions of the keys referenced in a KeyBundle.
+/// These public keys can be shared with other devices for secure communication.
+///
+/// - `ident_pk`: Public identity key for device identification
```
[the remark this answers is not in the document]
djohnson-spideroak: I think it is useful to have this information in the struct doc and member docs.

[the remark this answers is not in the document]
gknopf-aranya: True, although the amount of detail in each place doesn't need to be the same.
Once a key has been defined in detail, the rest of the docs can reference the original definition without needing to fully define them each time.

crates/aranya-keygen/src/keygen.rs
```diff
@@ -6,32 +6,121 @@ use aranya_crypto::{
 };
 use serde::{Deserialize, Serialize};
 
-/// A key bundle.
+/// A bundle of cryptographic keys for secure communication in Aranya.
+///
+/// A KeyBundle contains identifiers for three types of keys:
+/// - Device identity key: Used to identify a device in an Aranya team
```
gknopf-aranya, line 12: @djohnson-spideroak, @elagergren-spideroak do the keys perform encryption and decryption, signing and verification?
Or do the public and private components of the key pairs perform different operations?
This is not clear in the docs right now.

elagergren-spideroak: All three device keys are key pairs (i.e., they have a secret (private) half and a public half).

The secret half of the device identity key is a signing key: it creates digital signatures. The public half of the device identity key is a verifying key: it verifies signatures created by the secret half.

The secret half of the device encryption key is a decapsulation key: it decapsulates encapsulated KEM shared secrets. The public half of the device encryption key is an encapsulation key: it encapsulates KEM shared secrets.

The secret half of the device signing key is a signing key: it creates digital signatures. The public half of the device signing key is a verifying key: it verifies signatures created by the secret half.

---

The device identity key _is_ a device. It cannot be rotated or changed. We derive the device's ID from the public half of the device identity key. It is only used to sign Very Important administrative actions, like when the device rotates its encryption key or its signing key.

The device encryption key is used for key encapsulation. For example, AQC and AFC use it to securely send channel secrets to other devices. It _can_ be rotated, although we haven't yet built that functionality.

The device signing key is the most commonly used key. It signs each and every command published by a device. It _can_ be rotated, although we haven't yet built that functionality.

---

Note that we have plans to add something like a "RootIdentitySigningKey" so that one physical user can have N devices associated with them.

I also would like to split the device encryption key into separate keys for AFC and AQC. And I'd also like to scope the device signing key to only being for commands.

/cc: @mooreoak @ycarmy

crates/aranya-keygen/src/keygen.rs
```diff
@@ -6,32 +6,121 @@ use aranya_crypto::{
...
+    /// The encryption key is a public key used to encrypt
+    /// data to this device.
+    ///
     /// See [`EncryptionKey`].
     pub enc_id: EncryptionKeyId,
+
+    /// Identifier for the signing key.
+    ///
+    /// The signing key is used for creating cryptographic signatures that can be
+    /// verified using the corresponding public key.
```
ssweetney-spok, line 56: Seems like this could be clarified a bit more based on Eric's description above

gknopf-aranya:
@jdygert-spok, @elagergren-spideroak this just needs one more approval

gknopf-aranya:
@nikki-aranya, @ssweetney-spok could you please review this PR?

ssweetney-spok (approved):

nikki-aranya (approved):

PR #171, opened 2025-04-03
Persist Net Identifiers Across Daemon Reboots

nikki-aranya:
Updates the Daemon API Server to query the fact database and restore all previous NetIdentifier mappings when a daemon is rebooted.

### DIFF
--- a/crates/aranya-daemon/src/api.rs
+++ b/crates/aranya-daemon/src/api.rs
@@ -105,6 +105,7 @@ impl DaemonApiServer {
                 .collect::<Vec<_>>()
         };
 
+        #[allow(unused_mut)]
         let mut afc_peers = BTreeMap::new();
         #[cfg(any())]
         for graph_id in &graph_ids {


nikki-aranya:
I still need to add a test to make sure this actually works and then it should be ready to merge.

djohnson-spideroak:
Did you look into adding a new policy action/command to do this in a single query? I think the current approach of doing N+1 queries might be a bit expensive.

crates/aranya-client/tests/tests.rs
```diff
@@ -1088,3 +1091,186 @@ async fn test_afc_monotonic_seq() -> Result<()> {
 
     Ok(())
 }
+
+/// This tests a bug where if the daemon is killed after registering net identifiers, they're not
+/// reloaded upon reboot, which causes create_bidi_channel to fail with "unable to lookup peer".
```
gknopf-aranya, line 1096:
```suggestion
/// This test verifies that the daemon reloads network identifiers on restart.
```

crates/aranya-client/tests/tests.rs
```diff
@@ -1088,3 +1091,186 @@ async fn test_afc_monotonic_seq() -> Result<()> {
...
+#[cfg(feature = "afc")]
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_afc_persist_net_identifier() -> Result<()> {
+    let interval = Duration::from_millis(100);
+    let sync_config = SyncPeerConfig::builder().interval(interval).build()?;
+    let sleep_interval = interval * 6;
+
+    let tmp = tempdir()?;
+    let work_dir = tmp.path().to_path_buf();
```
gknopf-aranya, line 1105: nit: putting the boilerplate test code first makes tests a little more readable. 
```suggestion
    let tmp = tempdir()?;
    let work_dir = tmp.path().to_path_buf();

    let interval = Duration::from_millis(100);
    let sync_config = SyncPeerConfig::builder().interval(interval).build()?;
    let sleep_interval = interval * 6;
```
We aren't doing this in the other tests yet so feel free to ignore this. This could be cleaned up on a different PR.

crates/aranya-client/tests/tests.rs
```diff
@@ -1088,3 +1091,186 @@ async fn test_afc_monotonic_seq() -> Result<()> {
...
+
+    // create team.
+    let team_id = team
+        .owner
+        .client
+        .create_team()
+        .await
+        .expect("expected to create team");
+    info!(?team_id);
+    // TODO: implement add_team.
```
gknopf-aranya, line 1117: We don't need to copy this todo comment into all the tests. It just needs to be in one spot as a reminder.

crates/aranya-client/tests/tests.rs
```diff
@@ -1088,3 +1091,186 @@ async fn test_afc_monotonic_seq() -> Result<()> {
...
+    // Create a new channel, which should fail to lookup peer if we haven't re-populated afc_peers.
+    let _ = team
+        .membera
+        .client
+        .afc()
+        .create_bidi_channel(team_id, NetIdentifier(memberb_afc_addr.to_string()), label)
+        .await?;
+
+    // wait for ctrl message to be sent.
+    sleep(Duration::from_millis(100)).await;
```
gknopf-aranya, line 1273: I don't see the point of waiting for the peer to receive the ctrl message unless we're planning to test that data can be sent via the channel. We should send data over the channel to demonstrate that channel creation is not affected by a daemon restart (this is part of what proves the net identifiers were reloaded correctly).

crates/aranya-daemon/src/policy.rs
```diff
@@ -269,6 +276,7 @@ pub trait ActorExt {
         device_id: Id,
         net_identifier: String,
     ) -> Result<(), ClientError>;
+    fn query_afc_network_names(&mut self) -> Result<(), ClientError>;
```
gknopf-aranya, line 279: @nikki-aranya, @djohnson-spideroak do we want to add this query to the Rust/C application APIs?

crates/aranya-daemon/src/policy.md
```diff
@@ -1189,6 +1189,35 @@ command SetAfcNetworkName {
         }
     }
 }
+
+action query_afc_network_names() {
```
gknopf-aranya, line 1193: Let's add some docs for this new query

crates/aranya-daemon/src/aranya.rs
```diff
@@ -738,6 +743,36 @@ where
         })
         .in_current_span()
     }
+
+    /// Query assigned network identifiers.
+    #[allow(clippy::type_complexity)]
+    #[instrument(skip_all)]
+    #[cfg(feature = "afc")]
+    fn query_afc_network_names(
```
gknopf-aranya, line 751: We are using the `off_graph` naming convention for ephemeral session commands:
```suggestion
    fn query_afc_network_names_off_graph(
```

crates/aranya-daemon/src/api.rs
```diff
@@ -679,6 +744,11 @@ impl DaemonApi for DaemonApiHandler {
         node_id: NodeId,
         ctrl: AfcCtrl,
     ) -> ApiResult<(AfcId, NetIdentifier, Label)> {
+        if self.afc_peers_dirty.load(Ordering::Relaxed) {
```
gknopf-aranya, line 747: This code is sort of tested by waiting to receive the control message. That executes this code but doesn't prove it works as expected.
It would be better to verify the peer that receives the ctrl message sets up the AFC channel and can receive data.

crates/aranya-client/tests/tests.rs
nikki-aranya: I'd rather keep the historical reason why the test exists. If there wasn't an issue to test, there wouldn't be a test, and none of these are public-facing, they're only for us

crates/aranya-daemon/src/api.rs
```diff
@@ -636,6 +696,11 @@ impl DaemonApi for DaemonApiHandler {
     ) -> ApiResult<(AfcId, AfcCtrl)> {
         info!("create_afc_bidi_channel");
 
+        if self.afc_peers_dirty.load(Ordering::Relaxed) {
```
gknopf-aranya, line 699: The current implementation is reloading afc net identifiers whenever a new channel is created or a channel is received.
I'd rather load them when `DaemonApiServer` is initialized. Then there's no need to track if the list is out-of-date with a boolean.
If there is a reason to load them later (e.g. wanting to load them lazily for performance reasons), we could use an `Optional<>` field to track whether the value is set rather than tracking this with a separate boolean.

nikki-aranya: I'm unable to know the GraphId when DaemonApiServer is initialized to be able to run queries, that's why this fix took a day longer than expected

crates/aranya-client/tests/tests.rs
gknopf-aranya: We are not testing a bug. We are testing that a specific features works and that there are no regressions.
If keeping the historical reason is important, we could reference the issue number here that resolved the bug / added the feature.

crates/aranya-daemon/src/api.rs
nikki-aranya: Additionally, the user could create a new graph so making afc_peers an option doesn't really solve anything

gknopf-aranya: We could reload the net identifiers in `add_team()`. Then `add_team()` would have something there for the implementation. That seems like the first opportunity to know what the graph ID is.

nikki-aranya: add_team is also unimplemented and uncalled

gknopf-aranya: "Additionally, the user could create a new graph so making afc_peers an option doesn't really solve anything"
If the user creates a new graph, there are no net identifiers to reload for that graph. Reloading net identifiers only applies to existing graphs.

gknopf-aranya: "add_team is also unimplemented and uncalled"
I'm suggesting this could be a good reason to implement it and invoke it from tests/examples.

gknopf-aranya: "I'm unable to know the GraphId when DaemonApiServer is initialized to be able to run queries"
@nikki-aranya, @djohnson-spideroak, @jdygert-spok this could be a reason to make the team/graph IDs persistent.

crates/aranya-daemon/src/api.rs
```diff
@@ -106,6 +108,7 @@ impl DaemonApiServer {
                 pk,
                 peers,
                 afc_peers: Arc::default(),
+                afc_peers_dirty: Arc::new(true.into()),
```
gknopf-aranya, line 111: @nikki-aranya this won't work if there is more than 1 existing graph.
The first time a channel create/receive operation occurs for one graph, this flag would be cleared.
If another graph ID is used later, the net identifiers for that graph wouldn't be reloaded.

We would need to keep track of a flag for each graph ID for this to work.
Or make graph IDs persistent so we can load all the net identifiers up-front.

crates/aranya-client/tests/tests.rs
djohnson-spideroak: Given its a test, I think its fine to have both comments there. Historical reason, issue number, description of what it should verify.

crates/aranya-daemon/src/policy.rs
djohnson-spideroak: Yes, in a different PR as lower priority than MVP tasks.

crates/aranya-daemon/src/api.rs
djohnson-spideroak: I think its fine to lazy load if they are uninitialized when we get a relevant API call or effect. I do think though that eventually we are going to want an easier way to get a list of teams this device/daemon is involved in, and having the team IDs we plan to participate on on startup would make it easier to do a lot of this stuff. Lets look into that in a different issue/PR.

crates/aranya-daemon/src/policy.rs
gknopf-aranya: Here's an issue for it:
https://github.com/aranya-project/aranya/issues/174

crates/aranya-client/tests/tests.rs
```diff
@@ -1088,3 +1091,186 @@ async fn test_afc_monotonic_seq() -> Result<()> {
...
+    let memberb_afc_addr = team.memberb.afc_local_addr().await?;
+
+    // setup sync peers.
+    let mut owner_team = team.owner.client.team(team_id);
+    let mut admin_team = team.admin.client.team(team_id);
+    let mut operator_team = team.operator.client.team(team_id);
+    let mut membera_team = team.membera.client.team(team_id);
+    let mut memberb_team = team.memberb.client.team(team_id);
+
+    owner_team
```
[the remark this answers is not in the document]
nikki-aranya: #180

gknopf-aranya:
Since AFC is being removed from the MVP release, this change will be more valuable to have for AQC network identifiers.

However, it won't be possible to implement this change for AQC network identifiers until after this PR is merged (unless it's added to that PR).
https://github.com/aranya-project/aranya/pull/145
That's already a huge PR, so I'm not planning to increase scope any more.

Here's another issue for AQC net identifiers:
https://github.com/aranya-project/aranya/issues/184

Cargo.toml
```diff
@@ -35,23 +35,23 @@ unwrap_used = "warn"
...
-aranya-idam-ffi = { version = "0.5.0" }
-aranya-libc = { version = "0.1.1" }
-aranya-perspective-ffi = { version = "0.5.0" }
-aranya-policy-compiler = { version = "0.5.0" }
-aranya-policy-ifgen = { version = "0.5.0", features = ["serde"] }
-aranya-policy-ifgen-build = { version = "0.2.0" }
-aranya-policy-lang = { version = "0.2.0" }
-aranya-policy-vm = { version = "0.5.0", features = ["std"] }
-aranya-runtime = { version = "0.5.0", features = ["std", "libc"] }
+aranya-afc-util = { version = "0.6.0", git = "https://github.com/aranya-project/aranya-core", rev = "1ee86eada84042d329cb3c75f0d8c87e55b06803", features = ["alloc"] }
```
gknopf-aranya, line 38: todo: bump versions once these are released

crates/aranya-daemon/src/api.rs
```diff
@@ -91,6 +92,21 @@ impl DaemonApiServer {
     ) -> Result<Self> {
         info!("uds path: {:?}", daemon_sock);
         let device_id = pk.ident_pk.id()?;
+
+        let afc_peers = BiBTreeMap::new();
+        let mut aranya = client.aranya.lock().await;
+        #[allow(unused_variables)]
+        if let Some(graph_id) = aranya.provider().list_graph_ids()?.next() {
```
jdygert-spok, line 99: This should loop over all the graphs. But #200 is needed for correctness.

nikki-aranya: I specifically made it only take one graph, since multi-graph is a much larger fix than this, this PR was *supposed* to just be a simple bugfix but ballooned in the name of optimization :\ there's a reason I'm not holding this up for AQC-specific changes either

gknopf-aranya: That's valid. I'm hoping to have you working on something else soon too. Sorry this ended up being more work than anticipated.

nikki-aranya: I realized that even if we're only handling one graph, we still have to loop them all in case of a stale working directory so I probably should change this regardless, gonna have to port over #200's BTree<BiBTree<>> to do this correctly but eh

crates/aranya-daemon/src/api.rs
```diff
@@ -302,6 +363,7 @@ impl DaemonApiHandler {
                 Effect::QueryAqcNetIdentifierResult(_) => {}
                 Effect::QueriedLabelAssignment(_) => {}
                 Effect::QueryLabelExistsResult(_) => {}
+                Effect::QueryAqcNetworkNamesOutput(_) => {}
```
gknopf-aranya, line 366: This is closer to the naming convention we already have:
```suggestion
                Effect::QueryAqcNetworkNameResult(_) => {}
```
Only a single net identifier can be returned per effect.

nikki-aranya: Saving it for that separate PR we discussed, I mostly care about the functionality in this

crates/aranya-daemon/src/api.rs
```diff
@@ -93,6 +93,43 @@ impl DaemonApiServer {
     ) -> Result<Self> {
         info!("uds path: {:?}", daemon_sock);
         let device_id = pk.ident_pk.id()?;
+
+        let graph_ids = {
```
gknopf-aranya, line 97: Let's add a todo here to use the ideal approach once the `Send` fix is implemented in `aranya-core`

crates/aranya-daemon/src/api.rs
```diff
@@ -93,6 +93,43 @@ impl DaemonApiServer {
...
+                .aranya
+                .lock()
+                .await
+                .provider()
+                .list_graph_ids()?
+                .flatten()
+                .collect::<Vec<_>>()
+        };
+
+        let mut afc_peers = BTreeMap::new();
```
gknopf-aranya, line 108: This should be hidden unless the afc feature flag is enabled.

crates/aranya-daemon/src/api.rs
nikki-aranya: We've decided we're not implementing Send unless we absolutely have to, this honestly probably works better anyways since I can re-use the Vec twice

crates/aranya-daemon/src/api.rs
nikki-aranya: This is basically the same as Arc::default(), it won't get populated since the extending is #[cfg(any())]

crates/aranya-daemon/src/policy.md
```diff
@@ -2155,3 +2155,45 @@ command QueryAqcNetIdentifier {
...
+        }
+    }
+}
+
+effect QueryAqcNetworkNamesOutput {
+    net_identifier string,
+    device_id id,
+}
+
+command QueryAqcNetworkNamesCommand {
```
gknopf-aranya, line 2178: The other commands don't have `Command` in their names
```suggestion
command QueryAqcNetworkNames {
```

gknopf-aranya (approved):
lgtm other than a couple minor variable name nits

nikki-aranya:
> lgtm other than a couple minor variable name nits

https://github.com/aranya-project/aranya/issues/218

gknopf-aranya (approved):

djohnson-spideroak (approved):

PR #175, opened 2025-04-07
capi: device role query

gknopf-aranya:
Add API for querying device role.
Include method for converting the device role to a C string for display.

Blocked by ID-based roles:
https://github.com/aranya-project/aranya-docs/issues/91
https://github.com/aranya-project/aranya/issues/181

### DIFF
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -91,7 +91,7 @@
 #define ARANYA_DEVICE_ID_STR_LEN (((64 * 1375) / 1000) + 1)
 
 /**
- * The size in bytes of a [`Role`] converted to a human-readable string.
+ * The size in bytes of a `Role` converted to a human-readable string.
  */
 #define ARANYA_ROLE_STR_LEN 256
 
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -91,6 +91,11 @@
 #define ARANYA_DEVICE_ID_STR_LEN (((64 * 1375) / 1000) + 1)
 
 /**
+ * The size in bytes of a `DeviceId` converted to a human-readable base64 string.
+ */
+#define ARANYA_ROLE_STR_LEN 256
+
+/**
  * An error code.
  *
  * For extended error information, see [`AranyaExtError`](@ref AranyaExtError).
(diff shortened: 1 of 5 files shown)


gknopf-aranya:
Opened a PR here to release the new capi-codegen crate to unblock this PR:
https://github.com/aranya-project/aranya-core/pull/204

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1233,7 +1251,47 @@ pub fn device_id_to_str(
...
+) -> Result<Role, imp::Error> {
+    let client = client.deref_mut();
+    let role = client
+        .rt
+        .block_on(client.inner.queries(team.0).device_role(device.0))?;
+    Ok(role.into())
+}
+
+/// The size in bytes of a `DeviceId` converted to a human-readable base64 string.
+pub const ARANYA_ROLE_STR_LEN: u64 = 256;
```
ssweetney-spok, line 1275: I don't think this comment aligns with the usage of this variable? Also, 256 bytes is excessive. We should be fine going with the "string size" of the largest variant

gknopf-aranya: We'll want some margin in case the variants change.
In the future, custom role support will be added, so we'll need more room for whatever custom role strings users may add.

gknopf-aranya: Because this constant is part of the MVP API we should use a larger size than we need right now.
Otherwise, increasing this constant would change the MVP API later.

gknopf-aranya: Updated the comment.

ssweetney-spok: > In the future, custom role support will be added, so we'll need more room for whatever custom role strings users may add. Because this constant is part of the MVP API we should use a larger size than we need right now.
Otherwise, increasing this constant would change the MVP API later.

Assuming we enforce a character limit, wouldn't we want to change this value to match whatever character limit we set for custom roles?

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1765,6 +1770,49 @@ AranyaError aranya_device_id_to_str(struct AranyaDeviceId device,
...
+ * Writes the human-readable encoding of `role` to `str`.
+ *
+ * To always succeed, `str` must be at least `ARANYA_ROLE_STR_LEN` bytes long.
+ *
+ * @param role [`AranyaRole`](@ref AranyaRole).
+ * @param role string [`AranyaRole`](@ref AranyaRole).
+ *
+ * @relates AranyaError.
+ */
+AranyaError aranya_role_to_str(AranyaRole role, char *str, size_t *str_len);
```
djohnson-spideroak, line 1814: do we have any reason to implement string->role as well?

djohnson-spideroak (commented):
Please add some tests that check the returned value

elagergren-spideroak (changes requested):
We need at least some sort of basic spec for how we're implementing roles. They have the same issues as labels.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Maybe. We haven't decided what the character limit would be for custom roles yet since that hasn't been designed yet. 255 characters seems like it would be a safe limit.

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: Converting from role->string is useful for printing the role out in a human-readable way.
In general, we would want application code and our internal code using the role enum instead of strings.
If an application has a role string they want to convert to a role enum, it should already have the role enum available to it without needing to convert.

gknopf-aranya:
> Please add some tests that check the returned value

Will add these tests later. Stopping work on this PR until there is a spec for ID-based roles.

gknopf-aranya:
Closing for now since this is blocked by ID-based roles which needs a spec

PR #179, opened 2025-04-08
afc: remove policy

gknopf-aranya:
Remove AFC policy until we're ready to add AFC back with ID based labels we're adding for AQC.

Migrating AFC to ID based labels is blocked on some changes to AFC crypto and ffis.


crates/aranya-client-capi/cbindgen.toml
```diff
@@ -96,7 +96,7 @@ style = "both"
 "target_os = linux" = "__linux__"
 "target_os = vxworks" = "__VXWORKS__"
 "target_arch = x86_64" = "__x86_64__"
-"feature = afc" = "ENABLE_AFC"
+#"feature = afc" = "ENABLE_AFC"
```
nikki-aranya, line 99: This disables the feature flags below which means we're exposing it in the header which is bleh. It's fine to leave this and just not enable the feature when compiling

crates/aranya-daemon/src/api.rs
```diff
@@ -85,8 +81,8 @@ impl DaemonApiServer {
     pub fn new(
         client: Arc<Client>,
         local_addr: SocketAddr,
-        afc: Arc<Mutex<WriteState<CS, Rng>>>,
-        eng: CE,
+        _afc: Arc<Mutex<WriteState<CS, Rng>>>,
+        _eng: CE,
```
nikki-aranya, line 85: This is already gated by AFC, just change all cfg(feature = "afc") to #[cfg(any())]

crates/aranya-client-capi/cbindgen.toml
nikki-aranya: actually, with rebasing to cfg(any()), this is fine to leave out since it'll remove those functions from the header, but be sure to also remove afc from parse.expand.features below.

gknopf-aranya:
> This is needlessly destructive when the majority of it is already covered by the `#[cfg(feature = "afc")]` feature flag which we can change to `#[cfg(any())]` which will always evaluate to false, and then all we need to do is remove it as the default in the examples and remove the feature flag. I'd just rebase and start over, it'll be easier.

Pushed a new commit with these suggestions.

nikki-aranya (approved):

gknopf-aranya:
Cleaning up some compilation issues with the unit tests

gknopf-aranya:
Allowing some afc related fields to be dead code in api.rs seemed like a cleaner alternative to either deleting the code or adding an underscore in front of things.

nikki-aranya (approved):

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -352,64 +324,17 @@ typedef struct ARANYA_ALIGNED(8) AranyaSyncPeerConfig {
...
-     * This can be used to sort out-of-order messages.
-     */
-    uint64_t seq;
-    /**
-     * Peer's network socket address.
-     */
-    AranyaSocketAddr addr;
-} AranyaAfcMsgInfo;
-#endif
+typedef uint32_t AranyaLabel;
```
jdygert-spok, line 337: Shouldn't be here.

crates/aranya-client/tests/tests.rs
jdygert-spok, line 1: There are afc-specific tests here guarded by `feature = afc` still, they should just be changed to `cfg(any())` and have the body unchanged.

jdygert-spok:
There are a lot of `cfg(feature = "afc")` left. TBH I'd consider just renaming the feature to `_afc` or something and not making other changes. So we can still enable for testing purposes but signify that it's unstable/unusable. (We probably wanted to name it like that originally anyway)

gknopf-aranya:
> There are a lot of `cfg(feature = "afc")` left. TBH I'd consider just renaming the feature to `_afc` or something and not making other changes. So we can still enable for testing purposes but signify that it's unstable/unusable. (We probably wanted to name it like that originally anyway)

The problem is that if the policy is removed, there is no way to test with the AFC feature.
We're wanting to disable/remove AFC and AFC policy while making it as painless as possible to add back in once it supports ID-based labels.

crates/aranya-client/tests/tests.rs
gknopf-aranya: I intentionally only put the parts of the AFC tests that no longer work due to policy being removed behind `cfg(any())`. The rest of the afc tests that still work are behind the afc feature flag so they continue to be tested.

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: Put this behind afc feature flag.

jdygert-spok (approved):
I'd still like to rename the feature to either `_afc` or `unstable-afc`, but it can be on another PR if desired.

gknopf-aranya:
> I'd still like to rename the feature to either `_afc` or `unstable-afc`, but it can be on another PR if desired.

I'd prefer to do that on a different PR since this PR is on the critical path for the upcoming MVP release.

gknopf-aranya:
> I'd still like to rename the feature to either `_afc` or `unstable-afc`, but it can be on another PR if desired.

Here's an issue for renaming the feature:
https://github.com/aranya-project/aranya/issues/182

crates/aranya-daemon/src/api.rs
```diff
@@ -618,13 +653,24 @@ impl DaemonApi for DaemonApiHandler {
...
-    #[cfg(feature = "afc")]
+    #[instrument(skip(self))]
+    async fn revoke_afc_label(
+        self,
+        _: context::Context,
+        team: TeamId,
+        device: ApiDeviceId,
+        label: Label,
+    ) -> ApiResult<()> {
+        Err(anyhow!("Aranya Fast Channels is disabled for this daemon!").into())
```
djohnson-spideroak, line 670: maybe we want a specific error type for this?

gknopf-aranya: This API is an internal API between the Aranya client and daemon. So it does not affect MVP API stability and can be added pre- or post- MVP whenever we have time.

nikki-aranya:
> I'd still like to rename the feature to either `_afc` or `unstable-afc`, but it can be on another PR if desired.

I think the main thing is just excluding it from cargo all-features

djohnson-spideroak (approved):

gknopf-aranya:
> I think the main thing is just excluding it from cargo all-features, we could easily just ignore AFC if it wasn't being actively tested

Would this be another change on this PR?

jdygert-spok:
> > I'd still like to rename the feature to either `_afc` or `unstable-afc`, but it can be on another PR if desired.
> 
> I think the main thing is just excluding it from cargo all-features, we could easily just ignore AFC if it wasn't being actively tested

The naming is an orthogonal issue to what is happening in this PR. We still want to test it. We also want to signify that it is an unstable feature (i.e. the API is not bound to semver). This is a common naming scheme which is known by e.g. cargo-semver-checks.

PR #183, opened 2025-04-09
aranya-client-capi: access to bytes in ids

### LINKED ISSUE #85: client: ensure underlying bytes of IDs are accessible
A common issue we have had in the past with earlier versions of Aranya is unwieldy IDs, especially in C. We should make sure that bytes of these IDs are easily accessed by developers. Ideally, we also provide some easy way to serialize the IDs to a common format like the existing base58 format we currently use.

ssweetney-spok:
Closes #85

- Adds some utility methods to the C API for ID types
    - converting an ID into a string
    - converting a string into an ID
    - ~~Getting the bytes of an ID as a pointer~~ The bytes of the ID structs are accessible by their field(s)

### DIFF
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -86,9 +86,14 @@
 #define ARANYA_DURATION_NANOSECONDS 1
 
 /**
- * The size in bytes of a `DeviceId` converted to a human-readable base64 string.
+ * The size in bytes of an ID converted to a human-readable base58 string.
  */
-#define ARANYA_DEVICE_ID_STR_LEN (((64 * 1375) / 1000) + 1)
+#define ARANYA_ID_STR_LEN (((64 * 1375) / 1000) + 1)
+
+/**
+ * The size in bytes of an ID
+ */
+#define ARANYA_ID_LEN 64
 
 /**
  * An error code.
@@ -1156,16 +1161,81 @@ AranyaError aranya_query_devices_on_team_ext(struct AranyaClient *client,
 /**
  * Writes the human-readable encoding of `device` to `str`.
  *
- * To always succeed, `str` must be at least `ARANYA_DEVICE_ID_STR_LEN` bytes long.
+ * To always succeed, `str` must be at least `ARANYA_ID_STR_LEN` bytes long.
  *
  * @param device ID [`AranyaDeviceId`](@ref AranyaDeviceId).
- * @param device ID string [`AranyaDeviceId`](@ref AranyaDeviceId).
+ * @param str ID string [`AranyaDeviceId`](@ref AranyaDeviceId).
+ * @param str_len returns the length of `str`
  *
  * @relates AranyaError.
  */
 AranyaError aranya_device_id_to_str(struct AranyaDeviceId device,
                                     char *str,
                                     size_t *str_len);
+
+/**
+ * Returns the bytes of `device_id` as an array.
+ *
+ * @param device ID [`AranyaDeviceId`](@ref AranyaDeviceId).
+ *
+ */
+AranyaError aranya_device_id_to_bytes(const struct AranyaDeviceId *device_id,
+                                      uint8_t *bytes,
+                                      size_t bytes_len);
+
+/**
+ * Writes the human-readable encoding of `team` to `str`.
+ *
+ * To always succeed, `str` must be at least `ARANYA_ID_STR_LEN` bytes long.
+ *
+ * @param team ID [`AranyaTeamId`](@ref AranyaTeamId).
+ * @param str ID string [`AranyaTeamId`](@ref AranyaTeamId).
+ * @param str_len returns the length of `str`
+ *
+ * @relates AranyaError.
+ */
+AranyaError aranya_team_id_to_str(struct AranyaTeamId team,
+                                  char *str,
+                                  size_t *str_len);
+
+/**
+ * Returns the bytes of `team_id` as an array.
+ *
+ * @param device ID [`AranyaTeamId`](@ref AranyaTeamId).
+ *
+ */
+AranyaError aranya_team_id_to_bytes(const struct AranyaTeamId *team_id,
+                                    uint8_t *bytes,
+                                    size_t bytes_len);
+
+#if defined(ENABLE_AFC)
+/**
+ * Writes the human-readable encoding of `channel` to `str`.
+ *
+ * To always succeed, `str` must be at least `ARANYA_ID_STR_LEN` bytes long.
+ *
+ * @param channel ID [`AranyaChannelId`](@ref AranyaChannelId).
+ * @param str ID string [`AranyaChannelId`](@ref AranyaChannelId).
+ * @param str_len returns the length of `str`
+ *
+ * @relates AranyaError.
+ */
+AranyaError aranya_channel_id_to_str(struct AranyaChannelId channel,
+                                     char *str,
+                                     size_t *str_len);
+#endif
+
+#if defined(ENABLE_AFC)
+/**
+ * Returns the bytes of `channel_id` as an array.
+ *
+ * @param device ID [`AranyaChannelId`](@ref AranyaChannelId).
+ *
+ */
+AranyaError aranya_channel_id_to_bytes(const struct AranyaChannelId *channel_id,
+                                       uint8_t *bytes,
+                                       size_t bytes_len);
+#endif
 
 /**
  * Query device's keybundle.
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -1208,35 +1208,6 @@ AranyaError aranya_team_id_to_bytes(const struct AranyaTeamId *team_id,
                                     uint8_t *bytes,
                                     size_t bytes_len);
 
-#if defined(ENABLE_AFC)
-/**
- * Writes the human-readable encoding of `channel` to `str`.
- *
- * To always succeed, `str` must be at least `ARANYA_ID_STR_LEN` bytes long.
- *
- * @param channel ID [`AranyaChannelId`](@ref AranyaChannelId).
- * @param str ID string [`AranyaChannelId`](@ref AranyaChannelId).
- * @param str_len returns the length of `str`
- *
- * @relates AranyaError.
- */
-AranyaError aranya_channel_id_to_str(struct AranyaChannelId channel,
-                                     char *str,
-                                     size_t *str_len);
-#endif
-
-#if defined(ENABLE_AFC)
-/**
- * Returns the bytes of `channel_id` as an array.
- *
- * @param device ID [`AranyaChannelId`](@ref AranyaChannelId).
- *
- */
-AranyaError aranya_channel_id_to_bytes(const struct AranyaChannelId *channel_id,
-                                       uint8_t *bytes,
-                                       size_t bytes_len);
-#endif
-
 /**
  * Query device's keybundle.
  *
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -1153,34 +1153,19 @@ AranyaError aranya_query_devices_on_team_ext(struct AranyaClient *client,
                                              struct AranyaExtError *__ext_err);
 
 /**
- * Writes the human-readable encoding of `device` to `str`.
+ * Writes the human-readable encoding of `id` to `str`.
  *
  * To always succeed, `str` must be at least `ARANYA_ID_STR_LEN` bytes long.
  *
- * @param device ID [`AranyaDeviceId`](@ref AranyaDeviceId).
- * @param str ID string [`AranyaDeviceId`](@ref AranyaDeviceId).
+ * @param device ID [`AranyaId`](@ref AranyaId).
+ * @param str ID string [`AranyaId`](@ref AranyaId).
  * @param str_len returns the length of `str`
  *
  * @relates AranyaError.
  */
-AranyaError aranya_device_id_to_str(const struct AranyaDeviceId *device,
-                                    char *str,
-                                    size_t *str_len);
-
-/**
- * Writes the human-readable encoding of `team` to `str`.
- *
- * To always succeed, `str` must be at least `ARANYA_ID_STR_LEN` bytes long.
- *
- * @param team ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param str ID string [`AranyaTeamId`](@ref AranyaTeamId).
- * @param str_len returns the length of `str`
- *
- * @relates AranyaError.
- */
-AranyaError aranya_team_id_to_str(const struct AranyaTeamId *team,
-                                  char *str,
-                                  size_t *str_len);
+AranyaError aranya_id_to_str(const struct AranyaId *id,
+                             char *str,
+                             size_t *str_len);
 
 /**
  * Query device's keybundle.
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -77,6 +77,11 @@
 #endif /* ARANYA_PACKED */
 
 
+/**
+ * The size in bytes of an ID
+ */
+#define ARANYA_ID_LEN 64
+
 #define ARANYA_DURATION_SECONDS (1000 * ARANYA_DURATION_MILLISECONDS)
 
 #define ARANYA_DURATION_MILLISECONDS (1000 * ARANYA_DURATION_MICROSECONDS)
@@ -89,11 +94,6 @@
  * The size in bytes of an ID converted to a human-readable base58 string.
  */
 #define ARANYA_ID_STR_LEN (((64 * 1375) / 1000) + 1)
-
-/**
- * The size in bytes of an ID
- */
-#define ARANYA_ID_LEN 64
 
 /**
  * An error code.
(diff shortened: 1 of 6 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1211,15 +1211,23 @@ pub fn query_devices_on_team(
     Ok(())
 }
 
-/// The size in bytes of a `DeviceId` converted to a human-readable base64 string.
-pub const ARANYA_DEVICE_ID_STR_LEN: u64 = (64 * 1375) / 1000 + 1;
+/// The size in bytes of an ID converted to a human-readable base58 string.
+pub const ARANYA_ID_STR_LEN: u64 = (64 * 1375) / 1000 + 1;
+
+/// The size in bytes of an ID
+pub const ARANYA_ID_LEN: u64 = 64; // size_of::<aranya_fast_channels::crypto::aranya_crypto::Id>() as u64;
```
ssweetney-spok, line 1218: This (the commented out code) doesn't work for some reason. I see it's included in the code generated by capi-codegen but cbindgen seems to ignore it

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1233,6 +1241,59 @@ pub fn device_id_to_str(
     Ok(())
 }
 
+/// Returns the bytes of `device_id` as an array.
+///
+/// @param device ID [`DeviceId`].
+///
+/// cbindgen:ptrs-as-arrays=[[bytes; 64]]
```
ssweetney-spok, line 1248: Logically, it makes more sense for bytes arg to be an array. I'd like to use this but capi-codegen trims any doc [comments starting with cbindgen](https://github.com/aranya-project/aranya-core/blob/ac2f0c9e1a0f17e8fca8898acc51373cf9340fe2/crates/aranya-capi-codegen/src/syntax/doc.rs#L30)

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok, line 1: There's a lot of repetition that could be reduced with a macro but it seems like capi-codegen ignores macro calls in the global scope so I couldn't do something like


```Rust
macro_rules! id_to_str {
    ($([$ty:ty, $fn_name:ident]),+ $(,)?) => {
        $(
            pub fn $fn_name(
                id: $ty,
                str: &mut MaybeUninit<c_char>,
                str_len: &mut usize,
            ) -> Result<(), imp::Error> {
                let str = aranya_capi_core::try_as_mut_slice!(str, *str_len);
                aranya_capi_core::write_c_str(str, &id.0, str_len)?;
                Ok(())
            }
        )+
    };
}

id_to_str!([DeviceId, device_id_to_str]);
```

ssweetney-spok: It seems like the easiest way is to add a "dummy" macro to aranya_capi_macros that's similar to `aranya_capi_macro::builds`

ssweetney-spok (commented):
Wondering if others think it's worth updating the capi-codegen crates to accommodate the changes I want to make

cc: @jdygert-spok

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Yeah cbindgen doesn't evaluate those unfortunately, same reason the above is hardcoded to 64.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1233,6 +1241,59 @@ pub fn device_id_to_str(
...
+#[aranya_capi_core::no_ext_error]
+pub fn device_id_to_bytes(device_id: &DeviceId, bytes: &mut [u8]) -> Result<(), imp::Error> {
+    let src = device_id.0.as_array();
+    if bytes.len() >= src.len() {
+        bytes[0..src.len()].copy_from_slice(src);
+        Ok(())
+    } else {
+        Err(imp::Error::BufferTooSmall)
+    }
+}
```
jdygert-spok, line 1259: I think we should return a ptr rather than copying to a buf. Maybe just exposing the field of the ID type directly. Unsure.

gknopf-aranya: If we return a pointer to the underlying bytes of the ID, we should be clear in the docs that if the device ID is assigned a different value, the array returned from this function will point to the new bytes.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1233,6 +1241,59 @@ pub fn device_id_to_str(
...
 }
 
+/// Returns the bytes of `device_id` as an array.
+///
+/// @param device ID [`DeviceId`].
+///
+/// cbindgen:ptrs-as-arrays=[[bytes; 64]]
+// TODO: bytes param comment
+#[aranya_capi_core::no_ext_error]
+pub fn device_id_to_bytes(device_id: &DeviceId, bytes: &mut [u8]) -> Result<(), imp::Error> {
```
gknopf-aranya, line 1251: @ssweetney-spok, @jdygert-spok, @djohnson-spideroak the purpose of converting a device ID to bytes is so the application can pass the device ID over the wire to another device. This means we also need a `device_id_from_bytes` method so the application that receives the bytes can construct a `DeviceId` from the raw bytes.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Could that introduce a race condition if there are multiple readers/writers of the same underlying byte array?

elagergren-spideroak: We'll be returning `*const u8`, so that's implicit. Plus, we don't ever mutate IDs.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1233,6 +1241,59 @@ pub fn device_id_to_str(
...
+    let src = device_id.0.as_array();
```
gknopf-aranya, line 1252: @ssweetney-spok, @jdygert-spok do we want to use serde serialization/deserialization for the to/from bytes conversions?

crates/aranya-client-capi/src/api/defs.rs
elagergren-spideroak: > Could that introduce a race condition if there are multiple readers/writers of the same underlying byte array?

?

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: We should probably just expose the bytes directly then. I'm thinking something like this which will also remove the repetition of the multiple str methods.
```c
typedef struct { char bytes[64] } Id;
typedef struct { id Id } DeviceId;
typedef struct { id Id } TeamId;
```

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: No, it is just plain bytes.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: The application could mutate the DeviceId since they have access to the `DeviceId` type.
If we returned a pointer to data inside that type, then the application mutates the `DeviceId`, that could cause problems.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1233,6 +1241,59 @@ pub fn device_id_to_str(
...
+/// @param str ID string [`TeamId`].
+/// @param str_len returns the length of `str`
+///
+/// @relates AranyaError.
+#[aranya_capi_core::no_ext_error]
+pub fn team_id_to_str(
+    team: TeamId,
+    str: &mut MaybeUninit<c_char>,
+    str_len: &mut usize,
+) -> Result<(), imp::Error> {
```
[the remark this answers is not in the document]
elagergren-spideroak: It sorta precludes writing directly into existing memory. Unless we use LTO, the compiler probably isn't going to be able to "inline" the write, so there will always be extra copies to and then from the stack. Base58 encoding is ~138% larger, so for 32 byte IDs that's a 78+1 (null terminator) byte stack copy. It might be NBD, but it's not _wonderful_.

crates/aranya-client-capi/src/api/defs.rs
elagergren-spideroak: They pass in `*const AranyaDeviceId`. If they mutate that pointer for the duration of the function call then it's their own fault. There isn't anything we can do about it. And it's no different from every other function of ours that takes pointers.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: To my knowledge, there is no way to validate if the ID bytes are correct because it is a randomly generated binary array. If an incorrect DeviceID is used, policy checks would fail.

crates/aranya-client-capi/src/api/defs.rs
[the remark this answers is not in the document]
elagergren-spideroak: Also, base58 encoding isn't fixed size (it trims zeros), so the caller would still need to use strlen to figure out the length of the ID.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: "I think we should return a ptr rather than copying to a buf"
This is how the thread started. I'm talking about if a pointer was returned rather than giving the application it's own buffer. If DeviceId was modified after the function call, but before the application sends the DeviceId over the wire, they would end up sending a different DeviceId than what was expected unless we documented that behavior.

crates/aranya-client-capi/src/api/defs.rs
elagergren-spideroak: > Do we need some kind of validation check when converting from raw bytes? If not then maybe we should get rid of these methods and go with @jdygert-spok suggestion to just expose the field directly

Nope, no validation. IDs are just a bag of bytes. We use types to differentiate. Sometimes I wish we had a "type" byte prefix or something, but I also wasn't always a fan of that back in flow2.

gknopf-aranya: Now that I'm thinking about this, I think having some sort of wrapper around the bytes we use to validate the type/integrity of the data could be useful. Depends on how much we want to trust the application to transport the bytes correctly.

crates/aranya-client-capi/src/api/defs.rs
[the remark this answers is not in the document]
jdygert-spok: That's 45 bytes for 32 byte IDs, right?

You _would_ need an extra strlen call in some cases, but not if you're just passing it to some function that takes a cstring anyways.

I think these are not worth worrying about for `aranya`, maybe in `aranya-core`. We should do what makes the API nicest to use (within reason).

gknopf-aranya:
@ssweetney-spok could you please update this branch with the latest from `main`?

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: I think it's fine without. There should be no security issue here. But you could argue for a usability improvement of earlier detection of bad IDs. I don't think it's really worth the extra complexity.

crates/aranya-client-capi/src/api/defs.rs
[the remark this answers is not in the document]
elagergren-spideroak: > That's 45 bytes for 32 byte IDs, right?

Yeah, I'm bad at math sometimes.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -172,17 +172,71 @@ pub fn init_logging() -> Result<(), imp::Error> {
...
+#[derive(Copy, Clone, Debug)]
+pub struct Id {
+    bytes: [u8; ARANYA_ID_LEN],
+}
+
+impl AsRef<aranya_crypto::Id> for Id {
+    fn as_ref(&self) -> &aranya_crypto::Id {
+        unsafe { mem::transmute(self) }
+    }
+}
```
ssweetney-spok, line 193: This is Ok right? I wanted to avoid having to do the copy I did [here](https://github.com/aranya-project/aranya/pull/183/commits/beaddabd930969f68bc7b26209ad7d8e4e9e8c5d) when converting to a string

elagergren-spideroak: yeah, but it needs a safety comment, and we might want to consider using `zerocopy` for stuff like this

jdygert-spok: Yeah zerocopy would be nice.

Avoid transmute when there are alternatives, in this case by pointer casting. Also, being explicit can help avoid refactor errors.
```rust
unsafe {
    &*core::ptr::from_ref::<[u8; ARANYA_ID_LEN]>(&self.bytes).cast::<aranya_crypto::Id>()
}
```

jdygert-spok:
We will want `id_from_str` for human readable ID transfer.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1226,25 +1297,26 @@ pub fn query_devices_on_team(
     Ok(())
 }
 
-/// The size in bytes of a `DeviceId` converted to a human-readable base64 string.
-pub const ARANYA_DEVICE_ID_STR_LEN: u64 = (64 * 1375) / 1000 + 1;
+/// The size in bytes of an ID converted to a human-readable base58 string.
+pub const ARANYA_ID_STR_LEN: u64 = (64 * 1375) / 1000 + 1;
```
jdygert-spok, line 1301:
```suggestion
pub const ARANYA_ID_STR_LEN: usize = (ARANYA_ID_LEN * 1375) / 1000 + 1;
```

djohnson-spideroak: any potential issues around usize potentially being a different size?

jdygert-spok: Nah `usize` is correct here, although cbindgen makes it an untyped macro anyway.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: I was curious so I went to [godbolt](https://godbolt.org/z/jdehGfj8s) and they compile to the same assembly when optimizations are enabled

elagergren-spideroak: yeah, avoiding transmute is just b/c it's easier to do UB

crates/aranya-client-capi/src/api/defs.rs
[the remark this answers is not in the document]
ssweetney-spok: I'm leaning towards the way it's implemented currently because it could be useful for users to know exactly how many bytes are written. They could want write the string (without trailing zeroes) to a file

examples/c/example.c
```diff
@@ -424,12 +424,23 @@ AranyaError run(Team *t) {
...
+        err = aranya_id_from_str(device_str, &decodedId);
+        EXPECT("error decoding string into an ID", err);
+
         free(device_str);
+
+        if (!(memcmp(decodedId.bytes, device_result.id.bytes, ARANYA_ID_LEN) == 0)) {
+            fprintf(stderr, "application failed: Decoded ID doesn't match\r\n");
+            return EXIT_FAILURE;
+        }
+
```
djohnson-spideroak, line 443: Can we do something similar for the team ID?

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1154,18 +1153,29 @@ AranyaError aranya_query_devices_on_team_ext(struct AranyaClient *client,
...
- * Writes the human-readable encoding of `device` to `str`.
+ * Writes the human-readable encoding of `id` to `str`.
+ *
+ * To always succeed, `str` must be at least `ARANYA_ID_STR_LEN` bytes long.
+ *
+ * @param device ID [`AranyaId`](@ref AranyaId).
+ * @param str ID string [`AranyaId`](@ref AranyaId).
+ * @param str_len returns the length of `str`
+ *
+ * @relates AranyaError.
```
jdygert-spok, line 1164:
```suggestion
 * @relates AranyaId.
```

This makes it show up on the page for `AranyaId`, kinda like a method. Ditto for the existing one below.

examples/c/example.c
jdygert-spok: No harm but it is a bit redundant since we deduped the ID functions

djohnson-spideroak: Given pretty much every integration is going to require moving team IDs about, I think its important to include in the example.

crates/aranya-client-capi/src/api/defs.rs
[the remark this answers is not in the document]
jdygert-spok: It will be the full size 255/256 of the time. I've thought before it should probably just be a fixed size encoding for simplicity. I can imagine someone assuming it is fixed size and then rarely encountering a bug because of it.

examples/c/example.c
```diff
@@ -223,6 +223,23 @@ AranyaError init_team(Team *t) {
     err = aranya_create_team(&t->clients.owner.client, &t->id);
     EXPECT("error creating team", err);
 
+    // Test ID serialization and deserialization
+    size_t team_id_str_len        = ARANYA_ID_STR_LEN;
+    char *team_id_str             = malloc(team_id_str_len);
```
jdygert-spok, line 228: Just noticed these malloc unnecessarily, not sure if that is confusing or worth changing :shrug:

examples/c/example.c
ssweetney-spok: Like Dygert mentioned it is redundant but I do see your point about wanting to include it in the example. 

I've been wondering if it makes sense to have separate unit test cases for things like this to keep it separate from the Aranya management code. The example is getting kinda messy since "test" code is interleaved with the actual code for initializing and managing Aranya

examples/c/example.c
djohnson-spideroak: Can you elaborate?

crates/aranya-client-capi/src/api/defs.rs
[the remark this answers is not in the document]
elagergren-spideroak: We can always just make that fix to base58.

> It will be the full size 255/256 of the time.

Note that while IDs are uniformly random _in practice_, I believe there are some hash functions that meet our criteria (preimage and collision resistance) whose digests are _not_ uniformly random. But this is irrelevant to the discussion.

examples/c/example.c
ssweetney-spok: The size is known at compile time so we can just initialize an array

djohnson-spideroak (approved):

PR #191, opened 2025-04-15
WIP: aqc: finalize policy

gknopf-aranya:
This PR is based on the `rm-policy` branch where the `policy.md` file has been deleted from the `main` branch.
Note: this statement will be true after merging https://github.com/aranya-project/aranya/pull/145 and updating the `rm-policy` branch.

This was done so we can review the entire policy on this PR and identify any breaking changes from the original policy interface baseline from this PR:
https://github.com/aranya-project/aranya/pull/145

Once this PR is approved, it will be merged directly into the `main` branch.
`rm-policy` is used only for convenience so the entire policy can be reviewed rather than only the deltas from the initial baseline.


crates/aranya-daemon/src/policy.md
```diff
@@ -0,0 +1,1969 @@
...
+    policy {
+        let author = get_valid_device(envelope::author_id(envelope))
+        let device = check_unwrap find_existing_device(this.device_id)
+
+        // Only Owners and Operators can associate a network name.
+        check is_owner(author.role) || is_operator(author.role)
+        // Only Members can be associated a network name.
+        check is_member(device.role)
+
+        // TODO: check that the network identifier is valid.
```
gknopf-aranya, line 948: @djohnson-spideroak what kind of validity checks should we perform on AQC network identifiers here?

djohnson-spideroak: I dont know if there even are any checks we can do. It could be an IP (v4 or v6), domain name, or maybe even something else if the transport changes.

gknopf-aranya: It would be much easier to check format validity in Rust than in policy code.
I don't think string parsing in policy is very sophisticated yet.

gknopf-aranya: I'd prefer to remove this todo from the policy and open an issue to validate network identifiers with some Rust code.
Unless the person that added this todo has something else in mind.

gknopf-aranya: Removed this todo from the policy

djohnson-spideroak:
I think instead of rm'ing the full policy in the base branch, you need to revert it to post-afc-removal pre-aqc-addition. As it currently is, it shows the full file in the diff and we want to highlight the AQC changes.

crates/aranya-daemon/src/policy.md
mooreoak: > It would be much easier to check format validity in Rust than in policy code. I don't think string parsing in policy is very sophisticated yet.

If we ever do want to require that the address has a specific format we should do it in the policy but could use and ffi to do the check.

gknopf-aranya:
@djohnson-spideroak I've changed the base branch to `pre-aqc-policy`

crates/aranya-daemon/src/policy.md
```diff
@@ -941,8 +1048,875 @@ command UnsetAqcNetworkName {
...
+        // Per the AQC specification, this must be at least 32.
+        psk_length_in_bytes int,
+    }
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        let author = get_valid_device(envelope::author_id(envelope))
+        let peer = check_unwrap find_existing_device(this.peer_id)
```
djohnson-spideroak, line 1170: any reason we wouldn't want to use `get_valid_device` for both calls? The only difference is that `get_valid_device` checks TeamEnd then calls `find_existing_device`.

crates/aranya-daemon/src/policy.md
```diff
@@ -941,8 +1048,875 @@ command UnsetAqcNetworkName {
...
+        // The label must exist.
+        let label = check_unwrap query Label[label_id: this.label_id]
+
+        // Only Members can create AQC channels with other peer Members
+        check is_member(author.role)
+        check is_member(peer.role)
+
+        // Check that both devices are allowed to participate in
+        // this bidirectional channel.
+        check can_create_aqc_bidi_channel(author.device_id, peer.device_id, label.label_id)
```
djohnson-spideroak, line 1183: Might be good for clarity to document here that this function checks that the devices have permissions for the label, and that they have correct send/recv permissions.

crates/aranya-daemon/src/policy.md
```diff
@@ -941,8 +1048,875 @@ command UnsetAqcNetworkName {
...
+    // The label name.
+    label_name string,
+    // The ID of the device that created the label.
+    label_author_id id,
+}
+```
+
+###### Query Label Assignments
+
+Queries for a list labels assigned to a device.
```
djohnson-spideroak, line 1860: For another PR, we probably want query devices assigned to a label.

djohnson-spideroak (approved):
LGTM

crates/aranya-daemon/src/policy.md
```diff
@@ -941,8 +1048,875 @@ command UnsetAqcNetworkName {
...
+        // Verify that the label exists.
+        //
+        // This will happen in the `finish` block if we try to
+        // create an already true label, but checking first
+        // results in a nicer error (I think?).
+        let label = check_unwrap query Label[label_id: this.label_id]
+
+        finish {
+            // Cascade deleting the label assignments.
+            delete AssignedLabel[device_id: ?, label_id: label.label_id]
```
jdygert-spok, line 1557: `?` will need to be on right so it is a prefix query.

jdygert-spok: But that will break query_label_assignments...

jdygert-spok:
LGTM beyond open convos

crates/aranya-daemon/src/policy.md
mooreoak: Ya that's is not a valid delete. The way to do that is with an action that produces more then one delete command.

mooreoak: It is not valid because it could visit an unknown number of rows.

mooreoak: The reason that it is ok to have an unknown number of commands is that each transaction is bounded which makes it possible to reason about the amount of work that will happen when processing each command.

This really should just be GC any way as the policy should check that the label is valid not just that the user has the label. That will mean that even partial states across rows will not lead weird stats being accessable.

jdygert-spok: `delete AssignedLabel[label_id: label.label_id, device_id: ?]` would be valid, since it just writes one tombstone.

I think the way forward for now is to switch it, and have the query_label_assignments do an inefficient scan since that is just local in a query action.

```
map AssignedLabel[label_id: ?, device_id: ?] as f {
    if f.device_id == device_id { .. }
}
```

mooreoak: > `delete AssignedLabel[label_id: label.label_id, device_id: ?]` would be valid, since it just writes one tombstone.

So we have wild cars in tombstobes? That's a cool idea.

mooreoak: @jdygert-spok How dose wild cards in tombstobes in work when doing a binary search of the index. Couldn't you miss the tombstobe?

Assuming that the wild card acts like the biggest or smallest value when searching for a fully qualified key you could just miss it right?

crates/aranya-daemon/src/policy.md
gknopf-aranya: So we don't check `TeamEnd` twice?

gknopf-aranya: This seems more ergonomic:
```
check_team_exists()
let author = get_valid_device(envelope::author_id(envelope))
let peer = get_valid_device(this.peer_id)
```

crates/aranya-daemon/src/policy.md
gknopf-aranya: https://github.com/aranya-project/aranya/issues/211

crates/aranya-daemon/src/policy.md
gknopf-aranya: Clarifying this

crates/aranya-daemon/src/policy.md
djohnson-spideroak: I could imagine a situation where we remove the TeamEnd check in get_valid_device and fail to readd the check. I think if we also expect peer to be a valid device, we should use get_valid_device.

We dont need to check it twice, I think we just need to make sure that important checks like that are a bit more explicit. It might not make sense to check TeamEnd in get_valid_device anyways, maybe a different PR with a `check_team_state` function that does checks like that?

gknopf-aranya: I pushed a commit that adds a `check_team_exist` to each command except the team creation command.
The docs specify why that needs to be checked for each command.
It wasn't being checked previously for commands that weren't calling `get_valid_device` and now it is.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Changing the key order for `AssignedLabel`

gknopf-aranya:
Reminder that once this is approved, I'll need to rebase it off `main` before merging it in.

jdygert-spok (approved):

djohnson-spideroak (dismissed):

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok I've rebased the PR onto `main`.

djohnson-spideroak (approved):

jdygert-spok (approved):

crates/aranya-daemon/src/policy.md
```diff
@@ -159,11 +159,19 @@ function find_existing_device(device_id id) optional struct Device {
     return device
 }
 
-// Returns a valid Device after performing sanity checks per the stated invariants.
-function get_valid_device(device_id id) struct Device {
+// Checks if the team exists.
+// This should always be the first thing that is checked before executing a command on a team.
+// The only command that doesn't run this check first is `CreateTeam`.
+// Caller does not need to check the return bool, since it will always be true if the `check` succeeds.
+function check_team_exists() bool {
```
elagergren-spideroak, line 166: this plus having to use `let t = check_team_exists()` everywhere is sort of a strange construction. Why not just return a bool and have users call check themselves?

gknopf-aranya: I wanted to call the function without a return value but the compiler wouldn't let me.

elagergren-spideroak: I know. Just make it return true if the fact exists and false otherwise, then have the caller use `check`

mooreoak: > I wanted to call the function without a return value but the compiler wouldn't let me.

just wanted to point out that "check team_exists()" is the same number of key strokes as "check_team_exists()"

elagergren-spideroak: yeah, and `check team_exists()` gets rid of the weird `let t = check_team_exists()`

jdygert-spok: We may want to allow unit functions though, for e.g. `check_can_create_bidi_channel` which has many checks, having the checks inside the function will give more info on failure.

elagergren-spideroak: agreed, but until we add unit  functions we should get rid of the weirdness  :)

djohnson-spideroak (approved):

PR #193, opened 2025-04-16
WIP: aqc: keystore patch

gknopf-aranya:
Update AQC PSK key management to be more secure.


gknopf-aranya:
Not needed. Daemon will configure the AQC ffi to use a different keystore and pass that path to the Aranya client.

PR #197, opened 2025-04-16
Refactor `load_or_gen_key_wrap_key`

jdygert-spok:
Simplifies and generalizes the function into `load_or_gen_key`.

### DIFF
--- a/crates/aranya-util/src/util.rs
+++ b/crates/aranya-util/src/util.rs
@@ -2,6 +2,8 @@
 
 use std::{fs::Permissions, os::unix::fs::PermissionsExt, path::Path, str::FromStr};
 
+use anyhow::Context as _;
+use aranya_crypto::{import::Import, keys::SecretKey, Rng};
 use aranya_fast_channels::shm;
 use tokio::{fs, io};
 use tracing::warn;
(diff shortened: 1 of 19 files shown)


jdygert-spok:
I had an old WIP refactor of the function itself so went ahead and did this.

gknopf-aranya (approved):
Thank you @jdygert-spok

crates/aranya-util/src/util.rs
```diff
@@ -67,3 +69,26 @@ impl TryFrom<String> for ShmPathBuf {
         s.parse::<Self>()
     }
 }
+
+/// Loads a key from a file or generates and writes a new one.
+pub async fn load_or_gen_key<K: SecretKey>(path: impl AsRef<Path>) -> anyhow::Result<K> {
+    match fs::read(&path).await {
+        Ok(buf) => {
+            tracing::info!("loaded key");
```
elagergren-spideroak, line 77: probably want to include the path

crates/aranya-util/src/util.rs
```diff
@@ -67,3 +69,26 @@ impl TryFrom<String> for ShmPathBuf {
...
+            let key = Import::import(buf.as_slice()).context("unable to import key from file")?;
```
elagergren-spideroak, line 78: ditto about the path

crates/aranya-util/src/util.rs
elagergren-spideroak: also in the errors

elagergren-spideroak (dismissed):

gknopf-aranya (approved):

gknopf-aranya (approved):

gknopf-aranya (changes requested):
We won't need this PR anymore if the encrypted AQC PSK is passed via the UDS IPC instead of via a keystore.

jdygert-spok:
It still makes improvements so we should either merge as-is or I'll move the new version into the daemon.

gknopf-aranya:
Then we should put this in the daemon

gknopf-aranya (approved):

djohnson-spideroak (approved):

PR #198, opened 2025-04-16
WIP: aqc: psk keystore patch

gknopf-aranya:
For security purposes, use a different keystore for the AQC PSKs than the regular Aranya key bundle.

### DIFF
--- a/crates/aranya-client/src/client.rs
+++ b/crates/aranya-client/src/client.rs
@@ -134,8 +134,8 @@ impl Client {
             addr = ?afc.local_addr().map_err(Error::Afc)?,
             "bound AFC router",
         );
-        debug!("getting key store info");
-        let keystore_info = daemon.get_keystore_info(context::current()).await??;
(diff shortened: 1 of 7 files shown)


gknopf-aranya:
@elagergren-spideroak, @jdygert-spok now that different key stores are being used, the aqc ffi no longer has access to the encryption key in needs to create a channel:
```
2025-04-16T17:59:01.249108Z ERROR unable to find key: device encryption key
2025-04-16T17:59:01.249148Z ERROR 
IO: internal error at line 1048 col 5:
	let ch = aqc::create_bidi_channel(
        parent_cmd_id,
        author.enc_key_id,
        author_id,
        peer_enc_pk,
        peer_id,
        label_id,
    )
```
Will the Aranya client also need access to the encryption key when deriving the PSK from the encap?

gknopf-aranya:
@elagergren-spideroak has suggested encrypting the AQC PSK and sending it via the UDS IPC instead.

PR #200, opened 2025-04-16
daemon: split peer maps by graph

jdygert-spok:


crates/aranya-daemon/src/sync.rs
```diff
@@ -233,7 +235,7 @@ impl Syncer {
         };
         let n = effects.len();
         self.send_effects
-            .send(effects)
+            .send((*id, effects))
```
gknopf-aranya, line 238: This is sending a graph ID and effect for each effect that is processed which is inefficient.
We may want to set up a channel for each graph ID, so only the effect needs to be sent over the corresponding channel each time an effect is processed.
Memory vs. processing tradeoff.

jdygert-spok: It's one graph id per sync, not per effect.

gknopf-aranya: With regular sync intervals, it's common to only process a single effect per sync.

gknopf-aranya: If it works, let's go with it. Optimizing performance could be something done later. And most likely there would be some other bottleneck that's more important to optimize than this anyways.

crates/aranya-daemon/src/api.rs
```diff
@@ -202,13 +206,13 @@ struct DaemonApiHandler {
     /// AFC peers.
     #[cfg(feature = "afc")]
     #[allow(dead_code)]
-    afc_peers: Arc<Mutex<BiBTreeMap<api::NetIdentifier, DeviceId>>>,
+    afc_peers: Arc<Mutex<BTreeMap<GraphId, BiBTreeMap<api::NetIdentifier, DeviceId>>>>,
```
nikki-aranya, line 209: can we make this into a type = ; as well? I feel like the extremely long types should be aliased

nikki-aranya (dismissed):

nikki-aranya (approved):

PR #201, opened 2025-04-16
docs: aranya-util rustdocs

djohnson-spideroak:
Add rustdocs to aranya-util.


crates/aranya-util/src/addr.rs
```diff
@@ -34,10 +37,12 @@ macro_rules! const_assert {
     }
 }
 
-/// A 'host:port' network address.
+/// Represents a network address composed of a host (domain name, IPv4, or IPv6)
```
gknopf-aranya, line 40: Should keep `host:port` in these docs

crates/aranya-util/src/addr.rs
```diff
@@ -34,10 +37,12 @@ macro_rules! const_assert {
...
+/// and a port number.
 ///
-/// The 'host' portion can be a domain name, IPv4 address, or
-/// IPv6 address.
+/// `Addr` ensures that the host part is a syntactically valid domain name or IP address.
+/// It provides methods for DNS lookup, conversion to socket addresses, and serde
+/// serialization.
```
gknopf-aranya, line 45:
```suggestion
/// serialization/deserialization
```

crates/aranya-util/src/util.rs
```diff
@@ -1,34 +1,65 @@
...
+/// # Errors
+///
+/// Returns `io::Error` if the file cannot be written to (e.g., due to permissions
+/// or invalid path), but not if setting permissions fails.
 pub async fn write_file(path: impl AsRef<Path>, data: &[u8]) -> io::Result<()> {
     fs::write(path.as_ref(), data).await?;
     let perms = Permissions::from_mode(0o600);
     if let Err(err) = fs::set_permissions(&path, perms).await {
-        warn!(err = ?err, path = %path.as_ref().display(), "unable to set file perms");
+        warn!(err = ?err, path = %path.as_ref().display(), "unable to set file perms to 0o600");
```
gknopf-aranya, line 24: Let's reuse a mode variable so these don't get out of sync: 
```
let mode = 0o600
format!("unable to set file perms to ...", mode)
```

crates/aranya-util/src/util.rs
```diff
@@ -1,34 +1,65 @@
...
+/// # Errors
+///
+/// Returns `io::Error` if the directory cannot be created (e.g., due to permissions
+/// or invalid path), but not if setting permissions fails.
 pub async fn create_dir_all(path: impl AsRef<Path>) -> io::Result<()> {
     fs::create_dir_all(path.as_ref()).await?;
     let perms = Permissions::from_mode(0o700);
     if let Err(err) = fs::set_permissions(&path, perms).await {
-        warn!(err = ?err, path = %path.as_ref().display(), "unable to set directory perms");
+        warn!(err = ?err, path = %path.as_ref().display(), "unable to set directory perms to 0o700");
```
gknopf-aranya, line 45: Ditto

crates/aranya-util/src/util.rs
elagergren-spideroak: they're never going to change

crates/aranya-util/src/util.rs
```diff
@@ -1,34 +1,65 @@
...
     }
     Ok(())
 }
 
-/// An owned shared memory path.
+/// An owned, validated buffer representing a shared memory path string.
+///
+/// Shared memory paths often have specific syntax requirements (e.g., starting
+/// with '/' and containing only a single slash). The requirements for
```
gknopf-aranya, line 53: I've encountered a fairly small maximum length when setting the shm paths to longer values.
I'm not sure if this maximum length is imposed by the OS or Aranya somewhere.

elagergren-spideroak: All constraints are from the OS, not Aranya (although Aranya will enforce them on behalf of the OS b/c the OS never provides good errors).

crates/aranya-util/README.md
```diff
@@ -0,0 +1,7 @@
+# Aranya Utilities (`aranya-util`)
+
+This crate provides common utility functions and types used across various Aranya components.
```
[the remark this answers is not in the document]
djohnson-spideroak: Is it needed for the util crate to list the other crates currently using it? I dont know how useful that would be.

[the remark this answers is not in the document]
gknopf-aranya: It definitely shouldn't list all of them. That would be difficult to maintain as more crates use it.
Mentioning some of the key architectural components or the `aranya` repo in general would be useful though.
It shows our intent that this crate be used by other crates in `aranya` and not by crates in `aranya-core` or a user's application for example.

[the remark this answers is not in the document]
elagergren-spideroak: The docs should state that this is an internal crate without any backward compatibility guarantees.

gknopf-aranya (approved):

gknopf-aranya (approved):

ssweetney-spok (approved):

Issue #77, opened 2025-02-18
Evaluate `tarpc` Alternatives


nikki-aranya:
Coming from #70 / #59, we should be running the Example App under the `INFO` level, however our current approach really seems to spam the logs. After demoting the info lines in [Syncer::sync](https://github.com/aranya-project/aranya/blob/577235d6c61a40113e44d110c011c52a18ecfec6/crates/aranya-daemon/src/sync.rs#L172C1-L192C2), the next biggest offender making it difficult to see what's going on is in tarpc itself, which we could simply filter out but should put thought into if there's something better. Given we already use `postcard`, we should see if `postcard-rpc` works better for our needs. The bonus of `postcard-rpc` is that we can share it with the embedded platform. See also #76 for more work on this front.

elagergren-spideroak:
Relates to https://github.com/aranya-project/aranya/issues/164

PR #202, opened 2025-04-16
encrypt data between client and daemon

### LINKED ISSUE #199: aqc: send encrypted PSKs over IPC

elagergren-spideroak:
Encrypts a tarpc transport using HPKE.

While we're here, clean up some other APIs. The clean up work should really be done on another PR, but we're pressed for time so oh well.

Fixes #199


elagergren-spideroak:
@jdygert-spok @gknopf-aranya fyi, i decided to go the "encrypt everything" route b/c it ends up being a _little_ easier to implement

crates/aranya-daemon/src/daemon.rs
```diff
@@ -107,41 +115,25 @@ impl Daemon {
...
-                    },
-                    self.cfg.uds_api_path.clone(),
-                    Arc::new(pk),
-                    peers,
-                    recv_effects,
-                )
-                .context("Unable to start daemon API!")?
-            }
+            let afc = self.setup_afc()?;
+            DaemonApiServer::new(
```
gknopf-aranya, line 119: Where is the code that configures the IPC to use the daemon's encryption key? I was expecting to see the `ServerCodec` somewhere in `daemon.rs` to initialize the API so it encrypts data with the encryption key.
I see `ClientCodec` in the Aranya client's `client.rs` file when initializing the client connection as expected.

crates/aranya-daemon-api/src/crypto/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! TODO
```
gknopf-aranya, line 1: What's left todo for this?

gknopf-aranya (approved):

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -585,15 +487,23 @@ pub type ClientConfig = Safe<imp::ClientConfig>;
...
 /// @param cfg a pointer to the client config builder
 /// @param address a string containing the address
-pub fn client_config_builder_set_daemon_addr(
+pub fn client_config_builder_set_daemon_uds_path(
     cfg: &mut ClientConfigBuilder,
     address: *const c_char,
 ) {
-    cfg.daemon_addr(address);
+    cfg.set_daemon_addr(address);
+}
```
nikki-aranya, line 499: This won't always be a UDS path, I specifically kept it as "address" for that reason; in fact address might be too network-specific a term anyways

nikki-aranya:
....why are we removing AFC? The whole point was to stub it out so we can do some brief cleanup of the majority of the scaffolding once we're ready to introduce it back in as an option, unless we're fully abandoning AFC support

crates/aranya-client/src/aqc.rs
```diff
@@ -1,75 +1,43 @@
...
-        );
-        let eng = {
-            let key = load_or_gen_key(keystore_info.wrapped_key).await?;
-            CE::new(&key, Rng)
-        };
+    #[instrument(skip_all)]
+    pub(crate) async fn new() -> Result<Self, AqcError> {
+        Ok(Self {})
+    }
+}
```
nikki-aranya, line 24: Is there a reason this was reverted?

crates/aranya-client/src/client.rs
```diff
@@ -60,159 +54,158 @@ impl Labels {
...
 }
 
-/// List of AFC labels.
-#[cfg(feature = "afc")]
-pub struct AfcLabels {
-    data: Vec<AfcLabel>,
+/// Builds a [`Client`].
+pub struct ClientBuilder<'a> {
+    /// The UDS that the daemon is listening on.
+    uds_path: Option<&'a Path>,
```
nikki-aranya, line 60: This should also be unix-gated like the function is on the builder below

crates/aranya-daemon-api/src/crypto/txp.rs
nikki-aranya, line 1: This whole file could use additional documentation to explain what all the hell is happening inside but I won't hold up this PR on that

crates/aranya-client/src/aqc.rs
elagergren-spideroak: The code exists inside the daemon now. It was only in the client because we did not have want to send PSKs over an insecure IPC. The IPC is secure now, so the code was moved into the daemon.

crates/aranya-client-capi/src/api/defs.rs
elagergren-spideroak: I changed this to be "UDS path" so that we can add different configuration methods for different IPC types. It's (unfortunately) too optimistic to assume that all our IPC mechanisms can be represented as an "address."

nikki-aranya: Yeah, seems fine looking at the rest of the PR, I can work around it in the future

crates/aranya-daemon/src/daemon.rs
```diff
@@ -244,16 +204,48 @@ impl Daemon {
...
+        let path = self.cfg.daemon_api_pk_path();
+        match try_read_cbor::<PublicApiKey<E::CS>>(&path).await? {
+            Some(pk) => {
+                let id = pk.id()?;
+                let sk = store
+                    .get_key::<E, ApiKey<E::CS>>(eng, id.into())?
+                    // If the public API key exists then the
+                    // secret half should exist the keystore. If
+                    // not, then something deleted it frmo the
+                    // keystore.
```
ssweetney-spok, line 231: If this scenario does occur, the user has to delete the file in order for a new key to be generated right?

crates/aranya-daemon/src/api.rs
```diff
@@ -976,94 +560,81 @@ impl DaemonApi for DaemonApiHandler {
...
+                            .assume("`psk_length_in_bytes` is out of range")?,
+                    };
+                    let psk = self.aqc.bidi_channel_received(&info).await?;
+                    let net_id = self
+                        .aqc
+                        .find_net_id(graph, e.author_id.into())
+                        .await
+                        .context("missing net identifier for channel author")?;
+                    // TODO(eric): Why do we ignore the remaining
+                    // commands in `ctrl`?
```
jdygert-spok, line 607: The action should only produce one ephemeral command. We should probably be more explicit about that / handle it earlier.

crates/aranya-daemon-api/src/crypto/txp.rs
```diff
@@ -0,0 +1,887 @@
...
+            .into_raw_parts()
+            .assume("should be able to decompose `SendCtx`")?;
+
+        Ok(Self {
+            seal: SealCtx::new(&seal_key, &seal_nonce, Seq::ZERO)?,
+            open: OpenCtx::new(&open_key, &open_nonce, Seq::ZERO)?,
+        })
+    }
+
+    fn make_ad(base: &[u8; 14], seq: Seq) -> [u8; 8 + 14] {
```
ssweetney-spok, line 88: Could you create a constant for base ad length? I think it might make it easier to follow

Maybe something like

```Rust
const BASE_AD_LEN: usize = 14;

const _: () = {
    assert!(std::mem::size_of_val(SERVER_BASE_AD) == BASE_AD_LEN);
    assert!(std::mem::size_of_val(CLIENT_BASE_AD) == BASE_AD_LEN);
};

Comment (ssweetney-spok) [reply crates/aranya-daemon-api/src/crypto/txp.rs:88]: And possibly `std::mem::size_of::<Seq>()` instead of 8

File: crates/aranya-daemon-api/src/crypto/txp.rs [hunk 9]
```diff
@@ -0,0 +1,887 @@
...
+    /// For rekeying.
+    rng: R,
+    /// The server's public key.
+    pk: PublicApiKey<CS>,
+    /// The "info" parameter when rekeying.
+    info: Box<[u8]>,
+    ctx: Option<Ctx<CS>>,
+    /// The number of times we've rekeyed, including the initial
+    /// keying.
+    rekeys: usize,
```
ssweetney-spok, line 192: Is this solely for debugging purposes?

nikki-aranya:
> Outside of the verbosity of having to implement `Stream` and `Sink` for the new transport related types, creating a custom Tarpc transport doesn't seem too bad. Definitely feels like the right solution to this problem

This is just kicking replacing tarpc down the road but I'd rather have this scaffolding once we address #77 since it'll be easier to cook up an alternative, I'm mostly just worried about encoding tarpc's quirks inside all this.

crates/aranya-daemon-api/Cargo.toml
```diff
@@ -15,16 +15,27 @@ workspace = true
...
 anyhow = { workspace = true }
+bytes = { version = "1", features = ["serde"] }
+ciborium = { workspace = true }
+futures-util = { workspace = true }
+pin-project = { version = "1.1" }
 postcard = { workspace = true }
 serde = { workspace = true }
 tarpc = { workspace = true }
+tokio = { workspace = true }
+tokio-serde = { version = "0.9", features = ["messagepack"] }
```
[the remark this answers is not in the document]
nikki-aranya: I'm not convinced there's any alternative to postcard that isn't materially worse in some way, unless your goal is zero-copy in which case we have rkyv already somewhat integrated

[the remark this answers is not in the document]
ssweetney-spok: That's another good option. I think we'd have to write our own codec that uses postcard since it's not listed in the formats [here](https://docs.rs/tokio-serde/latest/tokio_serde/formats/index.html)

crates/aranya-daemon-api/src/crypto/txp.rs
```diff
@@ -0,0 +1,887 @@
...
+struct Data {
+    seq: u64,
+    ciphertext: BytesMut,
+    tag: Bytes,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+struct Rekey {
+    enc: Bytes,
+}
```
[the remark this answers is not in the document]
elagergren-spideroak: `ciphertext` is `BytesMut` because it simplifies the code (fewer conversions). Everything else is `Bytes` because it's the more appropriate type. But it doesn't really matter since this is all internal and the buffers aren't ever shared.

crates/aranya-daemon/src/config.rs
```diff
@@ -29,11 +29,12 @@ pub struct Config {
     pub sync_addr: Addr,
 
     /// AFC configuration.
-    pub afc: AfcConfig,
+    pub afc: Option<AfcConfig>,
+
+    /// AQC configuration.
+    pub aqc: AqcConfig,
```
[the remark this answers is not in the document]
elagergren-spideroak: I honestly don't know what AQC config options we'll have. That we don't have any right now is telling, I think, because AFC had a bunch of options right from the get go.

We could just make it optional ¯\_(ツ)_/¯

crates/aranya-daemon/src/api.rs
elagergren-spideroak: yeah, the code is definitely smelly without that context.

crates/aranya-client/src/client.rs
```diff
@@ -60,159 +54,158 @@ impl Labels {
...
+            .map_err(IpcError::new)?
+            .map_err(aranya_error)?;
+        let want = env!("CARGO_PKG_VERSION");
+        if got != want {
+            return Err(IpcError::new(io::Error::new(
+                io::ErrorKind::Unsupported,
+                format!("version mismatch: `{got}` != `{want}`"),
+            ))
+            .into());
+        }
```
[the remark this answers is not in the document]
elagergren-spideroak: Well, major and probably minor should match. Everything else is irrelevant. So, I guess I'll drop this then.

crates/aranya-daemon-api/Cargo.toml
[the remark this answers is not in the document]
elagergren-spideroak: > Is MessagePack better than Bincode for our use case?

I don't care which we use. I chose MessagePack b/c we were previously using JSON. But we can use whatever.

crates/aranya-daemon/src/daemon.rs
elagergren-spideroak: Yeah. The daemon public key setup kinda sucks right now. It's not the most user friendly.

crates/aranya-daemon-api/src/crypto/txp.rs
elagergren-spideroak: Mostly, yeah.

elagergren-spideroak:
> Outside of the verbosity of having to implement `Stream` and `Sink` for the new transport related types, creating a custom Tarpc transport doesn't seem too bad. Definitely feels like the right solution to this problem

We could probably simplify it a little bit more by taking a `tarpc::Transport` instead of `AsyncRead + AsyncWrite`, but I don't want to spend more time experimenting.

crates/aranya-daemon-api/src/crypto/txp.rs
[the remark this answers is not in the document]
jdygert-spok: Oh I missed that `tokio-serde` uses them.

crates/aranya-daemon-api/src/crypto/txp.rs
elagergren-spideroak: I've added some more docs. Lmk if they're sufficient.

elagergren-spideroak:
> EDIT: Looking at this further, with how much of a wrecking ball it is, it's probably better to just rewrite any AFC stuff anyways so meh

Ordinarily I'd split this up into multiple PRs, but I just don't have time right now. :/

jdygert-spok (approved):

crates/aranya-daemon-api/Cargo.toml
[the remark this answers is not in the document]
ssweetney-spok: We can decide if it's worth changing after we add benchmarks

crates/aranya-client-capi/src/imp/config.rs
jdygert-spok, line 1: Many added `#[repr(C)]`s here from merge

crates/aranya-daemon/src/main.rs
```diff
@@ -40,16 +40,25 @@ fn main() -> Result<()> {
     rt.block_on(async {
         let daemon = Daemon::load(cfg).await.context("unable to load daemon")?;
         info!("loaded Aranya daemon");
+
+        if flags.print_api_pk {
+            let pk = daemon.public_api_key().await?.encode()?;
+            println!("{}", hex::encode(pk));
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
elagergren-spideroak: fwiw I chose hex because it's much simpler to decode a hex string than it is base64, so i figured it'd be more user friendly (other than being much larger than necessary). Base64 is more normal for encoded public keys, though.

crates/aranya-daemon/src/config.rs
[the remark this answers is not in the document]
jdygert-spok: Let's make `aqc` optional for now.

crates/aranya-client/src/client.rs
```diff
@@ -48,91 +55,162 @@ impl Labels {
...
-        let aqc = AqcChannelsImpl::new(device_id, keystore_info).await?;
+        let got = daemon
+            .version(context::current())
+            .await
+            .map_err(IpcError::new)?
+            .map_err(aranya_error)?;
+        let want = Version::parse(env!("CARGO_PKG_VERSION"))
+            .map_err(|err| io::Error::new(io::ErrorKind::Other, err))
+            .map_err(IpcError::new)?;
+        if got.major != want.major || got.minor != want.minor {
```
nikki-aranya, line 157: What level of backcompat are we wanting to offer? If/when we hit 1.0, it might be nice to allow backcompat on minor versions even if it requires some manual conversion.

elagergren-spideroak: Yeah, once we're 1.x we should probably do that.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -165,26 +161,18 @@ enum AranyaError
...
      * Could not send request to daemon.
      */
-    ARANYA_ERROR_RPC,
+    ARANYA_ERROR_IPC,
     /**
-     * Daemon reported error.
+     * An Aranya error.
      */
-    ARANYA_ERROR_DAEMON,
+    ARANYA_ERROR_ARANYA,
```
gknopf-aranya, line 171: All these errors are aranya errors. What does `ARANYA_ERROR_ARANYA` mean when there's already a `ARANYA_ERROR_OTHER` for unclassified errors?

gknopf-aranya (approved):
lgtm. Thank you for refactoring the examples.

crates/aranya-client-capi/output/aranya-client.h
elagergren-spideroak: It's for policy errors. "Daemon" is wrong, "Aranya" is less wrong. We should figure out the correct name on another PR, though, to avoid going back and forth about it and blocking this PR.

ssweetney-spok (approved):

gknopf-aranya (approved):

crates/aranya-daemon/src/daemon.rs
```diff
@@ -266,7 +386,14 @@ mod tests {
...
             pid_file: work_dir.join("pid"),
             sync_addr: any,
-            afc: None,
+            afc: Some(AfcConfig {
+                shm_path: "/test_daemon1".to_owned(),
+                unlink_on_startup: true,
+                unlink_at_exit: true,
+                create: true,
+                max_chans: 100,
+            }),
```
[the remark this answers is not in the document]
gknopf-aranya: PR looks good to me after reverting this

gknopf-aranya:
@elagergren-spideroak this PR has been approved but cannot be merged due to the open conversations.

PR #209, opened 2025-04-18
WIP: aqc: delete chans

gknopf-aranya:
Delete AQC channels:
- Remove the PSK encap from the keystore
- Delete the AQC PSK (this will most likely be done on a different PR since right now the PSK isn't used anywhere so it's discarded right after it's derived from the encap)
- Remove channel objects from memory
- Cleanup any networking resources associated with the QUIC channels (this will most likely be added on Ben's AQC PR)

### DIFF
--- a/crates/aranya-client/src/aqc.rs
+++ b/crates/aranya-client/src/aqc.rs
@@ -1,6 +1,6 @@
 //! AQC support.
 
-use std::{io, net::SocketAddr, path::PathBuf};
+use std::{collections::BTreeMap, io, net::SocketAddr, path::PathBuf};
 
 use anyhow::{anyhow, bail, Context};
 use aranya_aqc_util::{
@@ -8,7 +8,10 @@ use aranya_aqc_util::{
 };
 use aranya_crypto::{
     aead::Aead,
-    aqc::{BidiChannelId, BidiPeerEncap, UniChannelId, UniPeerEncap},
+    aqc::{
+        BidiAuthorSecretId, BidiChannelId, BidiPeerEncap, UniAuthorSecretId, UniChannelId,
+        UniPeerEncap,
+    },
     default::DefaultEngine,
     generic_array::GenericArray,
     import::Import,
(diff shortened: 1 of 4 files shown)


crates/aranya-client/src/aqc.rs
```diff
@@ -205,25 +220,37 @@ impl<'a> AqcChannels<'a> {
     // It is an error if the channel does not exist
     #[instrument(skip_all, fields(chan = %chan))]
     pub async fn delete_bidi_channel(&mut self, chan: AqcBidiChannelId) -> crate::Result<()> {
-        let _ctrl = self
-            .client
+        let Some(secret_id) = self.bidi_chans.remove(&chan) else {
```
benz-spideroak, line 223: What's the purpose of this check? Is it to prevent calling delete_aqc_bidi_channel twice if delete_bidi_channel is called twice?

benz-spideroak: It seems like it'd be simpler to only delete the key in delete_aqc_bidi_channel if the key exists in the keystore.

gknopf-aranya: This removes the entry in the map where the channel secret is stored.
It could be used to cleanup any resources associated with the channel such as the channel id.

The keystore is in the daemon, so the client wouldn't know whether the PSK encap is in the keystore without querying first.

gknopf-aranya: Now that I'm thinking about this more, I'd rather delete the PSK encap from the daemon keystore after the encrypted PSK is sent to the client via the UDS IPC.
Since the keystore won't need to store that anymore after the client has it.
Then the client only needs to worry about deleting it's own copy of the PSK.

benz-spideroak: What I'm asking is why you're creating a map to store the channel secrets. Are they used for anything other than this check?

gknopf-aranya: Yes, they are used to make an IPC call into the daemon that tells it to delete the corresponding PSK encap from its keystore. That entry in the keystore was created when processing the channel creation commands.

crates/aranya-client/src/aqc.rs
```diff
@@ -68,11 +71,17 @@ impl AqcChannelsImpl {
 /// sending data between peers.
 pub struct AqcChannels<'a> {
     client: &'a mut crate::Client,
+    bidi_chans: BTreeMap<AqcBidiChannelId, BidiAuthorSecretId>,
```
benz-spideroak, line 74: This info belongs within the channel struct. AqcChannelSender for uni and AqcBidirectionalChannel for bidi. You'll call `close()` on that struct to call the channel and it'll hold the author secret. The user won't have the channel id they'll have the channel struct.

gknopf-aranya:
Closing this PR since it overlaps with AQC C API and networking changes.

PR #212, opened 2025-04-18
syncer: persist peer caches

### LINKED ISSUE #87: sync: persist PeerCache
Currently, the PeerCache is discarded after each sync. Implement the ability to persist the PeerCache.

https://github.com/aranya-project/aranya/blob/main/crates/aranya-daemon/src/aranya.rs#L116

cc @jdygert-spok @nikki-aranya @benz-spideroak

ssweetney-spok:
Resolves #87

Keep an `aranya_util::addr::Addr`->`aranya_runtime::sync::responder::PeerCache` map that is shared between the `aranya_daemon::aranya::Client` and `aranya_daemon::aranya::Server` instances


crates/aranya-daemon/src/aranya.rs
```diff
@@ -41,14 +41,18 @@ pub enum SyncResponse {
 pub struct Client<EN, SP, CE> {
     /// Thread-safe Aranya client reference.
     aranya: Arc<Mutex<ClientState<EN, SP>>>,
+    /// Thread-safe Peer Cache reference.
+    /// Lock must be aquired after [`Self::aranya`]
+    peer_cache: Arc<Mutex<PeerCache>>,
```
ssweetney-spok, line 46: I went back and forth trying to decide if it was worth putting the client state and peer cache in the same tuple:

```
type ClientStatePeerCache<EN, SP> = (ClientState<EN, SP>, PeerCache);
 
Arc<Mutex<ClientStatePeerCache<EN, SP>>>
 ```
 
 If `Client::sync_peer`and `Server::sync_respond` are ran around the same time, we want to ensure that the locks are acquired in the same order to prevent deadlocks. I'm thinking that since there's only two locks it should be easy to catch during code review but I'm curious as to what others think

Comment (ssweetney-spok) [reply crates/aranya-daemon/src/aranya.rs:46]: Also, the comment isn't completely accurate since we could just flip the order to have the peer cache lock be acquired first and then the client state lock afterwards

**Comment by gknopf-aranya** [comment]
Adding Ben as a reviewer since he implemented PeerCache initially.

File: crates/aranya-daemon/src/aranya.rs [hunk 2]
```diff
@@ -366,7 +382,7 @@ where
             .poll(
                 &mut buf,
                 client.lock().await.provider(),
-                &mut PeerCache::new(),
+                &mut *peer_cache.lock().await,
```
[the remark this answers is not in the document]
gknopf-aranya: Without some measurements on how long this normally takes, it's difficult to determine which approach is faster.

crates/aranya-daemon/src/aranya.rs
gknopf-aranya: I like keeping it as a separate field unless we use an alias for the tuple or add this to the `ClientState` definition.

gknopf-aranya:
If the client and server each had their own cache, we wouldn't need to use acquire a `Mutex` to use the cache which would improve performance. The problem with that is that the client and server cache could have information that the other doesn't. What if we resolve cache inconsistencies asynchronously when we aren't in the middle of syncing?

crates/aranya-daemon/src/aranya.rs
```diff
@@ -252,6 +257,9 @@ where
 pub struct Server<EN, SP> {
     /// Thread-safe Aranya client reference.
     aranya: Arc<Mutex<ClientState<EN, SP>>>,
+    /// Thread-safe Peer Cache reference.
+    /// Lock must be aquired after [`Self::aranya`]
+    peer_cache: Arc<Mutex<PeerCache>>,
```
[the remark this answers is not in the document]
ssweetney-spok: Oh, gotcha. Should the `PeerCache`s be shared between the sync server (`aranya_daemon::aranya::Server`) and aranya client (`aranya_daemon::aranya::Client`)?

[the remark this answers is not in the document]
ssweetney-spok: Wondering if it's even worth wrapping `BTreeMap<SocketAddr, PeerCache>` with `Arc<Mutex...`

crates/aranya-daemon/src/aranya.rs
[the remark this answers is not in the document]
jdygert-spok: Given that we have to lock the client anyways, there's no case where the peer cache would be under contention.

crates/aranya-daemon/src/aranya.rs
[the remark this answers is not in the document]
jdygert-spok: Yes, they should be shared.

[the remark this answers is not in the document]
gknopf-aranya: I think you'd want this so the entire map doesn't need to be locked:
`BTreeMap<SocketAddr, Arc<Mutex<PeerCache>>>`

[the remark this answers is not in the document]
ssweetney-spok: We'd probably want a concurrent hashmap then if we want something like that. I don't think we need this because you can't do anything useful with the rest of the `PeerCache`s since we only have a single client instance

crates/aranya-daemon/src/aranya.rs
```diff
@@ -41,14 +44,21 @@ pub enum SyncResponse {
 pub struct Client<EN, SP, CE> {
     /// Thread-safe Aranya client reference.
     aranya: Arc<Mutex<ClientState<EN, SP>>>,
+    /// Thread-safe Peer Cache reference.
+    /// Lock must be aquired after [`Self::aranya`]
+    peer_caches: Arc<Mutex<BTreeMap<Addr, PeerCache>>>,
```
gknopf-aranya, line 49: Since this is the only cache for this object, the field name could be more concise:
```suggestion
    cache: Arc<Mutex<BTreeMap<Addr, PeerCache>>>,
```

ssweetney-spok: Shortened it to "caches" since there are multiple. We're keeping [one per peer](https://github.com/aranya-project/aranya/pull/212#discussion_r2050987534)

gknopf-aranya (approved):

gknopf-aranya (approved):

crates/aranya-daemon/src/aranya.rs
```diff
@@ -298,9 +318,12 @@ where
             debug!(?addr, "received sync request");
 
             let client = Arc::clone(&self.aranya);
+            let peer_caches = Arc::clone(&self.caches);
```
benz-spideroak, line 321: nit: Simpler would be self.caches.clone()

jdygert-spok: It's an idiom to highlight that it is a cheap/sharing clone.

benz-spideroak (dismissed):

crates/aranya-daemon/src/aranya.rs
```diff
@@ -119,7 +131,7 @@ where
             if !cmds.is_empty() {
                 let mut client = self.aranya.lock().await;
                 let mut trx = client.transaction(id);
-                // TODO: save PeerCache somewhere.
+
                 client
                     .add_commands(&mut trx, sink, &cmds)
                     .context("unable to add received commands")?;
```
nikki-aranya, line 137: We need to `update_heads()` below for the PeerCache to actually do anything afaik, otherwise it never gets any locations stored inside it. Writing a test that actually checks the contents of the PeerCache would probably be really useful.

nikki-aranya (changes requested):
I'd like to see a test to make sure this actually does anything before this gets merged.

gknopf-aranya:
> I'd like to see a test to make sure this actually does anything before this gets merged.

Yeah, would be nice to know it works and improves performance first. Thinking it could be better to slate this after benchmarks are in.

nikki-aranya:
Honestly, setting up benchmarks once you write a test is basically....10 lines of code? Add divan as a toml dependency is two lines, create a benches/ folder and a .rs file, add
```rs
#[global_allocator]
static ALLOC: divan::AllocProfiler = divan::AllocProfiler::system();

fn main() {
    divan::main();
}
```
to it is 6 lines, and then mark any function, doesn't even have to be in benches, you want to test with `#[divan::bench]` is one line, and boom, you have a benchmark on a test. In order to test its performance, you'd probably have to make a function that takes a T, that's either PeerCache or a stubbed version, and measure the difference.

crates/aranya-daemon/src/aranya.rs
ssweetney-spok: > otherwise it never gets any locations stored inside it

It's seems like it does when `aranya_runtime::sync::responder::SyncResponder::poll` is called

https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-runtime/src/sync/responder.rs#L213-L217

crates/aranya-daemon/src/aranya.rs
```diff
@@ -254,6 +266,9 @@ where
 pub struct Server<EN, SP> {
     /// Thread-safe Aranya client reference.
     aranya: Arc<Mutex<ClientState<EN, SP>>>,
+    /// Thread-safe reference to an [`Addr`]->[`PeerCache`] map.
+    /// Lock must be aquired after [`Self::aranya`]
```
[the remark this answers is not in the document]
ssweetney-spok: I had a similar concern and posted about an alternative in a [comment](https://github.com/aranya-project/aranya/pull/212#discussion_r2050965022). With a tuple, we'd just be locking and unlocking both at the same time.

If we do add more locks and decide we need more fine-grained locking then we could could get fancy with something like [compile-time lock ordering](https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/src/connectivity/network/netstack3/core/lock-order/src/lib.rs)

gknopf-aranya:
Closing in favor of:
https://github.com/aranya-project/aranya/issues/296

We're moving away from a TCP syncer to the QUIC syncer.

PR #213, opened 2025-04-18
WIP: custom roles

gknopf-aranya:


crates/aranya-daemon/src/policy.md
```diff
@@ -111,13 +102,35 @@ struct KeyIds {
...
+}
 ```
 
 ### Facts
 
 ```policy
-// A device on the team.
-fact Device[device_id id]=>{role enum Role, sign_key_id id, enc_key_id id}
+// Devices on the team.
+fact Devices[device_id id]=>{device struct Device}
```
elagergren-spideroak, line 133: I don't understand the plural. I know it's because of struct Device, but plural doesn't make sense since this is one device, not many.

crates/aranya-daemon/src/policy.md
```diff
@@ -128,6 +141,16 @@ fact DeviceSignKey[device_id id]=>{key_id id, key bytes}
...
 
+// A ID-based role on the team.
+fact Roles[role_id id]=>{role struct Role}
+
+// Records that a role was assigned to a device.
+fact AssignedRole[role_id id, device_id id]=>{}
+
+// Records that a role was assigned permission to execute a certain command.
+// TODO: use enum permission key.
+fact PermRole[perm string]=>{role_id id}
```
elagergren-spideroak, line 152: CmdRequiresRole instead of PermRole?

crates/aranya-daemon/src/policy.md
```diff
@@ -242,6 +245,48 @@ function open_envelope(sealed_envelope struct Envelope) bytes {
...
+finish function assign_perm_role(perm string, role_id id) {
+    create PermRole[perm: perm]=>{role_id: role_id}
+}
+
+finish function revoke_perm_role(perm string) {
+    delete PermRole[perm: perm]
+}
+
+// Check if device has permission to execute the current command.
+function device_has_permission(device_id id, perm string) bool {
```
elagergren-spideroak, line 268: device_has_permission -> device_can_publish_cmd

crates/aranya-daemon/src/policy.md
```diff
@@ -242,6 +245,48 @@ function open_envelope(sealed_envelope struct Envelope) bytes {
...
+    return exists AssignedRole[role_id: role_id, device_id: device_id]
+}
+
+// Check if the author device has higher priority than the target device.
+// This means the author has permission to execute commands on the target.
+function author_dominates_target(author_id id, target_id id) bool {
+    // Check if the device has higher priority than the target device.
+    let author_device = unwrap query Devices[device_id: author_id]
+    let target_device = unwrap query Devices[device_id: target_id]
+    if author_device.device.priority >= target_device.device.priority {
```
elagergren-spideroak, line 284: I think we want > not >= ?

crates/aranya-daemon/src/policy.md
```diff
@@ -401,8 +446,41 @@ command CreateTeam {
...
+
+            // TODO: assign default permissions to owner role.
+            assign_perm_role("AddMember", role.role_id)
+            assign_perm_role("RemoveMember", role.role_id)
+            assign_perm_role("CreateRole", role.role_id)
+            assign_perm_role("DeleteRole", role.role_id)
+            assign_perm_role("AssignRole", role.role_id)
+            assign_perm_role("RevokeRole", role.role_id)
+            assign_perm_role("AssignRolePermission", role.role_id)
+            assign_perm_role("RevokeRolePermission", role.role_id)
```
elagergren-spideroak, line 480: Maybe this should be part of the CreateTeam command or something? Like in the data fields

crates/aranya-daemon/src/policy.md
gknopf-aranya: I needed to create a struct to store device info that called `struct Device`.
The name of the struct and fact need to be different so I renamed the fact to `Devices` since it stores the device IDs.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Changed it to `>` unless we have a use for devices executing commands on themselves.

crates/aranya-daemon/src/policy.md
djohnson-spideroak: does it need to be `struct Device`? maybe `Device` for the fact, and `DeviceInfo` for the struct?

crates/aranya-daemon/src/policy.md
gknopf-aranya: Mentioning commands makes sense in the context of the policy.
In the context of custom roles and RBAC, using permissions terminology helps to document that these are RBAC permission controls.

crates/aranya-client-capi/src/imp/label.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub fn get_name(&self) -> *const c_char {
+        self.name.as_ptr()
+    }
+}
+
+impl From<aranya_daemon_api::Label> for Label {
+    fn from(value: aranya_daemon_api::Label) -> Self {
+        Self {
+            id: value.id,
+            name: CString::new(value.name).expect("expected to create string"),
```
jdygert-spok, line 38: hmm, I'm realizing all these user provided strings could have interior null bytes and break the C API (even if we use `Result`, it still won't _work_).

I think we'd have to validate them in the policy command blocks to fully prevent that.

crates/aranya-daemon/src/policy.md
gknopf-aranya: What does that buy us? This will be about as verbose either way.

crates/aranya-daemon/src/policy.md
jdygert-spok: I think self revocation came up before as slightly problematic anyway due to braiding?

crates/aranya-client-capi/src/imp/label.rs
gknopf-aranya: If we need to do this, I'd prefer to tackle it on another PR rather than adding additional scope to this PR.

crates/aranya-client/tests/tests.rs
```diff
@@ -170,54 +172,150 @@ impl TeamCtx {
...
 
+    async fn create_all_roles(&mut self, team_id: TeamId) -> Result<()> {
+        let mut owner_team = self.owner.client.team(team_id);
+        let roles = RoleCtx {
+            admin: owner_team.create_role("admin".into()).await?,
+            operator: owner_team.create_role("operator".into()).await?,
+            member: owner_team.create_role("member".into()).await?,
+        };
+
+        owner_team
```
gknopf-aranya, line 183: It would be nicer to have a static map with these permissions:
```
static DEFAULT_PERMS: LazyLock<BTreeMap<&str, &str>> = LazyLock::new(|| {
    let mut m = BTreeMap::new();
    m.insert("SetAqcNetworkName", "operator");
    ...
    m
});
```

gknopf-aranya: Created a map for default perms that's used in the test and Rust example.

crates/aranya-daemon/src/policy.md
```diff
@@ -401,8 +446,41 @@ command CreateTeam {
...
+            create_role(role)
+
+            // Assign default permissions to owner role.
+            assign_perm_role("AddMember", role.role_id)
+            assign_perm_role("RemoveMember", role.role_id)
+            assign_perm_role("CreateRole", role.role_id)
+            assign_perm_role("DeleteRole", role.role_id)
+            assign_perm_role("AssignRole", role.role_id)
+            assign_perm_role("RevokeRole", role.role_id)
+            assign_perm_role("AssignRolePermission", role.role_id)
```
gknopf-aranya, line 479: If the owner was allowed to execute commands on itself, we could bootstrap the `AssignRolePermission` here, then allow the owner to assign additional permissions to itself outside of the policy.

crates/aranya-daemon/src/policy.md
nikki-aranya: I'd still prefer to encode it in Rust than in the policy for anything that we possibly can, for the simple fact that the policy takes up more memory and cpu time

examples/c/example.c
```diff
@@ -587,6 +674,208 @@ AranyaError run_aqc_example(Team *t) {
...
+    return err;
+}
+
+// Query device roles. Returned `roles` ptr must be freed.
+AranyaError query_device_roles(Team *t, AranyaDeviceId *device,
+                               AranyaRole **roles, size_t *roles_len) {
+    AranyaError err;
+
+    *roles_len = BUF_LEN;
+    *roles     = malloc(*roles_len * sizeof(AranyaDeviceId));
```
jdygert-spok, line 851:
```suggestion
    *roles     = malloc(*roles_len * sizeof(AranyaRole));
```

or use the expression version to help avoid this:

```suggestion
    *roles     = malloc(*roles_len * sizeof(**roles));
```

crates/aranya-client-capi/src/imp/label.rs
```diff
@@ -0,0 +1,46 @@
...
+    mem::MaybeUninit,
+};
+
+use aranya_capi_core::safe::{TypeId, Typed};
+use aranya_daemon_api::LabelId;
+
+/// A label that can be assigned to a device.
+pub struct Label {
+    id: LabelId,
+    name: CString,
```
jdygert-spok, line 12: You can just use `pub(crate)` fields for each of these types

crates/aranya-daemon/src/policy.md
```diff
@@ -711,38 +769,56 @@ command AssignAdmin{
...
-        check device.role != Role::Admin
+        check device_has_permission(author.device_id, "DeleteRole")
+
+        // Query role.
+        let role = check_unwrap query Roles[role_id: this.role_id]
 
         finish {
-            assign_role(device, Role::Admin)
+            // Revoke role from all devices.
+            delete AssignedRole[role_id: role.role.role_id, device_id: ?]
```
gknopf-aranya, line 780: @jdygert-spok, @elagergren-spideroak, @ycarmy after cascading the `AssignedRole` delete in `DeleteRole`, I expected the owner to no longer be able to delete any other roles after deleting its own role. However, it appears it can still delete other roles based on this C example output:
`delete AssignedRole[role_id: role.role.role_id, device_id: ?]`
```
deleting role: admin
2025-04-24T19:02:02.390964Z  INFO RPC{rpc.deadline=2025-04-24T19:02:12.390864125Z otel.kind="client" otel.name=DaemonApi.delete_role rpc.trace_id=00}: tarpc::client: SendRequest
2025-04-24T19:02:02.394243Z  INFO RPC{rpc.deadline=2025-04-24T19:02:12.390864125Z otel.kind="client" otel.name=DaemonApi.delete_role rpc.trace_id=00}: tarpc::client: ReceiveResponse
deleting role: owner
2025-04-24T19:02:02.394345Z  INFO RPC{rpc.deadline=2025-04-24T19:02:12.394286833Z otel.kind="client" otel.name=DaemonApi.delete_role rpc.trace_id=00}: tarpc::client: SendRequest
deleting role: member
2025-04-24T19:02:02.396205Z  INFO RPC{rpc.deadline=2025-04-24T19:02:12.394286833Z otel.kind="client" otel.name=DaemonApi.delete_role rpc.trace_id=00}: tarpc::client: ReceiveResponse
2025-04-24T19:02:02.396254Z  INFO RPC{rpc.deadline=2025-04-24T19:02:12.396230042Z otel.kind="client" otel.name=DaemonApi.delete_role rpc.trace_id=00}: tarpc::client: SendRequest
2025-04-24T19:02:02.398207Z  INFO RPC{rpc.deadline=2025-04-24T19:02:12.396230042Z otel.kind="client" otel.name=DaemonApi.delete_role rpc.trace_id=00}: tarpc::client: ReceiveResponse
deleting role: operator
2025-04-24T19:02:02.398266Z  INFO RPC{rpc.deadline=2025-04-24T19:02:12.398227042Z otel.kind="client" otel.name=DaemonApi.delete_role rpc.trace_id=00}: tarpc::client: SendRequest
cleaning up the Aranya team 
```

gknopf-aranya: Issue has been opened to resolve this bug:
https://github.com/aranya-project/aranya-core/issues/229

crates/aranya-client-capi/src/imp/label.rs
gknopf-aranya: Is that needed? Usually I'd only make something `pub` if the compiler complains because I tried to use it from another module or if I want it to be part of the API.

jdygert-spok: I mean instead of having all the getters/setters below.

crates/aranya-client-capi/src/imp/label.rs
gknopf-aranya: https://github.com/aranya-project/aranya/issues/231

crates/aranya-daemon/src/policy.md
gknopf-aranya: These currently need to be part of bootstrapping the owner because the `author_dominates_target` check will not allow the owner to invoke any commands on itself.

crates/aranya-client-capi/src/imp/label.rs
elagergren-spideroak: please use TryFrom and use `buggy` instead of `expect`

gknopf-aranya:
@elagergren-spideroak, @jdygert-spok, @djohnson-spideroak, @ycarmy, @ssweetney-spok, @nikki-aranya I'm ready for a final look at this PR

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -835,6 +881,64 @@ AranyaError aranya_get_device_id_ext(struct AranyaClient *client,
...
+ * Compare two device IDs.
+ *
+ * @param device_id_a first device ID to compare [`AranyaDeviceId`](@ref AranyaDeviceId).
+ * @param device_id_b second device ID to compare [`AranyaDeviceId`](@ref AranyaDeviceId).
+ *
+ * @param __output boolean representing whether the device IDs are equal.
+ *
+ * Returns true if device IDs match. Returns false otherwise.
+ */
+AranyaError aranya_cmp_device_ids(const struct AranyaDeviceId *device_id_a,
```
jdygert-spok, line 894: These can just be `memcmp` on the C side.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -925,25 +1106,48 @@ pub fn create_label(
...
-/// Permission to perform this operation is checked against the Aranya policy.
+/// @param role the role [`Role`].
 ///
-/// @param client the Aranya Client [`Client`].
-/// @param team the team's ID [`TeamId`].
-/// @param label_id the channel label ID [`LabelId`] to delete.
+/// Returns the role's ID [`RoleId`].
+pub fn get_role_id(role: &Role) -> Result<RoleId, imp::Error> {
+    Ok(role.id.into())
+}
```
jdygert-spok, line 1116: Could make each of these getters infallible so they don't use out params. Also I guess they should end up named as `aranya_role_get_id` etc

```rust
/// @relates AranyaRole.
pub fn role_get_id(role: Option<&Role>) -> Option<&RoleId> {
    role.map(|r| &r.id)
}

/// @relates AranyaRole.
pub fn role_get_name(role: Option<&Role>) -> *const c_char {
    role.map_or(ptr::null(), |r| r.name.as_ptr())
}
```

crates/aranya-client/src/client.rs
```diff
@@ -48,6 +48,52 @@ impl Labels {
...
+    m.insert("SetAqcNetworkName", "operator");
+    m.insert("UnsetAqcNetworkName", "operator");
+    m.insert("CreateLabel", "operator");
+    m.insert("AssignLabel", "operator");
+    m.insert("RevokeLabel", "operator");
+    m.insert("DeleteLabel", "admin");
+    m.insert("AqcCreateBidiChannel", "member");
+    m.insert("AqcCreateUniChannel", "member");
+    m
+});
```
jdygert-spok, line 95: Or just slice

```rust
// Default role commands.
pub static DEFAULT_CMDS: &[(&str, &str)] = &[...];
```

Maybe we'd rather have a setup_default_roles method/action? IDK

examples/c/example.c
```diff
@@ -407,6 +436,65 @@ AranyaError run(Team *t) {
...
+    for (size_t i = 0; i < roles_len; i++) {
+        AranyaRole role_result = roles[i];
+        const char *role_str   = NULL;
+        err                    = aranya_get_role_name(&role_result, &role_str);
+        EXPECT("unable to get role name", err);
+        printf("role: %s at index: %zu/%zu \r\n", role_str, i, roles_len);
+
+        // TODO: get_role_id()
+    }
+    free(roles);
```
ssweetney-spok, line 458: Like the `Cmd` type, the string data should also be de-allocated properly

crates/aranya-client/src/client.rs
ssweetney-spok: >  Maybe we'd rather have a setup_default_roles method/action?

Yea, I think it'd be better to keep that defined in the policy and have the owner call it

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: I think an arg with the type `Option<&T>` will make the function fallible anyway

gknopf-aranya: Is there a reason I can't return `-> Result<RoleId>` directly?

ssweetney-spok: I can't see a reason why you couldn't. Our pattern of returning results, e.g. `-> Result<T, E>`, generates a C function that looks kinda like `AranyaError fn_name(type_of_t *out_param)`

The API would be nicer though if we didn't use out params for the getters

crates/aranya-client/src/client.rs
gknopf-aranya: A role's ID is the ID of the command that created the role. So in order to initialize several roles in a `setup_default_roles` action, it would need to publish a different command for each role.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1325,36 +1478,73 @@ AranyaError aranya_create_label_ext(struct AranyaClient *client,
...
+ * Get name of role.
  *
- * Permission to perform this operation is checked against the Aranya policy.
+ * Returns a C string pointer to the role's name.
+ */
+AranyaError aranya_role_get_name(const struct AranyaRole *role,
+                                 const char **__output);
+
+/**
+ * Get ID of role.
```
djohnson-spideroak, line 1510: label

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1325,36 +1478,73 @@ AranyaError aranya_create_label_ext(struct AranyaClient *client,
...
+ * Returns the label's ID [`AranyaLabelId`](@ref AranyaLabelId).
  */
-AranyaError aranya_delete_label_ext(struct AranyaClient *client,
-                                    const struct AranyaTeamId *team,
-                                    const struct AranyaLabelId *label_id,
+AranyaError aranya_label_get_id(const struct AranyaLabel *label,
+                                struct AranyaLabelId *__output);
+
+/**
+ * Get ID of role.
```
djohnson-spideroak, line 1520: label

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -911,6 +1067,7 @@ pub unsafe fn aqc_remove_net_identifier(
 /// @param name label name string [`LabelName`].
 ///
 /// @relates AranyaClient.
+// TODO: return `Label`
```
djohnson-spideroak, line 1070: do we have an issue for this?

crates/aranya-client/tests/tests.rs
```diff
@@ -170,54 +172,132 @@ impl TeamCtx {
...
+        let role_list = [&roles.admin, &roles.operator, &roles.member];
+
+        // Assign commands to roles.
+        for (cmd, role_name) in DEFAULT_CMDS.iter() {
+            for role in &role_list {
+                if *role_name == role.name {
+                    owner_team.assign_role_cmd(role.id, cmd.to_string()).await?;
+                }
+            }
+        }
```
djohnson-spideroak, line 191: nit: seems a bit inefficient, could use a map or something

djohnson-spideroak: its in the tests though, so dont hold the PR up on this

crates/aranya-client/tests/tests.rs
djohnson-spideroak, line 1: do we cover all of the new functions in the tests?

crates/aranya-client/tests/tests.rs
gknopf-aranya: This is iterating over a hash map

gknopf-aranya: Although I'm currently deleting this code in favor of initializing default roles inside the policy.

crates/aranya-daemon/src/api.rs
```diff
@@ -380,17 +382,55 @@ impl DaemonApi for DaemonApiHandler {
...
+        role: api::RoleId,
+    ) -> api::Result<()> {
+        // TODO: include role name in error context.
+        self.client
+            .actions(&team.into_id().into())
+            .delete_role(role)
+            .await
+            .context("unable to delete role from team")?;
+        Ok(())
+    }
```
djohnson-spideroak, line 420: maybe need to check the effect here?

crates/aranya-daemon/src/policy.md
```diff
@@ -602,52 +692,26 @@ finish function remove_device(device_id id) {
 - Members can only be removed by Operators and Owners.
 - Removing non-Members requires revoking their higher role so the device is made into a Member first.
 
+## AssignPriority
 
-## AssignRole
+Assign priority to a device on a team.
```
djohnson-spideroak, line 697: Is this meant to be stubbed?

crates/aranya-daemon/src/policy.md
```diff
@@ -711,38 +774,56 @@ command AssignAdmin{
...
         finish {
-            assign_role(device, Role::Admin)
+            // Revoke role from all devices.
+            delete AssignedRole[role_id: role.role.role_id, device_id: ?]
 
-            emit AdminAssigned {
-                device_id: this.device_id,
+            // TODO: revoke command permissions.
+            // Not strictly required because a role cannot be used if it's been deleted and has no devices assigned to it.
+            // Cleans up unused data from the factdb.
```
djohnson-spideroak, line 789: why is this a TODO? related to the bug above?

crates/aranya-daemon/src/policy.md
```diff
@@ -754,24 +835,42 @@ command AssignOperator{
...
+    // Name of the role.
+    name string,
+    // ID of the device that assigned the role.
+    author_id id,
+}
 ```
 
 **Invariants**:
 
+TODO: fix these docs
```
djohnson-spideroak, line 873: do we have an issue for fixing the policy docs after merge?

crates/aranya-daemon/src/policy.md
```diff
@@ -886,38 +968,57 @@ command RevokeAdmin{
...
+        cmd: cmd,
+    }
 }
 
-command RevokeOperator{
+command RevokeRoleCmd {
     fields {
-        // The revoked device's ID.
-        device_id id,
+        // ID of the role to revoke command from.
```
djohnson-spideroak, line 1018: typo

crates/aranya-daemon/src/policy.md
```diff
@@ -927,32 +1028,40 @@ command RevokeOperator{
...
+        check device_can_publish_cmd(author.device_id, "RevokeRoleCmd")
 
-        finish {
-            revoke_role(device)
+        // Query role.
+        let role = check_unwrap query Role[role_id: this.role_id]
 
-            emit OperatorRevoked {
-                device_id: this.device_id,
+        if this.role_id == role.role.role_id {
```
djohnson-spideroak, line 1037: isnt this check redundant? we already fetch by role_id, so I would assume the role ID is correct.

djohnson-spideroak (commented):
I dont see any major issues aside from open comments.

crates/aranya-client/src/client.rs
gknopf-aranya: I've implemented default roles in policy by publishing a command for each role.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I don't believe so. It would affect the API.

crates/aranya-client/src/client.rs
```diff
@@ -48,6 +48,52 @@ impl Labels {
...
+    let mut m = BTreeMap::new();
+    m.insert("SetAqcNetworkName", "operator");
+    m.insert("UnsetAqcNetworkName", "operator");
+    m.insert("CreateLabel", "operator");
+    m.insert("AssignLabel", "operator");
+    m.insert("RevokeLabel", "operator");
+    m.insert("DeleteLabel", "admin");
+    m.insert("AqcCreateBidiChannel", "member");
+    m.insert("AqcCreateUniChannel", "member");
+    m
```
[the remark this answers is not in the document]
gknopf-aranya: It should cover everything except the owner role commands. Note that I've moved the default role initialization into the policy now.

crates/aranya-daemon/src/policy.md
gknopf-aranya: I ended up not implementing this.
Priorities are assigned to devices once when they are added to the team.

examples/c/example.c
```diff
@@ -407,6 +436,65 @@ AranyaError run(Team *t) {
...
+    }
+    for (size_t i = 0; i < cmds_len; i++) {
+        AranyaCmd cmd_result = cmds[i];
+        const char *cmd_str  = NULL;
+        err                  = aranya_get_cmd_name(&cmd_result, &cmd_str);
+        EXPECT("unable to get cmd name", err);
+        printf("cmd: %s at index: %zu/%zu \r\n", cmd_str, i, cmds_len);
+    }
+
+    free(cmds);
```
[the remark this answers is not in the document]
gknopf-aranya: How would we do this?

crates/aranya-client/tests/tests.rs
gknopf-aranya: Policy action for initializing default roles has been added.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I pushed a commit to return label/role IDs and name strings when they are created rather than only returning the IDs.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I pushed a commit that returns the `RoleId` directly without using `Result`.
In the C app, the function still returns an error. But at least from the perspective of `defs.rs` the method is infallible.

crates/aranya-client/src/client.rs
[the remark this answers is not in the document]
gknopf-aranya: Confirming that the union of commands assigned to the owner role during team creation and commands assigned to the other roles covers all the commands. With the exception of the query commands, which all roles have permission to execute.

crates/aranya-daemon/src/policy.md
gknopf-aranya: I'm working on updating these docs on this PR

crates/aranya-daemon/src/policy.md
gknopf-aranya: Yes, as long as there isn't a bug in the policy.

crates/aranya-daemon/src/policy.md
gknopf-aranya: We don't have a way to lookup the `CmdRequiresRole` facts to delete from the `role_id` because the ID isn't part of the key:
`fact CmdRequiresRole[cmd string]=>{role_id id}`

gknopf-aranya: btw, delete role will not be included in the public API until the prefix query fact deletion bug is resolved.

crates/aranya-client/tests/tests.rs
gknopf-aranya: We did until `setup_default_roles()` was moved into the policy. Now `assign_role_cmd()` is no longer covered.

gknopf-aranya: I solved this in the C example by creating a dummy role and assigning a dummy command to it.

gknopf-aranya: Added a test for this.

crates/aranya-daemon/src/api.rs
gknopf-aranya: We shouldn't need to check the effect here. If role creation fails, an error would be returned.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Do we have a use-case for changing a device's priority after the device has been added to the team?

crates/aranya-daemon/src/policy.md
gknopf-aranya: Added more docs to this TODO

gknopf-aranya:
Resolved most of the feedback except for the string deallocation and dynamic priority assignment ones.

examples/c/example.c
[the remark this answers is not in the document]
gknopf-aranya: afaict, the `CString` ownership for roles/labels stays in Rust. So the `Drop` impl should take care of deallocating the `CString` for those objects.
If we had transferred ownership from Rust to the C application with `into_raw`, then we'd need to transfer ownership back to Rust with `from_raw` before deallocating the memory (or deallocate it from the C application).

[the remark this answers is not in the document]
ssweetney-spok: > afaict, the `CString` ownership for roles/labels stays in Rust.

The data is written to an "out param"

```
pub fn query_role_cmds(
    client: &mut Client,
    team: &TeamId,
    role: &RoleId,
    cmds: Option<&mut MaybeUninit<Cmd>>,
    cmds_len: &mut usize,
) -> Result<(), imp::Error> {
```

This is like telling the compiler that it's the caller that actually owns the data. In this case, this function is being called by C code.

> So the Drop impl should take care of deallocating the CString for those objects

It will. We just need to trigger it manually

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -406,6 +414,51 @@ impl NetIdentifier {
...
+
+impl CmdName {
+    unsafe fn as_underlying(self) -> Result<String, imp::Error> {
+        // SAFETY: Caller must ensure the pointer is a valid C String.
+        let cstr = unsafe { core::ffi::CStr::from_ptr(self.0) };
+        Ok(String::from(cstr.to_str()?))
+    }
+}
+
+/// A device priority.
```
elagergren-spideroak, line 454: This needs a new name. Commands also have priorities.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -740,27 +803,157 @@ pub fn close_team(client: &mut Client, team: &TeamId) -> Result<(), imp::Error>
...
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param client the Aranya Client [`Client`].
+/// @param team the team's ID [`TeamId`].
+/// @param role_id the role ID [`RoleId`] to assign a command to.
+/// @param cmd the command to assign to the role [`Cmd`].
+///
+/// @relates AranyaClient.
+pub fn assign_role_cmd(
```
elagergren-spideroak, line 846: This needs a new name.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -740,27 +803,157 @@ pub fn close_team(client: &mut Client, team: &TeamId) -> Result<(), imp::Error>
...
+    client.rt.block_on(
+        client
+            .inner
+            .team(team.into())
+            .revoke_role_cmd(role_id.into(), cmd),
+    )?;
+    Ok(())
+}
+
+/// Setup default roles on team.
```
elagergren-spideroak, line 897: Needs some more docs. Eg, why should you call this method?

elagergren-spideroak (changes requested):
Still reviewing

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: This is the first time I'm hearing about command priorities. What are command priorities used for?

examples/c/example.c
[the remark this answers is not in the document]
gknopf-aranya: This seems to have worked:
```
/// Cleanup dynamically allocated strings in role.
///
/// @param role the role [`Role`].
pub unsafe fn role_cleanup(role: OwnedPtr<Role>) {
    // SAFETY: Caller must ensure `role` is a valid object.
    unsafe { role.drop_in_place(); }
}
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Changing to "device precedence"

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Is this changing to `assign_role_op` or `assign_role_operation`?

crates/aranya-daemon/src/policy.md
gknopf-aranya: Implemented a `AssignDevicePrecedence` command in the policy.

gknopf-aranya:
Operation enums have been added to the API so they when operations enums are implemented in the policy, the changes will be backward compatible with the API.

gknopf-aranya:
@chip-so, @elagergren-spideroak the custom roles PR is ready for review

gknopf-aranya:
Added some unit tests to cover the revocation cases

crates/aranya-client-capi/src/imp/op.rs
```diff
@@ -0,0 +1,17 @@
+use aranya_capi_core::safe::{TypeId, Typed};
+
+/// An operation that a role can perform.
```
elagergren-spideroak, line 3: I don't understand why you have Op inside defs.rs and also Operation in defs.rs and also Op here. imp::Op isn't needed.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -324,27 +357,95 @@ impl From<&AqcUniChannelId> for aranya_daemon_api::AqcUniChannelId {
     }
 }
 
-/// An enum containing team roles defined in the Aranya policy.
+/// Valid operations that roles can perform.
+#[aranya_capi_core::opaque(size = 24, align = 8)]
+pub type Operation = Safe<imp::Op>;
```
elagergren-spideroak, line 362: Why do you need both Operation and Op?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -740,27 +874,154 @@ pub fn close_team(client: &mut Client, team: &TeamId) -> Result<(), imp::Error>
...
+    let name = unsafe { name.as_underlying() }?;
+
+    let r = client
+        .rt
+        .block_on(client.inner.team(team.into()).create_role(name))?;
+    Safe::init(role, r.clone().try_into()?);
+    Ok(())
+}
+
+/// Assign role operation.
```
elagergren-spideroak, line 907: I don't know what this method does. It needs a new name and/or better docs.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -740,27 +874,154 @@ pub fn close_team(client: &mut Client, team: &TeamId) -> Result<(), imp::Error>
...
 /// @param team the team's ID [`TeamId`].
+/// @param precedence is the device's precedence [`DevicePrecedence`].
 /// @param keybundle serialized keybundle byte buffer `KeyBundle`.
 /// @param keybundle_len is the length of the serialized keybundle.
 ///
 /// @relates AranyaClient.
 pub unsafe fn add_device_to_team(
     client: &mut Client,
     team: &TeamId,
+    precedence: &DevicePrecedence,
```
elagergren-spideroak, line 1013: This should be by value not by ref.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -909,40 +1196,110 @@ pub unsafe fn aqc_remove_net_identifier(
...
-    label_id: &LabelId,
+/// @param role the role [`Role`].
+///
+/// Returns a C string pointer to the role's name.
+#[aranya_capi_core::no_ext_error]
+pub fn role_get_name(role: &Role) -> *const c_char {
+    role.name.as_ptr()
+}
+
+/// Cleanup dynamically allocated strings in role.
```
elagergren-spideroak, line 1238: "Releases any resources used by the [`Role`]"

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -909,40 +1196,110 @@ pub unsafe fn aqc_remove_net_identifier(
...
+///
+/// @param label the label [`Label`].
+///
+/// Returns a C string pointer to the label's name.
+#[aranya_capi_core::no_ext_error]
+pub fn label_get_name(label: &Label) -> *const c_char {
+    label.name.as_ptr()
+}
+
+/// Cleanup dynamically allocated strings in label.
```
elagergren-spideroak, line 1267: Ditto about just saying "releasing resources," not specifying what those resources are.

elagergren-spideroak, line 1267: Ditto about just saying "releasing resources," not specifying what those resources are.

crates/aranya-daemon/src/api.rs
```diff
@@ -380,17 +405,71 @@ impl DaemonApi for DaemonApiHandler {
...
+        _: context::Context,
+        team: api::TeamId,
+        name: String,
+    ) -> api::Result<api::Role> {
+        let effects = self
+            .client
+            .actions(&team.into_id().into())
+            .create_role(name.clone())
+            .await
+            .context(format!("unable to create role on team: {}", name))?;
```
elagergren-spideroak, line 436: Use with_context(|| format!(...))

crates/aranya-daemon/src/api.rs
```diff
@@ -402,16 +481,48 @@ impl DaemonApi for DaemonApiHandler {
...
+        role: api::RoleId,
+        op: api::Op,
     ) -> api::Result<()> {
         self.client
             .actions(&team.into_id().into())
-            .revoke_role(device.into_id().into(), role.into())
+            .assign_role_operation(role, op)
             .await
-            .context("unable to revoke device role")?;
+            .context(format!("unable to assign role operation: {}", op))?;
```
elagergren-spideroak, line 507: Ditto for with_context

elagergren-spideroak (changes requested):
Still reviewing

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -72,6 +72,14 @@ pub enum Error {
     /// Serialization error.
     #[capi(msg = "serialization")]
     Serialization,
+
+    /// CString allocation error.
+    #[capi(msg = "CString")]
+    CString,
```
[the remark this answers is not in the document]
gknopf-aranya: The error is occurring here if a `CString` cannot be created from a role `String`:
`name: CString::new(value.name).map_err(crate::imp::Error::CString)?,`

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -72,6 +72,14 @@ pub enum Error {
...
+
+    /// CString conversion error.
```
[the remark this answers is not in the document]
gknopf-aranya: This error is no longer used. Removing it.

crates/aranya-client-capi/src/api/defs.rs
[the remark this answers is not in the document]
elagergren-spideroak: I understand. It's that `CString` is a Rust type, so `ARANYA_ERROR_CSTRING` doesn't make much sense. And, more confusingly, the docs talk about allocations, but the error isn't about allocations, it's about an errant null byte. So, use `CString::new(...).assume("should not have null byte")` instead.

crates/aranya-client-capi/src/imp/op.rs
gknopf-aranya: Because `query_role_operations` methods needs to return an array of elements that are `Safe`.
I originally tried using the enum only but this did not work, so I made another `Safe` type for this.
This is what happens if I get rid of the safe type:
```
[cargo-make] INFO - Execute Command: "cargo" "build" "--release"
   Compiling aranya-client-capi v0.0.0 (/Users/gknopf/scm/aranya/crates/aranya-client-capi)
    --> crates/aranya-client-capi/src/api/defs.rs:1812:20
     |
1812 |         Safe::init(dst, (*src).into())
     |         ---------- ^^^ expected `&mut MaybeUninit<Safe<_>>`, found `&mut MaybeUninit<Op>`
     |         |
     |         arguments to this function are incorrect
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Closing since this is a duplicate of another comment.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Updating method name and docs.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Updating

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Updating

crates/aranya-client-capi/src/imp/op.rs
gknopf-aranya: I'll try setting the value like this:
`dst.write((*src).into());`

gknopf-aranya: Now the `write` doesn't know what type it's supposed to be. So I converted the type explicitly first.
```
    for (dst, src) in out.iter_mut().zip(data) {
        let op: Op = (*src).into();
        dst.write(op);
    }
```
But it thinks it's a `generated::AranyaOp` instead of a `defs::Op`.
```
9087 |       let ops: ::core::option::Option<&mut ::core::mem::MaybeUninit<_>> = {
     |  _________________________________________________________________________-
9088 | |         let ops = __capi::try_as_opt_mut!(ops);
9089 | |         __capi::to_inner!(ops)
9090 | |     };
     | |_____- here the type of `ops` is inferred to be `Option<&mut std::mem::MaybeUninit<generated::AranyaOp>>`
...
9094 |       match { crate::api::defs::query_role_operations(client, team, role, ops, ops_len) } {
     |               ---------------------------------------                     ^^^ expected `Option<&mut MaybeUninit<Op>>`, found `Option<&mut MaybeUninit<AranyaOp>>`
     |               |
     |               arguments to this function are incorrect
     |
     = note: expected enum `Option<&mut std::mem::MaybeUninit<defs::Op>>`
                found enum `Option<&mut std::mem::MaybeUninit<generated::AranyaOp>>`
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Updating this

crates/aranya-client-capi/src/imp/op.rs
gknopf-aranya: The conversion appears to work for `DeviceId`, so I'm thinking I may be able to get this to work if I wrap the enum in a struct. Is this a capi generator bug or am I doing something incorrectly?

crates/aranya-daemon/src/api.rs
gknopf-aranya: Updating

crates/aranya-client-capi/src/imp/op.rs
gknopf-aranya: Changing the param from `Op` to `AranyaOp` compiles. Seems strange though since we don't do this for any of the other params.
```
    ops: Option<&mut MaybeUninit<AranyaOp>>,
    ops_len: &mut usize,
```

elagergren-spideroak: Okay, after thinking about it a bit I guess I'm fine with the API that you originally had if you rename `defs::Op` to something like `defs::OpName`. Or just get rid of `OpName` and use strings instead.

It's just that having `Operation` and `Op` is confusing.

gknopf-aranya:
Resolved open feedback. Now need to resolve tons of merge conflicts from the IPC PR.

crates/aranya-daemon/src/policy.md
```diff
@@ -128,6 +153,16 @@ fact DeviceSignKey[device_id id]=>{key_id id, key bytes}
 // A device's public EncryptionKey.
 fact DeviceEncKey[device_id id]=>{key_id id, key bytes}
 
+// A ID-based role on the team.
+fact Role[role_id id]=>{role struct RoleInfo}
```
elagergren-spideroak, line 157: using `RoleInfo` here means we store the role ID twice. Please inline the fields here. Same with `Device` btw.

crates/aranya-daemon/src/policy.md
```diff
@@ -128,6 +153,16 @@ fact DeviceSignKey[device_id id]=>{key_id id, key bytes}
 // A device's public EncryptionKey.
 fact DeviceEncKey[device_id id]=>{key_id id, key bytes}
 
+// A ID-based role on the team.
```
elagergren-spideroak, line 156: nit: I don't know what "ID-based" means. Just say something like "An RBAC role" or whatever.

crates/aranya-daemon/src/policy.md
```diff
@@ -481,16 +573,130 @@ command TerminateTeam {
 - Only an Owner can create this event.
 - Once terminated, no further communication will occur over the team graph.
 
+## SetupDefaultRoles
+
+The `SetupDefaultRoles` command sets up default roles on the team.
+Operation can only be invoked by the device the team owner.
```
djohnson-spideroak, line 579: Its not clear in the current implementation where this permission check happens.

crates/aranya-daemon/src/policy.md
```diff
@@ -602,52 +818,84 @@ finish function remove_device(device_id id) {
...
+            emit DevicePrecedenceAssigned {
+                device_id: device.device_id,
+                precedence: this.precedence,
             }
         }
-        _ => { check false }
     }
 }
+
+// A precedence value to a device on the team.
```
djohnson-spideroak, line 869: nit: doc

crates/aranya-daemon/src/policy.md
```diff
@@ -657,51 +905,50 @@ command AssignOwner{
...
+
+        let role = RoleInfo {
+            role_id: role_id,
+            name: this.name,
+            author_id: author.device_id,
+        }
 
         finish {
-            assign_role(device, Role::Owner)
+            create_role(role)
```
djohnson-spideroak, line 921: we should avoid this naming collision. we have an action create_role, and a finish function with the same name.

crates/aranya-daemon/src/policy.md
```diff
@@ -657,51 +905,50 @@ command AssignOwner{
...
-    device_id id,
+// Delete a custom role.
+action delete_role(role_id id){
+    publish DeleteRole {
+        role_id: role_id,
+    }
 }
 
-command AssignAdmin{
+command DeleteRole {
```
djohnson-spideroak, line 948: should delete_role check that the author has presidence over all the devices with that role?

djohnson-spideroak (commented):
still early in review, will come back to it later today

crates/aranya-daemon/src/policy.md
gknopf-aranya: Adding permissions checks for this.
These commands won't explicitly check that the device is the owner.
The `CreateTeam` command will bootstrap the owner role with permission to execute these role setup commands.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Same for `delete_role`

crates/aranya-daemon/src/policy.md
gknopf-aranya: Possibly. Although the current policy implementation only allows the owner role to delete roles.
I'm not aware of plans to allow the admin or operator roles permission to delete roles.
The owner will have the highest precedence on the team.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -302,11 +299,83 @@ pub enum ChanOp {
...
+    AqcCreateBidiChannel,
+    AqcCreateUniChannel,
+}
+
+/// Converts [`Op`] to string.
+impl FromStr for Op {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "Operation::AddMember" => Ok(Self::AddMember),
```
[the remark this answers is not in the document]
gknopf-aranya: It's easier to search for operations in the policy this way. Without it, the search will return operations and commands.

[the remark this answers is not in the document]
elagergren-spideroak: That's fine. Right now we just have a bunch of needless runtime string comparisons. Maybe the compiler optimizes them out, maybe it doesn't. And IMO it makes the core more difficult to read because I have to scan past a bunch of `Operation::`s everywhere.

crates/aranya-daemon/src/policy.md
```diff
@@ -1663,8 +1962,8 @@ command AssignLabel {
         let author = get_valid_device(envelope::author_id(envelope))
         let target = get_valid_device(this.device_id)
 
-        // Only Owners and Operators can assign labels to Members.
-        check is_owner(author.role) || is_operator(author.role)
+        check device_can_execute_op(author.device_id, "AssignLabel")
+        check author_dominates_target(author.device_id, target.device_id)
```
elagergren-spideroak, line 1966: I don't think we care about dominance here. Why does it matter if, for example, an operator grants an admin permission to use a label?

crates/aranya-daemon/src/policy.md
```diff
@@ -481,16 +576,136 @@ command TerminateTeam {
...
+            role_id: role_id,
+            name: "admin",
+            author_id: author.device_id,
+        }
+
+        finish {
+            create_role_fact(role)
+            assign_op_role("DeleteLabel", role.role_id)
+
+            emit QueriedRole {
```
elagergren-spideroak, line 617: You want `RoleCreated`, not `QueriedRole`. Same for the other two default roles.

crates/aranya-daemon/src/policy.md
```diff
@@ -111,13 +104,45 @@ struct KeyIds {
...
+
+// Defines a role on the team.
+struct RoleInfo {
+    // ID of the role.
+    role_id id,
+    // Name of the role.
+    name string,
+    // ID of device that created the role.
+    author_id id,
+}
```
elagergren-spideroak, line 138: Only actions and effects are part of the policy's stable API. Using structs inside an effect seems like a violation of that.

crates/aranya-daemon/src/policy.md
```diff
@@ -242,6 +257,53 @@ function open_envelope(sealed_envelope struct Envelope) bytes {
...
+
+finish function delete_role_fact(role struct Role) {
+    delete Role[role_id: role.role_id]
+}
+
+finish function assign_op_role(op string, role_id id) {
+    create OpRequiresRole[op: op]=>{role_id: role_id}
+}
+
+finish function revoke_op_role(op string) {
```
elagergren-spideroak, line 280: This is only used in one place, just inline it there. (Plus, the naming isn't clear that it's deleting the `OpRequiresRole` fact.)

crates/aranya-daemon/src/policy.md
```diff
@@ -832,51 +1104,61 @@ command RevokeOwner{
...
-Revoke the `Admin` role from a device.
+Assign permission to execute a certain operation to the role.
+Note: currently each operation is only allowed to be executed by one role.
 
 ```policy
-// The Admin role was revoke from A device.
-effect AdminRevoked {
-    device_id id,
+
+action assign_operation_to_role(role_id id, op string) {
```
elagergren-spideroak, line 1149: It's the other way around. In order to perform an operation you must have been assigned a role. This should really be `require_role_for_operation`.

The language needs to be cleaned up elsewhere in the policy file.

elagergren-spideroak (changes requested):
still reviewing

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -727,273 +859,543 @@ pub fn sync_peer_config_builder_set_sync_now(cfg: &mut SyncPeerConfigBuilder) {
...
-/// @param label_id the channel label ID [`LabelId`] to delete.
 ///
 /// @relates AranyaClient.
-pub fn delete_label(
+pub unsafe fn setup_default_roles(
     client: &mut Client,
     team: &TeamId,
-    label_id: &LabelId,
+    roles: Option<&mut MaybeUninit<Role>>,
+    roles_len: &mut usize,
```
elagergren-spideroak, line 969: I don't think that this function should return the roles. For example, this code—which _looks_ correct—is actually incorrect:

```c
AranyaError err = aranya_setup_default_roles(&client, &team, &roles, &roles_len);
if (err == ARANYA_ERROR_BUFFER_TOO_SMALL) {
    roles = realloc(roles, roles_len * sizeof(AranyaRole));
    assert(new_roles != NULL);
    err = aranya_setup_default_roles(&client, &team, &roles, &roles_len);
}
if (err != ARANYA_ERROR_SUCCESS) { ... }
```

Calling `aranya_setup_default_roles` twice will return an error the second time.

The function should just return nothing and then users can immediately call `query_team_roles` (or whatever it's called) to learn about the default roles.

gknopf-aranya: We should document that the method should only be invoked once either way. It's supposed to be part of team setup, not something that's invoked multiple times.

crates/aranya-daemon/src/policy.md
```diff
@@ -1748,9 +2047,8 @@ command RevokeLabel {
         let author = get_valid_device(envelope::author_id(envelope))
         let target = get_valid_device(this.device_id)
 
-        // Only Owners, Admins, and Operators are allowed to revoke a label from a Member.
-        check is_owner(author.role) || is_admin(author.role) || is_operator(author.role)
-        check is_member(target.role)
+        check device_can_execute_op(author.device_id, "RevokeLabel")
+        check author_dominates_target(author.device_id, target.device_id)
```
mooreoak, line 2051: If we remove the one on assignment this should probably be removed too.

mooreoak (changes requested):

crates/aranya-client-capi/src/imp/label.rs
```diff
@@ -0,0 +1,27 @@
...
+
+/// A label that can be assigned to a device.
+pub struct Label {
+    pub(crate) id: LabelId,
+    pub(crate) name: CString,
+    _pad: MaybeUninit<[u8; 2 * (8 - size_of::<usize>())]>,
+}
+
+impl Typed for Label {
+    const TYPE_ID: TypeId = TypeId::new(0xbfafb41c);
```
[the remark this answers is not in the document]
elagergren-spideroak: Developers randomly generate them.

crates/aranya-client-capi/src/imp/op.rs
```diff
@@ -0,0 +1,17 @@
...
+#[derive(Copy, Clone, Debug)]
+pub struct Op {
+    pub(crate) op: crate::api::defs::Op,
+}
+
+impl Typed for Op {
+    const TYPE_ID: TypeId = TypeId::new(0xbfafe41c);
```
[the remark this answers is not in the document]
elagergren-spideroak: Two different `Typed` impls should not have the same `TYPE_ID`.

crates/aranya-client-capi/src/imp/role.rs
```diff
@@ -0,0 +1,27 @@
+use std::{ffi::CString, mem::MaybeUninit};
+
+use aranya_capi_core::safe::{TypeId, Typed};
+use aranya_daemon_api::RoleId;
+
+/// A role on the team.
+pub struct Role {
+    pub(crate) id: RoleId,
+    pub(crate) name: CString,
+    _pad: MaybeUninit<[u8; 2 * (8 - size_of::<usize>())]>,
```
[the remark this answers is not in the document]
elagergren-spideroak: `Role` (inside `../api/defs.rs`) is marked as `#[opaque(size = SIZE, align = ALIGN)]`, which generates the following C code:

```c
typedef struct Role __attribute__(aligned(ALIGN)) {
    char __for_size_only[SIZE];
} Role;
```

It is normally okay for `SIZE` to be _larger_ than `core::mem::size_of::<Role>()`, but not when `Role` is an element in a C array. This line of code pads `Role` out to exactly `SIZE`.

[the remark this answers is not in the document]
ssweetney-spok: I had similar question and I understand why the padding is necessary but not how it's done. Doesn't `2 * (8 - size_of::<usize>())` evaluate to 0 on 64 bit targets?

crates/aranya-client/src/client.rs
```diff
@@ -383,16 +471,6 @@ impl Team<'_> {
             .map_err(aranya_error)
     }
 
-    /// Delete a label.
```
gknopf-aranya, line 386: @elagergren-spideroak, @djohnson-spideroak, @mooreoak based on discussion today, we're going to add `delete_label` back into the API and allow the label assignments to be garbage after a label is deleted (until we support prefix query deletion).

crates/aranya-client-capi/src/imp/role.rs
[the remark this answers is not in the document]
jdygert-spok: > I had similar question and I understand why the padding is necessary but not how it's done. Doesn't `2 * (8 - size_of::<usize>())` evaluate to 0 on 64 bit targets?

Yes, the goal is to make it the same size on all targets. For these types we are using, we only have to add padding for 32 bit targets to account for the smaller pointers/usize. `CString == Box<[u8]>` has 2 ptr/usize so we add `2 *` the needed padding.

BTW this is just a hack until https://github.com/aranya-project/aranya-core/issues/227 is implemented. I had started it but it got confusing to thread it through the capi rewriting.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -727,273 +859,543 @@ pub fn sync_peer_config_builder_set_sync_now(cfg: &mut SyncPeerConfigBuilder) {
...
+/// Returns a C string pointer to the role's name.
+#[aranya_capi_core::no_ext_error]
+pub fn role_get_name(role: &Role) -> *const c_char {
+    role.name.as_ptr()
+}
+
+/// Releases any resources used by the [`Role`].
+///
+/// @param role the role [`Role`].
+pub unsafe fn role_cleanup(role: OwnedPtr<Role>) {
```
jdygert-spok, line 1242: `#[capi::derive(Cleanup)` on each of these?

ssweetney-spok: I [suggested](https://github.com/aranya-project/aranya/pull/213#discussion_r2064967962) that initially but the generated doc comment isn't quite right. In the future, it'd be nice if we could support custom doc comments with this proc macro 🤔

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: > I think an arg with the type `Option<&T>` will make the function fallible anyway

Oh, I see now that it also checks alignment, so yeah there's no way to take a reference infallibly. :/

djohnson-spideroak:
@elagergren-spideroak are you still planning on refactoring this PR?

PR #223, opened 2025-04-22
Remove AFC

nikki-aranya:
Removes Aranya Fast Channels for the time being, as re-introducing it will require a large rewrite anyways.


nikki-aranya:
This should massively simplify the diff in #202. It's mostly pulled straight from there, but I had a bunch more cleanup and made sure the C and Rust examples passed.

crates/aranya-daemon/src/config.rs
```diff
@@ -29,11 +29,9 @@ pub struct Config {
     pub sync_addr: Addr,
 
     /// AFC configuration.
-    pub afc: AfcConfig,
+    pub afc: Option<AfcConfig>,
```
jdygert-spok, line 32:
```suggestion
    #[serde(default)]
    pub afc: Option<AfcConfig>,
```

And I guess remove it from our example configs?

jdygert-spok (dismissed):

jdygert-spok (dismissed):

gknopf-aranya (changes requested):
lgtm after removing the remaining afc references from some files

crates/aranya-daemon/src/api.rs
```diff
@@ -14,7 +14,7 @@ use std::{
 use anyhow::{anyhow, Context, Result};
 use aranya_crypto::{Csprng, DeviceId, Rng};
 use aranya_daemon_api::{self as api, DaemonApi};
-use aranya_fast_channels::{Label, NodeId};
+use aranya_fast_channels::NodeId;
```
djohnson-spideroak, line 17: given this is used for AQC as well, in the future we want to move it out of aranya_fast_channels

jdygert-spok: `NodeId` is the afc u32 cheap local id thing, it should actually be removed entirely. (I think #202 does.)

nikki-aranya:
> > * `deny.toml`
> > * `audits.toml`
> > * `config.rs`
> > * `examples/c/run.bash`
> 
> aranya-daemon/src/config also still has AFC

I can nuke the config option and the dependency from the audit as well, but I thought it would be far more annoying to remove and have to re-add, those instances were left on purpose.

crates/aranya-daemon/src/api.rs
nikki-aranya: Yeah, I'm only concerned with removing AFC, not making any changes to the structure, 202 is covering all that.

jdygert-spok (approved):

PR #226, opened 2025-04-24
Share the same `SyncPeers` instance between `DaemonApiHanlder`s

ssweetney-spok:
Share the same `SyncPeers` instance between `DaemonApiHanlder`s

### DIFF
--- a/crates/aranya-daemon/src/api.rs
+++ b/crates/aranya-daemon/src/api.rs
@@ -106,7 +106,7 @@ impl DaemonApiServer {
                 client,
                 local_addr,
                 pk,
-                peers,
+                peers: Arc::new(Mutex::new(peers)),
                 keystore_info,
                 aqc_peers: Arc::new(Mutex::new(aqc_peers)),
             },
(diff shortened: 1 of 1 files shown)


ssweetney-spok:
Thinking through how to test this. The tests in "crates/aranya-client/tests/tests.rs" don't have access to the `DaemonApiHandler` that gets created in `aranya_daemon::daemon::Daemon::run` -> `aranya_daemon::api::DaemonApiServer::new`

I think testing that type specifically would be the simplest option 🤔

jdygert-spok:
`cfgs` should be in the `Syncer` instead. `SyncPeers` is meant to be just a cloneable handle to interact with the actual `Syncer`.

ssweetney-spok:
> `cfgs` should be in the `Syncer` instead. `SyncPeers` is meant to be just a cloneable handle to interact with the actual `Syncer`.

Gotcha. I'm closing this PR and created a new [issue](https://github.com/aranya-project/aranya/issues/233)

ssweetney-spok:
#233

PR #227, opened 2025-04-24
hotfix opaque slice with manual padding

jdygert-spok:


gknopf-aranya (approved):

gknopf-aranya:
Thank you

jdygert-spok:
> Confirming that this isn't needed for the `query_devices_on_team` query?

Yep, because IDs are a known fixed size.

PR #235, opened 2025-04-28
Cleanup `aranya-client-capi`, C Example

nikki-aranya:
This includes cleanup of the generated `aranya-client.h` definition order, as well as shortening some `example.c` paths to be easier to read.


nikki-aranya:
Note that I don't have any of the AQC stuff in yet really, this is mostly just rearranging the existing code to be more sane. I figured out that the type definitions at the top of `aranya-client.h` are based on First Use so it was a little tricky to get acceptable but I think this is good.

I need to do another pass on the C example, add AQC Bidi/Uni Channel/Stream objects, their accessor functions, and we should be good for review.

gknopf-aranya:
@nikki-aranya the diff noise on this PR is making it more difficult to review just the AQC related changes.

crates/aranya-client-capi/src/imp/aqc.rs
```diff
@@ -0,0 +1,9 @@
+use aranya_capi_core::safe::{TypeId, Typed};
+
+pub struct AqcBidiChannel {
+    pub(crate) inner: aranya_client::aqc_net::AqcBidirectionalChannel,
+}
+
+impl Typed for AqcBidiChannel {
+    const TYPE_ID: TypeId = TypeId::new(0x89D528DE);
```
gknopf-aranya, line 8: nit: the type id hex in other modules in this folder are lowercase.

examples/c/example.c
```diff
@@ -548,41 +552,40 @@ AranyaError run_aqc_example(Team *t) {
...
-    err = aranya_aqc_create_bidi_channel(&t->clients.membera.client, &t->id,
-                                         aqc_addrs[MEMBERB], &label1_id,
-                                         &chan_id);
-    EXPECT("error creating aqc bidi channel", err);
+    EXPECT(aranya_aqc_create_bidi_channel(&t->clients.membera.client, &t->id,
+                                          aqc_addrs[MEMBERB], &label1_id,
+                                          &chan_id),
+           "error creating bidirectional AQC channel");
 
     // TODO: send AQC data
```
gknopf-aranya, line 579: This todo can be removed on this PR when the AQC methods are added for sending/receiving data

nikki-aranya:
> @nikki-aranya the diff noise on this PR is making it more difficult to review just the AQC related changes.

After going down this path, I think it's better to make this its own PR and do AQC separately

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya, line 1: Where are the AQC API methods for sending/receiving data via AQC channels? That's an important part of having a working data plane.

nikki-aranya: Yes, that's why it was still in a draft form when you asked

gknopf-aranya:
> After going down this path, I think it's better to make this its own PR and do AQC separately

Agreed. Will be easier to review the changes.

examples/c/example.c
```diff
@@ -175,42 +167,42 @@ AranyaError init_client(Client *c, const char *name, const char *daemon_addr,
...
     char *team_id_str      = malloc(team_id_str_len);
-    aranya_id_to_str(&t->id.id, team_id_str, &team_id_str_len);
+    EXPECT(aranya_id_to_str(&t->id.id, team_id_str, &team_id_str_len),
+           "error converting ID to string");
     printf("Team ID: %s \r\n", team_id_str);
 
     AranyaId decodedId;
-    err = aranya_id_from_str(team_id_str, &decodedId);
-    EXPECT("error decoding string into an ID", err);
+    EXPECT(aranya_id_from_str(team_id_str, &decodedId),
```
gknopf-aranya, line 202: Should call this variable `decoded_id`:
```suggestion
    EXPECT(aranya_id_from_str(team_id_str, &decoded_id),
```

examples/c/example.c
```diff
@@ -266,81 +261,81 @@ AranyaError add_sync_peers(Team *t, AranyaSyncPeerConfig *cfg) {
...
 
     // initialize the Aranya team.
     printf("initializing team\r\n");
-    err = init_team(t);
-    EXPECT("error initializing team", err);
+    EXPECT(init_team(t), "error initializing team");
+
+    Client *owner = &t->clients.owner;
+    Client *admin = &t->clients.admin;
+    Client *operator= & t->clients.operator;
```
gknopf-aranya, line 276: nit: looks like there's an extra space before the `&` here:
```suggestion
    Client *operator= &t->clients.operator;
```

examples/c/example.c
```diff
@@ -266,81 +261,81 @@ AranyaError add_sync_peers(Team *t, AranyaSyncPeerConfig *cfg) {
...
-    EXPECT("error assigning admin role", err);
-
-    // upgrade role to operator.
-    err = aranya_assign_role(&t->clients.admin.client, &t->id,
-                             &t->clients.operator.id, ARANYA_ROLE_OPERATOR);
+    EXPECT(aranya_assign_role(&owner->client, &t->id, &admin->id,
+                              ARANYA_ROLE_ADMIN),
+           "error assigning admin role");
 
+    // upgrade role to operator. this includes testing that sync_now works.
```
gknopf-aranya, line 295: A lot of this example code will change when the custom roles PR is merged. E.g. the way sync now is tested and the way roles are assigned. There will be a method added to the C API to initialize default roles.

gknopf-aranya (dismissed):
lgtm

ssweetney-spok (approved):

gknopf-aranya (approved):

PR #236, opened 2025-04-29
Implement `addTeam` API

### LINKED ISSUE #167: implement add_team
`add_team()` is unimplemented and will panic if it is called because of a `todo!()` in the daemon API implementation.
Implement `add_team()`.

cc: @djohnson-spideroak, @jdygert-spok, @nikki-aranya

ssweetney-spok:
Resolves #167

TODO:
- [ ] C API (addTeam)
- [ ] RemoveTeam (C and Rust API)


Cargo.toml
```diff
@@ -78,6 +78,19 @@ tokio-util = { version = "0.7.12" }
...
+aranya-aqc-util = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
+aranya-crypto-ffi = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
+aranya-device-ffi = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
+aranya-envelope-ffi = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
+aranya-idam-ffi = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
+aranya-perspective-ffi = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
+aranya-crypto = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
+aranya-policy-module = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
+aranya-policy-ast = { git = 'https://github.com/aranya-project/aranya-core.git', branch = "init-command-from-new-graph" }
```
ssweetney-spok, line 92: These lines will be removed once https://github.com/aranya-project/aranya-core/pull/228 is merged and a new release of the crates are published

crates/aranya-client/tests/tests.rs
```diff
@@ -403,3 +403,33 @@ async fn test_query_functions() -> Result<()> {
...
+    // Set up our team context so we can run the test.
+    let work_dir = tempfile::tempdir()?.path().to_path_buf();
+    let mut team = TeamCtx::new("test_sync_now", work_dir).await?;
+
+    // Create the initial team, and get our TeamId and init command.
+    let cfg = TeamConfig::builder().build()?;
+    let (team_id, init_command) = team
+        .owner
+        .client
+        .create_team(cfg)
```
gknopf-aranya, line 420: We should use consistent naming. Either `add_team` or `create_team`.

gknopf-aranya: I see `add_team` farther down so they are different operations. That works then.

ssweetney-spok: Those are 2 different APIs. `create_team` is used by the graph owners. `add_team` is supposed to be used by the other members of the team

Cargo.toml
gknopf-aranya: It's more likely not to forget to change this if a comment like this is added to the file:
`TODO: use released crates before merging this PR`

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -576,12 +576,26 @@ pub fn get_device_id(client: &mut Client) -> Result<DeviceId, imp::Error> {
...
 
-#[aranya_capi_core::opaque(size = 16, align = 8)]
+#[aranya_capi_core::opaque(size = 32, align = 8)]
+#[aranya_capi_core::derive(Init, Cleanup)]
 pub type TeamConfigBuilder = Safe<imp::TeamConfigBuilder>;
 
+/// Attempts to construct a [`TeamConfig`], returning an `Error::Config`
+/// if there are invalid parameters.
+///
+/// @param cfg a pointer to the team config builder
```
gknopf-aranya, line 589: Docs missing init cmd param

crates/aranya-client/src/client.rs
```diff
@@ -120,7 +121,7 @@ impl Client {
     }
 
     /// Create a new graph/team with the current device as the owner.
-    pub async fn create_team(&mut self, cfg: TeamConfig) -> Result<TeamId> {
+    pub async fn create_team(&mut self, cfg: TeamConfig) -> Result<(TeamId, Vec<u8>)> {
```
gknopf-aranya, line 124: Type alias can be used here:
```suggestion
    pub async fn create_team(&mut self, cfg: TeamConfig) -> Result<(TeamId, InitCmd)> {
```

crates/aranya-client/src/config.rs
```diff
@@ -110,8 +120,17 @@ impl TeamConfigBuilder {
         Self::default()
     }
 
+    // TODO(Steve): better docs
+    /// Set init command.
+    pub fn init_command(mut self, bytes: &[u8]) -> Self {
```
gknopf-aranya, line 125:
```suggestion
    pub fn init_command(mut self, init_command: InitCmd) -> Self {
```

crates/aranya-client/tests/tests.rs
```diff
@@ -401,3 +403,43 @@ async fn test_query_functions() -> Result<()> {
...
+        .await
+        .expect("expected to create team");
+    info!(?team_id);
+    info!(?init_command);
```
gknopf-aranya, line 424:
```suggestion
    info!(?team_id, ?init_command);
```

crates/aranya-client/tests/tests.rs
```diff
@@ -401,3 +403,43 @@ async fn test_query_functions() -> Result<()> {
...
+        .await?;
+    team.operator
+        .client
+        .add_team(team_id, team_cfg.clone())
+        .await?;
+    team.membera
+        .client
+        .add_team(team_id, team_cfg.clone())
+        .await?;
+    team.memberb.client.add_team(team_id, team_cfg).await?;
```
gknopf-aranya, line 442: Let's verify the team was added properly by syncing some commands and verifying that effects can be processed as expected.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -252,7 +253,7 @@ pub trait DaemonApi {
     async fn remove_team(team: TeamId) -> Result<()>;
 
     /// Create a new graph/team with the current device as the owner.
-    async fn create_team(cfg: TeamConfig) -> Result<TeamId>;
+    async fn create_team(cfg: TeamConfig) -> Result<(TeamId, Vec<u8>)>;
```
gknopf-aranya, line 256:
```suggestion
    async fn create_team(cfg: TeamConfig) -> Result<(TeamId, InitCmd)>;
```

crates/aranya-daemon/src/aranya.rs
```diff
@@ -145,9 +145,9 @@ where
         &self,
         owner_keys: KeyBundle,
         nonce: Option<&[u8]>,
-    ) -> Result<(GraphId, Vec<Effect>)> {
+    ) -> Result<(GraphId, Vec<u8>, Vec<Effect>)> {
```
gknopf-aranya, line 148: nit: the effects don't need to be returned if they aren't used anywhere.
They aren't used in `create_team` in api.rs. Haven't checked if they're used in a unit test somewhere.

ssweetney-spok: Yea, it seems like we can get rid of them if they aren't being used

Cargo.toml
ssweetney-spok: An unresolved comment should block the PR from being merged

examples/c/example.c
```diff
@@ -198,8 +201,21 @@ AranyaError init_team(Team *t) {
...
-    aranya_team_config_builder_build(&build, &cfg);
-    err = aranya_create_team(&t->clients.owner.client, &cfg, &t->id);
+
+    err = aranya_team_config_builder_init(&build);
+    EXPECT("error initialzing the team config builder", err);
+    err = aranya_team_config_builder_build(&build, &cfg);
+    EXPECT("error building a team config", err);
+
+    t->init_cmd_len = 1;
+    t->init_cmd     = malloc(t->init_cmd_len);
```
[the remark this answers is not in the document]
gknopf-aranya: C doesn't automatically initialize/zeroize memory like Rust when it's allocated. So this helps to ensure the memory is in a consistent state each time the program runs.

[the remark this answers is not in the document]
gknopf-aranya: `malloc` could cause undefined behavior if the data is read from before it's initialized. It's difficult to guarantee there is no code reading that data before initializing it without reviewing all the code. Better to just initialize it.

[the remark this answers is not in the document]
gknopf-aranya: We should be using `calloc` instead of `malloc` in the C example everywhere, so could open a different PR to handle that in the entire example.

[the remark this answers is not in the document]
ssweetney-spok: In that case, should we just create another issue to replace the other `malloc` calls?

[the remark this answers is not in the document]
gknopf-aranya: Yep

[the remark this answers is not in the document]
gknopf-aranya: @ssweetney-spok could you please open that issue?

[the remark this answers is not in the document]
gknopf-aranya: Actually, we made a lot of those changes already. You should get them once you merge with upstream changes.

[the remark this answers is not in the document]
ssweetney-spok: Gotcha, if they're already being replaced then keeping things consistent makes sense

examples/c/example.c
```diff
@@ -449,6 +466,36 @@ AranyaError add_sync_peers(Team *t, AranyaSyncPeerConfig *cfg) {
...
+    err = aranya_team_config_builder_init(&build);
+    EXPECT("error initialzing the team config builder", err);
+
+    err = aranya_team_config_builder_init_command(&build, t->init_cmd, t->init_cmd_len);
+    EXPECT("error setting the init command on the team config builder", err);
+
+    err = aranya_team_config_build(&build, &cfg);
+    EXPECT("error building a team config", err);
+
+    // Storage exists due to a prior call to `sync_now`
```
ssweetney-spok, line 484: Question: Should receiving commands from a sync request still create a new storage if one doesn't exist? I assume it was that way since we didn't have a way to create one from existing commands before `addTeam`

gknopf-aranya: I'm not sure what the original intent was for this. Seems like it would be convenient to be able to add a graph either by syncing or using an existing command rather than limiting the user to a single way.

cc: @elagergren-spideroak, @djohnson-spideroak, @jdygert-spok

elagergren-spideroak: No, if you do not have existing storage then syncing should _not_ create new storage. You need to manually add the init command, _then_ sync. Otherwise, an attacker could force you to create arbitrary storage(s).

ssweetney-spok: Gotcha, then I just need to make an update to delete these [lines](https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-runtime/src/client/transaction.rs#L164-L168) and update any tests that rely on syncing to create a new storage

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -888,32 +897,54 @@ pub fn revoke_label(
 
 /// Create a new graph/team with the current device as the owner.
 ///
+/// Returns an `AranyaBufferTooSmall` error if the output buffer is too small hold the init command.
```
gknopf-aranya, line 900:
```suggestion
/// Returns an `AranyaBufferTooSmall` error if the output buffer is too small to hold the init command.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -888,32 +897,54 @@ pub fn revoke_label(
...
+/// Writes the number of bytes that would have been returned to `init_cmd_len`.
+/// The application can use `init_cmd_len` to allocate a larger buffer.
```
gknopf-aranya, line 902:
```suggestion
/// The application can use the returned `init_cmd_len` to allocate a larger buffer.
```

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -238,21 +238,92 @@ impl Default for SyncPeerConfigBuilder {
...
 
+impl TryFrom<&TeamConfig> for aranya_client::TeamConfig {
+    type Error = Error;
+
+    fn try_from(value: &TeamConfig) -> Result<Self, Self::Error> {
+        let mut builder = Self::builder();
+
+        if !value.init_command.is_null() {
+            let bytes = {
+                // This is a workaround because try_as_slice! expects the first arg to be an identifier
```
gknopf-aranya, line 258: Will this PR resolve the workaround?

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -238,21 +238,92 @@ impl Default for SyncPeerConfigBuilder {
...
+    ) -> Result<(), Error> {
+        if init_command.is_null() {
+            let e = Error::InvalidArg(InvalidArg::new(
+                "init_command",
+                "Tried to set the `init_command` to an invalid address",
+            ));
+            return Err(e);
+        }
+
+        if init_command_len == 0 {
```
gknopf-aranya, line 303: What if the init_command is not null and not zero length, but is still too short?
If the size of the init command is a known constant, we should be able to validate the length here before setting it.

crates/aranya-client/src/client.rs
```diff
@@ -204,7 +204,7 @@ impl Client {
     }
 
     /// Create a new graph/team with the current device as the owner.
-    pub async fn create_team(&mut self, cfg: TeamConfig) -> Result<TeamId> {
+    pub async fn create_team(&mut self, cfg: TeamConfig) -> Result<(TeamId, Box<[u8]>)> {
```
gknopf-aranya, line 207: Should use a type alias here instead of `Box<[u8]>` for the init command

crates/aranya-client/src/config.rs
```diff
@@ -110,8 +120,17 @@ impl TeamConfigBuilder {
         Self::default()
     }
 
+    // TODO(Steve): better docs
```
gknopf-aranya, line 123: Commenting here so this todo isn't forgotten

crates/aranya-client/tests/tests.rs
```diff
@@ -366,15 +367,16 @@ async fn test_query_functions() -> Result<()> {
     let work_dir = tempfile::tempdir()?.path().to_path_buf();
     let mut team = TeamCtx::new("test_query_functions", work_dir).await?;
 
-    // Create the initial team, and get our TeamId.
+    // Create the initial team, and get our TeamId and init command.
     let cfg = TeamConfig::builder().build()?;
-    let team_id = team
+    let (team_id, init_command) = team
```
gknopf-aranya, line 372: @ssweetney-spok how are the tests in this file still working? I thought there was a change being made so that a device must explicitly `add_team` rather than automatically getting new graphs from syncing. This means each test would perform operations in this order:
1. `create_team() -> (team_id, init_cmd)`
2. `add_team(team_id, init_cmd)` invoked by all devices other than the team owner
3. Then syncing and graph operations can work

ssweetney-spok: They aren't. I didn't push the commit that fixed them

crates/aranya-client/tests/tests.rs
```diff
@@ -411,3 +413,43 @@ async fn test_query_functions() -> Result<()> {
 
     Ok(())
 }
+
+/// Tests add_team() by showing that the init command of a graph created by an owner can be used
+/// by other peers to create a graph on their device
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_add_team() -> Result<()> {
```
gknopf-aranya, line 420: This is testing the opposite of what's needed.
`add_team` is the default path that will be tested by the other tests in this file.
What we need to add is a test that shows syncing and actions will no longer work for a graph that hasn't been added yet.

gknopf-aranya:
The `add_team` portion of this PR lgtm after the negative unit test without `add_team` is added and the Rust example is compiling again. I'll review again once the `remove_team` changes are added.

crates/aranya-client/src/client.rs
```diff
@@ -417,6 +424,16 @@ impl Team<'_> {
             .map_err(IpcError::new)?
             .map_err(aranya_error)
     }
+
+    /// Add a team to a device's local store
+    pub async fn add_team_to_device(&self, cfg: TeamConfig) -> Result<()> {
```
djohnson-spideroak, line 429: need docs on the difference between this and `add_team`

examples/c/example.c
```diff
@@ -449,6 +458,35 @@ AranyaError add_sync_peers(Team *t, AranyaSyncPeerConfig *cfg) {
...
+    err = aranya_add_team(&t->clients.operator.client, &t->id, &cfg);
+    EXPECT("error adding team for operator", err);
+    err = aranya_add_team(&t->clients.membera.client, &t->id, &cfg);
+    EXPECT("error adding team for membera", err);
+    err = aranya_add_team(&t->clients.memberb.client, &t->id, &cfg);
+    EXPECT("error adding team for memberb", err);
+
+    return ARANYA_ERROR_SUCCESS;
+exit:
+    return err;
```
djohnson-spideroak, line 487: Is this used?

PR #237, opened 2025-04-30
ci: fix sync now test flake

gknopf-aranya:
The sync now test should test without adding sync peers so that when `sync_now` is invoked, it can prove that it caused a sync to occur.

At some point, a PR introduced adding sync peers to this test, which caused a CI flake.
Could have been one of my PRs, haven't investigated which PR introduced it yet.


nikki-aranya (approved):

gknopf-aranya:
@chip-so, @elagergren-spideroak PR to resolve a CI flake is ready for review

elagergren-spideroak (approved):

PR #245, opened 2025-05-01
aqc: Implement Bidirectional C API 

nikki-aranya:
Implements a C API for bidirectional AQC channels.

### DIFF
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -96,6 +96,22 @@
 #define ARANYA_ID_STR_LEN (((ARANYA_ID_LEN * 1375) / 1000) + 1)
 
 /**
+ * An enum containing all [`AranyaAqcChannel`](@ref AranyaAqcChannel) variants.
+ */
+enum AranyaAqcChannelType
+#ifdef __cplusplus
+  : uint8_t
+#endif // __cplusplus
+ {
+    ARANYA_AQC_CHANNEL_TYPE_SENDER,
+    ARANYA_AQC_CHANNEL_TYPE_RECEIVER,
+    ARANYA_AQC_CHANNEL_TYPE_BIDIRECTIONAL,
+};
+#ifndef __cplusplus
+typedef uint8_t AranyaAqcChannelType;
+#endif // __cplusplus
+
+/**
  * Valid channel operations for a label assignment.
  */
 enum AranyaChanOp
@@ -392,6 +408,23 @@ typedef struct AranyaLabelId {
 } AranyaLabelId;
 
 /**
+ * A type to represent a span of time in nanoseconds.
+ */
+typedef uint64_t AranyaDuration;
+
+/**
+ * A type containing the response of receiving a channel.
+ */
+typedef struct ARANYA_ALIGNED(8) AranyaAqcChannel {
+    /**
+     * This field only exists for size purposes. It is
+     * UNDEFINED BEHAVIOR to read from or write to it.
+     * @private
+     */
+    uint8_t __for_size_only[192];
+} AranyaAqcChannel;
+
+/**
  * Bidirectional AQC Channel Object
  */
 typedef struct ARANYA_ALIGNED(8) AranyaAqcBidiChannel {
@@ -404,28 +437,35 @@ typedef struct ARANYA_ALIGNED(8) AranyaAqcBidiChannel {
 } AranyaAqcBidiChannel;
 
 /**
- * Channel ID for AQC bidi channel.
+ * Sender AQC Channel Object
  */
-typedef struct AranyaAqcBidiChannelId {
-    struct AranyaId id;
-} AranyaAqcBidiChannelId;
-
-/**
- * A type containing the response of receiving a channel.
- */
-typedef struct ARANYA_ALIGNED(8) AranyaAqcChannelType {
+typedef struct ARANYA_ALIGNED(8) AranyaAqcSenderChannel {
     /**
      * This field only exists for size purposes. It is
      * UNDEFINED BEHAVIOR to read from or write to it.
      * @private
      */
-    uint8_t __for_size_only[192];
-} AranyaAqcChannelType;
+    uint8_t __for_size_only[112];
+} AranyaAqcSenderChannel;
 
 /**
- * A type to represent a span of time in nanoseconds.
+ * Receiver AQC Channel Object
  */
-typedef uint64_t AranyaDuration;
+typedef struct ARANYA_ALIGNED(8) AranyaAqcReceiverChannel {
+    /**
+     * This field only exists for size purposes. It is
+     * UNDEFINED BEHAVIOR to read from or write to it.
+     * @private
+     */
+    uint8_t __for_size_only[88];
+} AranyaAqcReceiverChannel;
+
+/**
+ * Channel ID for AQC bidi channel.
+ */
+typedef struct AranyaAqcBidiChannelId {
+    struct AranyaId id;
+} AranyaAqcBidiChannelId;
 
 #ifdef __cplusplus
 extern "C" {
@@ -1449,113 +1489,6 @@ AranyaError aranya_revoke_label_ext(struct AranyaClient *client,
                                     struct AranyaExtError *__ext_err);
 
 /**
- * Releases any resources associated with `ptr`.
- *
- * `ptr` must either be null or initialized by `::aranya_aqc_bidi_channel_init`.
- *
- * @relates AranyaAqcBidiChannel
- */
-AranyaError aranya_aqc_bidi_channel_cleanup(struct AranyaAqcBidiChannel *ptr);
-
-/**
- * Releases any resources associated with `ptr`.
- *
- * `ptr` must either be null or initialized by `::aranya_aqc_bidi_channel_init`.
- *
- * @relates AranyaAqcBidiChannel
- */
-AranyaError aranya_aqc_bidi_channel_cleanup_ext(struct AranyaAqcBidiChannel *ptr,
-                                                struct AranyaExtError *__ext_err);
-
-/**
- * Create a bidirectional AQC channel between this device and another peer.
- *
- * Permission to perform this operation is checked against the Aranya policy.
- *
- * @param client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param peer the peer's network identifier [`AranyaNetIdentifier`](@ref AranyaNetIdentifier).
- * @param label_id the AQC channel label ID [`AranyaLabelId`](@ref AranyaLabelId) to create the channel with.
- * @param channel the AQC channel object [`AranyaAqcBidiChannel`](@ref AranyaAqcBidiChannel) that holds channel info.
- *
- * @relates AranyaClient.
- */
-AranyaError aranya_aqc_create_bidi_channel(struct AranyaClient *client,
-                                           const struct AranyaTeamId *team,
-                                           AranyaNetIdentifier peer,
-                                           const struct AranyaLabelId *label_id,
-                                           struct AranyaAqcBidiChannel *channel);
-
-/**
- * Create a bidirectional AQC channel between this device and another peer.
- *
- * Permission to perform this operation is checked against the Aranya policy.
- *
- * @param client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param peer the peer's network identifier [`AranyaNetIdentifier`](@ref AranyaNetIdentifier).
- * @param label_id the AQC channel label ID [`AranyaLabelId`](@ref AranyaLabelId) to create the channel with.
- * @param channel the AQC channel object [`AranyaAqcBidiChannel`](@ref AranyaAqcBidiChannel) that holds channel info.
- *
- * @relates AranyaClient.
- */
-AranyaError aranya_aqc_create_bidi_channel_ext(struct AranyaClient *client,
-                                               const struct AranyaTeamId *team,
-                                               AranyaNetIdentifier peer,
-                                               const struct AranyaLabelId *label_id,
-                                               struct AranyaAqcBidiChannel *channel,
-                                               struct AranyaExtError *__ext_err);
-
-/**
- * Delete a bidirectional AQC channel.
- *
- * @param client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param channel_id the AQC Channel [`AranyaAqcBidiChannel`](@ref AranyaAqcBidiChannel) to delete.
- *
- * @relates AranyaClient.
- */
-AranyaError aranya_aqc_delete_bidi_channel(struct AranyaClient *client,
-                                           const struct AranyaAqcBidiChannelId *channel);
-
-/**
- * Delete a bidirectional AQC channel.
- *
- * @param client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param channel_id the AQC Channel [`AranyaAqcBidiChannel`](@ref AranyaAqcBidiChannel) to delete.
- *
- * @relates AranyaClient.
- */
-AranyaError aranya_aqc_delete_bidi_channel_ext(struct AranyaClient *client,
-                                               const struct AranyaAqcBidiChannelId *channel,
-                                               struct AranyaExtError *__ext_err);
-
-/**
- * Releases any resources associated with `ptr`.
- *
- * `ptr` must either be null or initialized by `::aranya_aqc_channel_type_init`.
- *
- * @relates AranyaAqcChannelType
- */
-AranyaError aranya_aqc_channel_type_cleanup(struct AranyaAqcChannelType *ptr);
-
-/**
- * Releases any resources associated with `ptr`.
- *
- * `ptr` must either be null or initialized by `::aranya_aqc_channel_type_init`.
- *
- * @relates AranyaAqcChannelType
- */
-AranyaError aranya_aqc_channel_type_cleanup_ext(struct AranyaAqcChannelType *ptr,
-                                                struct AranyaExtError *__ext_err);
-
-AranyaError aranya_aqc_receive_channel(struct AranyaClient *client,
-                                       struct AranyaAqcChannelType *channel);
-
-AranyaError aranya_aqc_receive_channel_ext(struct AranyaClient *client,
-                                           struct AranyaAqcChannelType *channel,
-                                           struct AranyaExtError *__ext_err);
-
-/**
  * Configures how often the peer will be synced with.
  *
  * By default, the interval is not set. It is an error to call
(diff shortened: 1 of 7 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -61,6 +62,10 @@ pub enum Error {
     #[capi(msg = "AQC library error")]
     Aqc,
 
+    /// Tried to do something on a channel that was closed.
```
gknopf-aranya, line 65:
```suggestion
    /// Attempted to perform operation on a closed AQC channel.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1356,3 +1296,266 @@ pub fn query_labels(
     *labels_len = data.len();
     Ok(())
 }
+
+/// A type containing the response of receiving a channel.
```
gknopf-aranya, line 1300:
```suggestion
/// A type containing the AQC channel variant.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1356,3 +1296,266 @@ pub fn query_labels(
...
+
+/// Delete a bidirectional AQC channel.
+///
+/// @param client the Aranya Client [`Client`].
+/// @param channel_id the AQC Channel [`AqcBidiChannel`] to delete.
+///
+/// @relates AranyaClient.
+pub fn aqc_delete_bidi_channel(
+    client: &mut Client,
+    channel: &AqcBidiChannelId,
```
gknopf-aranya, line 1370: This arg doesn't match the docs. `channel_id` is more descriptive than `channel` so I'd pick that one.
```suggestion
    channel_id: &AqcBidiChannelId,
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1356,3 +1296,266 @@ pub fn query_labels(
...
+
+/// A type containing the response of receiving a channel.
+#[aranya_capi_core::derive(Cleanup)]
+#[aranya_capi_core::opaque(size = 192, align = 8)]
+pub type AqcChannel = Safe<imp::AqcChannelType>;
+
+/// An enum containing all [`AqcChannel`] variants.
+#[repr(u8)]
+#[derive(Copy, Clone, Debug)]
+pub enum AqcChannelType {
```
gknopf-aranya, line 1308: You can use this enum directly instead of defining a new `imp::AqcChannelType` type.

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: This needs to be changed in ben's PR to be a channel since we aren't using IDs, I'm leaving it like this so I can get it to compile while I wait

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1356,3 +1296,266 @@ pub fn query_labels(
...
+///
+/// # Example
+/// ```C
+/// AqcChannel channel;
+/// AqcBidiChannel bidi;
+/// AqcSenderChannel sender;
+/// AqcReceiverChannel receiver;
+/// switch (aranya_aqc_get_channel_type(channel)) {
+///     case ARANYA_AQC_CHANNEL_TYPE_BIDIRECTIONAL:
+///         return aranya_aqc_get_bidirectional_channel(channel, &bidi);
```
gknopf-aranya, line 1409: This example wouldn't work if the return types are different.

nikki-aranya: it's all technically AranyaError but fair

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1356,3 +1296,266 @@ pub fn query_labels(
...
+            );
+            return Err(e.into());
+        }
+    }
+
+    Ok(())
+}
+
+/*
+TODO(nikki): AQC uni support
```
gknopf-aranya, line 1509: We should uncomment or delete the unidirectional code before merging the PR.

nikki-aranya: I don't think there's any harm in keeping this commented out (unless we update this to also include uni support) since it'll immediately be followed by a PR that makes all the code useful, reduce code churn

nikki-aranya:
Note to self: still need to wait for the horrific rebasing on upstream, add example code to example.c, make sure CI actually passes, and we can probably implement unidirectional channels without too much pain on this PR. My TODO on refactoring to accessors can probably be kicked down the road

crates/aranya-client-capi/src/imp/aqc.rs
```diff
@@ -0,0 +1,89 @@
+use aranya_capi_core::safe::{TypeId, Typed};
+use aranya_client::aqc::net::{self as aqc};
+
+// TODO(nikki): decide if we want to refactor this to add accessors and remove
+// pub(crate) on inners? We also expose .inner in `client.rs`.
+
+pub struct AqcBidiChannel {
```
gknopf-aranya, line 7: Need docs for `pub` types

crates/aranya-client-capi/src/imp/error.rs
```diff
@@ -22,6 +22,12 @@ pub enum Error {
     #[error("buffer too small")]
     BufferTooSmall,
 
+    #[error("AQC server was closed")]
+    AqcServerClosed,
+
+    #[error("AQC stream was closed")]
+    AqcStreamClosed,
```
gknopf-aranya, line 29: @benz-spideroak, @nikki-aranya will we need an error for when the entire channel is closed?

crates/aranya-client-capi/src/imp/error.rs
gknopf-aranya, line 1: Let's add errors for creating a channel/stream if possible. As well as errors for when sending/receiving data via a stream fails.

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
 
     // Create channel using Member A's client
     printf("creating AQC channel \n");
-    AranyaAqcBidiChannelId chan_id;
+    AranyaAqcBidiChannel send_channel;
```
gknopf-aranya, line 768: Why is this variable called `send_channel` when a bidi channel can send or receive? Seems like it should be called `bidi_channel`.

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
...
+    EXPECT("error receiving aqc bidi channel", err);
+
+    aranya_aqc_get_channel_type(&channel, &channel_type);
+    switch (channel_type) {
+    case ARANYA_AQC_CHANNEL_TYPE_BIDIRECTIONAL:
+        aranya_aqc_get_bidirectional_channel(&channel, &recv_channel);
+        break;
+    case ARANYA_AQC_CHANNEL_TYPE_SENDER:
+    case ARANYA_AQC_CHANNEL_TYPE_RECEIVER:
+        fprintf(stderr, "somehow got the wrong AQC channel type");
```
gknopf-aranya, line 789: If we add a method for converting a channel type to a string, we could print out the incorrect channel type as a string in this error.

examples/c/example.c
nikki-aranya: We have two bidi channels in this test and I didn't feel like naming them 1 and 2, they're send and recv because I'm opening a uni stream and this is the send end of it

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
...
+                                         &send_channel);
     EXPECT("error creating aqc bidi channel", err);
 
-    // TODO: send AQC data
+    // Receive the channel on Member B's side
+    AranyaAqcChannel channel;
+    AranyaAqcChannelType channel_type;
+    AranyaAqcBidiChannel recv_channel;
+
+    err = aranya_aqc_receive_channel(&t->clients.memberb.client, &channel);
```
gknopf-aranya, line 779: We'll want to add a sleep between the `create` and `receive` methods to allow the peer time to sync new commands and receive the ephemeral `ctrl` message via the network.

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
...
+    case ARANYA_AQC_CHANNEL_TYPE_RECEIVER:
+        fprintf(stderr, "somehow got the wrong AQC channel type");
+        return ARANYA_ERROR_AQC;
+    }
+
+    // Open up a unidirectional stream from Member A to Member B
+    AranyaAqcSendStream send;
+    AranyaAqcSendStream send_invalid;
+    AranyaAqcReceiveStream recv;
+    bool valid_send;
```
gknopf-aranya, line 797: Instead of using a boolean parameter to check whether the send stream is valid, we should check whether `send_invalid` is `NULL`. That would be more idiomatic C.

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
...
+    err = aranya_aqc_bidi_create_uni_stream(&t->clients.membera.client,
+                                            &send_channel, &send);
+    EXPECT("error creating an aqc uni stream", err);
+
```
gknopf-aranya, line 801: Need a sleep here. Or a way to loop over `receive_stream` until it returns a valid stream.

gknopf-aranya: I prefer looping until valid over a sleep because it won't cause a race.

examples/c/example.c
gknopf-aranya: Or better yet, a way to loop over `receive_channel` until a valid channel is returned.

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
...
+    err = aranya_aqc_bidi_receive_stream(&t->clients.memberb.client,
+                                         &recv_channel, &send_invalid, &recv,
+                                         &valid_send);
+    EXPECT("error receiving an aqc uni stream", err);
+    // Validate that we never got a send stream since this is a uni stream.
+    if (valid_send == true) {
```
gknopf-aranya, line 807: Ditto on checking whether `send` is `NULL` and removing the `valid_send` parameter.

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
...
+    }
+
+    const char *string = "hello from aqc!";
+    uint8_t buffer[BUFFER_LEN];
+    size_t received_length;
+    aranya_aqc_send_data(&t->clients.membera.client, &send,
+                         (const uint8_t *)string, strlen(string));
+    aranya_aqc_receive_data(&t->clients.memberb.client, &recv, buffer,
+                            BUFFER_LEN, &received_length);
+    if (strcmp(string, (const char *)buffer)) {
```
gknopf-aranya, line 819: `strncmp`

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
...
+    if (valid_send == true) {
+        fprintf(stderr, "somehow received an AQC send stream");
+        return ARANYA_ERROR_AQC;
+    }
+
+    const char *string = "hello from aqc!";
+    uint8_t buffer[BUFFER_LEN];
+    size_t received_length;
+    aranya_aqc_send_data(&t->clients.membera.client, &send,
+                         (const uint8_t *)string, strlen(string));
```
gknopf-aranya, line 816: `strlen(string)` is the length of the string not including the null terminator, which means `strcmp` is comparing a null terminated string to a buffer that is not guaranteed to be null terminated.
There are a few options:
1. Send the string with the null terminator.
2. Add the null terminator on the receiving side.
3. Provide utility methods that can send either binary or string data.

examples/c/example.c
```diff
@@ -774,21 +765,70 @@ AranyaError run_aqc_example(Team *t) {
...
+        fprintf(stderr, "somehow got the wrong AQC channel type");
+        return ARANYA_ERROR_AQC;
+    }
+
+    // Open up a unidirectional stream from Member A to Member B
+    AranyaAqcSendStream send;
+    AranyaAqcSendStream send_invalid;
+    AranyaAqcReceiveStream recv;
+    bool valid_send;
+    err = aranya_aqc_bidi_create_uni_stream(&t->clients.membera.client,
```
gknopf-aranya, line 798: It looks like the example is not testing `aranya_aqc_bidi_create_bidi_stream` which is part of the Rust API.

nikki-aranya: I thought we agreed that the C example doesn't need to cover every single function to be useful

examples/c/example.c
gknopf-aranya: The naming is a bit confusing though because we have bidi/uni channels and send/recv stream objects.
How about `bidi_send`, `bidi_recv`?

examples/c/example.c
gknopf-aranya: That discussion applies once we have integration tests which are different from the examples. Right now, the examples are also serving as our integration tests. We need to include each C API method in the C example because they are not tested anywhere else. How would we know they work and haven't broken later if there are no tests for certain API methods running in CI?

examples/c/example.c
gknopf-aranya: We support optional args which translate to C pointers. See the query APIs.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// An enum containing all [`AqcChannel`] variants.
+#[repr(u8)]
+#[derive(Copy, Clone, Debug)]
+pub enum AqcChannelType {
+    Sender,
+    Receiver,
+    Bidirectional,
+}
+
+/// AQC Bidirectional Channel Object
```
gknopf-aranya, line 1378: nit: we have a convention of adding periods at the end of doc comments.

examples/c/example.c
nikki-aranya: That's an optional *input* arg, not an optional *output* arg, I have no way to tell the user that something went wrong since we can only return AranyaError

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
+    Ok(())
 }
 
+/* TODO(nikki): these will panic currently, so let's stub them
+
 /// Delete a bidirectional AQC channel.
 ///
 /// @param client the Aranya Client [`Client`].
-/// @param chan the AQC channel ID [`AqcBidiChannelId`] of the channel to delete.
+/// @param channel_id the AQC Channel [`AqcBidiChannel`] to delete.
```
gknopf-aranya, line 1472: `channel_id` in docs does not match `channel` param name

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
-        .block_on(client.inner.aqc().delete_bidi_channel(chan.into()))?;
+        .block_on(client.inner.aqc().delete_bidi_channel(channel.into()))?;
     Ok(())
 }
 
 /// Delete a unidirectional AQC channel.
 ///
 /// @param client the Aranya Client [`Client`].
-/// @param chan the AQC channel ID [`AqcUniChannelId`] of the channel to delete.
+/// @param chan the AQC Channel [`AqcSenderChannel`] to delete.
```
gknopf-aranya, line 1490: `chan` doc does not match `channel` param

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
     let client = client.deref_mut();
     client
         .rt
-        .block_on(client.inner.aqc().delete_uni_channel(chan.into()))?;
+        .block_on(client.inner.aqc().delete_uni_channel(channel.into()))?;
+    Ok(())
+}
+*/
+
+/// Waits until an AQC channel is received from the client.
```
gknopf-aranya, line 1505: This doc implies that the function will block until an AQC channel is received.
The way the Rust API works is this method returns a channel if it exists, or returns `None` if no channel was received.
In the C API, we should return a specific error if no new channels have been received.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// @relates AranyaClient.
+pub fn aqc_receive_channel(
+    client: &mut Client,
+    channel: &mut MaybeUninit<AqcChannel>,
+) -> Result<(), imp::Error> {
+    let client = client.deref_mut();
+    let chan = client
+        .rt
+        .block_on(client.inner.aqc().receive_channel())
+        .ok_or(imp::Error::AqcServerClosed)?;
```
gknopf-aranya, line 1521: This is a non-fatal error. If a channel is not received, the caller can try again.
```suggestion
        .ok_or(imp::Error::AqcChannelNotReceived)?;
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// @relates AranyaClient.
+pub fn aqc_get_channel_type(channel: &mut AqcChannel) -> AqcChannelType {
+    match channel.inner {
+        aqc::AqcChannelType::Bidirectional { .. } => AqcChannelType::Bidirectional,
+        aqc::AqcChannelType::Sender { .. } => AqcChannelType::Sender,
+        aqc::AqcChannelType::Receiver { .. } => AqcChannelType::Receiver,
+    }
+}
+
+/// Converts the [`AqcChannel`]` into an [`AqcBidiChannel`] for use sending/receiving data.
```
gknopf-aranya, line 1564: grammar:
```suggestion
/// Converts the [`AqcChannel`]` into an [`AqcBidiChannel`] for sending/receiving data.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
+        }
+        _ => Err(InvalidArg::new(
+            "channel",
+            "Tried to call get_receiver_channel with a `AqcChannel` that wasn't a receiver!",
+        )
+        .into()),
+    }
+}
+
+/// Create a bidirectional stream from an [`AqcBidiChannel`].
```
gknopf-aranya, line 1645:
```suggestion
/// Create a bidirectional stream from a [`AqcBidiChannel`].
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
+        .rt
+        .block_on(channel.inner.create_unidirectional_stream())?;
+
+    Safe::init(stream, imp::AqcSendStream::new(send));
+    Ok(())
+}
+
+/// Obtains the receive (and potentially send) ends of a stream.
+///
+/// Note that the send stream will only be initialized if this returns true.
```
gknopf-aranya, line 1692: I wonder if we should provide an explicit bool arg for whether send is initialized.
Someone could make the mistake of assuming this returning true just means a stream was received.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
+        .deref_mut()
+        .rt
+        .block_on(channel.inner.receive_unidirectional_stream())?
+        .ok_or(imp::Error::AqcStreamClosed)?;
+
+    Safe::init(stream, imp::AqcReceiveStream::new(recv));
+    Ok(())
+}
+
+/// Send some data over an Aranya QUIC Channel.
```
gknopf-aranya, line 1768:
```suggestion
/// Send some data over an Aranya QUIC Channel stream.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1361,445 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// @relates AranyaClient.
+pub fn aqc_send_data(
+    client: &mut Client,
+    stream: &mut AqcSendStream,
+    data: &[u8],
+) -> Result<(), imp::Error> {
+    Ok(client.deref_mut().rt.block_on(stream.inner.send(data))?)
+}
+
+/// Receive some data from an Aranya QUIC Channel.
```
gknopf-aranya, line 1784:
```suggestion
/// Receive some data from an Aranya QUIC Channel stream.
```

examples/c/example.c
djohnson-spideroak: I am not sure its needed, we could just do different things in the different switch arms.

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: not sure we'll finish channel deletion in time, not worrying about getting those right, but I'll fix this

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: That's because it does.
`If there are no messages in the channel's buffer, but the channel has not yet been closed, this method will sleep until a message is sent or the channel is closed. Note that if [close] is called, but there are still outstanding [Permits] from before it was closed, the channel is not considered closed by recv until the permits are released.`

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: If it returns None, it means the channel is closed, AqcClient::receive_channel clearly states that (and so does mpsc::channel)

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: I think you have to provide a bool pointer regardless due to how __output works, which they can technically provide NULL but idk how aranya handles that. 
```c
/**
 * Obtains the receive (and potentially send) ends of a stream.
 *
 * Note that the send stream will only be initialized if this returns true.
 *
 * @param client the Aranya Client [`AranyaClient`](@ref AranyaClient).
 * @param channel the AQC channel object [`AranyaAqcBidiChannel`](@ref AranyaAqcBidiChannel).
 * @param send_stream the sending side of a stream [`AranyaAqcSendStream`](@ref AranyaAqcSendStream).
 * @param recv_stream the receiving side of a stream [`AranyaAqcReceiveStream`](@ref AranyaAqcReceiveStream).
 * @param __output whether or not we received a `send_stream`.
 *
 * @relates AranyaClient.
 */
AranyaError aranya_aqc_bidi_receive_stream(struct AranyaClient *client,
                                           struct AranyaAqcBidiChannel *channel,
                                           struct AranyaAqcSendStream *send_stream,
                                           struct AranyaAqcReceiveStream *recv_stream,
                                           bool *__output);
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Just realized this while testing things out. I don't think we want a blocking API for AQC. This could cause the application to hang unexpectedly. I'd prefer that we implement all the AQC API methods in a non-blocking way.

cc: @benz-spideroak, @djohnson-spideroak, @elagergren-spideroak, @jdygert-spok, @mooreoak

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Yeah, at least the parameter would have a more user-friendly name than `__output`.

examples/c/example.c
nikki-aranya: re: blocking, this just blocks until it receives a stream, the only way create_stream errors is an unspecified error, so it's catastrophic if it errors

examples/c/example.c
```diff
@@ -40,6 +40,23 @@
...
+// EXPECT.
+#define TRY_RECEIVE(C, M)                                                      \
+    while (1) {                                                                \
+        err = (C);                                                             \
+        if (err == ARANYA_ERROR_SUCCESS) {                                     \
+            break;                                                             \
+        }                                                                      \
+        switch (err) {                                                         \
+        case ARANYA_ERROR_EMPTY:                                               \
+            sleep(1);                                                        \
```
gknopf-aranya, line 53: A 1 second sleep is long. 100ms would perform better without spinning too much.

gknopf-aranya:
The rust example is not running on this PR currently. I made fixes to that on my other AQC test PR.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1368,447 @@ pub unsafe fn aqc_remove_net_identifier(
...
+
+/// Converts the [`AqcChannel`]` into an [`AqcReceiverChannel`] for receiving data.
+///
+/// Returns `ARANYA_ERROR_INVALID_ARGUMENT` if called when the AqcChannel is the wrong type.
+///
+/// @param channel the AQC channel container [`AqcChannel`].
+/// @param receiver the AQC channel object [`AqcReceiverChannel`].
+///
+/// @relates AranyaClient.
+pub fn aqc_get_receiver_channel(
```
gknopf-aranya, line 1629: This method is not tested by the C example

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1368,447 @@ pub unsafe fn aqc_remove_net_identifier(
...
+
+/// Converts the [`AqcChannel`]` into an [`AqcSenderChannel`] for sending data.
+///
+/// Returns `ARANYA_ERROR_INVALID_ARGUMENT` if called when the AqcChannel is the wrong type.
+///
+/// @param channel the AQC channel container [`AqcChannel`].
+/// @param sender the AQC channel object [`AqcSenderChannel`].
+///
+/// @relates AranyaClient.
+pub fn aqc_get_sender_channel(
```
gknopf-aranya, line 1602: This method is not tested by the C example

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1368,447 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param client the Aranya Client [`Client`].
+/// @param team the team's ID [`TeamId`].
+/// @param peer the peer's network identifier [`NetIdentifier`].
+/// @param label_id the AQC channel label ID [`LabelId`] to create the channel with.
+/// @param channel the AQC channel object [`AqcSenderChannel`].
+///
+/// @relates AranyaClient.
+pub unsafe fn aqc_create_uni_channel(
```
gknopf-aranya, line 1453: This method is not tested by the C example

examples/c/example.c
gknopf-aranya: Added a test for this

examples/c/example.c
gknopf-aranya: We decided to today we might as well add this string conversion. Generally, having the ability to print out a string representation of enums is useful.

examples/c/example.c
```diff
@@ -774,21 +782,82 @@ AranyaError run_aqc_example(Team *t) {
...
+        "error receiving aqc bidi channel");
+
+    aranya_aqc_get_channel_type(&channel, &channel_type);
+    switch (channel_type) {
+    case ARANYA_AQC_CHANNEL_TYPE_BIDIRECTIONAL:
+        aranya_aqc_get_bidirectional_channel(&channel, &bidi_recv);
+        break;
+    case ARANYA_AQC_CHANNEL_TYPE_SENDER:
+    case ARANYA_AQC_CHANNEL_TYPE_RECEIVER:
+        fprintf(stderr, "somehow got the wrong AQC channel type");
```
gknopf-aranya, line 811: Added `fprintf`s are missing newline. E.g.
```
Trying to receive the bidi channel
somehow got the wrong AQC channel typeerror running aqc example
cleaning up the Aranya team 
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: After further inspection, it doesn't appear that it's possible to test this method.
`aqc_try_receive_channel` will only ever return a `AranyaAqcBidiChannel` or `AranyaAqcReceiverChannel` type because those are the only channel types that can be received.
A `AranyaAqcSenderChannel` type is created when creating a uni channel, but a channel of that type cannot be received by a peer.
There are no other methods in the API that return a `AqcChannel` type.

So, it's impossible for me to get a `AqcChannel` object of type `AqcSenderChannel` from the current API.

This leaves us with a couple options:
1. Remove `aqc_get_sender_channel()` from the API since it should never be used
2. Add a way to get a `AqcChannel` object of type `AqcSenderChannel`

I vote for 1. since it keeps the API simpler and removes unused code.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Added a test for this.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Added a test for this

examples/c/example.c
gknopf-aranya: Added newlines

gknopf-aranya:
The C example tests all the methods in the AQC C API now

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -385,18 +413,76 @@ typedef const char *AranyaAddr;
...
-    struct AranyaId id;
-} AranyaAqcBidiChannelId;
+typedef struct ARANYA_ALIGNED(8) AranyaAqcChannel {
+    /**
+     * This field only exists for size purposes. It is
+     * UNDEFINED BEHAVIOR to read from or write to it.
+     * @private
+     */
+    uint8_t __for_size_only[192];
+} AranyaAqcChannel;
```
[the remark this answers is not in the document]
gknopf-aranya: Here's a summary of the lifecycle:
`AranyaAqcBidiChannel` and `AranyaAqcSenderChannel` are initialized when creating a corresponding channel.
`AranyaAqcChannel` is a type returned when receiving a channel of unknown type.
`aqc_get_channel_type()` determines what type `AranyaAqcChannel` is, so `aranya_aqc_get_bidirectional_channel()` or `aranya_aqc_get_receiver_channel()` can return the corresponding `AranyaAqcReceiverChannel` or `AranyaAqcBidiChannel` type. An `AranyaAqcSenderChannel` can be created by creating an unidirectional channel, but can never be received.

Once you have a channel object (e.g. AranyaAqcSenderChannel, AranyaAqcReceiverChannel, AranyaAqcBidiChannel), you can create/receive streams and send/receive data as applicable based on the channel direction.

benz-spideroak:
Fix the [Correctness checks and linting](https://github.com/aranya-project/aranya/actions/runs/15126150428/job/42518476609?pr=245) and then it looks good to me.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1889,95 +1978,580 @@ AranyaError aranya_aqc_remove_net_identifier_ext(struct AranyaClient *client,
...
+ * Create a bidirectional AQC channel between this device and a peer.
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
  * @param client the Aranya Client [`AranyaClient`](@ref AranyaClient).
  * @param team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
  * @param peer the peer's network identifier [`AranyaNetIdentifier`](@ref AranyaNetIdentifier).
  * @param label_id the AQC channel label ID [`AranyaLabelId`](@ref AranyaLabelId) to create the channel with.
- * @param __output the AQC channel's ID [`AranyaAqcBidiChannelId`](@ref AranyaAqcBidiChannelId)
+ * @param channel the AQC channel object [`AranyaAqcBidiChannel`](@ref AranyaAqcBidiChannel).
```
djohnson-spideroak, line 2122: to be clear, this is initialized by calling this function? We should make sure the output params are clearly marked.

djohnson-spideroak (commented):
LGTM, once this PR and the AQC net PR is ready, we can merge them into main. The fmt error needs to be resolved, the cargo vet stuff will be resolved by the AQC net PR

crates/aranya-client-capi/output/aranya-client.h
nikki-aranya: Correct. I don't know what the consensus on taking a pointer vs returning a thing is, from what I saw from all the Safe::init stuff on the config builder we just return (). We're also unable to return a tuple so any params that need to initialize multiple things you just can't

benz-spideroak (approved):

crates/aranya-client-capi/output/aranya-client.h
jdygert-spok: We ought to add `[in] / [out] / [in,out]` to all params in the C api.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1350,69 +1369,418 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// @param[in]  buffer_len length of the target buffer.
+/// @param[out] __output the number of bytes written to the buffer.
+///
+/// @relates AranyaClient.
+pub fn aqc_try_receive_data(
+    stream: &mut AqcReceiveStream,
+    buffer: &mut [u8],
+) -> Result<usize, imp::Error> {
+    Ok(stream.inner.try_receive(buffer)?)
+}
```
jdygert-spok, line 1786: `buffer` should be an out param.

This is different from our other buffer related functions which use a single in/out param for the length, I don't have a strong opinion but we should be consistent either way.

This should take `&mut [MaybeUninit<u8>]`.

It should write out the needed length when the buffer is too short, although this needs changes upstream to not drop the message in that case.

Cargo.toml
```diff
@@ -100,3 +100,4 @@ tag-prefix = ""
 
 [patch.crates-io]
 rustls = { git = "https://github.com/elagergren-spideroak/rustls", rev = "fc478a5e9d8e2ab08b7d2d3b62f57b422033a18c" }
+aranya-capi-core = { git = "https://github.com/aranya-project/aranya-core", rev = "d0d5c84cc82138709678f0c7be8895141d741d83" }
```
djohnson-spideroak, line 103: does this need to be a git path?

gknopf-aranya: I don't see a newer version than 0.4.0 on crates.io, so we'll need another release.

cc: @jdygert-spok

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1349,71 +1367,552 @@ pub unsafe fn aqc_remove_net_identifier(
...
-    Ok(AqcBidiChannelId {
-        id: chan_id.aqc_id().into_id().into(),
-    })
+
+    AqcSenderChannel::init(channel, imp::AqcSenderChannel::new(chan));
+    Ok(())
 }
 
-/*
+/* TODO(nikki): fixme
```
djohnson-spideroak, line 1481: fix or rm

Cargo.toml
nikki-aranya: yes, this has a couple needed fixes for opaque, see https://github.com/aranya-project/aranya-core/tree/opaque-pub-branch

gknopf-aranya: I've opened a PR here to release the new aranya-core crates needed by this PR:
https://github.com/aranya-project/aranya-core/pull/258

jdygert-spok: The fixes were merged: https://github.com/aranya-project/aranya-core/pull/252

gknopf-aranya:
@nikki-aranya, @djohnson-spideroak here's an audit of AQC C API test coverage in the C example:
- [x] aqc_config_build
- [x] aqc_config_builder_set_address
- [x] aqc_assign_net_identifier
- [x] aqc_remove_net_identifier
- [x] aqc_create_bidi_channel
- [x] aqc_create_uni_channel
- [ ] aqc_delete_bidi_channel
- [ ] aqc_delete_uni_channel
- [x] aqc_try_receive_channel
- [x] aqc_get_bidi_channel
- [x] aqc_get_receive_channel
- [x] aqc_bidi_create_bidi_stream
- [x] aqc_bidi_stream_send
- [x] aqc_bidi_stream_try_recv
- [ ] aqc_bidi_create_uni_stream
- [x] aqc_bidi_try_receive_stream
- [ ] aqc_send_create_uni_stream
- [ ] aqc_recv_try_receive_uni_stream
- [x] aqc_send_stream_send
- [x] aqc_recv_stream_try_recv

examples/c/example.c
```diff
@@ -659,12 +674,263 @@ AranyaError run(Team *t) {
...
+
+    if (strncmp("hello from aqc uni memberb!", uni_buffer, BUFFER_LEN)) {
+        fprintf(stderr, "received string doesn't match\n");
+        err = ARANYA_ERROR_AQC;
+        goto exit;
+    }
+    printf("Received AQC bidi stream data: \"%s\"\n", uni_buffer);
+
+exit:
+    aranya_aqc_delete_bidi_channel(ctx->client, &bidi_chan);
```
gknopf-aranya, line 791: Do we need to check if this returns an error?

examples/c/example.c
```diff
@@ -9,6 +9,7 @@
 // Note: this file is formatted with `clang-format`.
 
 #include <ctype.h>
+#include <pthread.h>
```
gknopf-aranya, line 12: @djohnson-spideroak, @elagergren-spideroak is it okay to test with `pthread` here?
I thought we're wanting to support single-threaded operation with our API.

examples/c/example.c
```diff
@@ -659,12 +674,263 @@ AranyaError run(Team *t) {
...
+        fprintf(stderr, "received string doesn't match\n");
+        err = ARANYA_ERROR_AQC;
+        goto exit;
+    }
+    printf("Received AQC uni stream data: \"%s\"\n", uni_buffer);
+
+    // Send some data to make sure it works.
+    printf("Sending some data back from member b\n");
+    const char *bidi_string = "hello from aqc memberb!";
+    aranya_aqc_send_stream_send(ctx->client, &bidi_send_stream,
```
gknopf-aranya, line 899: Check for error here

gknopf-aranya (dismissed):
Overall this PR lgtm. It's getting difficult to review the example since there is so much code in there now that isn't organized under simpler test-cases. A lot of the variable and function names are so similar that it would be difficult to catch typos.

examples/c/example.c
djohnson-spideroak: This isnt part of the API, so I think its fine. The single thread restriction was for the actual library and to have the user drive the operations.

nikki-aranya:
> Overall this PR lgtm. It's getting difficult to review the example since there is so much code in there now that isn't organized under simpler test-cases. A lot of the variable and function names are so similar that it would be difficult to catch typos.

The type system will catch a lot of it, but also this is an excuse to prioritize non-example C tests.

djohnson-spideroak (approved):

gknopf-aranya (approved):

gknopf-aranya (dismissed):

examples/rust/Cargo.toml
```diff
@@ -33,15 +33,15 @@ aranya-client = { path = "../../crates/aranya-client" }
...
-tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
+anyhow = { version = "=1.0.97" }
+backon = { version = "=1.4.0" }
+buggy = { version = "=0.1.0" }
+bytes = { version = "=1.10.0" }
+hex = { version = "=0.4.3" }
+tempfile = { version = "=3.17.1" }
+tokio = { version = "=1.44.2", features = ["io-util", "macros", "net", "process", "sync"] }
+tracing = { version = "=0.1.41" }
+tracing-subscriber = { version = "=0.3.19", features = ["env-filter"] }
```
jdygert-spok, line 44: Not necessary

gknopf-aranya:
There are some merge conflicts to resolve with the AqcPeerStream changes.

crates/aranya-client-capi/src/imp/error.rs
```diff
@@ -22,6 +23,15 @@ pub enum Error {
     #[error("buffer too small")]
     BufferTooSmall,
 
+    #[error(transparent)]
+    Aqc(#[from] AqcError),
```
jdygert-spok, line 27: Maybe just use `Client` variant.

examples/rust/Cargo.toml
gknopf-aranya: The `=` in front of the versions?

crates/aranya-client-capi/src/imp/aqc.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+/// Container for an AQC Channel variant.
+///
+/// This needs to be destructured before it can be used, since C doesn't have
+/// dataful enums.
+pub struct AqcPeerChannel {
+    pub(crate) inner: aqc::AqcPeerChannel,
+}
```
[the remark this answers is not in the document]
gknopf-aranya: That shouldn't be needed on this PR

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1349,71 +1362,561 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// @param[in] data pointer to the data to send.
+/// @param[in] data_len length of the data to send.
+///
+/// @relates AranyaClient.
+pub fn aqc_bidi_stream_send(
+    client: &mut Client,
+    stream: &mut AqcBidiStream,
+    data: &[u8],
+) -> Result<(), imp::Error> {
+    let data = Bytes::from(Vec::from(data));
```
jdygert-spok, line 1659: nit: `Bytes::copy_from_slice(data)`

jdygert-spok (approved):

gknopf-aranya:
Capturing future AQC C API work here:
https://github.com/aranya-project/aranya/issues/274

djohnson-spideroak (dismissed):

jdygert-spok (approved):

jdygert-spok (approved):

gknopf-aranya (dismissed):

djohnson-spideroak (dismissed):

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1349,71 +1361,474 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// @relates AranyaClient.
+pub fn aqc_bidi_stream_try_recv(
+    stream: &mut AqcBidiStream,
+    mut buffer: &mut [MaybeUninit<u8>],
+) -> Result<usize, imp::Error> {
+    let mut written = 0;
+    while {
+        written += consume_bytes(&mut buffer, &mut stream.data);
+        !buffer.is_empty()
+    } {
```
elagergren-spideroak: this is a little too confusing

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1349,71 +1361,474 @@ pub unsafe fn aqc_remove_net_identifier(
...
+/// @relates AranyaClient.
+pub fn aqc_recv_stream_try_recv(
+    stream: &mut AqcReceiveStream,
+    mut buffer: &mut [MaybeUninit<u8>],
+) -> Result<usize, imp::Error> {
+    let mut written = 0;
+    while {
+        written += consume_bytes(&mut buffer, &mut stream.data);
+        !buffer.is_empty()
+    } {
```
elagergren-spideroak: ditto

crates/aranya-client-capi/src/imp/aqc.rs
```diff
@@ -0,0 +1,138 @@
...
+
+/// Writes bytes into `buffer` from `bytes`, advancing them both.
+///
+/// Returns the number of bytes written.
+pub(crate) fn consume_bytes(buffer: &mut &mut [MaybeUninit<u8>], bytes: &mut Bytes) -> usize {
+    let len = core::cmp::min(bytes.len(), buffer.len());
+
+    // SAFETY: &[T] and &[MaybeUninit<T>] have the same layout.
+    let src = unsafe { mem::transmute::<&[u8], &[MaybeUninit<u8>]>(&bytes[..len]) };
```
elagergren-spideroak: we might not have the lint enabled, but this should probably be a pointer cast instead

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -59,6 +61,14 @@ pub enum Error {
     #[capi(msg = "AQC library error")]
     Aqc,
 
+    /// Tried to poll an endpoint but nothing received yet.
+    #[capi(msg = "no response ready yet")]
+    Empty,
```
elagergren-spideroak: i know this matches what AQC currently uses, but i think it might be more familiar as WouldBlock (or similar)

elagergren-spideroak (dismissed):

crates/aranya-client-capi/src/imp/aqc.rs
gknopf-aranya: Should we enable the lint?

gknopf-aranya: @elagergren-spideroak I'm not sure what change to make here. Are you able to push a fix to this PR for it?

ssweetney-spok: There's an example in "crates/aranya-client-capi/src/api/defs.rs".

```
impl AsRef<aranya_crypto::Id> for Id {
    fn as_ref(&self) -> &aranya_crypto::Id {
        // SAFETY: Each type is a struct with a single field containing an array of 64 bytes
        unsafe { &*ptr::from_ref::<[u8; ARANYA_ID_LEN]>(&self.bytes).cast::<aranya_crypto::Id>() }
    }
}
```

gknopf-aranya: Which lint would catch this?

gknopf-aranya: @ssweetney-spok mentioned he received similar feedback on a previous PR and is going to help rewrite this

elagergren-spideroak (approved):

ssweetney-spok (approved):

crates/aranya-client-capi/src/imp/aqc.rs
jdygert-spok: Pointer `cast<U>()` requires `U: Sized`. Oh, but we can use `as`.

I stole it from `slice::write_copy_of_slice` :shrug:

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1820,10 +1818,8 @@ pub fn aqc_recv_stream_try_recv(
     mut buffer: &mut [MaybeUninit<u8>],
 ) -> Result<usize, imp::Error> {
     let mut written = 0;
-    while {
+    while !buffer.is_empty() {
         written += consume_bytes(&mut buffer, &mut stream.data);
```
jdygert-spok: This change is slightly different in that it calls `try_receive` again when the buffer is already full, but that is probably fine.

crates/aranya-client-capi/src/imp/aqc.rs
ssweetney-spok: I wonder if `transmute` was more appropriate here since I think it'll actually retain the correct length when `src` becomes `&[MaybeUninit<u8>`. I'm not sure if casting with `as` does though...

Edit: Actually it may

PR #249, opened 2025-05-05
aqc: update rust example for AQC

gknopf-aranya:
Adds AQC example to the Rust example.

### DIFF
--- a/examples/rust/run.bash
+++ b/examples/rust/run.bash
@@ -15,8 +15,8 @@ current_dir="$(pwd)"
 
 cargo build \
     --release \
-    --manifest-path "examples/rust/Cargo.toml" \
-    --locked
+    --manifest-path "examples/rust/Cargo.toml" #\
+    #--locked
 
 cargo build \
     --release \
--- a/examples/rust/run.bash
+++ b/examples/rust/run.bash
@@ -15,8 +15,8 @@ current_dir="$(pwd)"
 
 cargo build \
     --release \
-    --manifest-path "examples/rust/Cargo.toml" #\
-    #--locked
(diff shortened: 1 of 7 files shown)


examples/rust/src/main.rs
```diff
@@ -371,12 +378,27 @@ async fn main() -> Result<()> {
 
     // TODO: send AQC ctrl via network
     info!("creating acq bidi channel");
-    let _aqc_id1 = membera
+    let created_aqc_chan = membera
```
gknopf-aranya, line 381: @benz-spideroak I'm seeing this error when attempting to create a bidi channel:
```
2025-05-05T20:44:15.447486Z  INFO creating acq bidi channel
Error: AQC error: The connection was closed because the handshake took longer than the max handshake duration of 10s

Caused by:
    0: The connection was closed because the handshake took longer than the max handshake duration of 10s
    1: The connection was closed because the handshake took longer than the max handshake duration of 10s
Error while executing command, exit code: 1
```

examples/rust/run.bash
```diff
@@ -15,8 +15,8 @@ current_dir="$(pwd)"
 
 cargo build \
     --release \
-    --manifest-path "examples/rust/Cargo.toml" \
-    --locked
+    --manifest-path "examples/rust/Cargo.toml" #\
```
gknopf-aranya, line 18: Revert this before merging

examples/rust/src/main.rs
gknopf-aranya: Resolve it. This happened because `local_addr` was returning the client aqc address rather than the server address.

PR #256, opened 2025-05-08
sync: refactor syncer code to prepare for quic syncer

gknopf-aranya:
Reorganizes syncer code into more modules to prepare for adding the quic syncer:
- `task.rs` - responsible for periodically syncing with peers
- `tcp.rs` - the TCP client/server syncer implementation
- `actions.rs` - the graph actions/effects API

Future work:
Implement QUIC syncer in new `quic.rs` module.


crates/aranya-daemon/src/sync/tcp.rs
```diff
@@ -0,0 +1,330 @@
...
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Creates the team.
+    /// Creates a new graph, adds the `CreateTeam` command to the root of the graph.
+    /// Returns the [`GraphId`] of the newly created graph.
+    #[instrument(skip_all)]
```
gknopf-aranya, line 141: I want to keep `sync_peer` in this file but possibly move `create_team` and `actions` to the `actions.rs` module.

gknopf-aranya: The session methods probably belong in `actions.rs` too. This file should only include TCP related network code for the sync protocol.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -45,8 +44,8 @@ pub(crate) type SP = LinearStorageProvider<FileManager>;
 /// EF = Policy Effect
 pub(crate) type EF = policy::Effect;
 
-pub(crate) type Client = aranya::Client<EN, SP, CE>;
-type Server = aranya::Server<EN, SP>;
+pub(crate) type Client = crate::sync::tcp::Client<EN, SP, CE>;
+type Server = crate::sync::tcp::Server<EN, SP>;
```
djohnson-spideroak, line 48: note for later: this will need to be changed for multi-protocol support.

crates/aranya-daemon/src/sync/task.rs
djohnson-spideroak, line 1: do we need to update some comments here with the rename?

gknopf-aranya: Docs at the top of the file still appear relevant:
```
//! Aranya sync task.
//! A task for syncing with Aranya peers at specified intervals.
//! A [`DelayQueue`] is used to retrieve the next peer to sync with at the specified interval.
//! [`SyncPeers`] handles adding/removing peers for the [`Syncer`].
//! [`Syncer`] syncs with the next available peer from the [`DelayQueue`].
//! [`SyncPeers`] and [`Syncer`] communicate via mpsc channels so they can run independently.
//! This prevents the need for an `Arc<<Mutex>>` which would lock until the next peer is retrieved from the [`DelayQueue`]
```

crates/aranya-daemon/src/sync/actions.rs
ssweetney-spok, line 1: Looks good mostly but I don't see why the graph actions/effects API is a part of the sync module. You can perform graph operations without having to do any sync operations and vice-versa.

gknopf-aranya: Yeah, the graph client and network client should be separated into different types.

gknopf-aranya: Pushed a commit that moves action client code out of the `sync` folder and implements different types for each with their own APIs.

crates/aranya-daemon/src/daemon.rs
gknopf-aranya: Yep, I'm moving `tcp.rs` out of the way so we can add a `quic.rs` and just delete `tcp.rs` on the upcoming quic syncer PR.

crates/aranya-daemon/src/sync/tcp.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// Failure.
+    Err(String),
+}
+
+/// Aranya sync client.
+pub struct Client<EN, SP, CE> {
+    /// Thread-safe Aranya client reference.
+    pub(crate) aranya: Arc<Mutex<ClientState<EN, SP>>>,
+    _eng: PhantomData<CE>,
+}
```
ssweetney-spok, line 38: Nit: Rather than having to define this type in 2 places, using newtypes would be more idiomatic

Something like

```Rust
struct Client {
    data: usize
}


mod actions {
    pub struct ActionClient(super::Client);
    
    impl ActionClient {  }
}

mod sync {
    pub struct SyncClient(super::Client);
    
    impl SyncClient {  }
}
```

Another option: we could just keep it as a single type and split it out into 2 impl blocks 🤔 

The graph related methods on `Client` can go in the actions module while the sync related methods can go in the sync module.

crates/aranya-daemon/src/sync/tcp.rs
```diff
@@ -0,0 +1,264 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum SyncResponse {
+    /// Success.
+    Ok(Box<[u8]>),
+    /// Failure.
+    Err(String),
+}
+
+/// Aranya sync client.
+pub struct Client<EN, SP, CE> {
```
djohnson-spideroak, line 34: Do we want to rename these?

crates/aranya-daemon/src/sync/tcp.rs
gknopf-aranya: What if we want to add different fields to these clients in the future?
They are completely different clients with different APIs.
The primary thing they have in common is they share a `ClientState`.
I think it's okay to use different types for these as it gives us more flexibility in the future.
E.g. we may add PSK/rustls related fields to the QUIC client type when implementing the QUIC syncer.

crates/aranya-daemon/src/sync/tcp.rs
gknopf-aranya: We can, but I'm not too worried about maintaining `tcp.rs` if it will be deleted shortly when `quic.rs` is added.
`use sync::tcp::Client` is clear and we could use `use sync::tcp::Client as TcpSyncClient` if it's confusing.

gknopf-aranya: Updated the naming for these types in `daemon.rs`

crates/aranya-daemon/src/sync/tcp.rs
gknopf-aranya: Marking as resolved since I want to keep these types flexible for the upcoming QUIC syncer implementation.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -105,6 +106,9 @@ impl Daemon {
             (client, local_addr)
         };
 
+        // Initialize Aranya actions client.
+        let actions_client = ActionsClient::new(Arc::clone(&client.aranya));
+
```
djohnson-spideroak, line 111: It isnt clear to me why we need this new client?

crates/aranya-daemon/src/sync/tcp.rs
gknopf-aranya: I do think we could group `aranya` and `_eng` fields into a type which becomes a field of the quic client and actions client.

crates/aranya-daemon/src/daemon.rs
gknopf-aranya: This PR is refactoring a module which combined the actions client and tcp syncer client/server into a single module.
They are 2 different clients implemented in a single type.
By separating these clients, the code is more organized and I'm able to modify the tcp->quic syncer code without touching actions related code.

gknopf-aranya: E.g. if someone changes policy/action code, we shouldn't need to resolve merge conflicts in the syncer modules.

crates/aranya-daemon/src/api.rs
```diff
@@ -57,8 +57,8 @@ pub(crate) struct DaemonApiServer {
     sk: ApiKey<CS>,
     /// The UDS path we serve the API on.
     uds_path: PathBuf,
-
-    client: Arc<Client>,
+    /// The Aranya actions client.
+    client: Arc<ActionsClient>,
```
jdygert-spok, line 61:
```suggestion
    client: ActionsClient,
```

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -132,7 +132,7 @@ type EffectSender = mpsc::Sender<(GraphId, Vec<EF>)>;
 /// Receives added/removed peers from [`SyncPeers`] via mpsc channels.
 pub struct Syncer {
     /// Aranya client to allow syncing the Aranya graph with another peer.
-    client: Arc<Client>,
+    client: Arc<TcpSyncClient>,
```
jdygert-spok, line 135:
```suggestion
    client: TcpSyncClient,
```

crates/aranya-daemon/src/sync/tcp.rs
jdygert-spok: We might just want to get rid of the `CE` param.

crates/aranya-daemon/src/sync/task.rs
ssweetney-spok: I changed this as part of my refactoring commit. I wasn't sure if I wanted to do it on the "refactor-syncer-code" branch so that this PR would get updated too but it would've been tricky to update since the other quic syncer branch is much farther along

crates/aranya-daemon/src/sync/tcp.rs
ssweetney-spok: I went ahead and removed it. I only saw the bound used once when specifying an associated type on the Engine trait but it doesn't seem like it needs to be on the `Client` type

crates/aranya-daemon/src/sync/task.rs
gknopf-aranya: @ssweetney-spok I was hoping to merge this refactoring PR in first before the QUIC syncer PR. Which means refactoring changes would be added to this branch rather than the QUIC syncer branch.

gknopf-aranya: If you're able to finalize this PR before we release AQC that would be ideal, then we can release the QUIC syncer in a future release.

crates/aranya-daemon/src/daemon.rs
ssweetney-spok: I handled this by making the Syncer a generic type and separating the impl blocks for the aranya client type into different files

crates/aranya-daemon/test-util/src/lib.rs
```diff
@@ -108,33 +129,40 @@ impl Deref for TestDevice {
...
-    pub owner: &'a TestDevice,
-    pub admin: &'a TestDevice,
-    pub operator: &'a TestDevice,
-    pub membera: &'a TestDevice,
-    pub memberb: &'a TestDevice,
+    pub owner: &'a mut TestDevice,
+    pub admin: &'a mut TestDevice,
+    pub operator: &'a mut TestDevice,
+    pub membera: &'a mut TestDevice,
+    pub memberb: &'a mut TestDevice,
```
djohnson-spideroak, line 147: why do these need to be mutable now?

ssweetney-spok: Yea the mutability propagated up from `Syncer::sync` so I updated the signature of `TestDevice::sync` to take a mutable reference. `TestDevice::sync` is now calling `Syncer::sync` which requires a mutable reference to the `Syncer` type.

So these need to be mutable in order to call that method.

crates/aranya-daemon/test-util/src/lib.rs
```diff
@@ -108,33 +129,40 @@ impl Deref for TestDevice {
...
 impl<'a> TestTeam<'a> {
-    pub fn new(clients: &'a [TestDevice]) -> Self {
+    pub fn new(clients: &'a mut [TestDevice]) -> Self {
         assert!(clients.len() >= 5);
+        // TODO: Replace with https://doc.rust-lang.org/std/primitive.slice.html#method.get_disjoint_mut when we upgrade to 1.86
+        let (owner, rest) = clients.split_at_mut(1);
+        let (admin, rest) = rest.split_at_mut(1);
+        let (operator, rest) = rest.split_at_mut(1);
+        let (membera, rest) = rest.split_at_mut(1);
+        let (memberb, _) = rest.split_at_mut(1);
```
jdygert-spok, line 158:
```suggestion
        let [owner, admin, operator, membera, memberb, ..] = clients else {
            panic!("need at least 5 clients");
        };
```

crates/aranya-daemon/src/daemon.rs
```diff
@@ -99,15 +100,15 @@ impl Daemon {
                 )
                 .await?;
             let local_addr = server.local_addr()?;
-            let client = Arc::new(client);
+            let client = client;
```
jdygert-spok, line 103:
```suggestion
```

gknopf-aranya (commented):
lgtm

nikki-aranya (dismissed):

gknopf-aranya:
lgtm. Needs approval from @djohnson-spideroak, @elagergren-spideroak, or @jdygert-spok
I can't approve since I originally opened the PR.

crates/aranya-daemon/src/sync/tcp.rs
```diff
@@ -0,0 +1,234 @@
...
+        sink: &mut S,
+        peer: &Addr,
+    ) -> impl Future<Output = Result<()>> + Send
+    where
+        S: Sink<<crate::EN as Engine>::Effect> + Send,
+    {
+        async move {
+            // send the sync request.
+
+            // TODO: Real server address.
```
djohnson-spideroak: does this need to happen on this PR?

djohnson-spideroak (dismissed):

crates/aranya-daemon/src/sync/tcp.rs
```diff
@@ -24,83 +25,86 @@ use crate::aranya::Client as AranyaClient;
...
     /// Syncs with the peer.
     /// Aranya client sends a `SyncRequest` to peer then processes the `SyncResponse`.
-    async fn sync_impl<S>(
+    fn sync_impl<S>(
         syncer: &mut super::task::Syncer<Self>,
         id: GraphId,
         sink: &mut S,
         peer: &Addr,
-    ) -> Result<()>
+    ) -> impl Future<Output = Result<()>> + Send
```
jdygert-spok: You can mix and match `async fn` and `fn () -> impl Future` between trait and impl BTW. Using `-> impl Future` for the trait lets you specify additional bounds and then you don't need to specify again in the impl.

crates/aranya-daemon/src/sync/tcp.rs
jdygert-spok: It was already there, only moved.

jdygert-spok (dismissed):

crates/aranya-daemon/src/sync/tcp.rs
jdygert-spok: Reverted this change

crates/aranya-daemon/src/sync/tcp.rs
gknopf-aranya: Yep, this PR should only refactor code not change functionality

nikki-aranya (approved):

crates/aranya-daemon/src/sync/tcp.rs
ssweetney-spok: Thanks. TIL

PR #260, opened 2025-05-12
aqc: pair PSKs with cipher suites

elagergren-spideroak:
This implementation assumes that the client only supports the cipher suites listed in [RFC 8446]. This is a fine assumption for now, but we might need to change it in the future if we allow users to customize the TLS implementation.

To change this, the caller could supply the list of supported cipher suites. Or the daemon could send the `BidiSecret`/`UniSecret` to the client, but that's trickier since `BidiSecret` and `UniSecret` are currently wrappers around the HPKE context, which can't reasonably be `Serialize`/`Deserialize`.

Fixes #259

[RFC 8446]: https://datatracker.ietf.org/doc/html/rfc8446#appendix-B.4

### DIFF
--- a/crates/aranya-daemon-api/src/service.rs
+++ b/crates/aranya-daemon-api/src/service.rs
@@ -23,7 +23,7 @@ use aranya_crypto::{
 use aranya_util::Addr;
 use buggy::Bug;
 pub use semver::Version;
-use serde::{de, Deserialize, Serialize};
+use serde::{Deserialize, Serialize};
 use tracing::error;
 
 /// CE = Crypto Engine
@@ -235,7 +235,7 @@ macro_rules! psk_map {
             pub fn try_from_fn<I, E, F>(id: I, mut f: F) -> anyhow::Result<Self>
             where
                 I: Into<Id>,
-                E: error::Error + fmt::Display + fmt::Debug + Send + Sync + 'static,
+                anyhow::Error: From<E>,
                 F: FnMut(CipherSuiteId) -> Result<$psk, E>,
             {
                 let id = id.into();
@@ -248,23 +248,6 @@ macro_rules! psk_map {
                     psks.insert(CsId(suite), psk);
                 }
                 Ok(Self { id, psks })
-            }
-        }
-
-        impl FromIterator<$psk> for $name {
-            fn from_iter<I>(iter: I) -> Self
-            where
-                I: IntoIterator<Item = $psk>,
-            {
-                let psks = HashMap::from_iter(iter.into_iter().map(|psk| {
-                    let suite = psk.cipher_suite();
-                    (CsId(suite), psk)
-                }));
-                Self {
-                    // TODO
-                    id: Id::default(),
-                    psks,
-                }
             }
         }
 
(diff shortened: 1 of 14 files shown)


crates/aranya-daemon-api/src/service.rs
```diff
@@ -385,19 +409,22 @@ pub trait DaemonApi {
         team: TeamId,
         peer: NetIdentifier,
         label_id: LabelId,
-    ) -> Result<(AqcCtrl, AqcBidiPsk)>;
+    ) -> Result<(AqcCtrl, Box<[AqcBidiPsk]>)>;
```
jdygert-spok, line 412: Could use a separate type to make it more explicit that these are mappings from `CipherSuiteId`

```rust
struct PskMap<T>(...);
impl Index<CipherSuiteId> for PskMap<T> { ... }
```

elagergren-spideroak: i like this

jdygert-spok (approved):

crates/aranya-daemon-api/src/service.rs
```diff
@@ -185,6 +196,180 @@ impl Drop for Secret {
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next().map(|(k, v)| (k.0, v))
+    }
+}
+
+// TODO(eric): Get rid of this once `CipherSuiteId` implements
+// `Hash`, `Serialize`, and `Deserialize`.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+struct CsId(CipherSuiteId);
```
jdygert-spok, line 343:
```suggestion
// TODO(eric): Get rid of this once `CipherSuiteId` implements `Hash`.
#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
#[serde(transparent)]
struct CsId(CipherSuiteId);
```

crates/aranya-daemon-api/src/service.rs
```diff
@@ -185,6 +196,180 @@ impl Drop for Secret {
...
+                    let suite = psk.cipher_suite();
+                    (CsId(suite), psk)
+                }));
+                Self {
+                    // TODO
+                    id: Id::default(),
+                    psks,
+                }
+            }
+        }
```
jdygert-spok, line 269: TODO, seems unused anyway.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -185,6 +196,180 @@ impl Drop for Secret {
...
+            pub fn get(&self, suite: CipherSuiteId) -> Option<&$psk> {
+                self.psks.get(&CsId(suite))
+            }
+
+            /// Creates a PSK map from a function that generates
+            /// a PSK for a cipher suite.
+            pub fn try_from_fn<I, E, F>(id: I, mut f: F) -> anyhow::Result<Self>
+            where
+                I: Into<Id>,
+                E: error::Error + fmt::Display + fmt::Debug + Send + Sync + 'static,
```
jdygert-spok, line 238: nit: `anyhow::Error: From<E>,` since you have anyhow in the API already.

jdygert-spok (approved):

djohnson-spideroak (dismissed):

PR #261, opened 2025-05-13
Use opaque exact size fix

jdygert-spok:
Integrates https://github.com/aranya-project/aranya-core/pull/239


jdygert-spok:
Solved by #260.

PR #265, opened 2025-05-21
chore: update to Rust 1.85

jdygert-spok:


djohnson-spideroak (dismissed):

djohnson-spideroak (approved):

ssweetney-spok:
Not sure if it matters but the [Ferrocene version](https://public-docs.ferrocene.dev/main/release-notes/25.05.0.html) that supports Rust 1.85 hasn't released yet.

Also, could we upgrade to the 2024 edition?

jdygert-spok:
It should release next week AFAIK and we don't actually use it at the moment so I think we're fine. We'll do the edition upgrade in another PR to keep them small and reviewable. (Probably less of an issue for this repo than aranya-core).

PR #266, opened 2025-05-27
aqc: addl tests

djohnson-spideroak:
add additional tests for AQC including:

- label not authorized for creating bidi channel (sender)
- label not authorized for creating bidi channel (receiver) 
- send/recv on closed send stream (behavior may change in base branch to consume stream)
- send/recv on deleted channel


crates/aranya-client/tests/aqc.rs
```diff
@@ -237,3 +237,451 @@ async fn test_aqc_chans() -> Result<()> {
...
+        team.memberb
+            .client
+            .aqc()
+            .delete_bidi_channel(bidi_chan2)
+            .await?;
+
+        // wait for ctrl message to be sent.
+        sleep(Duration::from_millis(100)).await;
+
+        // try sending after channels are closed
```
gknopf-aranya, line 678: Should we test that receiving on a deleted channel also fails?

djohnson-spideroak: added

crates/aranya-client/tests/aqc.rs
```diff
@@ -237,3 +243,620 @@ async fn test_aqc_chans() -> Result<()> {
...
+        team.memberb
+            .client
+            .aqc()
+            .delete_bidi_channel(bidi_chan2)
+            .await?;
+
+        // wait for ctrl message to be sent.
+        sleep(Duration::from_millis(100)).await;
+
+        bidi1_2
```
gknopf-aranya, line 855: Should we verify that send/recv operations fail on both sides of the channel after the bidi channel is deleted?

PR #271, opened 2025-05-28
aqc: additional tests

djohnson-spideroak:
add additional tests for AQC including:

- label not authorized for creating bidi channel (sender)
- label not authorized for creating bidi channel (receiver)
- send/recv on closed send stream 
- send/recv on deleted channel

original PR: https://github.com/aranya-project/aranya/pull/266


crates/aranya-client/tests/aqc.rs
```diff
@@ -237,3 +247,485 @@ async fn test_aqc_chans() -> Result<()> {
...
+
+        // wait for ctrl message to be sent.
+        sleep(Duration::from_millis(100)).await;
+
+        // try sending after channels are closed
+        let msg2 = Bytes::from_static(b"hello2");
+        bidi1_2
+            .send(msg2.clone())
+            .await
+            .expect_err("expected to fail sending");
```
ssweetney-spok, line 721: Can we do an assertion on the type of error. I think `AqcError::ConnectionError` is supposed to be returned in this case?

`AqcError::ConnectionClosed` seems like it only happens when trying to receive data

gknopf-aranya: Yep, we should be checking that the error type matches the expected error.

jdygert-spok: Use `assert!(matches!(err, MyError::Variant(_)))`

gknopf-aranya (approved):

crates/aranya-client/tests/common/mod.rs
```diff
@@ -114,11 +114,11 @@ pub struct TeamCtx {
...
-        let owner = DeviceCtx::new(name, "owner", work_dir.join("owner"), 9001).await?;
-        let admin = DeviceCtx::new(name, "admin", work_dir.join("admin"), 9002).await?;
-        let operator = DeviceCtx::new(name, "operator", work_dir.join("operator"), 9003).await?;
-        let membera = DeviceCtx::new(name, "membera", work_dir.join("membera"), 9004).await?;
-        let memberb = DeviceCtx::new(name, "memberb", work_dir.join("memberb"), 9005).await?;
+        let owner = DeviceCtx::new(name, "owner", work_dir.join("owner"), 0).await?;
+        let admin = DeviceCtx::new(name, "admin", work_dir.join("admin"), 0).await?;
+        let operator = DeviceCtx::new(name, "operator", work_dir.join("operator"), 0).await?;
+        let membera = DeviceCtx::new(name, "membera", work_dir.join("membera"), 0).await?;
+        let memberb = DeviceCtx::new(name, "memberb", work_dir.join("memberb"), 0).await?;
```
nikki-aranya, line 121: Why are the ports all removed?

gknopf-aranya: Looking into this more. The timeout before being allowed to bind to a port again after it was used appears to be for TCP rather than UDP. It's still a good idea to use random available ports for testing in case another test or application on the machine may try to use the same port, which would cause the port bind to fail.

nikki-aranya (approved):

crates/aranya-client/tests/common/mod.rs
jdygert-spok: Need to update `DeviceCtx::new` to grab the resolved address.

PR #272, opened 2025-05-29
Return `PeerStream` from `try_receive_stream` and add `split` method.

jdygert-spok:


gknopf-aranya (approved):
@jdygert-spok, @nikki-aranya do we want to add the `split()` method to the AQC C API? So that users can derive send/recv streams from a bidi stream?

djohnson-spideroak:
> @jdygert-spok, @nikki-aranya do we want to add the `split()` method to the AQC C API? So that users can derive send/recv streams from a bidi stream?

Since this is a new function, we can push adding split to the API until next release.

elagergren-spideroak (approved):

ssweetney-spok (approved):

PR #275, opened 2025-05-30
cleanup: rm unused DeviceCtx fields

gknopf-aranya:


gknopf-aranya:
@jdygert-spok, @elagergren-spideroak we should be able to merge this in tomorrow now that the AQC release is complete.

ssweetney-spok (approved):

PR #279, opened 2025-06-03
Fix `docsrs` cfg and enable `doc_cfg` feature

jdygert-spok:
Not necessary to enable in all the crates but might as well.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -208,7 +208,7 @@ install_crate = { crate_name = "cargo-machete", version = "0.8.0", binary = "car
 category = "docs"
 description = "Generate docs with rust nightly"
 toolchain = "${NIGHTLY}"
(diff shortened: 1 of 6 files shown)


gknopf-aranya:
> Not necessary to enable in all the crates but might as well.

True, it ensures we won't forget to enable before a release.

gknopf-aranya:
@jdygert-spok do we need something similar for aranya-core crates?

jdygert-spok:
> > Not necessary to enable in all the crates but might as well.
> 
> True, it ensures we won't forget to enable before a release.

Now that it is set properly in the docs generation run by CI, it would be caught if we didn't

> @jdygert-spok do we need something similar for aranya-core crates?

https://github.com/aranya-project/aranya-core/pull/260

gknopf-aranya (approved):

djohnson-spideroak (approved):

PR #285, opened 2025-06-03
aqc: update rust example readme

gknopf-aranya:
Rust example README is out-of-date after the AQC release.
This PR gets the README up-to-date and makes it less likely to become out of date in the future by removing implementation specific references/links from the README.


recluse-so (approved):
Looks great

jdygert-spok (approved):

PR #286, opened 2025-06-05
Refactor daemon

### LINKED ISSUE #35: daemon: handle `JoinSet` of sync server
We only ever spawn to it so it just grows indefinitely.

jdygert-spok:
- Move more initialization up into constructors so serving is infallible
- Better waiting for tasks. Closes #35.
- Daemon tracing all be within same span, with name
- Remove unix socket before binding to support reinitialization
- Reduce visibility `pub -> pub(crate) -> private` and remove a few unused things


crates/aranya-daemon/src/main.rs
```diff
@@ -42,12 +42,14 @@ fn main() -> Result<()> {
         info!("loaded Aranya daemon");
 
         if flags.print_api_pk {
-            let pk = daemon.public_api_key().await?.encode()?;
+            let pk = daemon.public_api_key().encode()?;
```
jdygert-spok: The daemon now goes through the entire startup before printing the key and exiting, but this will be removed by #295.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -49,134 +49,182 @@ pub(crate) type EF = policy::Effect;
...
+/// Dropping this will abort the daemon's tasks.
+#[clippy::has_significant_drop]
+pub struct DaemonHandle {
+    set: JoinSet<()>,
+}
+
+impl DaemonHandle {
+    /// Wait for the daemon to finish.
+    ///
+    /// Panics if the daemon's tasks panic.
```
gknopf-aranya: Does this panic if any of the daemon tasks panic?
```suggestion
    /// Panics if any of the daemon tasks panic.
```

gknopf-aranya (approved):

crates/aranya-util/src/task.rs
```diff
@@ -0,0 +1,192 @@
+use core::{any::Any, future::Future, panic::AssertUnwindSafe};
+use std::panic::resume_unwind;
+
+use futures_util::FutureExt as _;
+use tokio::sync::mpsc;
+use tokio_util::task::TaskTracker;
+
+// TODO(jdygert): Abort all tasks on drop?
```
gknopf-aranya, line 8: Is the alternative waiting for all the tasks to complete first which could take an indefinite amount of time?

jdygert-spok: The tasks will either continue to run to completion or be dropped when the runtime is shut down. Nothing will wait on them. (There's currently not even a `wait_for_all_tasks` on `tokio::Runtime`.)

gknopf-aranya (dismissed):

crates/aranya-daemon/src/daemon.rs
```diff
@@ -48,117 +49,159 @@ pub(crate) type EF = policy::Effect;
...
+            })
+        }
+        .instrument(info_span!(parent: span_id, "load"))
+        .await
+    }
 
-        Ok(())
+    /// The daemon's entrypoint.
+    pub fn spawn(mut self) -> DaemonHandle {
+        let _guard = self.span.enter();
```
ssweetney-spok, line 174: Can events actually be registered in this span if this guard is dropped when the function returns?

crates/aranya-util/src/task.rs
```diff
@@ -0,0 +1,192 @@
...
+
+/// Creates a scope for spawning scoped async tasks for structured concurrency.
+///
+/// The given function will be provided a [`Scope`] through which tasks can be
+/// [`spawned`][Scope::spawn]. The resulting future will pend until all tasks have finished.
+/// The tracked tasks can free their memory as soon as they finish, making this well suited for
+/// spawning many short lived tasks over time.
+///
+/// Unlike [`std::thread::scope`], this function does not let you spawn tasks which borrow from the
+/// local scope, since there is no safe way to do so.
```
ssweetney-spok, line 18: Nit: the naming confused me slightly since borrowing from the parent scope is probably the most useful thing about scoped threads. This (the scope function) seems like a variation `tokio::join` that's better in some scenarios

crates/aranya-daemon/src/daemon.rs
jdygert-spok: This makes it be the parent span for each of the `info_span!`s created below. When those tasks are polled the child spans will be entered.

But TBH I'm not sure if theres a difference in entering a child span when the parent has or hasn't been entered. It does at least use the daemon span in the logs, but I'm not sure whether e.g. it would count as time "inside" the daemon span.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -48,117 +49,159 @@ pub(crate) type EF = policy::Effect;
...
+#[clippy::has_significant_drop]
+pub struct DaemonHandle {
+    set: JoinSet<()>,
+}
+
+impl DaemonHandle {
+    /// Wait for the daemon to finish.
+    ///
+    /// Panics if any of the daemon's tasks panic.
+    pub async fn join(mut self) -> Result<(), Bug> {
```
jdygert-spok, line 63: Maybe this should be instrumented with the daemon span too.

crates/aranya-util/src/task.rs
jdygert-spok: Fair, but it is still a very similar pattern since you get an object to spawn tasks on and "exiting" the scope will "block" until all tasks are joined.

crates/aranya-daemon/src/daemon.rs
ssweetney-spok: Ah, I see. You used the ID of the span in `Daemon::load`. I guess that's doing something similar

crates/aranya-daemon/src/daemon.rs
jdygert-spok: Eh, I could see either way.

gknopf-aranya:
@jdygert-spok, @djohnson-spideroak are we hoping to get this PR into the MVP release? I like these change but am hesitant to resolve too many upstream merge conflicts when MVP features are still under development.

jdygert-spok:
At the very least, the removal of the socket before binding is needed.

ssweetney-spok (approved):

PR #288, opened 2025-06-05
Refactor aqc net

jdygert-spok:


crates/aranya-client/src/aqc/api.rs
```diff
@@ -1,161 +1,17 @@
...
-            .create_uni_channel(peer_addr, label_id, psks)
-            .await
-    }
-
-    /// Receives a channel.
-    pub async fn receive_channel(&mut self) -> crate::Result<AqcPeerChannel> {
-        self.client.receive_channel().await
-    }
-
-    /// Attempts to receive a channel.
```
gknopf-aranya, line 153: Were some of these methods unused? I don't see them redefined somewhere else in the PR.

jdygert-spok: They all just forwarded to the inner client so I replaced the use of this with that inner client.

jdygert-spok: (The methods were not actually public but only used by the api type below)

PR #289, opened 2025-06-06
bugfix: aqc ctrl message stream close flake

gknopf-aranya:
Resolves an AQC flake where the QUIC connection that sends the AQC ctrl message was closed before `stream.close()` was called, which resulted in an error being returned.

- Use `read_to_end()` to ensure all data is received before closing the QUIC connection that sends the AQC ctrl message
- Log warning if `stream.close()` fails rather than returning an error


crates/aranya-client/src/aqc/net.rs
```diff
@@ -259,11 +259,11 @@ impl AqcClient {
             .quic_client
             .connect(Connect::new(addr).with_server_name(addr.ip().to_string()))
             .await?;
-        conn.keep_alive(true)?;
```
gknopf-aranya, line 307: Using `keep_alive` for sending the ctrl message doesn't seem necessary. Makes more sense to keep a connection alive that will have more streams opened on it later. When sending the ctrl message, the connection is only used for a single ctrl msg and ack.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -294,7 +294,7 @@ impl AqcClient {
                     .send(Bytes::from(ack_bytes))
                     .await
                     .map_err(AqcError::from)?;
-                stream.close().await.map_err(AqcError::from)?;
```
gknopf-aranya, line 342: My theory is that if the peer already closed the stream/connection by dropping the objects after calling `send_ctrl()`, the `stream.close()` may fail here because it is an error to attempt to close a stream that is already closed or that does not have an open connection. The connection used for sending the ctrl message is short-lived, while the connection used for creating AQC QUIC streams later is longer lived. Instead of returning an error, let's make closing the stream a best-effort attempt.

.github/workflows/tests.yml
gknopf-aranya, line 1: Revert changes in this file before merging.
Need to run test in ci a bunch of times to check that the flake is resolved.
The flake occurred about 50% of the time so no failures in 10 runs is good.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -339,7 +340,7 @@ impl AqcClient {
...
+                let _ = stream.close().await;
```
djohnson-spideroak: Lets maybe warn on this, I dont like eating the error completely.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -304,18 +304,19 @@ impl AqcClient {
...
         stream.send(Bytes::from_owner(msg_bytes)).await?;
+        let _ = stream.close().await;
         if let Some(msg_bytes) = stream.receive().await? {
             let msg = postcard::from_bytes::<AqcAckMessage>(&msg_bytes).map_err(AqcError::Serde)?;
             match msg {
                 AqcAckMessage::Success => (),
                 AqcAckMessage::Failure(e) => return Err(AqcError::CtrlFailure(e)),
             }
         }
+        while stream.receive().await?.is_some() {}
```
djohnson-spideroak: What is the purpose of this?

gknopf-aranya: Making sure the stream is done receiving before dropping the QUIC connection at the end of this function.
I'm planning to use `read_to_end()` instead.

djohnson-spideroak (approved):

crates/aranya-client/src/error.rs
```diff
@@ -131,6 +131,10 @@ pub enum AqcError {
     #[error("stream error: {0}")]
     StreamError(#[from] s2n_quic::stream::Error),
 
+    /// Async I/O read error.
+    #[error("failed async read: {0}")]
+    AsyncReadError(#[from] io::Error),
```
ssweetney-spok, line 136: Nit: `std::io::Error` isn't just limited to async or read errors so I'd remove `#[from]` here and have users construct this variant explicitly. It doesn't change much but at least it doesn't implicitly create it like:

```Rust
fn foo(w: impl Write, buf: &mut[u8]) -> Result<(), AqcError> {
    w.write(buf)?; // May return Err(AqcError::AsyncReadError)
    
    Ok(())
}
```

`w.write(buf).map_err(AqcError::AsyncReadError)?` wouldn't be right either but users can see why it's wrong

ssweetney-spok (approved):
👍

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: @jdygert-spok do you want to include this `keep_alive` change?
It's not related to the bugfix directly but is something I noticed that doesn't seem to be needed for such a short-lived connection.

jdygert-spok: I have the change on mine. Seems good.

gknopf-aranya:
Closing in favor of:
https://github.com/aranya-project/aranya/pull/291

PR #291, opened 2025-06-06
Fix ctrl recv error causing flake

jdygert-spok:
Read to the end of the stream and ignore close error when acking.

Kind of silly to me what make this fail AFAICT.
- Peer is reading all data until stream is finished
- We close the stream, which marks it as finished and "flushes" to wait for ack of all data
- The peer finishes reading all the data and closes the stream
- Our flushing fails because the stream is closed, even though the peer couldn't have closed it until after we called close
- This really feels like bad behavior to me.

### DIFF
--- a/crates/aranya-client/src/aqc/net.rs
+++ b/crates/aranya-client/src/aqc/net.rs
@@ -351,6 +351,11 @@ impl AqcClient {
                     AqcAckMessage::Failure(format!("Failed to deserialize AqcCtrlMessage: {e}"));
                 let ack_bytes = postcard::to_stdvec(&ack_msg).assume("can serialize")?;
                 stream.send(Bytes::from(ack_bytes)).await.ok();
+                if let Err(err) = stream.close().await {
+                    if !is_close_error(err) {
+                        error!(%err, "error closing stream after ctrl failure");
+                    }
+                }
                 return Err(AqcError::Serde(e).into());
             }
         }
@@ -442,6 +447,11 @@ enum AqcAckMessage {
     Failure(String),
 }
 
+/// Read all of a stream until it has finished.
+///
+/// A bit more efficient than going through the `AsyncRead`-based impl,
+/// especially if there was only one chunk of data. Also avoids needing to
+/// convert/handle an `io::Error`.
 async fn read_to_end(stream: &mut BidirectionalStream) -> Result<Bytes, s2n_quic::stream::Error> {
     let Some(first) = stream.receive().await? else {
         return Ok(Bytes::new());
@@ -461,6 +471,7 @@ async fn read_to_end(stream: &mut BidirectionalStream) -> Result<Bytes, s2n_quic
     Ok(buf.freeze())
 }
 
+/// Indicates whether the stream error is "connection closed without error".
 fn is_close_error(err: s2n_quic::stream::Error) -> bool {
     matches!(
         err,
--- a/crates/aranya-client/src/aqc/net.rs
+++ b/crates/aranya-client/src/aqc/net.rs
@@ -14,7 +14,7 @@ use aranya_daemon_api::{
     AqcBidiPsks, AqcCtrl, AqcPsks, AqcUniPsks, DaemonApiClient, LabelId, TeamId,
 };
 use buggy::{Bug, BugExt as _};
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use channels::AqcPeerChannel;
 use s2n_quic::{
     self,
@@ -26,6 +26,7 @@ use s2n_quic::{
             rustls::{server::PresharedKeySelection, ClientConfig, ServerConfig},
         },
     },
+    stream::BidirectionalStream,
     Client, Connection, Server,
 };
 use tarpc::context;
@@ -304,18 +305,20 @@ impl AqcClient {
             .quic_client
             .connect(Connect::new(addr).with_server_name(addr.ip().to_string()))
             .await?;
-        conn.keep_alive(true)?;
         let mut stream = conn.open_bidirectional_stream().await?;
+
         let msg = AqcCtrlMessage { team_id, ctrl };
         let msg_bytes = postcard::to_stdvec(&msg).assume("can serialize")?;
-        stream.send(Bytes::from_owner(msg_bytes)).await?;
-        if let Some(msg_bytes) = stream.receive().await? {
-            let msg = postcard::from_bytes::<AqcAckMessage>(&msg_bytes).map_err(AqcError::Serde)?;
-            match msg {
-                AqcAckMessage::Success => (),
-                AqcAckMessage::Failure(e) => return Err(AqcError::CtrlFailure(e)),
-            }
+        stream.send(Bytes::from(msg_bytes)).await?;
+        stream.finish()?;
+
+        let ack_bytes = read_to_end(&mut stream).await?;
+        let ack = postcard::from_bytes::<AqcAckMessage>(&ack_bytes).map_err(AqcError::Serde)?;
+        match ack {
+            AqcAckMessage::Success => (),
+            AqcAckMessage::Failure(e) => return Err(AqcError::CtrlFailure(e)),
         }
+
         Ok(())
     }
 
@@ -325,10 +328,7 @@ impl AqcClient {
             .await
             .map_err(AqcError::ConnectionError)?
             .ok_or(AqcError::ConnectionClosed)?;
-        let Ok(Some(ctrl_bytes)) = stream.receive().await else {
-            error!("Failed to receive control message or stream closed");
-            return Err(AqcError::ConnectionClosed.into());
-        };
+        let ctrl_bytes = read_to_end(&mut stream).await.map_err(AqcError::from)?;
         match postcard::from_bytes::<AqcCtrlMessage>(&ctrl_bytes) {
             Ok(ctrl) => {
                 self.process_ctrl_message(ctrl.team_id, ctrl.ctrl).await?;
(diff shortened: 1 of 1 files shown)


crates/aranya-client/src/aqc/net.rs
```diff
@@ -304,18 +305,20 @@ impl AqcClient {
...
         let msg_bytes = postcard::to_stdvec(&msg).assume("can serialize")?;
-        stream.send(Bytes::from_owner(msg_bytes)).await?;
-        if let Some(msg_bytes) = stream.receive().await? {
-            let msg = postcard::from_bytes::<AqcAckMessage>(&msg_bytes).map_err(AqcError::Serde)?;
-            match msg {
-                AqcAckMessage::Success => (),
-                AqcAckMessage::Failure(e) => return Err(AqcError::CtrlFailure(e)),
-            }
+        stream.send(Bytes::from(msg_bytes)).await?;
+        stream.finish()?;
```
jdygert-spok, line 313: Only need to mark as finished on the sender side since we don't need to wait for an ACK to start trying to read.

elagergren-spideroak:
where's the flake?

crates/aranya-client/src/aqc/net.rs
```diff
@@ -339,15 +339,18 @@ impl AqcClient {
                     .send(Bytes::from(ack_bytes))
                     .await
                     .map_err(AqcError::from)?;
-                stream.close().await.map_err(AqcError::from)?;
+                if let Err(err) = stream.close().await {
+                    if !is_close_error(err) {
+                        return Err(AqcError::from(err).into());
+                    }
+                }
```
jdygert-spok, line 346: Kind of silly to me what make this fail AFAICT.

- Peer is reading all data until stream is finished
- We close the stream, which marks it as finished and "flushes" to wait for ack of all data
- The peer finishes reading all the data and closes the stream
- Our flushing fails because the stream is closed, even though the peer couldn't have closed it until after we called close

This really feels like bad behavior to me.

gknopf-aranya:
We should test this fix in CI about about 10 times. Passing CI once isn't enough times to prove the flake was fixed.

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: I thought the problem was that `send_ctrl` was dropping the `conn` object before the peer had a chance to send the ack back and call `stream.close()`. Since the connection itself no longer exists, there's no way to call `stream.close()` without an error.

jdygert-spok: No, now that `send_ctrl` reads until the end, it won't drop the connection until after we mark it finished in the close here.

gknopf-aranya:
I opened a similar PR that also fixed the flake:
https://github.com/aranya-project/aranya/pull/289
We just need to decide which approach to take.

jdygert-spok:
#292 for repeated testing.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -438,3 +441,32 @@ enum AqcAckMessage {
     /// The failure message.
     Failure(String),
 }
+
+async fn read_to_end(stream: &mut BidirectionalStream) -> Result<Bytes, s2n_quic::stream::Error> {
```
gknopf-aranya, line 455: What's the benefit of implementing this custom `read_to_end()` method compared to the one already provided by `sn2-quic`?

crates/aranya-client/src/aqc/net.rs
```diff
@@ -339,15 +339,18 @@ impl AqcClient {
                     .send(Bytes::from(ack_bytes))
                     .await
                     .map_err(AqcError::from)?;
-                stream.close().await.map_err(AqcError::from)?;
+                if let Err(err) = stream.close().await {
+                    if !is_close_error(err) {
```
gknopf-aranya, line 343: I think we'll still want to log a warning if there's a close error even if it's not returned. That way we can tell when unexpected s2n-quic library behavior is happening.

crates/aranya-client/src/aqc/net.rs
jdygert-spok: That uses the stream's `tokio::io::AsyncRead` impl which isn't really a perfect API mapping and does extra buffering/copying.

This also avoids needing to do something with an `io::Error` since you instead only deal with the quic stream error.

crates/aranya-client/src/aqc/net.rs
jdygert-spok: This means "closed by peer with no error", I don't think there's much point to log it since it doesn't really mean anything and happens so often.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -339,15 +339,18 @@ impl AqcClient {
...
+                }
             }
             Err(e) => {
                 error!("Failed to deserialize AqcCtrlMessage: {}", e);
                 let ack_msg =
                     AqcAckMessage::Failure(format!("Failed to deserialize AqcCtrlMessage: {e}"));
                 let ack_bytes = postcard::to_stdvec(&ack_msg).assume("can serialize")?;
-                let _ = stream.send(Bytes::from(ack_bytes)).await;
-                let _ = stream.close().await;
+                stream.send(Bytes::from(ack_bytes)).await.ok();
```
gknopf-aranya, line 353: Why was `stream.close()` removed here? Shouldn't we call it and handle the error the same way here as above?

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: This may be something to document in the code since I wouldn't have known this just by reading the function implementation.

crates/aranya-client/src/aqc/net.rs
jdygert-spok: The stream still gets closed on drop. We're already returning the deserialize error, but I guess we could log the close error. It just didn't seem very noteworthy to me for "if the peer sends us junk and then doesn't ack our reply that it is junk, then log an error"

jdygert-spok: Logged error

crates/aranya-client/src/aqc/net.rs
jdygert-spok: added doc

gknopf-aranya (approved):

djohnson-spideroak (approved):

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: I was about to resolve this but realized the original comment is important information for reviewers that probably belongs in the PR description.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -304,18 +305,20 @@ impl AqcClient {
...
-        if let Some(msg_bytes) = stream.receive().await? {
-            let msg = postcard::from_bytes::<AqcAckMessage>(&msg_bytes).map_err(AqcError::Serde)?;
-            match msg {
-                AqcAckMessage::Success => (),
-                AqcAckMessage::Failure(e) => return Err(AqcError::CtrlFailure(e)),
-            }
+        stream.send(Bytes::from(msg_bytes)).await?;
+        stream.finish()?;
+
+        let ack_bytes = read_to_end(&mut stream).await?;
```
ssweetney-spok, line 315: Is it flushing that causes an issue? I'm trying to understand why you wouldn't want to call `flush` before trying to read

ssweetney-spok: Or does it not matter since you're reading all the data from the stream with `read_to_end` now?

gknopf-aranya: My understanding is that the sender doesn't need to call `flush` because it will block on `read_to_end` until it receives a response. The recipient is guaranteed to get the ctrl msg and ack whether or not `flush` is called.

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: Updated PR description with this info.

ssweetney-spok (approved):

crates/aranya-client/src/aqc/net.rs
```diff
@@ -438,3 +446,38 @@ enum AqcAckMessage {
...
+        if let Some(even_more) = stream.receive().await? {
+            more = even_more;
+        } else {
+            break;
+        }
+    }
+    Ok(buf.freeze())
+}
+
+/// Indicates whether the stream error is "connection closed without error".
```
elagergren-spideroak, line 474: lol

gknopf-aranya: Are you requesting a change here?

crates/aranya-client/src/aqc/net.rs
elagergren-spideroak, line 313: were we just relying on its drop glue before..?

elagergren-spideroak:
I'm not quite sure I understand the bug. I mean, I understand what @jdygert-spok wrote, but either we were not using the API correctly or something else funky is happening. Or it's 3 PM on a Friday.

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: I believe so. The stream and connection objects were both dropped at the end of this function.
It's probably not very common to open a QUIC connection that's immediately dropped after only sending/receiving a single message on a stream.

gknopf-aranya:
Looks like we hit another flake we didn't know we had in one of the query tests:
https://github.com/aranya-project/aranya/actions/runs/15499996117/job/43645572515?pr=292

gknopf-aranya:
I looked to see if there were any open issues filed against `s2n-quic` for this.
The closest one I could find is this one:
https://github.com/aws/s2n-quic/issues/2514

Since there's no way to tell from a stream if a connection was closed, that could cause issues when performing operations on the stream after the connection is closed. This could possibly impact error handling for `close()`.

crates/aranya-client/src/aqc/net.rs
jdygert-spok: Previously we were just reading one data chunk in each direction, so not really relying on closing the streams at all but also potentially not reading all of the message.

PR #295, opened 2025-06-07
daemon: write API public key to file system

### LINKED ISSUE #294: daemon: write public key to same directory as IPC socket
The `aranya-daemon` crate contains the Aranya Daemon. The `aranya-client` crate contains the Aranya Client. The Client and Daemon communicate over an IPC API that uses Unix Domain Sockets. Data sent over that IPC is encrypted with the Daemon's public key (see the `Daemon::public_api_key` method).

Currently, the Client needs to be manually configured with the Daemon's public key (see the `Client::with_daemon_api_pk` method).

The Daemon currently creates a Unix Domain Socket at `Config::uds_api_path`. Update the Daemon to write its public key to the same directory that contains the Unix Domain Socket. For example, if `Config::uds_api_path` is `/var/run/daemon/ipc.sock`, the Daemon should write its public key to `/var/run/daemon/api.pk`. The Daemon must write its public key *before* it creates the Unix Domain Socket.

Then, instead of having to manually configure the Client, the Client will load the Daemon's public key from that path before it connects to the Daemon (see `Client::connect`).

elagergren-spideroak:
The client can load it from there.

Fixes #294


crates/aranya-client/src/client.rs
```diff
@@ -84,21 +83,36 @@ impl ClientBuilder<'_> {
...
-                "must specify the daemon's public key",
-            ))
-            .into());
+
+        // Load the public key from the UDS directory
+        let uds_dir = sock
+            .parent()
+            .context("UDS path must have a parent directory")
+            .map_err(IpcError::new)?;
+        let api_pk_path = uds_dir.join("api.pk");
```
djohnson-spideroak: Given we are just looking for a static filename, this might collide if two daemons have UDS sockets in the same directory.

jdygert-spok: Probably shouldn't have multiple in one directory anyway, but yeah I'd rather do `{uds_api_path}.pk`

jdygert-spok:
LGTM other than changing the path. Need to update example lockfile and C header.

crates/aranya-client/src/client.rs
gknopf-aranya: Have we documented anywhere that the user is responsible for configuring the correct permissions for the UDS socket path and public key path?

crates/aranya-daemon/src/daemon.rs
```diff
@@ -87,6 +87,9 @@ impl Daemon {
         let mut local_store = self.load_local_keystore().await?;
         let api_sk = self.load_or_gen_api_sk(&mut eng, &mut local_store).await?;
 
+        // Write the daemon's public API key to the UDS directory before creating the socket
+        self.write_api_key_to_uds_dir(&api_sk).await?;
```
gknopf-aranya: Let's call this method something else to indicate it's writing the public key and not a private key:
```suggestion
        self.write_api_pk_to_uds_dir(&api_sk).await?;
```

crates/aranya-client/src/client.rs
```diff
@@ -84,21 +83,36 @@ impl ClientBuilder<'_> {
...
+
+        let pk_bytes = match try_read_cbor::<PublicApiKey<CS>>(&api_pk_path).await {
```
gknopf-aranya: The code for loading the public key could be moved into a method such as `load_daemon_uds_pk()`

crates/aranya-client/src/client.rs
elagergren-spideroak: I do not believe so.

elagergren-spideroak: I wonder if we should change `Config::uds_api_path` to be a directory, not a file.

gknopf-aranya: Yeah, and probably call it `Config::uds_api_dir` so it's clear it's a directory and not a file path

gknopf-aranya: I wonder what the best place would be to add those docs.
They could be added in the client Rust/C API docs, or possibly the getting started guide.

elagergren-spideroak: https://github.com/aranya-project/aranya/pull/298

/cc: @jdygert-spok @djohnson-spideroak

crates/aranya-daemon/src/daemon.rs
```diff
@@ -325,7 +326,10 @@ impl Daemon {
 
 impl Drop for Daemon {
     fn drop(&mut self) {
-        let _ = std::fs::remove_file(self.cfg.uds_api_sock());
+        use std::fs;
+
+        let _ = fs::remove_file(self.cfg.api_pk_path());
+        let _ = fs::remove_file(self.cfg.uds_api_sock());
```
jdygert-spok, line 332: FYI this is never run since we don't implement graceful shutdown

elagergren-spideroak: ¯\_(ツ)_/¯

jdygert-spok: Changing it to delete before here: https://github.com/aranya-project/aranya/pull/286/files#diff-139a1c39ccd408ac5c57662dc7e7da41592f4454b91a275553e834d461ebeea4R229

crates/aranya-client/src/client.rs
gknopf-aranya: Issue to document permissions:
https://github.com/aranya-project/aranya-project.github.io/issues/56

PR #303, opened 2025-06-09
daemon: generate fresh API key at startup

### LINKED ISSUE #301: daemon: generate a new API key at startup
Now that #294 is done, the daemon does not need to store its API key on disk. It can just create a new API key at startup.

elagergren-spideroak:
Fixes #301

### DIFF
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -5,7 +5,7 @@ use aranya_crypto::{
     default::DefaultEngine,
     import::Import,
     keys::SecretKey,
-    keystore::{fs_keystore::Store, KeyStore, KeyStoreExt},
+    keystore::{fs_keystore::Store, KeyStore},
     Engine, Rng,
 };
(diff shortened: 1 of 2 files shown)


crates/aranya-daemon/src/daemon.rs
```diff
@@ -84,8 +73,11 @@ impl Daemon {
             .load_or_gen_public_keys(&mut eng, &mut aranya_store)
             .await?;
 
-        let mut local_store = self.load_local_keystore().await?;
-        let api_sk = self.load_or_gen_api_sk(&mut eng, &mut local_store).await?;
+        // Currently unused after #294.
+        let mut _local_store = self.load_local_keystore().await?;
```
gknopf-aranya, line 77: Will the unused key store be resolved on this PR or a future PR?

elagergren-spideroak: It doesn't need to be "resolved." It'll be used in the near future.

djohnson-spideroak (approved):

PR #305, opened 2025-06-10
quic syncer: save peer cache

gknopf-aranya:
Save QUIC syncer peer cache in memory.

### DIFF
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -35,7 +35,7 @@ use crate::{
     policy,
     sync::task::{
         quic::{PskStore, State as QuicSyncState},
-        Syncer,
+        PeerCacheMap, Syncer,
     },
     util::{load_team_psk_pairs, SeedDir},
     vm_policy::{PolicyEngine, TEST_POLICY_1},
@@ -57,6 +57,14 @@ pub(crate) type EF = policy::Effect;
 
 pub(crate) type Client = aranya::Client<EN, SP>;
 pub(crate) type SyncServer = crate::sync::task::quic::Server<EN, SP>;
+
+/// Sync configuration for setting up Aranya.
+struct SyncParams {
+    psk_store: Arc<PskStore>,
+    active_team_rx: Receiver<TeamId>,
+    caches: PeerCacheMap,
+    external_sync_addr: Addr,
+}
 
 mod invalid_graphs {
     use std::{
@@ -169,6 +177,9 @@ impl Daemon {
             let (psk_store, active_team_rx) = PskStore::new(initial_keys);
             let psk_store = Arc::new(psk_store);
 
+            // Create a shared PeerCacheMap
+            let caches: PeerCacheMap = Arc::new(Mutex::new(BTreeMap::new()));
+
             // Initialize Aranya client.
             let (client, sync_server) = Self::setup_aranya(
                 &cfg,
@@ -177,9 +188,12 @@ impl Daemon {
                     .try_clone()
                     .context("unable to clone keystore")?,
                 &pks,
-                qs_config.addr,
-                Arc::clone(&psk_store),
-                active_team_rx,
+                SyncParams {
+                    psk_store: Arc::clone(&psk_store),
+                    active_team_rx,
+                    caches,
+                    external_sync_addr: qs_config.addr,
+                },
             )
             .await?;
             let local_addr = sync_server.local_addr()?;
@@ -329,9 +343,7 @@ impl Daemon {
         eng: CE,
         store: AranyaStore<KS>,
         pk: &PublicKeys<CS>,
-        external_sync_addr: Addr,
-        psk_store: Arc<PskStore>,
-        active_team_rx: Receiver<TeamId>,
+        sync_params: SyncParams,
     ) -> Result<(Client, SyncServer)> {
         let device_id = pk.ident_pk.id()?;
 
@@ -344,12 +356,13 @@ impl Daemon {
 
         let client = Client::new(Arc::clone(&aranya));
 
-        info!(addr = %external_sync_addr, "starting QUIC sync server");
+        info!(addr = %sync_params.external_sync_addr, "starting QUIC sync server");
         let server = SyncServer::new(
             client.clone(),
-            &external_sync_addr,
-            psk_store,
-            active_team_rx,
+            &sync_params.external_sync_addr,
+            sync_params.psk_store,
+            sync_params.active_team_rx,
+            sync_params.caches,
         )
         .await
         .context("unable to initialize QUIC sync server")?;
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -191,7 +191,7 @@ impl Daemon {
                 SyncParams {
                     psk_store: Arc::clone(&psk_store),
                     active_team_rx,
-                    caches,
+                    caches: caches.clone(),
                     external_sync_addr: qs_config.addr,
                 },
             )
@@ -203,8 +203,14 @@ impl Daemon {
 
             let invalid_graphs = InvalidGraphs::default();
             let state = QuicSyncState::new(psk_store.clone())?;
-            let (syncer, peers) =
-                Syncer::new(client.clone(), send_effects, invalid_graphs.clone(), state);
+            let (syncer, peers) = Syncer::new(
+                client.clone(),
+                send_effects,
+                invalid_graphs.clone(),
+                state,
+                local_addr.into(),
+                caches,
+            );
 
             let graph_ids = client
                 .aranya
--- a/crates/aranya-daemon/src/sync/task.rs
+++ b/crates/aranya-daemon/src/sync/task.rs
@@ -132,7 +132,7 @@ type EffectSender = mpsc::Sender<(GraphId, Vec<EF>)>;
 
 /// Key for looking up syncer peer cache in map.
 #[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Debug)]
-pub(crate) struct PeerCacheKey {
+pub struct PeerCacheKey {
     /// The peer address.
     pub addr: Addr,
     /// The Aranya graph ID.
@@ -324,7 +324,7 @@ impl<ST: SyncState> Syncer<ST> {
 
     /// Get peer caches.
     #[cfg(test)]
-    pub fn get_peer_caches(&self) -> PeerCacheMap {
+    pub(crate) fn get_peer_caches(&self) -> PeerCacheMap {
         self.caches.clone()
     }
 
--- a/crates/aranya-daemon/src/sync/task.rs
+++ b/crates/aranya-daemon/src/sync/task.rs
@@ -24,11 +24,7 @@ use tokio_util::time::{delay_queue::Key, DelayQueue};
 use tracing::{instrument, trace, warn};
 
 use super::Result as SyncResult;
-use crate::{
-    daemon::{Client, EF},
-    vm_policy::VecSink,
-    InvalidGraphs,
-};
+use crate::{daemon::EF, vm_policy::VecSink, InvalidGraphs};
 
 pub mod quic;
 
@@ -50,9 +46,8 @@ enum Msg {
 type Request = (Msg, oneshot::Sender<Reply>);
 type Reply = SyncResult<()>;
 
-/// Sync Peer
-///
-/// Uniquely identifies a sync peer, consisting of
+/// A sync peer.
+/// Contains the information needed to sync with a single peer:
 /// - network address
 /// - Aranya graph id
 #[derive(Debug, Clone, Ord, Eq, PartialOrd, PartialEq, Hash)]
@@ -195,7 +190,7 @@ pub trait SyncState: Sized {
 
 impl<ST> Syncer<ST> {
     /// Creates a new [`Syncer`] and [`SyncPeers`] pair.
-    pub fn new(
+    pub(crate) fn new(
         client: crate::aranya::Client<crate::EN, crate::SP>,
         send_effects: EffectSender,
         invalid: InvalidGraphs,
@@ -320,11 +315,13 @@ impl<ST: SyncState> Syncer<ST> {
         self.caches.clone()
     }
 
+    /// Returns a reference to the Aranya client.
     #[cfg(test)]
     pub fn client(&self) -> &crate::aranya::Client<crate::EN, crate::SP> {
         &self.client
     }
 
+    /// Returns a mutable reference to the Aranya client.
     #[cfg(test)]
     pub fn client_mut(&mut self) -> &mut crate::aranya::Client<crate::EN, crate::SP> {
         &mut self.client
--- a/crates/aranya-daemon/src/sync/task.rs
+++ b/crates/aranya-daemon/src/sync/task.rs
@@ -148,7 +148,7 @@ impl PeerCacheKey {
 /// Thread-safe map of peer caches
 /// For a given peer, there's should only be one cache. If separate caches are used
 /// for the server and state it will reduce the efficiency of the syncer.
-pub type PeerCacheMap = Arc<Mutex<BTreeMap<PeerCacheKey, PeerCache>>>;
+pub(crate) type PeerCacheMap = Arc<Mutex<BTreeMap<PeerCacheKey, PeerCache>>>;
 
 /// Syncs with each peer after the specified interval.
 /// Uses a [`DelayQueue`] to obtain the next peer to sync with.
--- a/crates/aranya-daemon/src/sync/task.rs
+++ b/crates/aranya-daemon/src/sync/task.rs
@@ -6,16 +6,20 @@
 //! [`SyncPeers`] and [`Syncer`] communicate via mpsc channels so they can run independently.
 //! This prevents the need for an `Arc<<Mutex>>` which would lock until the next peer is retrieved from the [`DelayQueue`]
 
-use std::{collections::HashMap, future::Future};
+use std::{
+    collections::{BTreeMap, HashMap},
+    future::Future,
+    sync::Arc,
+};
 
 use anyhow::Context;
 use aranya_daemon_api::SyncPeerConfig;
-use aranya_runtime::{storage::GraphId, Engine, Sink};
+use aranya_runtime::{storage::GraphId, Engine, PeerCache, Sink};
 use aranya_util::Addr;
 use buggy::BugExt;
 use futures_util::StreamExt;
 use serde::{Deserialize, Serialize};
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::{mpsc, oneshot, Mutex};
 use tokio_util::time::{delay_queue::Key, DelayQueue};
 use tracing::{instrument, trace, warn};
 
@@ -46,14 +50,23 @@ enum Msg {
 type Request = (Msg, oneshot::Sender<Reply>);
 type Reply = SyncResult<()>;
 
-/// A sync peer.
-/// Contains the information needed to sync with a single peer:
+/// Sync Peer
+///
+/// Uniquely identifies a sync peer, consisting of
 /// - network address
 /// - Aranya graph id
 #[derive(Debug, Clone, Ord, Eq, PartialOrd, PartialEq, Hash)]
-struct SyncPeer {
+pub(crate) struct SyncPeer {
     addr: Addr,
     graph_id: GraphId,
+}
+
+impl SyncPeer {
+    /// Creates a new `SyncPeer`.
+    #[cfg(test)]
+    pub fn new(addr: Addr, graph_id: GraphId) -> Self {
+        Self { addr, graph_id }
+    }
 }
 
 /// Handles adding and removing sync peers.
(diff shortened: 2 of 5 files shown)


crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -104,6 +104,9 @@ pub struct State {
     /// Address -> Connection map used for re-using connections
     /// when making outgoing sync requests
     conns: BTreeMap<Addr, Connection>,
+    /// Thread-safe reference to an [`Addr`]->[`PeerCache`] map.
+    /// Lock must be acquired after [`Self::client`]
+    caches: Arc<Mutex<BTreeMap<Addr, PeerCache>>>,
```
benz-spideroak: I believe this needs to be for `Addr`, and `GraphId`. One cache isn't valid across multiple graphs.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -555,6 +569,8 @@ where
     #[instrument(skip_all)]
     async fn sync_respond(
         client: AranyaClient<EN, SP>,
+        caches: Arc<Mutex<BTreeMap<Addr, PeerCache>>>,
+        addr: Addr,
```
benz-spideroak, line 600: This isn't the `Addr` you want use for the cache. You want to use the one in SyncType::Poll. The cache should use the server address of the peer. This Addr is the peer's client address.

benz-spideroak: This will require sending the server's address when making requests. Currently a dummy address is sent.

crates/aranya-daemon/src/sync/task/quic.rs
gknopf-aranya: On the server-side, how do we know what `GraphId` a sync request/response is for?

gknopf-aranya: Found it. It's part of the `SyncRequest` type.

crates/aranya-daemon/src/sync/task/quic.rs
gknopf-aranya: Seems like we could also get the address directly from the QUIC connection/stream

gknopf-aranya: How do I send the server address when making a sync request? `syncer.poll()` currently does not take the server address as input.
```
syncer
    .poll(&mut send_buf, client.provider(), cache)
    .context("sync poll failed")?
```

gknopf-aranya: I see now, the `SyncRequester` can be initialized with the server address.

gknopf-aranya: Should the peer cache of the client also use the server address of the peer as the key?

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -373,6 +391,9 @@ pub struct Server<EN, SP> {
     set: JoinSet<()>,
     /// Identity Receiver.
     _identity_rx: mpsc::Receiver<Vec<u8>>,
+    /// Thread-safe reference to an [`Addr`]->[`PeerCache`] map.
+    /// Lock must be acquired after [`Self::aranya`]
+    caches: Arc<Mutex<BTreeMap<PeerCacheKey, PeerCache>>>,
```
benz-spideroak: There should only be one cache per peer. Currently you have two caches. One here and one in State.

benz-spideroak:
It's easy to mess this up and use the wrong addr which invalidates the cache. It would be good to have a test to make sure the cache is working. I did this is core by testing the number of commands received. That allowed me to see that the cache was working. https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-runtime/src/testing/dsl.rs#L587

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -97,13 +97,29 @@ pub enum Error {
...
 /// QUIC syncer state used for sending sync requests and processing sync responses
 pub struct State {
     /// QUIC client to make sync requests and handle sync responses.
     client: QuicClient,
     /// Address -> Connection map used for re-using connections
     /// when making outgoing sync requests
     conns: BTreeMap<Addr, Connection>,
+    /// Thread-safe reference to an [`Addr`]->[`PeerCache`] map.
+    /// Lock must be acquired after [`Self::client`]
+    caches: Arc<Mutex<BTreeMap<PeerCacheKey, PeerCache>>>,
```
ssweetney-spok: This should probably be moved into the `Syncer` type directly since the peer caches would be useful for other types of syncers. This `State` type is usually reserved for any state that is specific to one kind of a syncer.

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: To confirm, are you saying that they should be **sharing** the same the cache map rather than creating their own map in the `new` methods?

benz-spideroak: Yes. You get information to update the cache when sending and receiving sync requests. Both should share the same cache. It doesn't make sense to have more than one cache per peer and graph.

benz-spideroak: Either have both places share the same Map or probably better is only to have the Map in one place.

crates/aranya-daemon/test-util/src/lib.rs
ssweetney-spok, line 1: Just want to point out that this file got removed as part of https://github.com/aranya-project/aranya/pull/307. The new test is useful so I think it should stay

crates/aranya-daemon/src/test.rs
```diff
@@ -0,0 +1,406 @@
...
+            effect_recv,
+        })
+    }
+}
+
+impl TestDevice {
+    /// Syncs with a device and expects a certain number of commands to be received.
+    ///
+    /// Returns the effects that were received.
+    pub async fn sync_expect(
```
gknopf-aranya, line 112: This appears to be validating that the client's peer cache is updated properly.
How are we validating that the server's peer cache has been updated?

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -254,21 +276,23 @@ impl Syncer<State> {
...
     where
         A: Serialize + DeserializeOwned + Clone,
     {
-        info!("client sending sync request to QUIC sync server");
-        let mut send_buf = vec![0u8; MAX_SYNC_MESSAGE_SIZE];
+        info!(?peer, "sending sync request to");
+        // Must lock aranya then caches to prevent deadlock.
+        let mut aranya = self.client.aranya.lock().await;
+        let key = PeerCacheKey::new(*peer, id);
+        let mut caches = self.caches.lock().await;
```
gknopf-aranya: May want to group these fields into a new type so a single lock can be used on aranya and the cache.
This would avoid the need to lock in a particular order.

crates/aranya-daemon/src/test.rs
ssweetney-spok: They share the same peer cache instance. See ~~"crates/aranya-daemon/src/daemon.rs"~~

Edit: `TestCtx::new_client` in this file is where the cache is created

It'd be kind of awkward but The `TestDevice`s could keep a reference to the sync server too if we want to check both

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: I thought a doc comment might be sufficient since there's only 2 locks to worry about. `aranya_daemon::aranya::Client` uses a mutex internally so code would look something like

```Rust
let (ref client, ref mut caches) = client_caches.lock().await;
let mut aranya = client.aranya.lock().await;
```

There's still 2 locks but at least the ordering problem is solved

crates/aranya-daemon/src/test.rs
gknopf-aranya: Yeah, I understand that. The question is more in reference to how we can prove the server is using the cache and improving performance because of it.
Is there an empirical way to test/verify this other than code inspection?

ssweetney-spok: If we want to prove that the server is using the cache then off the top of my head:

1. Have a reference to the server. (Could be more challenging than I originally though since the server loops to respond to incoming requests indefinitely. Not sure if there ever would be a time the check in the test would be able to acquire the lock to inspect the server)

2. Use channels to notify when the cache has been updated. (Seems easier to implement but it doesn't seem that useful for production code so we'd take a performance hit just for the tests. Could put that code behind a feature flag but we probably don't want to do that.)

gknopf-aranya: Does the server benefit at all from having a peer cache if we haven't implemented bidirectional sync yet?

ssweetney-spok: The server calls `aranya_runtime::sync::SyncResponder::poll` to update the peer cache when receving a sync request.

I'd say so. If I'm understanding correctly, bidirectional syncing is just making a sync request back after receiving a sync request. Syncing in both directions can still occur as a result of syncing periodically over intervals.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -85,6 +89,26 @@ pub enum Error {
...
+impl PeerCacheKey {
+    fn new(addr: Addr, id: GraphId) -> Self {
+        Self { addr, id }
+    }
+}
+
+/// Thread-safe map of peer caches
+/// For a given peer, there's should only be one cache. If separate caches are used
+/// for the server and state it will reduce the efficiency of the syncer.
+pub type PeerCacheMap = Arc<Mutex<BTreeMap<PeerCacheKey, PeerCache>>>;
```
ssweetney-spok: Moving this up to the module above this one since it's applicable to all types of syncers

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok, @ssweetney-spok I think we can merge this PR in soon. There are a few discussions on it that I'd be interested in getting Ben's input on.
afaict, this is using the peer cache to improve sync performance. The added tests verify that commands do not need to be synced again due to the peer cache.

crates/aranya-client/src/client.rs
```diff
@@ -362,6 +362,26 @@ impl Team<'_> {
             .map_err(aranya_error)
     }
 
+    /// Add a team to local device storage.
```
gknopf-aranya: Merge error. This should be reverted.

ssweetney-spok (approved):

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -133,18 +165,26 @@ pub struct Syncer<ST> {
...
 /// object.
 pub trait SyncState: Sized {
     /// Syncs with the peer.
+    ///
+    /// Returns the number of commands that were received and successfully processed.
     fn sync_impl<S>(
         syncer: &mut Syncer<Self>,
         id: GraphId,
         sink: &mut S,
+        server_addr: Addr,
```
jdygert-spok: This is available through `syncer: &mut Syncer<Self>`

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -203,12 +247,7 @@ impl<ST: SyncState> Syncer<ST> {
                         self.sync(&peer).await?;
                     },
                     Msg::AddPeer { peer, cfg } => {
-                        let mut result = Ok(());
-                        if cfg.sync_now {
-                            result = self.sync(&peer).await;
-                        }
                         self.add_peer(peer, cfg);
-                        result?;
```
jdygert-spok: bad merge

crates/aranya-daemon/src/daemon.rs
```diff
@@ -321,8 +339,7 @@ impl Daemon {
         store: AranyaStore<KS>,
         pk: &PublicKeys<CS>,
         external_sync_addr: Addr,
-        psk_store: Arc<PskStore>,
-        active_team_rx: Receiver<TeamId>,
+        sync_params: SyncParams,
```
jdygert-spok: Move `external_sync_addr` into `sync_params`?

crates/aranya-daemon/src/test.rs
```diff
@@ -0,0 +1,403 @@
...
+            .actions()
+            .add_member(DeviceKeyBundle::try_from(&memberb.pk)?)
+            .await
+            .context("unable to add memberb member")?;
+        memberb.sync_check_cache(admin, Some(3)).await?;
+
+        owner.sync_check_cache(operator, Some(2)).await?;
+        admin.sync_check_cache(operator, Some(5)).await?;
+        membera.sync_check_cache(operator, Some(5)).await?;
+        memberb.sync_check_cache(operator, Some(5)).await?;
```
jdygert-spok: This all passes without any peer cache. I guess the only thing this might test is that we didn't screw up the cache making us not send commands that they don't have.

gknopf-aranya: I wonder if we should check if there are any commands stored in the peer cache.

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -203,12 +246,10 @@ impl<ST: SyncState> Syncer<ST> {
                         self.sync(&peer).await?;
                     },
                     Msg::AddPeer { peer, cfg } => {
-                        let mut result = Ok(());
+                        self.add_peer(peer.clone(), cfg.clone());
                         if cfg.sync_now {
-                            result = self.sync(&peer).await;
+                            self.sync(&peer).await?;
                         }
```
jdygert-spok: revert

crates/aranya-daemon/src/test.rs
gknopf-aranya: I tried checking the size of the caches, but they were empty:
https://github.com/aranya-project/aranya/pull/363

benz-spideroak:
I ended up using Geoff's cache size test. A more comprehensive set of tests would be ideal, but I'm not sure it's worth doing. It gets pretty complicated.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -136,6 +144,8 @@ impl Daemon {
         let span_id = span.id();
 
         async move {
+            // Create a shared PeerCacheMap
+            let caches = Arc::new(Mutex::new(BTreeMap::new()));
```
jdygert-spok: nit: move this down, probably below psk_store.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -246,21 +247,23 @@ impl Syncer<State> {
...
         A: Serialize + DeserializeOwned + Clone,
     {
-        info!("client sending sync request to QUIC sync server");
-        let mut send_buf = vec![0u8; MAX_SYNC_MESSAGE_SIZE];
+        info!(?peer, "sending sync request to QUIC sync server");
+        // Must lock aranya then caches to prevent deadlock.
+        let mut aranya = self.client.aranya.lock().await;
+        let key = PeerCacheKey::new(*peer, id);
+        let mut caches = self.caches.lock().await;
+        let cache = caches.entry(key).or_default();
```
jdygert-spok: These locks are now held for the whole method instead of just the time they are needed.

jdygert-spok (dismissed):

gknopf-aranya:
@benz-spideroak, @jdygert-spok could I get a summary of what the fix was for commands to show up in the peer cache?

crates/aranya-daemon/src/test.rs
```diff
@@ -0,0 +1,413 @@
...
+
+        let operator_caches = operator.syncer.get_peer_caches();
+        let operator_cache_size = operator_caches
+            .lock()
+            .await
+            .get(&owner_key)
+            .unwrap()
+            .heads()
+            .len();
+        assert!(operator_cache_size > 0);
```
gknopf-aranya, line 385: Should we check that cache size is greater than zero for each device?

gknopf-aranya (commented):
lgtm, I can't approve because I opened the PR originally.

gknopf-aranya:
Now that this PR has been merged, we'll want to split the peer caches based on team ID:
https://github.com/aranya-project/aranya/pull/416

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -118,6 +131,26 @@ impl SyncPeers {
...
+impl PeerCacheKey {
+    fn new(addr: Addr, id: GraphId) -> Self {
+        Self { addr, id }
+    }
+}
+
+/// Thread-safe map of peer caches
+/// For a given peer, there's should only be one cache. If separate caches are used
+/// for the server and state it will reduce the efficiency of the syncer.
+pub type PeerCacheMap = Arc<Mutex<BTreeMap<PeerCacheKey, PeerCache>>>;
```
gknopf-aranya:
```suggestion
pub(crate) type PeerCacheMap = Arc<Mutex<BTreeMap<PeerCacheKey, PeerCache>>>;
```

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -118,6 +131,26 @@ impl SyncPeers {
 
 type EffectSender = mpsc::Sender<(GraphId, Vec<EF>)>;
 
+/// Key for looking up syncer peer cache in map.
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Debug)]
+pub struct PeerCacheKey {
```
gknopf-aranya, line 135:
```suggestion
pub(crate) struct PeerCacheKey {
```

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -516,60 +530,62 @@ where
...
-                    warn!(team = %active_team, "missing requested graph, we likely have not synced yet");
-                    Ok(0)
-                } else {
-                    Err(err)
-                }
-            })
-            .context("sync resp poll failed")?;
+        let len = {
+            // Must lock aranya then caches to prevent deadlock.
+            let mut aranya = client.aranya.lock().await;
```
gknopf-aranya, line 623: Moving the lock code to a method could help to guarantee these are always locked in the correct order.

crates/aranya-daemon/src/test.rs
```diff
@@ -0,0 +1,413 @@
...
+    sync::{
+        self,
+        task::{quic::PskStore, PeerCacheKey, PeerCacheMap, SyncPeer},
+    },
+    vm_policy::{PolicyEngine, TEST_POLICY_1},
+    AranyaStore, InvalidGraphs,
+};
+
+// Aranya graph client for testing.
+pub type TestClient =
```
gknopf-aranya: These test types shouldn't need to be `pub` unless they are used from a different module/crate.

crates/aranya-daemon/src/test.rs
```diff
@@ -0,0 +1,413 @@
...
+    LinearStorageProvider<FileManager>,
+>;
+
+pub struct TestDevice {
+    /// Aranya sync client.
+    pub syncer: TestSyncer,
+    /// The Aranya graph ID.
+    pub graph_id: GraphId,
+    /// The address that the server is listening on.
+    pub local_addr: Addr,
```
gknopf-aranya:
```suggestion
    /// The address that the sync server is listening on.
    pub sync_local_addr: Addr,
```

crates/aranya-daemon/src/test.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        let operator_caches = operator.syncer.get_peer_caches();
+        let operator_cache_size = operator_caches
+            .lock()
+            .await
+            .get(&owner_key)
+            .unwrap()
+            .heads()
+            .len();
+        assert!(operator_cache_size > 0);
```
gknopf-aranya, line 385: Could hide the complexity in a test method like this:
```suggestion
        let operator_cache_size = operator.syncer.get_cache_size(&owner_key);
        assert!(operator_cache_size > 0);
```

crates/aranya-daemon/src/test.rs
```diff
@@ -0,0 +1,415 @@
...
+        server: TestServer,
+        local_addr: Addr,
+        pk: PublicKeys<DefaultCipherSuite>,
+        graph_id: GraphId,
+        psk_store: Arc<PskStore>,
+        caches: PeerCacheMap,
+    ) -> Result<Self> {
+        let waiter = ready::Waiter::new(1);
+        let notifier = waiter.notifier();
+        let handle = task::spawn(async { server.serve(notifier).await }).abort_handle();
```
gknopf-aranya, line 95: Should wait for the notification that the task has started:
```suggestion
        let handle = task::spawn(async { server.serve(notifier).await }).abort_handle();
        waiter.wait().await?;
```

gknopf-aranya (commented):
lgtm pending a few comments

crates/aranya-daemon/src/test.rs
ssweetney-spok: ➕

crates/aranya-daemon/src/test.rs
```diff
@@ -0,0 +1,417 @@
...
+        membera.sync_expect(admin, Some(3)).await?;
+        operator
+            .actions()
+            .add_member(DeviceKeyBundle::try_from(&memberb.pk)?)
+            .await
+            .context("unable to add memberb member")?;
+        memberb.sync_expect(admin, Some(3)).await?;
+
+        owner.sync_expect(operator, Some(2)).await?;
+        admin.sync_expect(operator, Some(5)).await?;
```
ssweetney-spok, line 401: I thought this would've been 4 here instead of a 5. I'm guessing the braid algorithm created a merge command during this sync?

benz-spideroak: Yes.

jdygert-spok (approved):

PR #307, opened 2025-06-12
tests: rm obsolete aranya-daemon tests

gknopf-aranya:
The tests in `aranya-daemon` predate the integration tests in `aranya-client` and are now obsolete.
They currently don't offer any additional test coverage that isn't already handled by the Aranya client integration tests that interact with the daemon via the UDS API.
In additional to that, the tests require extra work to maintain with no value added.
Let's remove them unless we have a valid test scenario that can't be covered by other tests.

This will simplify some test updates that were added on the QUIC syncer PR because there will be less test code to update after this code is removed.


crates/aranya-daemon/tests/tests.rs
```diff
@@ -1,98 +0,0 @@
...
-    let mut ctx = TestCtx::new()?;
-
-    ctx.new_team().await.context("unable to create team")?;
-    Ok(())
-}
-
-/// Tests removing members.
-#[test(tokio::test(flavor = "multi_thread"))]
-#[serial]
-async fn test_remove_members() -> Result<()> {
```
djohnson-spideroak, line 41: just to confirm, we have another test for remove_members right?

ssweetney-spok: Oh, good catch. It seems like this is associated with the `remove_device_from_team` client API and there isn't a test for that in another crate AFAICT

gknopf-aranya: I'll add a test for that in `aranya-client`

crates/aranya-client/tests/tests.rs
```diff
@@ -78,6 +78,45 @@ async fn test_sync_now() -> Result<()> {
...
+    owner.remove_device_from_team(team.operator.id).await?;
+    owner.revoke_role(team.admin.id, Role::Admin).await?;
+    owner.remove_device_from_team(team.admin.id).await?;
+    owner.revoke_role(team.owner.id, Role::Owner).await?;
+    owner
+        .remove_device_from_team(team.owner.id)
+        .await
+        .expect_err("owner should not be able to remove itself from team");
+
+    Ok(())
```
djohnson-spideroak: need to verify the devices on team after removal and sync

gknopf-aranya: Used the `devices_on_team` query to confirm that the number of devices on the team decreases each time a device is removed.

djohnson-spideroak (approved):

jdygert-spok (approved):

PR #310, opened 2025-06-12
capi-example: use the abs path like the daemon does

### LINKED ISSUE #309: capi: c example silently fails in multiple ways

jdygert-spok:
Closes #309.

### DIFF
--- a/crates/aranya-daemon/src/api.rs
+++ b/crates/aranya-daemon/src/api.rs
@@ -3,12 +3,12 @@
 
 #![allow(clippy::expect_used, clippy::panic, clippy::indexing_slicing)]
 
-use core::{future, net::SocketAddr, ops::Deref};
-use std::{path::PathBuf, pin::pin, sync::Arc};
+use core::{future, net::SocketAddr, ops::Deref, pin::pin};
+use std::{path::PathBuf, sync::Arc};
 
 use anyhow::{anyhow, Context as _, Result};
 use aranya_crypto::{Csprng, Rng};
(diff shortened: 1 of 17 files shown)


jdygert-spok:
I don't think so, I printed out the key in both client and daemon.

elagergren-spideroak:
> I don't think so, I printed out the key in both client and daemon.

ah, the `info` parameters are differnt

elagergren-spideroak:
> > I don't think so, I printed out the key in both client and daemon.
> 
> ah, the `info` parameters are differnt

the client is using a relative path, the daemon is using an abs path

crates/aranya-daemon/src/api.rs
```diff
@@ -141,7 +141,7 @@ impl DaemonApiServer {
         let server = {
             let listener = UnixListener::bind(&self.uds_path)?;
             info!(
-                addr = ?listener
+                path = ?listener
```
elagergren-spideroak: I assume `local_addr` will return the exact same path you gave to `bind`, but it's probably a bit cleaner to just log `uds_path`.

jdygert-spok: Yeah see #286

elagergren-spideroak: especially b/c that's what's used to create `info`

jdygert-spok:
I guess it still has a leak in CI that isn't detected locally

gknopf-aranya (dismissed):

gknopf-aranya (approved):

ssweetney-spok (approved):

PR #314, opened 2025-06-12
issue/308: suggestions

gknopf-aranya:
Some suggested changes for issue/308

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -8,7 +8,7 @@ use std::{ffi::OsStr, os::unix::ffi::OsStrExt, str::FromStr};
 use anyhow::Context as _;
 use aranya_capi_core::{opaque::Opaque, prelude::*, ErrorCode, InvalidArg};
 use aranya_client::aqc::{self, AqcPeerStream};
-use aranya_crypto::hex;
+use aranya_crypto::dangerous::spideroak_crypto::hex;
 use bytes::Bytes;
 use tracing::error;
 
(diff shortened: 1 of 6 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -496,7 +496,9 @@ pub fn init_logging() -> Result<(), imp::Error> {
 /// @param dst the output buffer
 /// @param src the input hexadecimal string
 pub fn decode_hex(dst: &mut [u8], src: &[u8]) -> Result<usize, imp::Error> {
```
jdygert-spok: this function is now unneeded after the api pk changes

crates/aranya-daemon/src/daemon.rs
```diff
@@ -319,6 +318,7 @@ async fn load_or_gen_key<K: SecretKey>(path: impl AsRef<Path>) -> Result<K> {
             }
             Err(err) if err.kind() == io::ErrorKind::NotFound => {
                 tracing::info!("generating key");
+                // TODO: not compiling
                 let key = K::new(&mut Rng);
```
jdygert-spok: K::random I think

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I'm not planning to merge this PR. It is a test to see if everything compiles after this fix: https://github.com/aranya-project/aranya-core/pull/284

gknopf-aranya: @jdygert-spok can I just delete the whole thing, or do we need to replace it with something else in the API?

elagergren-spideroak:
already have these

Issue #317, opened 2025-06-13
quic sync: update C api and example


gknopf-aranya:
After implementing the QUIC syncer here:
https://github.com/aranya-project/aranya/pull/258

There are some changes needed to the C API and C example.
Primarily with `create_team()`, `add_team()`, and `remove_team()`.

Create a new PR targeting the original QUIC syncer PR to make these updates.

cc: @ssweetney-spok, @djohnson-spideroak, @jdygert-spok, @elagergren-spideroak

djohnson-spideroak:
completed by #318 and #258

PR #302, opened 2025-06-09
sync: implement remove_team on quic syncer PR

gknopf-aranya:
Implement `remove_team()` on the QUIC syncer PR.
Calls `remove_graph()` in the runtime to remove the graph from storage.

PR #258, opened 2025-05-08
WIP: sync: quic syncer

gknopf-aranya:
Currently just an outline of the QUIC syncer implementation.

PR #318, opened 2025-06-13
quic sync: update c api and example

gknopf-aranya:
Updates the following methods in the C API based on the QUIC syncer implementation:
- `create_team()` - Generated an team PSK for the QUIC syncer
- `add_team()` - Passes in an encrypted PSK seed for the QUIC syncer
- `remove_team()` - Tests the API method in the C example now that's it's implemented by https://github.com/aranya-project/aranya/pull/302

Methods added to the C API on this PR:
- `get_psk_seed_encrypted()` - Gets a PSK seed encrypted for a specific device

The team config has been updated to include the QUIC syncer PSK seed (will be changed to an encrypted PSK seed soon once that's unblocked).

When creating a team with `create_team()`, a PSK seed is generated to be used by all QUIC syncer client/servers when syncing for that team.
The PSK seed should be encrypted with a device's public encryption key before it is sent to the device to be used when adding a team with `add_team()`.

Uses the updated/new methods in the C example app.

This PR is currently blocked on the daemon encrypting/decrypting PSK seeds and adding methods to get/set the PSK seed to the daemon IPC:
https://github.com/aranya-project/aranya/pull/258

Solves:
https://github.com/aranya-project/aranya/issues/317
https://github.com/aranya-project/aranya/issues/167


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -916,13 +955,23 @@ pub fn revoke_label(
...
+) -> Result<(), imp::Error> {
     let client = client.imp();
     let cfg: &imp::TeamConfig = cfg.deref();
     // FIXME(Steve): Return id and psk with out params
-    let CreateTeamResponse { team_id, .. } =
+    let CreateTeamResponse { team_id: id, .. } =
         client.rt.block_on(client.inner.create_team(cfg.into()))?;
-    Ok(team_id.into())
+
+    // TODO: use seed generated by create_team()
```
gknopf-aranya: Looking for a way to convert from the `Box<u8>` returned by `create_team()` to an array of bytes `[u8; 64]`

jdygert-spok: `mybox.as_ref().try_into()`

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -260,6 +261,13 @@ impl From<aranya_crypto::Id> for Id {
     }
 }
 
+/// The size in bytes of an ID
+pub const ARANYA_SEED_LEN: usize = 64;
```
gknopf-aranya: unused

crates/aranya-client/src/config.rs
```diff
@@ -85,7 +85,7 @@ impl Default for SyncPeerConfigBuilder {
 
 #[derive(Clone)]
 pub struct QuicSyncConfig {
-    seed_mode: SeedMode,
+    pub seed_mode: SeedMode,
```
jdygert-spok: revert

crates/aranya-client/src/config.rs
```diff
@@ -96,12 +96,17 @@ impl QuicSyncConfig {
 
 #[derive(Clone, Default)]
 pub struct QuicSyncConfigBuilder {
-    seed_mode: SeedMode,
+    pub seed_mode: SeedMode,
```
jdygert-spok: revert

crates/aranya-client/src/config.rs
```diff
@@ -96,12 +96,17 @@ impl QuicSyncConfig {
...
 
 impl QuicSyncConfigBuilder {
-    /// Sets the seed mode to 'Generate'.
-    /// This is the default mode.
+    /// Sets the PSK seed mode.
+    pub fn mode(mut self, mode: SeedMode) -> Self {
+        self.seed_mode = mode;
+        self
+    }
+
```
jdygert-spok: revert

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -260,6 +260,17 @@ impl From<aranya_crypto::Id> for Id {
     }
 }
 
+/// The size in bytes of a PSK seed IKM.
+// TODO: compute this?
+pub const ARANYA_WRAPPED_SEED_LEN: usize = 304;
```
jdygert-spok: Hmm, not a fan of this. Maybe we should do the "write out length needed if it is too short" thing? IDK

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -260,6 +260,17 @@ impl From<aranya_crypto::Id> for Id {
...
+
+/// The size in bytes of a PSK seed IKM.
+pub const ARANYA_SEED_LEN: usize = 32;
```
jdygert-spok: nit: `ARANYA_SEED_IKM_LEN`, it is not the length of the seed.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -649,13 +660,64 @@ pub fn client_config_builder_set_aqc_config(cfg: &mut ClientConfigBuilder, aqc_c
...
+}
+
+/// Attempts to set raw seed IKM value on [`QuicSyncConfigBuilder`].
+///
+/// This function consumes and releases any resources associated
+/// with the memory pointed to by `cfg`.
+///
+/// @param cfg a pointer to the quic sync config builder
+/// @param seed a pointer the raw PSK seed IKM
+pub fn quic_sync_config_raw_seed(
```
jdygert-spok: seed_ikm not raw_seed

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -649,13 +660,64 @@ pub fn client_config_builder_set_aqc_config(cfg: &mut ClientConfigBuilder, aqc_c
...
 
-#[aranya_capi_core::opaque(size = 64, align = 8)]
+#[aranya_capi_core::opaque(size = 288, align = 8)]
 pub type QuicSyncConfig = Safe<imp::QuicSyncConfig>;
 
+#[aranya_capi_core::opaque(size = 280, align = 8)]
+pub type EncapSeed = Safe<imp::EncapSeed>;
+
+#[aranya_capi_core::opaque(size = 312, align = 8)]
+pub type WrappedSeedForPeer = Safe<imp::WrappedSeedForPeer>;
```
jdygert-spok: Unused

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -918,9 +992,65 @@ pub fn create_team(client: &mut Client, cfg: &TeamConfig) -> Result<TeamId, imp:
...
+    team_id: &mut MaybeUninit<TeamId>,
+    seed: &mut MaybeUninit<EncapSeed>,
+) -> Result<(), imp::Error> {
+    let peer_seed = imp::peer_seed_deserialize(psk_bytes)?;
+
+    team_id.write(peer_seed.get_team_id().into());
+    EncapSeed::init(seed, peer_seed.get_encap_seed());
+
+    Ok(())
+}
```
jdygert-spok: Do we want a separate function and `EncapSeed` type? We could just pass in the serialized bytes to the cfg.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -242,37 +351,55 @@ impl Typed for QuicSyncConfig {
...
+            .mode(value.mode)
             .build()
             .expect("All fields are set")
     }
 }
 
 #[derive(Default)]
 pub struct QuicSyncConfigBuilder {
-    seed: Option<Box<[u8]>>,
+    mode: Option<SeedMode>,
```
jdygert-spok: `Option` not needed, generate is the default.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -672,13 +734,25 @@ pub fn quic_sync_config_build(
...
 
+/// Configures QUIC syncer for [`TeamConfigBuilder`].
+///
+/// By default, the QUIC syncer config is not set. It is an error to call
+/// [`team_config_build`] before setting the interval with
+/// this function
+///
+/// @param cfg a pointer to the builder for a team config
+/// @param quic Set the QUIC syncer config
+pub fn team_config_builder_set_quic_syncer(cfg: &mut TeamConfigBuilder, quic: &QuicSyncConfig) {
```
jdygert-spok: Maybe take `OwnedPtr<QuicSyncConfig>` instead of cloning.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: We do this elsewhere. Start with an initial guess for the buffer size, then reallocate if it isn't large enough.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -649,13 +660,64 @@ pub fn client_config_builder_set_aqc_config(cfg: &mut ClientConfigBuilder, aqc_c
...
+/// Attempts to set raw seed IKM value on [`QuicSyncConfigBuilder`].
+///
+/// This function consumes and releases any resources associated
+/// with the memory pointed to by `cfg`.
+///
+/// @param cfg a pointer to the quic sync config builder
+/// @param seed a pointer the raw PSK seed IKM
+pub fn quic_sync_config_raw_seed(
+    cfg: &mut QuicSyncConfigBuilder,
+    seed: &[u8],
```
```suggestion
    seed: &[u8; ARANYA_SEED_IKM_LEN],
```

or maybe via a `struct SeedIkm { bytes: [u8; 32] }`

jdygert-spok: If we do the struct we shouldn't need the ARANYA_SEED_IKM_LEN const

gknopf-aranya: Doesn't the application still need to know how many bytes to allocate before passing in the IKM buffer?

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: How do I add it to the `TeamConfigBuilder` after changing the type?

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: You can use `sizeof` when you need the length, but in this case you just pass in the `&SeedIkm` itself.

examples/c/example.c
```diff
@@ -268,26 +281,70 @@ AranyaError init_team(Team *t) {
...
+    AranyaQuicSyncConfigBuilder owner_quic_build;
+    err = aranya_quic_sync_config_builder_init(&owner_quic_build);
+    if (err != ARANYA_ERROR_SUCCESS) {
+        fprintf(stderr, "unable to init `AranyaQuicSyncConfigBuilder`\n");
+        return err;
+    }
+
+    uint8_t *seed = calloc(ARANYA_SEED_LEN, 1);
+    if (t->seed_mode == RAW_IKM) {
+        err = aranya_quic_sync_config_raw_seed(&owner_quic_build, seed,
```
jdygert-spok: This is using an ikm of all zeros. OK-ish for testing but not good to show in the example.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I tried removing the `EncapSeed` type, but then we end up with `Vec<u8>` in a bunch of places in defs.rs and config.rs. We could use an alias instead of a type struct.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: `unsafe { quic.read() }`

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: I don't really see any more vecs: 910a6eb52618a3670453547b9718a9a63fd4861c

crates/aranya-client/src/config.rs
gknopf-aranya: I'm using that method here:
```
impl From<QuicSyncConfig> for aranya_client::QuicSyncConfig {
    fn from(value: QuicSyncConfig) -> Self {
        Self::builder()
            .mode(value.mode)
            .build()
            .expect("All fields are set")
    }
}
```

jdygert-spok: We don't want to expose `SeedMode`. I guess we can either `doc(hidden)` this method or match on the mode to call the corresponding builder method.

ssweetney-spok: Yea, or if we do want to keep this method then a new type should be defined in `aranya-client` instead of using the one in `aranya-daemon-api`. `match`'ing is probably ideal for now

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -242,37 +331,49 @@ impl Typed for QuicSyncConfig {
...
+        self.mode = mode;
+    }
+
+    /// Sets mode to generate PSK seed.
+    pub fn generate(&mut self) {
+        self.mode = SeedMode::Generate;
+    }
+
+    /// Sets wrapped PSK seed.
+    pub fn wrapped_seed(&mut self, encap_seed: EncapSeed) -> Result<(), Error> {
```
jdygert-spok:
```suggestion
    pub fn wrapped_seed(&mut self, encap_seed: &[u8]) -> Result<(), Error> {
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Added buffer reallocation to the example if it isn't large enough.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Sounds like we need a new method to initialize `SeedIkm` from a buffer or something.

examples/c/example.bash
```diff
@@ -0,0 +1,101 @@
...
+    cat <<EOF >"${example}/configs/${device}-config.json"
+{
+    "name": "${device}",
+    "runtime_dir": "${out}/${device}/run",
+    "state_dir": "${out}/${device}/state",
+    "cache_dir": "${out}/${device}/cache",
+    "logs_dir": "${out}/${device}/log",
+    "config_dir": "${out}/${device}/config",
+    "sync_addr": "127.0.0.1:${port}",
+    "quic_sync": {},
```
djohnson-spideroak, line 57: is this consistent with the config value name in the main QUIC syncer PR?

crates/aranya-client/src/config.rs
gknopf-aranya: Used match for this. The annoying part is I needed `wrapped_seed()` and `wrapped_seed_from_bytes()` method to avoid deserializing a `WrappedSeed` type only to serialize it again.

examples/c/example.bash
gknopf-aranya: It matches the `example.json` file from that PR

crates/aranya-client/src/config.rs
jdygert-spok: Exposing `WrappedSeed` is the same issue.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: That's just memcpy.

gknopf-aranya: Defining a new type that needs to be initialized with a buffer before calling `quic_sync_config_raw_seed()` is more complex than passing the buffer into `quic_sync_config_raw_seed()` directly.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -649,13 +656,65 @@ pub fn client_config_builder_set_aqc_config(cfg: &mut ClientConfigBuilder, aqc_c
...
+    cfg.wrapped_seed(encap_seed.to_vec())?;
+    Ok(())
+}
+
+/// Raw PSK seed IKM for QUIC syncer.
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug)]
+pub struct SeedIkm {
+    bytes: [u8; ARANYA_SEED_IKM_LEN],
+}
```
jdygert-spok: Use `repr(C)` so it becomes a struct on the C side instead of a type alias.

crates/aranya-client/src/config.rs
gknopf-aranya: Pushed commit to hide the docs

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -260,6 +260,13 @@ impl From<aranya_crypto::Id> for Id {
     }
 }
 
+/// The size in bytes of a PSK seed IKM.
+pub const ARANYA_SEED_IKM_LEN: usize = 32;
+
+// PSK Seed
+#[aranya_capi_core::opaque(size = 56, align = 8)]
+pub type Seed = Safe<imp::Seed>;
```
gknopf-aranya: unused

examples/c/example.c
```diff
@@ -268,26 +281,72 @@ AranyaError init_team(Team *t) {
...
+    err = aranya_quic_sync_config_builder_init(&owner_quic_build);
+    if (err != ARANYA_ERROR_SUCCESS) {
+        fprintf(stderr, "unable to init `AranyaQuicSyncConfigBuilder`\n");
+        return err;
+    }
+
+    AranyaSeedIkm ikm;
+    for (uint8_t i = 0; i < sizeof(AranyaSeedIkm); i++) {
+        ikm.bytes[i] = i;
+    }
```
ssweetney-spok: There is a public method on the `Client` type for generating random bytes.

```Rust
    /// Generate random bytes from a CSPRNG.
    /// Can be used to generate IKM for a generating a PSK seed.
    pub async fn rand(&self, buf: &mut [u8]) {
        <Rng as Csprng>::fill_bytes(&mut Rng, buf);
    }
```

We can update the example on a later PR

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -672,13 +714,30 @@ pub fn quic_sync_config_build(
...
+///
+/// @param cfg a pointer to the builder for a team config
+/// @param quic set the QUIC syncer config
+pub fn team_config_builder_set_quic_syncer(
+    cfg: &mut TeamConfigBuilder,
+    quic: OwnedPtr<QuicSyncConfig>,
+) {
+    // SAFETY: the user is responsible for passing in a valid QuicSyncConfig pointer.
+    let quic = unsafe { quic.read() };
+    cfg.quic(&quic);
```
jdygert-spok: The point of `OwnedPtr` was to pass it in by value

examples/c/example.c
djohnson-spideroak: we shouldnt do this in the example, we should use actual random bytes from the API/CSPRNG. At a minimum, we should warn in a comment that using an insecure IKM like this isnt ideal.

crates/aranya-client/src/config.rs
```diff
@@ -100,8 +100,14 @@ pub struct QuicSyncConfigBuilder {
...
 
 impl QuicSyncConfigBuilder {
-    /// Sets the seed mode to 'Generate'.
-    /// This is the default mode.
+    /// Sets the PSK seed mode.
+    #[doc(hidden)]
+    pub fn mode(mut self, mode: SeedMode) -> Self {
+        self.seed_mode = mode;
+        self
+    }
```
ssweetney-spok, line 108: Does this mean `aranya_daemon_api::SeedMode` can't be updated if we want to guarantee API stability?

jdygert-spok: No, we are hiding this function to indicate it is not part of the public API. I'd probably also name it `__mode`.

examples/c/example.c
```diff
@@ -315,6 +374,152 @@ AranyaError init_team(Team *t) {
...
+                return err;
+            }
+            err = aranya_quic_sync_config_wrapped_seed(
+                &quic_build, received_seed, received_seed_len);
+            if (err != ARANYA_ERROR_SUCCESS) {
+                fprintf(stderr,
+                        "unable to set `AranyaQuicSyncConfigBuilder` wrapped "
+                        "seed\n");
+                return err;
+            }
```
djohnson-spideroak: Lets create an issue to update the example, especially around this part. This is a bit difficult to follow and we need comments/docs here. Things like "here is where you would transfer the wrapped seed..." etc.

crates/aranya-client/src/config.rs
ssweetney-spok: From skimming the [API guidelines](https://rust-lang.github.io/api-guidelines/documentation.html#rustdoc-does-not-show-unhelpful-implementation-details-c-hidden) it seems to be used differently. The doc attribute doesn't affect visibility but is that the common way of indicating that something isn't part of the public API?

There's also [this](https://rust-lang.github.io/api-guidelines/necessities.html#c-stable)

> A crate cannot be stable (>=1.0.0) without all of its public dependencies being stable. Public dependencies are crates from which types are used in the public API of the current crate.

jdygert-spok: Yes, it is still physically public but indicated as not part of the public API. It is a common practice. See e.g. https://predr.ag/blog/checking-semver-for-doc-hidden-items/

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -918,9 +977,74 @@ pub fn create_team(client: &mut Client, cfg: &TeamConfig) -> Result<TeamId, imp:
...
+/// The PSK seed will be encrypted using the public encryption key of the specified device on the team.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team_id the team's ID [`TeamId`].
+/// @param[in] keybundle serialized keybundle byte buffer `KeyBundle`.
+/// @param[out] seed the serialized, encrypted PSK seed.
+/// @param[out] seed_len the number of bytes written to the seed buffer.
+///
+/// @relates AranyaClient.
+pub unsafe fn psk_seed_encrypt_for_peer(
```
jdygert-spok: nit: `encrypt_psk_seed_for_peer`

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -918,9 +977,74 @@ pub fn create_team(client: &mut Client, cfg: &TeamConfig) -> Result<TeamId, imp:
...
+/// @param[in] psk_bytes the serialized PSK bytes received from a peer.
+/// @param[out] team_id the team's ID [`TeamId`].
+/// @param[out] seed the encrypted PSK seed.
+/// @param[out] seed_len the size in bytes of the encrypted PSK seed.
+pub unsafe fn psk_seed_receive_from_peer(
+    psk_bytes: &[u8],
+    team_id: &mut MaybeUninit<TeamId>,
+    seed: *mut MaybeUninit<u8>,
+    seed_len: &mut usize,
+) -> Result<(), imp::Error> {
```
jdygert-spok: I'm not a fan of this API. If the goal is to make the team config transferable in one thing, then we should add something for serializing/deserializing the config. Since this method has a param for each thing we can't continue to use this as we add more stuff to the config.

IMO for now we should have `encrypt_psk_seed_for_peer` produce just the serialized `WrappedSeed`, and later we can add something for serializing the whole config.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I'm following the convention of starting with the type first.

examples/c/example.c
gknopf-aranya: I added a comment about when the owner would transfer the PSK to a peer.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: This would work with the current team config, because it only has the ability to configure the QUIC syncer PSK. What if we add a field to the team config in the future that we want the peer to configure differently from the owner? If the owner serializes the team config before sending it to the peer, the peer will not have the ability to customize its own team config. Handling the PSK transmission this way is more backward compatible for future changes to the team config.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: It's not really operating on a psk seed. e.g. something more like this would

```rust
impl PskSeed {
    fn encrypt_for_peer(&mut self, ...) {}
}
```

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: If we're going to need more objects later, then it doesn't make sense to put the team ID in with the psk seed. It is not a property of the seed or even the quic sync config, it is a property of the team.

jdygert-spok: We can either have the owner set up the whole config for each user, or we could deserialize into the builder so they can continue to change it. We'd probably want to tweak the API in that case to allow modifying the quic sync config builder from the team config builder.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -918,9 +977,74 @@ pub fn create_team(client: &mut Client, cfg: &TeamConfig) -> Result<TeamId, imp:
...
+    let peer_seed = imp::peer_seed_deserialize(psk_bytes)?;
+
+    team_id.write(peer_seed.get_team_id().into());
+
+    let data = peer_seed.get_encap_seed();
+    if *seed_len < data.len() {
```
gknopf-aranya: We've used similar code in a few places in the C API. Created an issue to define the logic in one spot:
https://github.com/aranya-project/aranya/issues/340

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: We discussed previously that we are going to send the team ID and PSK in one type for the following reasons:
1. It simplifies the API and application because the application only needs to send a single type over for `AddTeam()`.
2. It's less likely for the application to accidentally mix up the team ID and PSK that are passed into `AddTeam()`.

cc: @djohnson-spideroak, @elagergren-spideroak

djohnson-spideroak: Yeah, lets make any changes we need on a different PR.

gknopf-aranya: "If we're going to need more objects later, then it doesn't make sense to put the team ID in with the psk seed. It is not a property of the seed or even the quic sync config, it is a property of the team."

Perhaps calling this `WrappedKey` is not a good idea since what we're really doing is sending information about the team to the peer. We could call the object `TeamInfoForPeer` or something to indicate that the owner is sending information about the team to the peer. The peer can use that info to configure its own team config.

gknopf-aranya: I think we got stuck with this odd naming because we started out thinking we only wanted to send the PSK seed, then realized it would be more efficient to send all the data needed by `AddTeam()` over in one shot.

gknopf-aranya: I'll open a new issue for this.

ssweetney-spok: Shouldn't there be parity between either passing seed IKM or the wrapped seed? There's a new type being added that bundles the team ID and wrapped seed but there isn't one for the team ID and IKM

Also, I think it'd make more sense to leave them separate. Even in the example, the `team_id_from_peer` variable isn't read.

gknopf-aranya: Here is the issue:
https://github.com/aranya-project/aranya/issues/341

djohnson-spideroak: I will bring this up at the rust call.

gknopf-aranya: "Even in the example, the team_id_from_peer variable isn't read."

The reason the `team_id_from_peer` isn't read is because the team ID is already available because the owner created the team. If we were running this on multiple machines, the team ID would need to be passed to the peer with the PSK before it calls `add_team()`

gknopf-aranya: "Shouldn't there be parity between either passing seed IKM or the wrapped seed? There's a new type being added that bundles the team ID and wrapped seed but there isn't one for the team ID and IKM"

The original use-case for the IKM suggested by @mooreoak  was that each peer already has an IKM it wants to use for the team. The IKM was generated ahead of time and passed out-of-band through some secure mechanism. Passing the raw IKM to a peer is not something I'd want to support via the API because the IKM is a secret.

ssweetney-spok: Yea, that's tricky part about these examples. They provide a "top down" view of Aranya when in some cases an example of a how a real application integrates with the client library would be more suitable.

gknopf-aranya: I can change the example to use `team_id_from_peer` instead so that part is more clear.

jdygert-spok (approved):

ssweetney-spok (approved):

djohnson-spideroak (approved):

PR #319, opened 2025-06-17
daemon: finalization error handling

gknopf-aranya:
Handle graph finalization error in the daemon.
- Remove sync peers for a team if the graph has had a finalization error
- Notify the daemon IPC API when a finalization error has occurred for a graph (via an `mpsc` channel)
- Return an error from the daemon IPC API indicating the graph has had a finalization error if any operations are attempted for that team ID

### DIFF
--- a/crates/aranya-daemon/src/api.rs
+++ b/crates/aranya-daemon/src/api.rs
@@ -6,7 +6,7 @@
 use core::{future, net::SocketAddr, ops::Deref, pin::pin};
 use std::{path::PathBuf, sync::Arc};
 
-use anyhow::{anyhow, Context as _, Result};
+use anyhow::{anyhow, Context as _};
 use aranya_crypto::{Csprng, Rng};
 pub(crate) use aranya_daemon_api::crypto::ApiKey;
 use aranya_daemon_api::{
@@ -80,7 +80,7 @@ impl DaemonApiServer {
         recv_effects: EffectReceiver,
         invalid: InvalidGraphs,
         aqc: Aqc<CE, KS>,
-    ) -> Result<Self> {
+    ) -> anyhow::Result<Self> {
         let listener = UnixListener::bind(&uds_path)?;
         let aqc = Arc::new(aqc);
         let effect_handler = EffectHandler {
@@ -165,7 +165,7 @@ struct EffectHandler {
 impl EffectHandler {
     /// Handles effects resulting from invoking an Aranya action.
     #[instrument(skip_all, fields(%graph, effects = effects.len()))]
-    async fn handle_effects(&self, graph: GraphId, effects: &[Effect]) -> Result<()> {
+    async fn handle_effects(&self, graph: GraphId, effects: &[Effect]) -> anyhow::Result<()> {
         trace!("handling effects");
 
         use Effect::*;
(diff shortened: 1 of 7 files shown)


crates/aranya-daemon/src/sync/task.rs
```diff
@@ -241,27 +262,45 @@ impl<ST: SyncState> Syncer<ST> {
...
-            <ST as SyncState>::sync_impl(self, *id, &mut sink, peer)
+            if let Err(e) = <ST as SyncState>::sync_impl(self, peer.graph_id, &mut sink, &peer.addr)
                 .await
-                .context("sync_peer error")
-                .inspect_err(|err| error!("{err:?}"))?;
+                .inspect_err(|err| error!("{err:?}"))
+            {
+                // If a finalization error has occurred, remove all sync peers for that team.
+                if let SyncerError::ClientError(ClientError::ParallelFinalize) = e {
+                    let mut v = Vec::new();
```
gknopf-aranya: `self.peers.retain()`

crates/aranya-daemon/src/api.rs
```diff
@@ -154,6 +179,62 @@ impl DaemonApiServer {
...
 }
 
+/// Keeps track of which graphs have had a finalization error.
+/// Once a finalization error has occurred for a graph,
+/// the graph error is permanent.
+/// The API will prevent subsequent operations on the invalid graph.
+#[derive(Debug, Clone)]
+struct InvalidGraphs {
+    /// HashSet for graph IDs of graphs with finalization errors.
+    graphs: HashSet<GraphId>,
```
jdygert-spok: Lets make this an `Arc<RwLock<HashSet<GraphId>>>` and share `InvalidGraphs` instead of using a channel.

jdygert-spok: Hmm although I guess it depends on where we want to handle the other stuff we might do later.

gknopf-aranya: Just pushed a commit to share the graph data

gknopf-aranya: We don't need to handle the rest of the logic in the API itself.

crates/aranya-daemon/src/sync/tcp.rs
jdygert-spok, line 1: Let's move the two error changes to another PR

crates/aranya-daemon/src/api.rs
```diff
@@ -332,11 +355,15 @@ impl DaemonApi for Api {
...
         cfg: api::TeamConfig,
     ) -> api::Result<()> {
+        self.check_team_valid(team).await?;
+
         todo!()
     }
 
     #[instrument(skip(self))]
     async fn remove_team(self, _: context::Context, team: api::TeamId) -> api::Result<()> {
+        self.check_team_valid(team).await?;
```
jdygert-spok: Allow?

crates/aranya-daemon/src/api.rs
```diff
@@ -316,6 +337,8 @@ impl DaemonApi for Api {
         peer: Addr,
         team: api::TeamId,
     ) -> api::Result<()> {
+        self.check_team_valid(team).await?;
```
jdygert-spok, line 340: Allow? (remove_sync_peer)

gknopf-aranya: All the sync peers for the team are removed when the finalization error happens.
So, we don't need to remove any more sync peers for the team.

crates/aranya-daemon/src/sync/tcp.rs
gknopf-aranya: `tcp.rs` is going to be deleted on the QUIC syncer PR.
Doesn't seem worth it to put any time into changing a file that's going to be deleted in a couple days.
These are minimal changes to get this PR to compile.

jdygert-spok: Oh, you're using it to look for the finalize error. Except it uses the Other variant anyway. I'll revert this and change it to use downcasting for now.

gknopf-aranya: I was planning to update the error to use the new QUIC syncer errors depending on what order the PRs are merged in.

jdygert-spok (approved):

gknopf-aranya:
@jdygert-spok good catch on removing the sync peer from that queue

jdygert-spok (approved):

ssweetney-spok (approved):
👍

PR #322, opened 2025-06-18
Update to latest aranya-core with policy text changes

jdygert-spok:


jdygert-spok:
Passing tests after https://github.com/aranya-project/aranya-core/pull/312

examples/rust/src/main.rs
```diff
@@ -5,6 +5,7 @@ use std::{
     time::Duration,
 };
 
+use aranya_policy_text::text;
```
jdygert-spok: Probably want to re-export this

jdygert-spok:
Hitting that `OnceCell: !Sync` issue we noted from caching IDs since we share the `PublicKeys` in the daemon API.

crates/aranya-daemon/src/api.rs
```diff
@@ -223,8 +226,10 @@ struct ApiInner {
     client: Client,
     /// Local socket address of the API.
     local_addr: SocketAddr,
+    /// ID of current device.
+    device_id: DeviceId,
     /// Public keys of current device.
-    pk: PublicKeys<CS>,
+    pk: KeyBundle,
```
jdygert-spok: Fixed `!Sync` by storing serialized key bundle and device ID since that is all we need them for.

gknopf-aranya (approved):

crates/aranya-daemon/src/api.rs
jdygert-spok: ~Ugh we will need the encryption key for the psk seeds.~ Secret not public. But I guess we need to use the enc id to look it up...

gknopf-aranya: Yep, we need the sender peer's public encryption key to decrypt a PSK seed

jdygert-spok: Meant our encryption key but yeah that is also needed. Reverted this and wrapped it in a mutex for now.

gknopf-aranya (approved):

elagergren-spideroak:
I think the APIs that take `Text` or similar should probably be generic over `TryInto<Text>` for ease of use.

PR #325, opened 2025-06-18
Add Initial Metrics

nikki-aranya:
This adds support for collecting metrics with `metrics-rs`, benchmarks with `divan`, and reporting to an external backend using Prometheus, DataDog, or TCP.

Future work includes adding more metrics, CI regression testing, and better integrations for customers.


nikki-aranya:
Still on the TODO, I need to figure out our Prometheus dashboard situation, write a script so we can better test this since it needs a prometheus server (or just add a custom Recorder instead of exporting to Prometheus), and probably write up some Markdown documentation on this, and probably fix some of my metrics calculations (e.g. high watermark and such).

This also doesn't include all the TODOs I've sprinkled in that I've forgotten, think the main one is reporting each individual daemon's metrics using KV pairs.

Cargo.toml
```diff
@@ -63,15 +63,20 @@ bytes = { version = "1.0" }
 backon = { version = "1.2.0" }
 ciborium = { version = "0.2" }
 clap = { version = "4", features = ["derive"] }
+divan = { version = "0.1", default-features = false }
```
gknopf-aranya, line 69: Is there a way to define dependencies as dev-only in the top-level `Cargo.toml`?
Would be nice to categorize them somehow so they don't show up here with the rest of the deps.

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,37 @@
...
+#[global_allocator]
+static ALLOC: AllocProfiler = AllocProfiler::system();
+
+fn main() {
+    divan::main();
+}
+
+// NOTE: divan currently requires sync functions to work, so we spawn a runtime and block on async.
+#[divan::bench]
+fn daemon_startup() -> Result<()> {
```
gknopf-aranya: In order to collect metrics for both client and daemon, we should create an instance of the client and daemon (connected via the IPC API) in the test.
This appears to only be creating an instance of the daemon without the client.
It could make sense to run the metrics tests from a crate outside of `aranya-client` and `aranya-daemon` so it has access to both crates simultaneously.

gknopf-aranya: I just needed to read more of the PR. That is exactly what the `aranya-metrics` crate is already doing :)

Makefile.toml
```diff
@@ -257,3 +257,32 @@ category = "examples"
...
+cleanup() {
+    echo "Cleaning up..."
+    kill $PUSHGATEWAY_PID 2>/dev/null
+    wait $PUSHGATEWAY_PID 2>/dev/null
+}
+trap cleanup EXIT
+
+echo "Running metrics collection..."
+cargo run --release --bin aranya-metrics -- $(pwd)/target/release/aranya-daemon
+'''
```
gknopf-aranya: This is complex enough to move into its own script under `crates/aranya-metrics` so it doesn't clutter the makefile too much:
```suggestion
cwd = crates/aranya-metrics
script = 'run.bash'
```

Makefile.toml
```diff
@@ -257,3 +257,32 @@ category = "examples"
...
+category = "metrics"
+description = "Run Metrics Crate"
+script = '''
+#!/usr/bin/env bash
+echo "Building our binaries..."
+cargo build --bin aranya-daemon --release
+cargo build --bin aranya-metrics --release
+
+echo "Starting pushgateway..."
+pushgateway &
```
gknopf-aranya: Since `pushgateway` is needed to run the metrics, let's make sure the script returns an error informing the caller to install it if it isn't already installed.

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    let cfg = Config {
+        name: "daemon".into(),
+        runtime_dir: work_dir.join("run"),
+        state_dir: work_dir.join("state"),
+        cache_dir: work_dir.join("cache"),
+        logs_dir: work_dir.join("log"),
+        config_dir: work_dir.join("config"),
+        sync_addr: addr_any,
+        afc: None,
```
gknopf-aranya: I thought afc was supposed to be completely removed from the codebase

nikki-aranya: It's still in the config so we don't break any old configs that still have it as a key

Cargo.toml
nikki-aranya: afaik no, the rationale is it doesn't really make sense for workspace dependencies since they're not actually "used"

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,66 @@
...
+As with any project that grows large enough, Aranya is now able to collect various metrics to measure performance and find areas to target for improvement. This is accomplished in several ways:
+
+-   Benchmarks that use `divan` to measure runtime and memory usage
+-   The `aranya-metrics` crate that adds a harness on top of Aranya code that's representative of typical operation in order to see how the entire stack performs in that scenario. This includes cpu time, memory usage, and disk usage.
+
+Future plans to support more information include:
+
+-   Adding `metrics-rs` into more areas of Aranya to measure performance directly, with an optional recorder enabled to actually collect and display those values. Note that `metrics-rs` is designed to be lightweight enough that without a recorder, the performance impact is negligible.
+-   Adding more granular metrics, using key-value pairs to report the data from specific daemon instances.
+-   Using the above `divan` benchmarks, as well as [flame graphs](https://www.brendangregg.com/flamegraphs.html) to narrow down specific functions to augment with `#[fastrace::trace]` for measuring hot spots for improvements.
```
gknopf-aranya: Could use the phrase "identify performance bottlenecks" here in relation to the flame graph

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,66 @@
...
+-   Adding a `metrics-rs` recorder and `fastrace` reporter to the daemon binary behind a feature flag to enable collecting metrics in production.
+
+# Installation Instructions
+
+Make sure [homebrew](https://brew.sh/) is installed, and then install the following packages:
+
+-   [brew install prometheus](https://formulae.brew.sh/formula/prometheus), which installs [Prometheus](https://prometheus.io/), an open source metrics collection backend.
```
gknopf-aranya: Are `prometheus`, `grafana`, and `pushgateway` all required for metics collection?
What if I want to collect some metrics locally without these?

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,66 @@
...
+Future plans to support more information include:
+
+-   Adding `metrics-rs` into more areas of Aranya to measure performance directly, with an optional recorder enabled to actually collect and display those values. Note that `metrics-rs` is designed to be lightweight enough that without a recorder, the performance impact is negligible.
+-   Adding more granular metrics, using key-value pairs to report the data from specific daemon instances.
+-   Using the above `divan` benchmarks, as well as [flame graphs](https://www.brendangregg.com/flamegraphs.html) to narrow down specific functions to augment with `#[fastrace::trace]` for measuring hot spots for improvements.
+-   Adding a `metrics-rs` recorder and `fastrace` reporter to the daemon binary behind a feature flag to enable collecting metrics in production.
+
+# Installation Instructions
+
+Make sure [homebrew](https://brew.sh/) is installed, and then install the following packages:
```
gknopf-aranya: This step only applies for MacOS

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+-   [brew install prometheus](https://formulae.brew.sh/formula/prometheus), which installs [Prometheus](https://prometheus.io/), an open source metrics collection backend.
+-   [brew install grafana](https://formulae.brew.sh/formula/grafana), which will install [Grafana](https://grafana.com/), an open source frontend to better visualize our metrics.
+
+Download the correct package for the [`pushgateway`](https://prometheus.io/download/#pushgateway) since it's not available on `homebrew`, extract it, and `sudo cp pushgateway /usr/local/bin/` to install it as a tool.
+
+Modify the Prometheus config file at `/opt/homebrew/etc/prometheus.yml` (or `/usr/local/etc/prometheus.yml`) to the following, to add `pushgateway` and `node_exporter` support:
+
+```yaml
```
gknopf-aranya: I would rather manage this file as its own file in this repo that's referenced from the README, rather than inline.
Then the reader can easily run something like this:
`cp <aranya repo>/crates/aranya-metrics/prometheus.yml /opt/homebrew/etc/prometheus.yml`
Version controlling the file separately from the README is a bit more manageable.

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,66 @@
...
+-   `brew services start prometheus`
+-   `brew services start grafana`
+
+Grafana can be accessed at `https://localhost:3000/`, sign-in by default is `admin:admin`.
+
+# Collecting Metrics
+
+The easiest way to do an `aranya-metrics` run is to simply call `cargo make metrics`, which will compile the daemon and this crate, spin up a `pushgateway`, run the example, and close everything.
+
+After that's done and you have the job name (i.e. `aranya_demo_1751415368`), the easiest way to view the data is to go to Prometheus (`localhost:9090`), query for the job (`{job="aranya_demo_1751415368"}`), go to the graph tab, adjust the timescale to one where you can see the whole run, and you can click and drag to select a smaller range. Make sure to change the resolution to whatever your scrape interval is (the above config sets it to 100ms). You can then selectively look at categories using the list below the graph.
```
gknopf-aranya: We will want a way to view metrics in CI without opening a Prometheus GUI. Printing the values to the command line is something we can reference later after the CI run completes.

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+# Technical Details
+
+Due to `sysinfo` only providing percentages of CPU usage at each refresh instead of raw user time/system time, we use `proc_pidinfo` on MacOS (falling back to `rusage` which only distinguishes between "self" and a generic "children") to grab CPU time and memory usage, using sysinfo for file I/O. `proc_pidinfo` and `rusage` both capture information "at the current moment" which means there isn't really a delta, whereas sysinfo's bytes_read/written is "since last refresh" so we have to aggregate it along with all previous deltas.
```
gknopf-aranya: Does this mean that metrics on non-MacOS will be less accurate?

crates/aranya-metrics/README.md
gknopf-aranya: Is there a way to save Prometheus metrics somewhere so we can access them after an ephemeral GitHub runner has been destroyed?

crates/aranya-metrics/src/main.rs
gknopf-aranya, line 1: CI indicates this file needs to be formatted

crates/aranya-metrics/run.bash
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -e # Exit on any errors
+
+if ! command -v pushgateway >/dev/null 2>&1; then
```
gknopf-aranya: This is checking if `pushgateway` is installed, but not if `prometheus` and `graphana` are installed.
Aren't they all currently required for metrics collection?

crates/aranya-metrics/run.bash
```diff
@@ -0,0 +1,33 @@
...
+    fi
+    sleep 0.25
+done
+
+cleanup() {
+    echo "Cleaning up..."
+    kill $PUSHGATEWAY_PID 2>/dev/null
+    wait $PUSHGATEWAY_PID 2>/dev/null
+}
+trap cleanup EXIT
```
gknopf-aranya: We've used this in several of our bash scripts to cleanup all child processes. This is easier to maintain as more child processes are added:
```suggestion
cleanup() {
    jobs -p | xargs -I{} kill {} || true
}
trap 'cleanup' EXIT
trap 'trap - SIGTERM && cleanup && kill -- -$$ || true' SIGINT SIGTERM EXIT
```

crates/aranya-metrics/README.md
nikki-aranya: No, it just means they need their own implementations (I still need to hook `/proc/<pid>/*`)

gknopf-aranya:
@jdygert-spok, @djohnson-spideroak could you please take a look at this PR when you get a chance? Okay to wait until after Thursday's release if you're booked.

crates/aranya-metrics/README.md
nikki-aranya: As discussed in I forget what meeting, this is the bare minimum to get this up and running, I'm not expecting anyone other than tech people (preferably only our team) who should know that homebrew is MacOS-only to read this

gknopf-aranya:
The options are nice-to-have. If we only support tracing/logging at first and add an option for datadog/prometheus later, that's okay too. Whatever is simplest to complete soon.

gknopf-aranya:
> Prometheus is already fully implemented, I just had to do the above refactoring to shove it in an enum, tracing is next and doesn't look too bad, dogstatsd just requires another couple structs for its config which we'll see how annoying that turns out to be, might just skip it for the moment

Datadog can be added in a follow-up PR. We just need to land a baseline for benchmarks on `main` without adding additional scope.

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,399 @@
...
+                let disk_read_rate = (current.total_disk_read_bytes as f64) / time_delta;
+                let disk_write_rate = (current.total_disk_write_bytes as f64) / time_delta;
+
+                gauge!("cpu_user_utilization_rate").set(cpu_user_rate);
+                gauge!("cpu_system_utilization_rate").set(cpu_system_rate);
+                gauge!("cpu_total_utilization_rate").set(cpu_user_rate + cpu_system_rate);
+                gauge!("memory_allocation_rate").set(memory_alloc_rate);
+                // TODO(nikki): standardize this on "per second", even if we tick more/less frequently?
+                gauge!("disk_read_bytes_rate").set(disk_read_rate);
+                gauge!("disk_write_bytes_rate").set(disk_write_rate);
```
jdygert-spok: Usually these are calculated further up the stack using the totals instead of being reported directly.

crates/aranya-metrics/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "aranya-metrics"
+version.workspace = true
```
jdygert-spok:
```suggestion
publish = false
```

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,42 @@
...
+    let cfg = Config {
+        name: "daemon".into(),
+        runtime_dir: work_dir.join("run"),
+        state_dir: work_dir.join("state"),
+        cache_dir: work_dir.join("cache"),
+        logs_dir: work_dir.join("log"),
+        config_dir: work_dir.join("config"),
+        sync: SyncConfig {
+            quic: Toggle::Enabled(QuicSyncConfig {
+                addr: Addr::from((Ipv4Addr::UNSPECIFIED, 4321)),
```
gknopf-aranya: This won't work if something already bound to port 4321. Better to bind to any available port:
```suggestion
                addr: Addr::from((Ipv4Addr::UNSPECIFIED,0)),
```

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,66 @@
+# Aranya Metrics Collection
+
+As with any project that grows large enough, Aranya is now able to collect various metrics to measure performance and find areas to target for improvement. This is accomplished in several ways:
```
gknopf-aranya: A profiler would be better at identifying areas for improvement. These metrics mostly show the numbers as they are without explaining why or identifying bottlenecks.

crates/aranya-metrics/example.hjson
gknopf-aranya, line 1: Is this filename supposed to be `example.json`? Not familiar with .hjson extension.

crates/aranya-metrics/example.hjson
```diff
@@ -0,0 +1,9 @@
+{
+    /*
```
gknopf-aranya: Why is this commented out?

crates/aranya-metrics/src/export/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+    #[default]
+    Total,
+    PerProcess,
+}
+
+/// Configuration for metrics collection and exporting
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default)]
+pub struct MetricsConfig {
+    pub mode: MetricsMode,
```
gknopf-aranya: Docs needed for `pub` fields. Should probably enable more lints for this crate.

crates/aranya-metrics/src/export/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub interval: Duration,
+    pub job_name: String,
+    pub debug_logs: DebugLogType,
+}
+
+impl MetricsConfig {
+    pub fn install(&self) -> Result<()> {
```
gknopf-aranya: Docs needed for `pub` methods

crates/aranya-metrics/src/export/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+            ),
+            debug_logs: DebugLogType::Total,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub enum MetricsMode {
+    Prometheus(prometheus::PrometheusConfig),
+    //DataDog(DataDogConfig),
```
gknopf-aranya:
```suggestion
    //TODO: DataDog(DataDogConfig),
```

crates/aranya-metrics/src/export/prometheus.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn format_push_gateway_url(base_url: &str, job_name: &str) -> String {
+        match Url::parse(base_url) {
+            Ok(mut url) => {
+                url.set_path(&format!("/metrics/job/{job_name}"));
+                url.set_query(None);
+                url.set_fragment(None);
+                url.to_string()
+            }
+            Err(_) => {
+                warn!("Failed to parse push gateway URL `{base_url}`");
```
gknopf-aranya: This method should probably return an error if the `base_url` parameter can't be parsed.

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    fn collect_aggregated_metrics(&mut self) -> Result<AggregatedMetrics> {
+        let mut metrics = AggregatedMetrics {
+            timestamp: Instant::now(),
+            total_cpu_user_time_us: 0,
+            total_cpu_system_time_us: 0,
+            total_memory_bytes: 0,
+            total_disk_read_bytes: 0,
+            total_disk_write_bytes: 0,
+            //total_network_rx_bytes: 0,
```
gknopf-aranya: It should be enough to only have the TODO in the struct definition. Not needed here.

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,408 @@
...
+
+        Ok(())
+    }
+
+    fn update_prometheus_metrics(&self, current: &AggregatedMetrics) -> Result<()> {
+        // Update all absolute metrics.
+        // TODO(nikki): add tags for individual PIDs so we can track each daemon?
+        let total = &self.total_metrics;
+
+        #[allow(clippy::cast_precision_loss)]
```
gknopf-aranya, line 384: @nikki-aranya, @jdygert-spok is there a way to not cast these to float? This doesn't seem right.

jdygert-spok: No, the metrics crate (and prometheus) expects floats.

crates/aranya-metrics/example.hjson
nikki-aranya: we use(d) deser_hjson which is a superset of json for our config parsing, but dygert recently transitioned to toml and I haven't yet ported those changes, that's also on the TODO to get this ready. It's commented out since I technically have default values already added from serde(default), it's mostly just to show what you'd write if you wanted to change the defaults

gknopf-aranya:
@nikki-aranya I added a checklist to this PR so it's easier to track what's left. Please let me know if there are any other todos we should track.

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,42 @@
...
+    let cfg = Config {
+        name: "daemon".into(),
+        runtime_dir: work_dir.join("run"),
+        state_dir: work_dir.join("state"),
+        cache_dir: work_dir.join("cache"),
+        logs_dir: work_dir.join("log"),
+        config_dir: work_dir.join("config"),
+        sync: SyncConfig {
+            quic: Toggle::Enabled(QuicSyncConfig {
+                addr: Addr::from((Ipv4Addr::UNSPECIFIED, 0)),
```
gknopf-aranya: Better to use `LOCALHOST` here. `UNSPECIFIED` will bind to external interfaces.

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,42 @@
...
+            }),
+        },
+        aqc: Toggle::Enabled(AqcConfig {}),
+    };
+
+    let rt = Runtime::new()?;
+    let _daemon = rt.block_on(Daemon::load(cfg))?;
```
gknopf-aranya: This is loading a daemon object from a config file but not starting up the daemon.
```suggestion
    let _daemon = rt.block_on(Daemon::load(cfg).await?.spawn())?;
```

supply-chain/audits.toml
gknopf-aranya, line 1: @djohnson-spideroak, @jdygert-spok, @nikki-aranya I don't think we want to trust specific Rust maintainers without reviewing the `cargo vet` diffs. We've added in trust for our own crates because those are guaranteed to go through our internal peer review process. 3rd party deps should be manually certified (unless they are already certified by existing 3rd party supply chain audits).

crates/aranya-metrics/src/backend/datadog.rs
gknopf-aranya, line 1: I thought we were only going to add the `prometheus` exporter in this initial PR

crates/aranya-metrics/src/main.rs
```diff
@@ -0,0 +1,466 @@
...
+
+    info!("revoking label from membera");
+    operator.revoke_label(ctx.membera.id, label3).await?;
+    info!("revoking label from memberb");
+    operator.revoke_label(ctx.memberb.id, label3).await?;
+
+    admin.sync_now(operator_addr.into(), None).await?;
+
+    info!("deleting label");
+    admin.delete_label(label3).await?;
```
gknopf-aranya, line 459: After adding the final command to the graph, having each daemon `sync_now()` with the admin would result in a stable state. Then this sleep wouldn't be needed at the end of the test:
`sleep(Duration::from_millis(25)).await;`

crates/aranya-metrics/src/main.rs
gknopf-aranya, line 1: @nikki-aranya, @djohnson-spideroak, @jdygert-spok keeping this up-to-date as the Rust example is changed will be a lot of work. It's already caused extra work on the current PR.

crates/aranya-metrics/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#![warn(clippy::missing_docs_in_private_items, missing_docs)]
```
gknopf-aranya: Can this be moved after the file docs?

crates/aranya-metrics/run.bash
```diff
@@ -0,0 +1,35 @@
...
+
+echo "Building our binaries..."
+cargo build --bin aranya-daemon --release
+cargo build --bin aranya-metrics --release
+
+echo "Starting pushgateway..."
+pushgateway &
+PUSHGATEWAY_PID=$!
+
+for i in {1..24}; do
```
gknopf-aranya:
```suggestion
# Wait for `pushgateway` server endpoint
for i in {1..24}; do
```

crates/aranya-metrics/run.bash
```diff
@@ -0,0 +1,35 @@
...
+cleanup() {
+    jobs -p | xargs -I{} kill {} || true
+}
+trap 'cleanup' EXIT
+trap 'trap - SIGTERM && cleanup && kill -- -$$ || true' SIGINT SIGTERM EXIT
+
+cd crates/aranya-metrics
+
+echo "Running metrics collection..."
+cd ../..
```
gknopf-aranya: This changes into the `<repo>/crates/aranya-metrics` folder then back into `<repo>`. Might as well stay at the repo level.
```suggestion
echo "Running metrics collection..."
```

crates/aranya-metrics/src/backend/datadog.rs
```diff
@@ -0,0 +1,155 @@
...
+use std::{net::SocketAddr, time::Duration};
+
+use anyhow::{Context as _, Result};
+use metrics::Label;
+use metrics_exporter_dogstatsd::DogStatsDBuilder;
+use tracing::info;
+
+/// Configuration info for the DogStatsD exporter.
+///
+/// This includes:
```
gknopf-aranya: The documentation for individual fields should be right above each field. Struct documentation is meant to be a high-level overview of the struct.

crates/aranya-metrics/src/backend/datadog.rs
```diff
@@ -0,0 +1,155 @@
...
+/// * Optional prefixes and additional labels that can be attached to metrics being sent
+/// * Whether to enable additional telemetry, which lets the Datadog Agent see additional info about
+///   how the exporter is performing
+/// * Whether to use reservoir sampling to represent arbitrarily large data using a smaller array
+/// * Whether to forward histograms as distributions, which allows the Datadog Agent to process the
+///   data which allows for richer insights
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(default)]
+pub struct DataDogConfig {
+    // Current backend is always synchronous, so we don't bother with it here.
```
gknopf-aranya: Is this doc for the `remote_addr` field or something else?

crates/aranya-metrics/src/backend/mod.rs
```diff
@@ -0,0 +1,117 @@
+//! This module contains all the backend code needed to export metrics to a remote server, as well
+//! as various other configuration parameters that handle things such as logging to the console.
+//!
+//! The currently supported backends are:
+//! * [Prometheus], using either [Prometheus](https://prometheus.io/) scraping, or a `pushgateway`.
+//! * [Datadog] using [DogStatsD], provided as a service as part of a Datadog Agent.
+//! * [TCP Server], used for more advanced metrics collection and processing, sent using [`protobuf`].
```
gknopf-aranya:
```suggestion
//! * [TCP Server], a lightweight TCP server for more customized metrics collection and processing, sent using [`protobuf`].
```

crates/aranya-metrics/src/backend/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub job_name: String,
+    /// Whether to log metrics to the console using [`tracing::debug!()`](tracing::debug).
+    pub debug_logs: DebugLogType,
+}
+
+impl MetricsConfig {
+    /// Sets up the selected exporter using the provided configuration info.
+    pub fn install(&self) -> Result<()> {
+        match &self.mode {
+            MetricsMode::Prometheus(prometheus) => {
```
gknopf-aranya: If you define a trait that requires each metrics backend to implement `install(config)`, this code can be simplified by calling `self.mode.install(self)` on the generic backend.

crates/aranya-metrics/src/backend/prometheus.rs
```diff
@@ -0,0 +1,199 @@
...
+use anyhow::{Context as _, Result};
+use metrics_exporter_prometheus::PrometheusBuilder;
+use metrics_util::MetricKindMask;
+use serde::Deserialize;
+use tracing::info;
+
+/// Configuration info for the Prometheus exporter.
+///
+/// This includes:
+/// * The current mode (either exporting to Prometheus via Unix Domain Socket or normal HTTP, or
```
gknopf-aranya: Docs for specific fields should be moved right above the field.

crates/aranya-metrics/src/backend/prometheus.rs
```diff
@@ -0,0 +1,199 @@
...
+pub struct PrometheusConfig {
+    /// The current mode that the Prometheus exporter is operating as.
+    mode: PrometheusMode,
+
+    /// Sets the idle timeout for metrics.
+    idle_timeout: (MetricsKindMask, Option<Duration>),
+    /// Sets whether a unit suffix is added to metric names.
+    enable_unit_suffix: bool,
+
+    /// Allows only certain addresses to access the scrape endpoint.
```
gknopf-aranya: Include whitelist from above here. That's a good description of this.

crates/aranya-metrics/src/backend/prometheus.rs
```diff
@@ -0,0 +1,199 @@
...
+            .context("Failed to install Prometheus exporter")?;
+
+        Ok(())
+    }
+}
+
+impl Default for PrometheusConfig {
+    fn default() -> Self {
+        Self {
+            mode: PrometheusMode::PushGateway {
```
gknopf-aranya, line 140: Should document that if the `PushGateway` server is hosted anywhere other than localhost, a username and password should be set.

crates/aranya-metrics/src/backend/prometheus.rs
```diff
@@ -0,0 +1,199 @@
...
+/// * The timeout for a metric before it stops being logged. For long running tasks, this is ideal
+///   to tune since it will give a nice cutoff after a specific task might have finished.
+/// * A whitelist of addresses allowed to connect to the exporter
+/// * The list of quantiles (0.0, 0.5, 0.9, 0.95, 0.99, 0.999, and 1.0) used when rendering
+///   histograms
+/// * The number and size of buckets, which allow for greater flexibility in Prometheus to derive
+///   quantiles
+/// * How often the upkeep task runs, which cleans out metrics that Prometheus has already scraped
+/// * A set of global labels that can be applied to all metrics
+#[derive(Debug, Clone, Deserialize)]
```
gknopf-aranya: Why is there a `Deserialize` with no `Serialize`? How would someone end up with a serialized object to deserialize without a `Serialize` implementation?

crates/aranya-metrics/src/backend/tcp_server.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// The address to listen on for TCP connections.
+    listen_addr: SocketAddr,
+    /// The size of the internal buffer used for processing metrics.
+    buffer_size: Option<usize>,
+}
+
+impl Default for TcpConfig {
+    fn default() -> Self {
+        Self {
+            listen_addr: SocketAddr::from(([0, 0, 0, 0], 5000)),
```
gknopf-aranya: Default should be localhost. 0.0.0.0 will bind to external interfaces. We should leave a note that the TCP exporter does not encrypt metrics.

crates/aranya-metrics/src/backend/tcp_server.rs
gknopf-aranya, line 1: Note that `protobuf` must be installed to build the metrics even if the tcp exporter isn't used:
```
error: failed to run custom build command for `metrics-exporter-tcp v0.11.1`

Caused by:
  process didn't exit successfully: `/Users/gknopf/scm/aranya/target/debug/build/metrics-exporter-tcp-610375c683a84cda/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=proto/event.proto

  --- stderr

  thread 'main' panicked at /Users/gknopf/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/metrics-exporter-tcp-0.11.1/build.rs:5:69:
  called `Result::unwrap()` on an `Err` value: Custom { kind: NotFound, error: "Could not find `protoc`. If `protoc` is installed, try setting the `PROTOC` environment variable to the path of the `protoc` binary. To install it on macOS, run `brew install protobuf`. It is also available at https://github.com/protocolbuffers/protobuf/releases  For more information: https://docs.rs/prost-build/#sourcing-protoc" }
```

gknopf-aranya: We may want to put the tcp exporter backend behind a build flag so `protobuf` doesn't need to be installed in ci and all dev machines. @jdygert-spok, @djohnson-spideroak thoughts on this?

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,341 @@
...
+    //total_network_rx_bytes: u64,
+    //total_network_tx_bytes: u64,
+    /// The current number of processes we're collecting metrics about.
+    process_count: usize,
+}
+
+impl Default for AggregatedMetrics {
+    fn default() -> Self {
+        Self {
+            timestamp: Instant::now(),
```
gknopf-aranya, line 88: nit: it's not ideal to define a default for all the fields when only the timestamp field doesn't have a default implementation already. Could define a new type for the timestamp where the default is `Instant::now()`.

```
#[derive(Debug)]
struct NewInstant(Instant);

/// Contains the metrics collected for all processes.
#[derive(Debug, Default)]
struct AggregatedMetrics {
    /// The moment we start collecting metrics, used to calculate the delta of how long it took.
    timestamp: NewInstant,
    /// The total time the CPU spent in userspace.
    total_cpu_user_time_us: u64,
    /// The total time the CPU spent processing syscalls.
    total_cpu_system_time_us: u64,
    /// The total number of bytes used all processes.
    total_memory_bytes: u64,
    /// The total number of bytes read from disk by all processes.
    total_disk_read_bytes: u64,
    /// The total number of bytes written to disk by all processes.
    total_disk_write_bytes: u64,

    // TODO(nikki): hook the TCP/QUIC syncer with a metrics macro.
    //total_network_rx_bytes: u64,
    //total_network_tx_bytes: u64,
    /// The current number of processes we're collecting metrics about.
    process_count: usize,
}

impl Default for NewInstant {
    fn default() -> Self {
        Self(Instant::now())
    }
}
```

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Collects metrics for a specific process, using a number of syscalls.
+    #[cfg(not(target_os = "macos"))]
+    #[allow(dead_code)]
+    fn collect_process_metrics(&self, _pid: u32, _metrics: &mut AggregatedMetrics) -> Result<()> {
+        Err(anyhow!("Unsupported target_os!"))
```
gknopf-aranya: Include the target OS and supported OS in the error so it's easier to troubleshoot.

crates/aranya-metrics/README.md
gknopf-aranya, line 1: If `protobuf` always needs to be installed to build the tcp exporter, we should mention this in the Aranya project's top-level `README.md`

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,68 @@
...
+-   Adding a `metrics-rs` recorder and `fastrace` reporter to the daemon binary behind a feature flag to enable collecting metrics in production.
+
+# Installation Instructions
+
+Make sure [homebrew](https://brew.sh/) is installed, and then install the following packages:
+
+-   [brew install prometheus](https://formulae.brew.sh/formula/prometheus), which installs [Prometheus](https://prometheus.io/), an open source metrics collection backend.
+-   [brew install grafana](https://formulae.brew.sh/formula/grafana), which will install [Grafana](https://grafana.com/), an open source frontend to better visualize our metrics.
+
+Download the correct package for the [`pushgateway`](https://prometheus.io/download/#pushgateway) since it's not available on `homebrew`, extract it, and `sudo cp pushgateway /usr/local/bin/` to install it as a tool.
```
gknopf-aranya: After installing `pushgateway` and attempting to run it I got this error:
“pushgateway” cannot be opened because the developer cannot be verified.

We should add a note about allowing this to run on MacOS: https://support.apple.com/en-us/102445

crates/aranya-metrics/src/backend/prometheus.rs
nikki-aranya: Because the only way to create this right now is a toml file we deserialize into the struct

crates/aranya-metrics/run.bash
nikki-aranya: oh, well that would explain why I couldn't force cwd when I was trying to fix it yesterday

crates/aranya-metrics/src/main.rs
nikki-aranya: The sleep is more for the metrics to grab a couple more frames of the end steady state

crates/aranya-metrics/src/backend/datadog.rs
nikki-aranya: That was only in your head, I said multiple times it wasn't hard to do and that I really should get it done now

supply-chain/audits.toml
nikki-aranya: We depend on third party audits which already trust specific maintainers so it's a moot point, you'd have to start a new audit from scratch if you want to do that

crates/aranya-metrics/src/main.rs
nikki-aranya: That'll be addressed in the follow-up PR to this hopefully, need to unify the example and metrics code, not that it matters too much since I only stole the example code cuz it was the most representative sample of a typical aranya workload we currently have

supply-chain/audits.toml
gknopf-aranya: By that line of reasoning we could trust the maintainers for all our deps and save ourselves the hassle of certifying each version change. I think we need more discussion before trusting all updates from certain maintainers.

crates/aranya-metrics/src/backend/datadog.rs
nikki-aranya: The [DogStatsD exporter](https://docs.rs/metrics-exporter-dogstatsd/latest/metrics_exporter_dogstatsd/struct.DogStatsDBuilder.html#method.with_synchronous_backend) has this which doesn't actually do anything right now, leaving it as a note in case that ever changes:
```
pub fn with_synchronous_backend(self) -> Self
Use a synchronous backend for forwarding metrics.
A background OS thread will be spawned to handle forwarding metrics to the remote server.
Defaults to true.
```

gknopf-aranya: I would mention this in the struct docs rather than the fields docs so it doesn't appear to be related to the field below it.

supply-chain/audits.toml
nikki-aranya: I'll leave you to figure that out then, that's the behavior that cargo vet pushes you into:
```
    cargo vet diff libc 0.2.67 0.2.172                    rust-lang-owner  cc, mio, rand, ring, divan, and 21 others                     255 files changed, 114083 insertions(+), 56161 deletions(-)
      NOTE: mozilla, cargo-vet, and bytecode-alliance trust rust-lang-owner - consider cargo vet trust libc rust-lang-owner
```

gknopf-aranya: @mooreoak, @elagergren-spideroak do you have thoughts on this? We added `cargo vet` to improve our dependency security posture, so I want to make sure we're in line with the original vision for this.

crates/aranya-metrics/src/backend/tcp_server.rs
nikki-aranya: build flag won't help for CI unless we specifically exclude that permutation in cargo-all-features which just means it won't get any testing, I already added the package to CI so it can run properly

running locally however, I could see an argument for, doesn't seem like too big a deal to slap 3 feature flags on

crates/aranya-metrics/src/backend/prometheus.rs
nikki-aranya: I guess since it's public there's no harm slapping on Serialize, I need to make sure all my instances are `serde::Deserialize` for consistency with the rest of the codebase

supply-chain/audits.toml
mooreoak: What is the sugested change here?

gknopf-aranya: I'm suggesting that we should continue to manually certify deps unless we've agreed to trust a certain maintainer.

crates/aranya-metrics/src/backend/datadog.rs
gknopf-aranya: The point that I've made previously is that starting with a single backend would've helped us land this PR sooner. Implementing 3 backends has increased the scope of the PR, the amount of code to be reviewed, and required an additional dep to be installed in CI.

gknopf-aranya:
`clippy` found some incorrect args here:
https://github.com/aranya-project/aranya/actions/runs/16653633994/job/47132762598?pr=325
Looks like CI checks need some work to get them all passing.

Cargo.toml
```diff
@@ -66,32 +66,39 @@ bytes = { version = "1.0" }
...
 [profile.dev]
 opt-level = 1
 
 [profile.release]
-lto = true
 opt-level = "z"
 strip = true
-
+lto = true
+panic = "abort"
```
ssweetney-spok, line 101: Do we want to abort? We use some types that zeroize their data as part of their `Drop` impls.

Makefile.toml
```diff
@@ -257,3 +257,23 @@ category = "examples"
...
+# Run `aranya-metrics`
+[tasks.metrics]
+category = "metrics"
+description = "Run Metrics Crate"
+script = '''
+#!/usr/bin/env bash
+set -e
+cargo build --bin aranya-daemon --release
+cargo build --bin aranya-metrics --release
+CONFIG_PATH=crates/aranya-metrics/metrics.toml $(pwd)/target/release/aranya-metrics $(pwd)/target/release/aranya-daemon
```
gknopf-aranya:
```suggestion
CONFIG_PATH=$(pwd)/crates/aranya-metrics/metrics.toml $(pwd)/target/release/aranya-metrics $(pwd)/target/release/aranya-daemon
```

Makefile.toml
```diff
@@ -257,3 +257,23 @@ category = "examples"
...
+'''
+
+# Run `aranya-metrics` using prometheus
+[tasks.metrics-prometheus]
+category = "metrics"
+description = "Run Metrics Crate w/ a Push Gateway"
```
gknopf-aranya:
```suggestion
description = "Run aranya-metrics crate w/ Prometheus exporter and a Push Gateway"
```

crates/aranya-daemon/benches/metrics.rs
gknopf-aranya: @nikki-aranya I thought this feedback was included but it appears to have been reverted at some point. `Daemon::load(cfg)` loads a daemon from a config. `spawn()` needs to be called to run the daemon.

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,55 @@
...
+-   Augmenting existing metrics with more information using key-value pairs, as well as reporting metrics for individual daemons.
+-   Extending the above metrics harness with benchmarks using `divan` and [flame graphs](https://www.brendangregg.com/flamegraphs.html) to identify performance bottlenecks and functions that need to be annotated with `#[fastrace::trace]` for better measurement of hotspots.
+
+## Prometheus Exporter
+
+One of the supported metrics backends allows for communicating with a [Prometheus](https://prometheus.io/) instance. Note that with the way that Prometheus works, a datapoint is only created once the data is scraped, so any data created locally that gets overwritten before being scraped is lost. This can either be letting the main Prometheus backend scrape data from the exporter which is meant for longer running data, or using a `pushgateway` with a unique job name to easily filter out specific runs, which is meant more for short runs.
```
gknopf-aranya:
```suggestion
One of the supported metrics backends allows for communicating with a [Prometheus](https://prometheus.io/) instance. Note that with the way that Prometheus works, a datapoint is only created once the data is scraped, so any data created locally that gets overwritten before being scraped is lost. This can either be letting the main Prometheus backend scrape data from the exporter which is meant for longer running data, or using a `pushgateway` with a unique job name to easily filter out specific runs, which is meant for shorter runs.
```

crates/aranya-metrics/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Technical Details
+
+Due to `sysinfo` only providing percentages of CPU usage at each refresh instead of raw user time/system time, we use `proc_pidinfo` on MacOS to grab the raw CPU time and memory usage, using sysinfo for file I/O. `proc_pidinfo` captures information "at the current moment" which means there isn't really a delta, whereas sysinfo's bytes_read/written is "since last refresh" so we have to aggregate it along with all previous deltas.
+
+The PIDs being tracked include "client", which is an `aranya-client` instance (plus metrics overhead), as well as five `aranya-daemon` instances ("owner", "admin", "operator", "member_a", and "member_b"), which are created as sub-processes using `Command::spawn()` (`proc_pidinfo` can do weird stuff if you don't have permissions to access another process's info).
```
gknopf-aranya: Typically we would run 1 client per daemon instance rather than a single client that interacts with multiple daemons.

crates/aranya-metrics/metrics.toml
gknopf-aranya, line 1: Could call this file `example-metrics.toml` so it's clear that it is an example.

crates/aranya-metrics/README.md
nikki-aranya: Oh, I probably misunderstood what ClientCtx is doing then, I think you're right.

Cargo.toml
nikki-aranya: I talked about this with Dygert in DMs, aranya-core already does panic="abort", this just makes things consistent.

crates/aranya-metrics/README.md
gknopf-aranya: A client connects to a single daemon IPC via `connect()`. Each client has it's own AQC server running to support communication with peers via AQC channels.

crates/aranya-metrics/src/backend/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use anyhow::Result;
+
+#[cfg(feature = "datadog")]
+pub mod datadog;
+#[cfg(feature = "prometheus")]
+pub mod prometheus;
+#[cfg(feature = "tcp_server")]
+pub mod tcp_server;
```
gknopf-aranya: nit: I think these are all running a server. `tcp` is a backend just like `datadog` and `prometheus` so there's no need to specify that it's a server here.
```suggestion
#[cfg(feature = "tcp")]
pub mod tcp;
```

crates/aranya-metrics/src/backend/prometheus.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn format_pushgateway_url(endpoint: &str, job_name: &str) -> String {
+        // Try to parse the Uri, or else set it to our default.
+        let uri = endpoint
+            .parse::<http::Uri>()
+            .unwrap_or(http::Uri::from_static("http://localhost:9091/"));
+        let scheme = uri.scheme_str().unwrap_or("http");
+        let authority = uri
+            .authority()
+            .map(|a| a.as_str())
+            .unwrap_or("localhost:9091");
```
gknopf-aranya: Could define the default URL in one spot:
```suggestion
        let default_url = "localhost:9091";
        let uri = endpoint
            .parse::<http::Uri>()
            .unwrap_or(http::Uri::from_static(format!("http://{}/", default_url)));
        let scheme = uri.scheme_str().unwrap_or("http");
        let authority = uri
            .authority()
            .map(|a| a.as_str())
            .unwrap_or(default_url);
```

crates/aranya-metrics/src/backend/prometheus.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        Ok(())
+    }
+}
+
+impl Default for PrometheusConfig {
+    fn default() -> Self {
+        Self {
+            mode: PrometheusMode::PushGateway {
+                endpoint: SocketAddr::from(([127, 0, 0, 1], 9091)).to_string(),
```
gknopf-aranya: The default URL can be defined in one place then reused everywhere else

crates/aranya-metrics/src/main.rs
```diff
@@ -0,0 +1,465 @@
...
+            owner,
+            admin,
+            operator,
+            membera,
+            memberb,
+        },
+    ))
+}
+
+// TODO(nikki): make sure this is using sync_now
```
gknopf-aranya: Looks like this todo can be removed since sync_now is being used.

gknopf-aranya:
@nikki-aranya, @djohnson-spideroak, @jdygert-spok overall this PR looks good to me. Commented on a few minor changes. I'll approve once `cargo vet` is complete.

crates/aranya-metrics/prometheus.toml
```diff
@@ -0,0 +1,5 @@
+#interval = "1s"
+#job_name = "this_is_a_test"
```
djohnson-spideroak: rm

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,361 @@
...
+            total_disk_read_bytes: self.total_metrics.total_disk_read_bytes
+                + current.total_disk_read_bytes,
+            total_disk_write_bytes: self.total_metrics.total_disk_write_bytes
+                + current.total_disk_write_bytes,
+        };
+
+        match self.config.debug_logs {
+            DebugLogType::None => (),
+            _ => debug!("Total Metrics (cumulative): {:?}", self.total_metrics),
+        }
```
djohnson-spideroak: nit: we could just set the log level to info if we dont want debug

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,361 @@
...
+    /// Collects metrics for a specific process, using a number of syscalls.
+    #[cfg(not(target_os = "macos"))]
+    #[allow(dead_code)]
+    fn collect_process_metrics(
+        &self,
+        _pid: (&'static str, u32),
+        _metrics: &mut AggregatedMetrics,
+    ) -> Result<()> {
+        Err(anyhow!(
+            "We don't currently support {} for metrics, sorry!",
```
djohnson-spideroak, line 301: do we only support macos? or linux too?

djohnson-spideroak (commented):
LGTM pending Geoff's open comments.

crates/aranya-metrics/src/harness.rs
nikki-aranya: Linux support hasn't been implemented yet, no, just MacOS, I set up a VM but I haven't done any testing there or dug into how sysinfo does it yet

crates/aranya-daemon/benches/metrics.rs
nikki-aranya: We don't want to run the daemon, that's the point, we're specifically benchmarking how long it takes to create/initialize, see the comment I added to the start of the function as well as the function name itself.

run means it'll keep spinning and that's unhelpful for benchmarks because it both makes the measurement inaccurate and doesn't have an obvious "end", initializing has an obvious end. These aren't normal tests

gknopf-aranya: Startup suggests that this is measuring how long it takes for the daemon to startup. The following is not currently included in the measurement:
1. How long it takes to initialize the daemon IPC server
2. How long it takes to start up the QUIC sync server

Measuring how long it takes to load the daemon's config file is not a very useful measurement.

nikki-aranya: Have you looked at the daemon code recently, Dygert split basically all the initialization into load and run is just a loop now, load includes creating a graph and all the other stuff needed to actually run it, I don't see how that's a bad representation of the daemon startup.

Also, regardless of what it does, it doesn't change the fact that if I call run, I then have to choose a point to kill the task because otherwise it'll never exit which tells us nothing, I call run and sleep 500 milliseconds and then we have to mentally subtract 500 milliseconds every time we run benchmarks. Benchmarks are for well defined functions with a known start and end, and the daemon spinning is not that.

If it's that big a deal, then I'd rather just remove this entirely since if I add run, the entire measurement is meaningless.

gknopf-aranya (dismissed):

gknopf-aranya (approved):

crates/aranya-metrics/src/backend/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// increase network processing overhead.
+    ///
+    /// [`interval`]: MetricsConfig::interval
+    #[cfg(feature = "datadog")]
+    DataDog(datadog::DataDogConfig),
+
+    /// Configures a TCP server that listens for connections and streams metrics using [`protobuf`].
+    ///
+    /// [`protobuf`]: https://protobuf.dev/
+    #[cfg(feature = "tcp")]
```
djohnson-spideroak, line 117: we probably want to prefix the bench features with `bench-`

gknopf-aranya: Since the features are in the `aranya-metrics` crate, the context implies the features are metrics related.

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+        sync: SyncConfig {
+            quic: Toggle::Enabled(QuicSyncConfig {
+                addr: Addr::from((Ipv4Addr::LOCALHOST, 0)),
+            }),
+        },
+        aqc: Toggle::Enabled(AqcConfig {}),
+    };
+
+    let rt = Runtime::new()?;
+    let _daemon = rt.block_on(Daemon::load(cfg))?;
```
jdygert-spok: This is failing because the cfg'd directories don't exist. `divan` doesn't seem to handle `Result` return values specially. It just allows any return value to allow you to exclude its drop impl from the timing.

crates/aranya-metrics/src/main.rs
jdygert-spok, line 1: I know the idea is to unify this somehow. But as-is this would probably be better suited as a cargo example, i.e. in `crates/aranya-metrics/examples`. Then `aranya-metrics` could move many of its dependencies to dev-dependencies.

jdygert-spok: Or even as an additional crate. IDK. This can be done in #437. Point just being that it shouldn't really be here...

nikki-aranya: That would work, yeah, would also make it easier to add more examples once Geoff starts adding more to the rust stuff

crates/aranya-metrics/src/backend/mod.rs
nikki-aranya: I'll have to do that when I add those feature flags to the daemon but I think here it's fine since the entire point of the crate is for metrics

crates/aranya-metrics/prometheus.toml
```diff
@@ -0,0 +1,5 @@
...
+debug_logs = "PerProcess"
+
+[mode.Prometheus]
```
jdygert-spok: nit: `mode.Prometheus` doesn't make much sense when only one can be enabled, I recommend using `#[serde(tag = "type")]`. Also exporter might be a nicer name.

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+
+    /// Register a description for all the gauges, histograms, and counters we use.
+    fn register_metrics() {
+        // Accumulated values from across the run
+        describe_gauge!(
+            "cpu_user_time_microseconds_total",
+            "Total User CPU time consumed by all monitored processes in microseconds"
+        );
```
jdygert-spok, line 157: Is there a reason we are reporting the total instead of per-process with a label?

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,52 @@
...
+
+fn main() {
+    divan::main();
+}
+
+// NOTE: divan currently requires sync functions to work, so we spawn a runtime and block on async.
+/// Benchmarks how long the daemon takes to set itself up to be spawned.
+#[divan::bench]
+fn daemon_startup() {
+    let work_dir = tempfile::tempdir().unwrap().path().to_path_buf();
```
jdygert-spok: If you intend to consume the `TempDir` so the directory doesn't get deleted afterward, use `into_path`. Otherwise, make a separate variable for the `TempDir` so it gets dropped at the end of the function instead.

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,52 @@
...
+        &cfg.cache_dir,
+        &cfg.logs_dir,
+        &cfg.config_dir,
+    ] {
+        rt.block_on(aranya_util::create_dir_all(dir))
+            .expect("should be able to create directory");
+    }
+
+    let _daemon = rt.block_on(Daemon::load(cfg)).unwrap();
+}
```
ssweetney-spok: Do we want the creation and removal of directories in the benchmark? If not then maybe use something like:

```suggestion
#[divan::bench]
fn daemon_startup(bencher: divan::Bencher) {
    let temp_dir = tempfile::tempdir().unwrap();
    let work_dir = temp_dir.path().to_path_buf();
    let rt = Runtime::new().unwrap();

    let cfg = Config {
        name: "test-daemon-run".into(),
        runtime_dir: work_dir.join("run"),
        state_dir: work_dir.join("state"),
        cache_dir: work_dir.join("cache"),
        logs_dir: work_dir.join("logs"),
        config_dir: work_dir.join("config"),
        sync: SyncConfig {
            quic: Toggle::Enabled(QuicSyncConfig {
                addr: Addr::from((Ipv4Addr::LOCALHOST, 0)),
            }),
        },
        aqc: Toggle::Enabled(AqcConfig {}),
    };

    for dir in [
        &cfg.runtime_dir,
        &cfg.state_dir,
        &cfg.cache_dir,
        &cfg.logs_dir,
        &cfg.config_dir,
    ] {
        rt.block_on(aranya_util::create_dir_all(dir))
            .expect("should be able to create directory");
    }

    bencher.bench(move || {
        let _daemon = rt.block_on(Daemon::load(cfg)).unwrap();
    });

}
```

crates/aranya-metrics/src/harness.rs
nikki-aranya: I want to do both but I have yet to do the former

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,374 @@
...
+    /// All child process PIDs for tracking
+    pids: Vec<(&'static str, u32)>,
+    /// System struct for sysinfo fallback
+    system: System,
+    /// Collection start time for rate calculations
+    _start_time: Instant,
+    /// Total cumulative metrics for this run
+    total_metrics: ProcessMetrics,
+    /// Total cumulative metrics for individual processes
+    individual_metrics: HashMap<u32, ProcessMetrics>,
```
gknopf-aranya: Would be better to give the process IDs a more human-readable name in the code:
```suggestion
    individual_metrics: HashMap<Pid, ProcessMetrics>,
```
If there isn't an existing type from one of the metrics libraries, we could define an alias for it.

crates/aranya-metrics/src/harness.rs
```diff
@@ -0,0 +1,374 @@
...
+
+use crate::backend::MetricsConfig;
+
+/// Collector for process metrics, uses native APIs when possible.
+#[derive(Debug)]
+pub struct ProcessMetricsCollector {
+    /// Config options for the current run
+    config: MetricsConfig,
+    /// All child process PIDs for tracking
+    pids: Vec<(&'static str, u32)>,
```
gknopf-aranya:
```suggestion
    pids: Vec<(&'static str, Pid)>,
```

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -0,0 +1,64 @@
...
+
+// NOTE: divan currently requires sync functions to work, so we spawn a runtime and block on async.
+/// Benchmarks how long the daemon takes to set itself up to be spawned.
+#[divan::bench]
+fn daemon_startup(bencher: divan::Bencher<'_, '_>) {
+    bencher
+        .with_inputs(|| {
+            // Spawn a new dir that will live for one run, since we want to measure "cold boot" perf
+            let tmp_dir = tempfile::tempdir().expect("We should be able to create directories");
```
djohnson-spideroak, line 25: does this tempdir() get dropped at the end of this block?

nikki-aranya: I'm pretty sure it does but either way I don't know how much it matters. The intention is that this chain runs many times (in this case the default is 100 to get a good sampling) and each time, the Daemon gets a new/clean tempdir, runtime, and config, so we eliminate a lot of the run to run variance of those elements.

gknopf-aranya: A temp dir will be dropped when it goes out of scope. The directory on the filesystem will be deleted when it is dropped.

gknopf-aranya: We just need it to be held as long as the daemon is running.

crates/aranya-metrics/src/harness.rs
```diff
@@ -10,30 +10,66 @@
...
-use sysinfo::{Pid, ProcessRefreshKind, ProcessesToUpdate, System};
+use sysinfo::{ProcessRefreshKind, ProcessesToUpdate, System};
 use tracing::{debug, warn};
 
 use crate::backend::MetricsConfig;
 
+/// Helper type for PIDs
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[repr(transparent)]
+pub struct Pid((&'static str, sysinfo::Pid));
```
gknopf-aranya: This would be more readable as a struct with named fields.
Accessing values with `self.0.1` is harder to read.

crates/aranya-metrics/src/harness.rs
```diff
@@ -10,30 +10,66 @@
...
+
+impl Pid {
+    /// Constructs a Pid from a u32 and str.
+    pub fn from_u32(pid: u32, name: &'static str) -> Self {
```
gknopf-aranya, line 33: Are we guaranteed to always have a name that corresponds to a Pid? Wonder if the name field should be optional.

nikki-aranya: I'm gonna kick this down the road, if someone needs to make it optional then they can do that in a future PR but I don't want to add an abstraction just because right now.

crates/aranya-metrics/src/harness.rs
nikki-aranya: I couldn't find a good combination of named fields that didn't require me making them functions instead of pub, I'll play with it some more.

gknopf-aranya: `self.name` `self.pid` or something like that.

gknopf-aranya (approved):

djohnson-spideroak:
> @nikki-aranya, @jdygert-spok, @djohnson-spideroak should we take virtual memory out of the output until we figure out why the number reported is too high? `2025-08-08T22:33:39.869296Z DEBUG Total Metrics: User Time: 5.213ms, System Time: 2.701ms, Physical Memory: 117.94MiB, Virtual Memory: 392.33GiB, Disk Reads: 148.00KiB, Disk Writes: 3.20MiB`

I think so. We can just remove it from the output, no need to remove it.

djohnson-spideroak (approved):

PR #326, opened 2025-06-24
Move sync peer config map into Syncer

### LINKED ISSUE #233: (aranya-daemon) Move config map from SyncPeers to Syncer
See this [comment](https://github.com/aranya-project/aranya/pull/226#issuecomment-2828855733)

Sync configs should be saved after a call to `AddSyncPeer` but they aren't. This can be resolved by moving the config map into the Syncer struct

ssweetney-spok:
Resolves #233

Fixes the sync peer configs not being persisted


jdygert-spok (approved):

jdygert-spok:
Sorry for hijacking, noticed some related cleanup. Some was undoing the `&->&mut` which previously introduced a mutex and some was just general refactoring.

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -208,25 +174,17 @@ impl<ST> Syncer<ST> {
...
-            })
-            .or_insert(PeerInfo {
-                interval: cfg.interval,
-                key,
-            });
+    fn add_peer(&mut self, peer: SyncPeer, cfg: SyncPeerConfig) {
+        let new_key = self.queue.insert(peer.clone(), cfg.interval);
+        if let Some((_, old_key)) = self.peers.insert(peer, (cfg, new_key)) {
+            self.queue.remove(&old_key);
+        }
```
jdygert-spok, line 181: I believe this is equivalent and the right behavior. Would appreciate a second look. I don't recall having tests which overwrite existing config.

jdygert-spok (approved):

gknopf-aranya:
```
error[E0433]: failed to resolve: use of undeclared type `Mutex`
   --> crates/aranya-daemon/src/api.rs:105:21
    |
105 |             crypto: Mutex::new(crypto),
    |                     ^^^^^ use of undeclared type `Mutex`
```

crates/aranya-daemon/src/sync/task.rs
ssweetney-spok: Seems good to me. It's also more concise

jdygert-spok (approved):

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -67,76 +73,43 @@ pub(crate) enum SyncResponse {
...
-            .context("unable to add sync now peer")
-        {
-            error!(?e, "error adding sync now peer to syncer");
-            return Err(e);
-        }
-        Ok(())
+        cfg: Option<SyncPeerConfig>,
+    ) -> Result<(), Bug> {
+        let peer = SyncPeer { addr, graph_id };
+        self.send(Msg::SyncNow { peer, cfg }).await
```
gknopf-aranya, line 112: This PR removed some of the error context we used to have. Let's add it back unless there's a good reason for removing it.

ssweetney-spok: I'm not 100% sure but I think the goal was to reduce redundancy and move the "error context" to `SyncPeers::send`.

jdygert-spok: It's fine, there's no need to distinguish between the three operations, it can only fail if the syncer is dropped. I also plan to log response errors higher up in the API.

gknopf-aranya (approved):

PR #356, opened 2025-06-30
capi: fix potentially invalid borrow expansion

### LINKED ISSUE #204: c-api: fix potentially invalid borrow expansion
Our current way of creating slice from a reference in `aranya-client-capi` are questionable, violating the experimental stacked borrow model. 

The TLDR is you can't "expand" from `&arr[0]` to `&arr[..]` 

Run "Tools > Miri" on this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ba02c0b88584f107d83308158117e60a

> this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
> see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information

We can just change the rust function signatures to use pointers directly. (And make sure the capi tooling itself doesn't rely on this anywhere).

cc: @elagergren-spideroak @gknopf-aranya

jdygert-spok:
Closes #204.

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -163,9 +163,9 @@ pub type ExtError = Safe<imp::ExtError>;
 /// @param msg_len length of the message buffer.
 ///
 /// @relates AranyaExtError.
-pub fn ext_error_msg(
+pub unsafe fn ext_error_msg(
     err: &ExtError,
-    msg: &mut MaybeUninit<c_char>,
+    msg: *mut MaybeUninit<c_char>,
     msg_len: &mut usize,
 ) -> Result<(), imp::Error> {
     let msg = aranya_capi_core::try_as_mut_slice!(msg, *msg_len);
@@ -536,9 +536,9 @@ pub const ARANYA_ID_STR_LEN: usize = (ARANYA_ID_LEN * 1375) / 1000 + 1;
 ///
 /// @relates AranyaId.
 #[aranya_capi_core::no_ext_error]
-pub fn id_to_str(
+pub unsafe fn id_to_str(
     id: &Id,
-    str: &mut MaybeUninit<c_char>,
+    str: *mut MaybeUninit<c_char>,
     str_len: &mut usize,
 ) -> Result<(), imp::Error> {
     let str = aranya_capi_core::try_as_mut_slice!(str, *str_len);
@@ -1276,10 +1276,10 @@ pub unsafe fn sync_now(
 /// @param devices_len returns the length of the devices list [`DeviceId`].
 ///
 /// @relates AranyaClient.
-pub fn query_devices_on_team(
+pub unsafe fn query_devices_on_team(
     client: &mut Client,
     team: &TeamId,
-    devices: Option<&mut MaybeUninit<DeviceId>>,
+    devices: *mut MaybeUninit<DeviceId>,
     devices_len: &mut usize,
 ) -> Result<(), imp::Error> {
     let client = client.imp();
@@ -1287,10 +1287,6 @@ pub fn query_devices_on_team(
         .rt
         .block_on(client.inner.team(team.into()).queries().devices_on_team())?;
     let data = data.__data();
-    let Some(devices) = devices else {
-        *devices_len = data.len();
-        return Err(imp::Error::BufferTooSmall);
-    };
     let out = aranya_capi_core::try_as_mut_slice!(devices, *devices_len);
     if *devices_len < data.len() {
         *devices_len = data.len();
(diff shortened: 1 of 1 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1276,21 +1276,17 @@ pub unsafe fn sync_now(
...
-    devices: Option<&mut MaybeUninit<DeviceId>>,
+    devices: *mut MaybeUninit<DeviceId>,
     devices_len: &mut usize,
 ) -> Result<(), imp::Error> {
     let client = client.imp();
     let data = client
         .rt
         .block_on(client.inner.team(team.into()).queries().devices_on_team())?;
     let data = data.__data();
-    let Some(devices) = devices else {
```
gknopf-aranya, line 1277: If this is removed, the caller won't have a way to reallocate the buffer to the correct size

gknopf-aranya: This change would also impact code in `example.c` that performs a reallocation after receiving a `BufferTooSmall` error.

jdygert-spok: Look below, it still checks the length. And if the ptr is null the length must be zero or youll get an invalid arg error.

gknopf-aranya: Should the example be updated to perform reallocation on the invalid arg error instead of the buffer too small error?

jdygert-spok: No, the invalid arg is if you pass `(nullptr, 42)` which is bad behavior.

But we could change them to pass `(nullptr, 0)`

gknopf-aranya: If the pointer is null, we still want to set the length so the caller knows what size to reallocate the buffer to:
`*devices_len = data.len();`

jdygert-spok: Yes, if the ptr is null and len is 0, try_as_mut_slice will give an empty slice, so the length will be too small and it will write out the needed length.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1349,11 +1345,11 @@ pub unsafe fn query_device_keybundle(
 /// @param labels_len returns the length of the labels list [`LabelId`].
 ///
 /// @relates AranyaClient.
-pub fn query_device_label_assignments(
+pub unsafe fn query_device_label_assignments(
```
gknopf-aranya, line 1333: Docs haven't been updated to mention the `NULL` ptr case

jdygert-spok: It was nullable before too

gknopf-aranya: Oh, we probably should've have been documenting that case before.

jdygert-spok: It's the same for every method with the outbuf pattern. Any suggestion for wording? I'd rather not make the docs too verbose, maybe we can just mention it in one place at the start.

gknopf-aranya: Something like this?
For output buffers following this pattern:
```
buf: *mut MaybeUninit<type>,
buf_len: &mut usize,
 ```
1. If `buf` is NULL, and `buf_len` is 0, ...
2. If `buf` is NULL and `buf_len` != 0, ...
3. If `buf` is not NULL, ...

It's probably okay to document this in one place at the top of the header file and refer to it from the rest of the functions using the pattern.

Comment (jdygert-spok) [reply crates/aranya-client-capi/src/api/defs.rs:1333]: Let's do it on a followup. Docs are prone to bikeshedding and not sure how best to do it so doxygen shows it well. #367.

File: crates/aranya-client-capi/src/api/defs.rs [hunk 3]
```diff
@@ -162,9 +162,9 @@ pub type ExtError = Safe<imp::ExtError>;
 /// @param[in,out] msg_len length of the message buffer.
 ///
 /// @relates AranyaExtError.
-pub fn ext_error_msg(
+pub unsafe fn ext_error_msg(
```
djohnson-spideroak, line 165: what are the implications of making these functions unsafe?

djohnson-spideroak (approved):

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: They are unsafe because they now take pointers, meaning it is possible to pass in bad pointers and cause segfault/UB. This is always possible from C anyways, and we never call them from Rust so not very significant but this is just a bit safer.

PR #357, opened 2025-06-30
Initialize a team config from existing data

### LINKED ISSUE #341: Initialize an AddTeamConfigBuilder from existing data
Improve QUIC syncer C API from a usability, naming, and backward compatibility standpoint.

Here is some discussion related to it on the QUIC syncer C API PR:
https://github.com/aranya-project/aranya/pull/318#discussion_r2169392631

cc: @jdygert-spok, @djohnson-spideroak, @elagergren-spideroak

Edit:

There is data that must be "correct" in order to join a team correctly. That data will be created as a type that is versioned for backwards compatability.

The `TeamConfigBuilder`s in "aranya-client" and "aranya-client-capi" should be able to be initialized from that data type.

ssweetney-spok:
Resolves #341

### DIFF
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -364,6 +364,13 @@ typedef struct ARANYA_ALIGNED(8) AranyaTeamConfigBuilder {
     uint8_t __for_size_only[288];
 } AranyaTeamConfigBuilder;
 
+/**
+ * Team ID.
+ */
+typedef struct AranyaTeamId {
+    struct AranyaId id;
+} AranyaTeamId;
+
 typedef struct ARANYA_ALIGNED(8) AranyaTeamConfig {
     /**
      * This field only exists for size purposes. It is
@@ -401,13 +408,6 @@ typedef struct ARANYA_ALIGNED(8) AranyaSyncPeerConfig {
  * A type to represent a span of time in nanoseconds.
  */
 typedef uint64_t AranyaDuration;
-
-/**
- * Team ID.
- */
-typedef struct AranyaTeamId {
-    struct AranyaId id;
-} AranyaTeamId;
 
 /**
  * An AQC label name.
@@ -1145,11 +1145,91 @@ AranyaError aranya_team_config_builder_cleanup_ext(struct AranyaTeamConfigBuilde
                                                    struct AranyaExtError *__ext_err);
 
 /**
+ * Serializes team information into a byte buffer.
+ *
+ * Creates a TeamInfo V1 struct from the wrapped seed and team ID,
+ * then serializes it.
+ *
+ * Returns an `AranyaBufferTooSmall` error if the output buffer is too small to hold the serialized data.
+ * Writes the number of bytes that would have been returned to `buf_len`.
+ * The application can use `buf_len` to allocate a larger buffer.
+ *
+ * @param[in] wrapped_seed the serialized representation of a wrapped seed.
+ * @param[out] team_id the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
+ * @param[out] buf buffer to write the serialized team info to.
+ * @param[out] buf_len the size of the buffer, updated with the actual bytes written.
+ */
+AranyaError aranya_to_team_info_v1_bytes(const uint8_t *wrapped_seed,
+                                         size_t wrapped_seed_len,
+                                         const struct AranyaTeamId *team_id,
+                                         uint8_t *buf,
+                                         size_t *buf_len);
+
+/**
+ * Serializes team information into a byte buffer.
+ *
+ * Creates a TeamInfo V1 struct from the wrapped seed and team ID,
+ * then serializes it.
+ *
+ * Returns an `AranyaBufferTooSmall` error if the output buffer is too small to hold the serialized data.
+ * Writes the number of bytes that would have been returned to `buf_len`.
+ * The application can use `buf_len` to allocate a larger buffer.
+ *
+ * @param[in] wrapped_seed the serialized representation of a wrapped seed.
+ * @param[out] team_id the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
+ * @param[out] buf buffer to write the serialized team info to.
+ * @param[out] buf_len the size of the buffer, updated with the actual bytes written.
+ */
+AranyaError aranya_to_team_info_v1_bytes_ext(const uint8_t *wrapped_seed,
+                                             size_t wrapped_seed_len,
+                                             const struct AranyaTeamId *team_id,
+                                             uint8_t *buf,
+                                             size_t *buf_len,
+                                             struct AranyaExtError *__ext_err);
+
+/**
+ * Initializes team and QUIC sync config builders from serialized team info.
+ *
+ * Deserializes team information data and creates initialized builders that can be used
+ * to build a team config.
+ *
+ * @param[in] team_info_bytes serialized team information bytes (created by [`aranya_to_team_info_v1_bytes`](@ref aranya_to_team_info_v1_bytes)).
+ * @param[out] team_cfg_builder the uninitialized team config builder [`AranyaTeamConfigBuilder`](@ref AranyaTeamConfigBuilder).
+ * @param[out] qs_cfg_builder the uninitialized QUIC sync config builder [`AranyaQuicSyncConfigBuilder`](@ref AranyaQuicSyncConfigBuilder).
+ * @param[out] team_id returns the team's ID [`AranyaTeamId`](@ref AranyaTeamId) extracted from the team info.
+ *
+ * @relates AranyaTeamConfigBuilder.
+ */
+AranyaError aranya_team_config_builder_init_from_team_info_v1_bytes(const uint8_t *team_info_bytes,
+                                                                    size_t team_info_bytes_len,
+                                                                    struct AranyaTeamConfigBuilder *team_cfg_builder,
+                                                                    struct AranyaQuicSyncConfigBuilder *qs_cfg_builder,
+                                                                    struct AranyaTeamId *team_id);
+
+/**
+ * Initializes team and QUIC sync config builders from serialized team info.
+ *
+ * Deserializes team information data and creates initialized builders that can be used
+ * to build a team config.
+ *
+ * @param[in] team_info_bytes serialized team information bytes (created by [`aranya_to_team_info_v1_bytes`](@ref aranya_to_team_info_v1_bytes)).
+ * @param[out] team_cfg_builder the uninitialized team config builder [`AranyaTeamConfigBuilder`](@ref AranyaTeamConfigBuilder).
+ * @param[out] qs_cfg_builder the uninitialized QUIC sync config builder [`AranyaQuicSyncConfigBuilder`](@ref AranyaQuicSyncConfigBuilder).
+ * @param[out] team_id returns the team's ID [`AranyaTeamId`](@ref AranyaTeamId) extracted from the team info.
+ *
+ * @relates AranyaTeamConfigBuilder.
+ */
+AranyaError aranya_team_config_builder_init_from_team_info_v1_bytes_ext(const uint8_t *team_info_bytes,
+                                                                        size_t team_info_bytes_len,
+                                                                        struct AranyaTeamConfigBuilder *team_cfg_builder,
+                                                                        struct AranyaQuicSyncConfigBuilder *qs_cfg_builder,
+                                                                        struct AranyaTeamId *team_id,
+                                                                        struct AranyaExtError *__ext_err);
+
+/**
  * Configures QUIC syncer for [`AranyaTeamConfigBuilder`](@ref AranyaTeamConfigBuilder).
  *
- * By default, the QUIC syncer config is not set. It is an error to call
- * [`aranya_team_config_build`](@ref aranya_team_config_build) before setting the interval with
- * this function
+ * By default, the QUIC syncer config is not set.
  *
  * @param cfg a pointer to the builder for a team config
  * @param quic set the QUIC syncer config
@@ -1160,9 +1240,7 @@ AranyaError aranya_team_config_builder_set_quic_syncer(struct AranyaTeamConfigBu
 /**
  * Configures QUIC syncer for [`AranyaTeamConfigBuilder`](@ref AranyaTeamConfigBuilder).
  *
- * By default, the QUIC syncer config is not set. It is an error to call
- * [`aranya_team_config_build`](@ref aranya_team_config_build) before setting the interval with
- * this function
+ * By default, the QUIC syncer config is not set.
  *
  * @param cfg a pointer to the builder for a team config
  * @param quic set the QUIC syncer config
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -7,7 +7,10 @@ use std::{ffi::OsStr, ops::Deref, os::unix::ffi::OsStrExt, str::FromStr};
 
 use anyhow::Context as _;
 use aranya_capi_core::{opaque::Opaque, prelude::*, ErrorCode, InvalidArg};
-use aranya_client::aqc::{self, AqcPeerStream};
+use aranya_client::{
+    aqc::{self, AqcPeerStream},
+    config::TeamInfo,
+};
 use aranya_crypto::dangerous::spideroak_crypto::hex;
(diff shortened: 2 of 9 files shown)


crates/aranya-client/src/config.rs
```diff
@@ -146,6 +159,21 @@ impl QuicSyncConfigBuilder {
     }
 }
 
+/// Represents team configuration information in various versions.
+///
+/// This enum contains versioned team data that can be serialized and used to
+/// reconstruct team configs on different devices.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[non_exhaustive]
+pub enum TeamInfo {
```
ssweetney-spok, line 168: Postcard doesn't support this `serde(tag = "version")`.

Not a big deal but that might've been useful for users that wanted to serialize to some format like JSON

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -752,11 +755,78 @@ pub type TeamConfig = Safe<imp::TeamConfig>;
...
+) -> Result<(), imp::Error> {
+    let team_info = postcard::from_bytes(team_info_bytes)?;
+    let (team_builder, qs_builder, id) = imp::TeamConfigBuilder::from_team_info_v1(team_info)?;
+
+    TeamConfigBuilder::init(team_cfg_builder, team_builder);
+    QuicSyncConfigBuilder::init(qs_cfg_builder, qs_builder);
+    *team_id = id;
+
+    Ok(())
+}
```
ssweetney-spok, line 825: I wanted the C API to be similar to the Rust API in how you can call `TeamConfigBuilder::quic_sync` to get a reference to the inner `QuicSyncConfigBuilder` but that seems challenging since `Safe<T>` and `T` have different layouts.

So having a single function that initialized both builders seemed like the least awkward option.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -349,6 +350,26 @@ impl Typed for TeamConfigBuilder {
...
+            let mut qs_builder = QuicSyncConfig::builder();
+            qs_builder.wrapped_seed(&wrapped_seed)?;
+
+            team_cfg_builder.quic(qs_builder.clone().build()?); // Remove?
+
+            (team_cfg_builder, qs_builder)
+        };
+
+        Ok((team_builder, qs_builder, id.into()))
+    }
```
ssweetney-spok, line 371: I'm thinking that this line: `team_cfg_builder.quic(qs_builder.clone().build()?);` should be removed. 

I added it originally because I thought it might be convienient for users to call `aranya_team_config_builder_init_from_team_info_v1_bytes` and then call `aranya_team_config_build` without having to call `aranya_team_config_builder_set_quic_syncer`. Now I think it may be confusing since it's not clear that the returned `TeamConfigBuilder` is partially initialized with a quic sync config.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1206,12 +1206,92 @@ AranyaError aranya_team_config_builder_cleanup(struct AranyaTeamConfigBuilder *p
...
+ * Returns an `AranyaBufferTooSmall` error if the output buffer is too small to hold the serialized data.
+ * Writes the number of bytes that would have been returned to `buf_len`.
+ * The application can use `buf_len` to allocate a larger buffer.
+ *
+ * @param[in] wrapped_seed the serialized representation of a wrapped seed.
+ * @param[out] team_id the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
+ * @param[out] buf buffer to write the serialized team info to.
+ * @param[out] buf_len the size of the buffer, updated with the actual bytes written.
+ */
+AranyaError aranya_to_team_info_v1_bytes(const uint8_t *wrapped_seed,
```
gknopf-aranya, line 1224: This method won't be backward compatible if more fields are added to the `TeamInfo` struct.
Wondering if we should use a builder pattern to construct `TeamInfo` so it will be backward compatible when fields are changed.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1206,12 +1206,92 @@ AranyaError aranya_team_config_builder_cleanup(struct AranyaTeamConfigBuilder *p
...
+ * to build a team config.
+ *
+ * @param[in] team_info_bytes serialized team information bytes (created by [`aranya_to_team_info_v1_bytes`](@ref aranya_to_team_info_v1_bytes)).
+ * @param[out] team_cfg_builder the uninitialized team config builder [`AranyaTeamConfigBuilder`](@ref AranyaTeamConfigBuilder).
+ * @param[out] qs_cfg_builder the uninitialized QUIC sync config builder [`AranyaQuicSyncConfigBuilder`](@ref AranyaQuicSyncConfigBuilder).
+ * @param[out] team_id returns the team's ID [`AranyaTeamId`](@ref AranyaTeamId) extracted from the team info.
+ *
+ * @relates AranyaTeamConfigBuilder.
+ */
+AranyaError aranya_team_config_builder_init_from_team_info_v1_bytes(const uint8_t *team_info_bytes,
```
gknopf-aranya, line 1265: `serde` can determine the version of the `TeamInfo` struct at runtime when it is deserialized. So we shouldn't need to tie this method to a particular version of the `TeamInfo` struct.
https://ivanbyte.medium.com/backward-compatible-data-de-serialization-with-serde-flow-in-rust-c87a2e8bc9ea

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: Yea, I didn't think it was worth exposing another type for the C API which is why you initialize the builders directly from raw bytes. My thinking was that you'd do something like `aranya_to_team_info_v*_bytes` for each version

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -752,11 +755,78 @@ pub type TeamConfig = Safe<imp::TeamConfig>;
...
+///
+/// @relates AranyaTeamConfigBuilder.
+pub fn team_config_builder_init_from_team_info_v1_bytes(
+    team_info_bytes: &[u8],
+    team_cfg_builder: &mut MaybeUninit<TeamConfigBuilder>,
+    qs_cfg_builder: &mut MaybeUninit<QuicSyncConfigBuilder>,
+    team_id: &mut TeamId,
+) -> Result<(), imp::Error> {
+    let team_info = postcard::from_bytes(team_info_bytes)?;
+    let (team_builder, qs_builder, id) = imp::TeamConfigBuilder::from_team_info_v1(team_info)?;
```
gknopf-aranya, line 818: I would implement separate getter methods for each field of `TeamInfo`. That way, fetching fields can be backward compatible with older types.

crates/aranya-client/src/config.rs
```diff
@@ -102,16 +108,23 @@ pub struct QuicSyncConfigBuilder {
 impl QuicSyncConfigBuilder {
     /// Sets the PSK seed mode.
     #[doc(hidden)]
-    pub fn mode(mut self, mode: SeedMode) -> Self {
+    pub fn mode(&mut self, mode: SeedMode) -> &mut Self {
```
gknopf-aranya, line 111: Builder methods are supposed to return a new copy of the object with the change rather than mutating the original object.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -247,14 +248,13 @@ impl Typed for QuicSyncConfig {
 
 impl From<QuicSyncConfig> for aranya_client::QuicSyncConfig {
     fn from(value: QuicSyncConfig) -> Self {
-        Self::builder()
-            .mode(value.mode)
-            .build()
-            .expect("All fields are set")
+        let mut builder = Self::builder();
+        builder.mode(value.mode);
+        builder.build().expect("All fields are set")
```
gknopf-aranya, line 253: I think this change should be reverted

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: That's a good point. I'm wondering how things will work if future versions have different fields. Where do we document which fields of the `TeamConfigBuilder`, `QuicSyncConfigBuilder` (and future builders) are intialized?

crates/aranya-client/src/config.rs
ssweetney-spok: This change was made so that you can mutate the `QuicSyncConfigBuilder` from a `TeamConfigBuilder`.

FWIW, this pattern is also present in the Rust standard [library](https://doc.rust-lang.org/stable/std/fs/struct.OpenOptions.html)

jdygert-spok: I've considered this change previously, it would also let us define the C builders in terms of the rust builders which would be nice.

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: Adding new methods to the API for each version seems cumbersome.
I'd rather upgrade from the `v1` to `v2` type internally and only require the user to initialize new fields on the type if they choose to.
We can use postcard to deserialize the type and match to any version of the type inside the API without the user needing to know which version of the type they are using. The getter/setter methods could return an error if a field is not valid for a certain version of the type.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Can't we have a `Safe<QuicsyncConfigBuilder>` as a field of the `TeamConfigBuilder`?

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: That would probably be documented in the type conversions between `TeamInfo` and `TeamConfigBuilder`

crates/aranya-client/src/config.rs
jdygert-spok: > define the C builders in terms of the rust builders

Except not if `Safe<T>` wrapping is needed, and some other types might not line up :shrug:

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: I like the idea of defining a `From` impl for each type conversion to make the conversions very well-defined.
Rather than hoping the right builder methods are called.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: Yea, that works if I remove the derived `Clone` impl from `TeamConfigBuilder`. It felt kind of odd for the inner type to be be wrapped with Safe but not the type itself.

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: I don't think the docs for those `From` impls show up in the header file.

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: Wouldn't we need to add new ~~fields~~ methods to the API for each version anyway? If there's a new field then there's needs to be a new method to initialize it (assuming there's a TeamInfo builder)

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: We could document which versions of the type are valid for the getter/setter methods on the builder.

ssweetney-spok: Also, I intentionally did not write a `From` impl for `TeamInfo` -> `TeamConfigBuilder` since the conversion isn't [value preserving](https://doc.rust-lang.org/stable/std/convert/trait.From.html#when-to-implement-from), i.e. the team info means something different then the team config builder.

crates/aranya-client-capi/output/aranya-client.h
jdygert-spok: Is there a reason we need separate "team info" and can't just use the existing config/builder? e.g. you serialize the builder (or maybe config), and then the other user deserializes into the builder.

ssweetney-spok: I thought the benefit of separating it was so that it could be "versioned". We could de-/serialize the existing TeamConfigBuilder but then it may become less clear which fields are actually initialized since the builders are opaque

gknopf-aranya: > Is there a reason we need separate "team info" and can't just use the existing config/builder? e.g. you serialize the builder (or maybe config), and then the other user deserializes into the builder.

One reason I can think of is if the team config of a non-owner device is supposed to be configured differently than the owner device.
We could work around this by calling methods on the builder after it is received and deserialized from a peer. But, that would mean the order of operations could result in a different team config.

gknopf-aranya: > I thought the benefit of separating it was so that it could be "versioned". We could de-/serialize the existing TeamConfigBuilder but then it may become less clear which fields are actually initialized since the builders are opaque

The purpose of versioning the type is backward compatibility. If we can achieve backward compatibility a different way, the version may not be needed.

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: If the `From` impl doesn't always succeed, we could use `TryFrom` instead. Although I don't see a reason why the conversion would fail if all the methods on the team config builder are called optionally.

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: > Wouldn't we need to add new ~fields~ methods to the API for each version anyway? If there's a new field then there's needs to be a new method to initialize it (assuming there's a TeamInfo builder)

For the most part, yes. A new version of a type could also occur if a field was removed, renamed, made optional, etc.
I think adding a method like `team_info_set_new_field()` is more meaningful to the reader than `team_info_init_v2()` because it tells the reader something about which field was added.

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: I'm not sure what you mean? It's not about the whether the conversion is fallible but about what the values represent

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: If the reader is the one creating the `TeamInfo` type then yea I agree. The builder approach for `TeamInfo` makes it harder for the user receiving the data to actually know which fields get initialized on the `TeamConfigBuilder`. They'd probably have to check at runtime?

With versioned APIs, users on the receiving end only have to look in a single place (the docs for the initialization function) to know which fields on the `TeamConfigBuilder` were set

ssweetney-spok: I'm also realizing that the versioning approach is inflexible with respect to new syncer types. V1 has data that's specific to the QUIC syncer but we may need similar data for another syncer type and using V2 for that doesn't seem right since it's not really an "upgrade".

With that in mind, maybe we should just serialize the existing `TeamConfigBuilder` because there'd probably be a bunch of duplication between `TeamInfoBuilder` and `TeamConfigBuilder` otherwise 🤔.

Edit:

Then that means the IKM for the seed could be visible if the quic sync config was set to the "IKM" mode...

Edit 2: We could get around that with a custom implementation of `Serialize` that fails if the "IKM" mode is set

crates/aranya-client-capi/src/imp/config.rs
ssweetney-spok: https://github.com/aranya-project/aranya/pull/357#discussion_r2175940770

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: I would make the quic syncer config and PSK optional. That way if we decide to use a different syncer or certs in the future, it will be backward compatible.

gknopf-aranya:
@ssweetney-spok let's close this PR for now since it's blocked by:
https://github.com/aranya-project/aranya/pull/368
There will likely be a lot of merge conflicts to resolve by the time that's merged, and it may be easier to open a new PR by the time this is ready to be picked up again.

gknopf-aranya:
@ssweetney-spok now that this work has been unblocked, are you planning to reuse this PR or open a new one?

PR #364, opened 2025-07-01
psk ikm: don't log secret bytes

gknopf-aranya:
Implement custom `Debug` for `SeedMode`'s `Ikm` that doesn't show the secret bytes.
This prevents the plaintext IKM from being printed by the logger.

Resolve same issue for `Secret` type.

### DIFF
--- a/crates/aranya-daemon-api/src/service.rs
+++ b/crates/aranya-daemon-api/src/service.rs
@@ -181,7 +181,7 @@ impl fmt::Display for NetIdentifier {
 pub type AqcCtrl = Vec<Box<[u8]>>;
 
 /// A secret.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize, Hash)]
 pub struct Secret(Box<[u8]>);
 
 impl Secret {
--- a/crates/aranya-daemon-api/src/service.rs
+++ b/crates/aranya-daemon-api/src/service.rs
@@ -180,8 +180,48 @@ impl fmt::Display for NetIdentifier {
 /// A serialized command for AQC.
 pub type AqcCtrl = Vec<Box<[u8]>>;
 
+/// A PSK IKM.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Ikm([u8; 32]);
+
+impl Ikm {
+    /// Provides access to the raw IKM bytes.
+    #[inline]
+    pub fn raw_ikm_bytes(&self) -> &[u8] {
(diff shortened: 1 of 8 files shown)


crates/aranya-daemon-api/src/service/quic_sync.rs
```diff
@@ -45,6 +51,21 @@ impl Default for SeedMode {
     }
 }
 
+impl fmt::Debug for SeedMode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            SeedMode::Generate => f.debug_struct("SeedMode::Generate").finish(),
+            SeedMode::IKM(ikm) => {
+                // TODO: use aranya-crypto hasher.
+                let mut hasher = DefaultHasher::new();
```
gknopf-aranya: Use a hasher from one of our crypto libraries instead

crates/aranya-daemon/src/sync/task/quic/psk.rs
```diff
@@ -36,9 +36,14 @@ impl PskSeed {
         Self(aranya_crypto::tls::PskSeed::new(rng, &group))
     }
 
-    pub(crate) fn import_from_ikm(ikm: &[u8; SEED_IKM_SIZE], team: TeamId) -> Self {
+    pub(crate) fn import_from_ikm(ikm: &Ikm, team: TeamId) -> Result<Self> {
         let group = GroupId::from(team.into_id());
-        Self(aranya_crypto::tls::PskSeed::import_from_ikm(ikm, &group))
+        Ok(Self(aranya_crypto::tls::PskSeed::import_from_ikm(
```
gknopf-aranya: Revert

crates/aranya-daemon-api/src/service.rs
```diff
@@ -180,8 +180,48 @@ impl fmt::Display for NetIdentifier {
...
+    /// Provides access to the raw IKM bytes.
+    #[inline]
+    pub fn raw_ikm_bytes(&self) -> &[u8; SEED_IKM_SIZE] {
+        &self.0
+    }
+}
+
+impl<T> From<T> for Ikm
+where
+    T: Into<[u8; 32]>,
```
jdygert-spok: Just do `From<[u8; 32]>`, better to be more explicit to prevent accidental conversions.

gknopf-aranya:
Merged. @elagergren-spideroak let me know if this is important enough for a patch release. Otherwise, it will be included with the upcoming MVP release.

PR #366, opened 2025-07-01
capi: consistent length params

gknopf-aranya:
Use consistent length output params in C API.
Return the length written from the same param that the buffer length is provided with rather than a separate param.

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -129,7 +129,7 @@ impl From<&InvalidArg<'static>> for Error {
 ///
 /// The resulting pointer must NOT be freed.
 ///
-/// @param err `u32` error code from `AranyaError`.
+/// @param[in] err `u32` error code from `AranyaError`.
 ///
 /// @relates AranyaError.
 #[aranya_capi_core::no_ext_error]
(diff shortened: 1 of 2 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1840,24 +1824,34 @@ pub fn aqc_bidi_stream_send(
 ///
 /// @param[in]  stream the receiving side of a stream [`AqcBidiStream`].
 /// @param[out] buffer pointer to the target buffer.
-/// @param[in] buffer_len length of the target buffer.
-/// @param[out] __output the number of bytes written to the buffer.
+/// @param[out] buffer_len length of the target buffer.
```
jdygert-spok: `in,out`, ditto on other

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1986,22 +1980,32 @@ pub fn aqc_send_stream_send(
...
+    buffer: *mut MaybeUninit<u8>,
+    buffer_len: &mut usize,
+) -> Result<(), imp::Error> {
+    if *buffer_len == 0 {
+        return Err(InvalidArg::new(
+            "stream",
+            "Tried to call aqc_recv_stream_try_recv with an empty buffer",
+        )
+        .into());
+    }
```
```suggestion
    if buffer.is_null() || *buffer_len == 0 {
        return Err(InvalidArg::new(
            "buffer",
            "Tried to call aqc_recv_stream_try_recv with an empty buffer",
        )
        .into());
    }
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1840,24 +1824,34 @@ pub fn aqc_bidi_stream_send(
...
+    buffer: *mut MaybeUninit<u8>,
+    buffer_len: &mut usize,
+) -> Result<(), imp::Error> {
+    if *buffer_len == 0 {
+        return Err(InvalidArg::new(
+            "stream",
+            "Tried to call aqc_bidi_stream_try_recv with an empty buffer",
+        )
+        .into());
+    }
```
```suggestion
    if buffer.is_null() || *buffer_len == 0 {
        return Err(InvalidArg::new(
            "buffer",
            "Tried to call aqc_bidi_stream_try_recv with an empty buffer",
        )
        .into());
    }
```

PR #368, opened 2025-07-01
Split TeamConfig and TeamConfigBuilder into separate types

### LINKED ISSUE #342: aranya-client: move `Queries` object under `Team`
It's more convenient when you can hold onto the team object, but currently you have to drop it if you want to do queries.

ssweetney-spok:
Resolves #342

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -673,7 +673,7 @@ pub type QuicSyncConfig = Safe<imp::QuicSyncConfig>;
 
 /// A builder for initializing a [`QuicSyncConfig`].
 ///
-/// The [`QuicSyncConfig`] is an optional part of initializing a [`TeamConfig`].
+/// The [`QuicSyncConfig`] is an optional part of initializing an [`AddTeamConfig`] or [`CreateTeamConfig`].
 #[aranya_capi_core::derive(Init, Cleanup)]
 #[aranya_capi_core::opaque(size = 288, align = 8)]
 pub type QuicSyncConfigBuilder = Safe<imp::QuicSyncConfigBuilder>;
@@ -741,29 +741,38 @@ pub fn quic_sync_config_build(
     Ok(())
 }
 
-/// Team configuration.
+/// Team configuration used when joining a team.
 ///
-/// Use a [`TeamConfigBuilder`] to construct this object.
-#[aranya_capi_core::opaque(size = 288, align = 8)]
-pub type TeamConfig = Safe<imp::TeamConfig>;
+/// Use an [`AddTeamConfigBuilder`] to construct this object.
+#[aranya_capi_core::opaque(size = 320, align = 8)]
+pub type AddTeamConfig = Safe<imp::AddTeamConfig>;
 
-/// A builder for initializing a [`TeamConfig`].
+/// A builder for initializing an [`AddTeamConfig`].
+#[aranya_capi_core::derive(Init, Cleanup)]
+#[aranya_capi_core::opaque(size = 328, align = 8)]
+pub type AddTeamConfigBuilder = Safe<imp::AddTeamConfigBuilder>;
+
+/// Team configuration used when creating a team.
+///
+/// Use an [`CreateTeamConfigBuilder`] to construct this object.
+#[aranya_capi_core::opaque(size = 288, align = 8)]
+pub type CreateTeamConfig = Safe<imp::CreateTeamConfig>;
+
+/// A builder for initializing a [`CreateTeamConfig`].
 #[aranya_capi_core::derive(Init, Cleanup)]
 #[aranya_capi_core::opaque(size = 288, align = 8)]
-pub type TeamConfigBuilder = Safe<imp::TeamConfigBuilder>;
+pub type CreateTeamConfigBuilder = Safe<imp::CreateTeamConfigBuilder>;
 
-/// Configures QUIC syncer for [`TeamConfigBuilder`].
+/// Configures QUIC syncer for [`AddTeamConfigBuilder`].
 ///
-/// By default, the QUIC syncer config is not set. It is an error to call
-/// [`team_config_build`] before setting the interval with
-/// this function
+/// By default, the QUIC syncer config is not set.
 ///
-/// @param cfg a pointer to the builder for a team config [`TeamConfigBuilder`]
+/// @param cfg a pointer to the builder for a team config [`AddTeamConfigBuilder`]
 /// @param quic set the QUIC syncer config [`QuicSyncConfig`]
 ///
-/// @relates AranyaTeamConfigBuilder.
-pub fn team_config_builder_set_quic_syncer(
-    cfg: &mut TeamConfigBuilder,
+/// @relates AranyaAddTeamConfigBuilder.
+pub fn add_team_config_builder_set_quic_syncer(
+    cfg: &mut AddTeamConfigBuilder,
     quic: OwnedPtr<QuicSyncConfig>,
 ) {
     // SAFETY: the user is responsible for passing in a valid QuicSyncConfig pointer.
@@ -771,18 +780,65 @@ pub fn team_config_builder_set_quic_syncer(
     cfg.quic(quic.imp());
 }
 
-/// Attempts to construct a [`TeamConfig`].
+/// Configures team ID field for [`AddTeamConfigBuilder`].
+///
+/// By default, the ID is not set.
+///
+/// @param cfg a pointer to the builder for a team config [`AddTeamConfigBuilder`]
+/// @param id a pointer to a [`TeamId`]
+///
+/// @relates AranyaAddTeamConfigBuilder.
+pub fn add_team_config_builder_set_id(cfg: &mut AddTeamConfigBuilder, id: &TeamId) {
+    cfg.id(*id);
+}
+
+/// Attempts to construct an [`AddTeamConfig`].
 ///
 /// This function consumes and releases any resources associated
 /// with the memory pointed to by `cfg`.
 ///
-/// @param cfg a pointer to the team config builder [`TeamConfigBuilder`]
-/// @param out a pointer to write the team config to [`TeamConfig`]
+/// @param cfg a pointer to the team config builder [`AddTeamConfigBuilder`]
+/// @param out a pointer to write the team config to [`AddTeamConfig`]
 ///
-/// @relates AranyaTeamConfigBuilder.
-pub fn team_config_build(
-    cfg: OwnedPtr<TeamConfigBuilder>,
-    out: &mut MaybeUninit<TeamConfig>,
+/// @relates AranyaAddTeamConfigBuilder.
+pub fn add_team_config_build(
+    cfg: OwnedPtr<AddTeamConfigBuilder>,
+    out: &mut MaybeUninit<AddTeamConfig>,
+) -> Result<(), imp::Error> {
+    // SAFETY: No special considerations.
+    unsafe { cfg.build(out)? }
+    Ok(())
+}
+
+/// Configures QUIC syncer for [`CreateTeamConfigBuilder`].
+///
+/// By default, the QUIC syncer config is not set.
+///
+/// @param cfg a pointer to the builder for a team config [`CreateTeamConfigBuilder`]
+/// @param quic set the QUIC syncer config [`QuicSyncConfig`]
+///
+/// @relates AranyaCreateTeamConfigBuilder.
+pub fn create_team_config_builder_set_quic_syncer(
+    cfg: &mut CreateTeamConfigBuilder,
+    quic: OwnedPtr<QuicSyncConfig>,
+) {
+    // SAFETY: the user is responsible for passing in a valid QuicSyncConfig pointer.
+    let quic = unsafe { quic.read() };
+    cfg.quic(quic.imp());
+}
+
+/// Attempts to construct a [`CreateTeamConfig`].
+///
+/// This function consumes and releases any resources associated
+/// with the memory pointed to by `cfg`.
+///
+/// @param cfg a pointer to the team config builder [`CreateTeamConfigBuilder`]
+/// @param out a pointer to write the team config to [`CreateTeamConfig`]
+///
+/// @relates AranyaCreateTeamConfigBuilder.
+pub fn create_team_config_build(
+    cfg: OwnedPtr<CreateTeamConfigBuilder>,
+    out: &mut MaybeUninit<CreateTeamConfig>,
 ) -> Result<(), imp::Error> {
     // SAFETY: No special considerations.
     unsafe { cfg.build(out)? }
@@ -1014,14 +1070,13 @@ pub fn revoke_label(
 /// Create a new graph/team with the current device as the owner.
 ///
 /// @param client the Aranya Client [`Client`].
-/// @param cfg the Team Configuration [`TeamConfig`].
+/// @param cfg the Team Configuration [`CreateTeamConfig`].
 /// @param __output the team's ID [`TeamId`].
 ///
 /// @relates AranyaClient.
-#[allow(unused_variables)] // TODO(nikki): once we have fields on TeamConfig, remove this for cfg
-pub fn create_team(client: &mut Client, cfg: &TeamConfig) -> Result<TeamId, imp::Error> {
+pub fn create_team(client: &mut Client, cfg: &CreateTeamConfig) -> Result<TeamId, imp::Error> {
     let client = client.imp();
-    let cfg: &imp::TeamConfig = cfg.deref();
+    let cfg: &imp::CreateTeamConfig = cfg.deref();
     let team_id = client
         .rt
         .block_on(client.inner.create_team(cfg.into()))?
@@ -1095,17 +1150,13 @@ pub unsafe fn encrypt_psk_seed_for_peer(
 /// Add a team to the local device store.
 ///
 /// @param client the Aranya Client [`Client`].
-/// @param team the team's ID [`TeamId`].
-/// @param cfg the Team Configuration [`TeamConfig`].
+/// @param cfg the Team Configuration [`AddTeamConfig`].
 ///
 /// @relates AranyaClient.
-#[allow(unused_variables)] // TODO(nikki): once we have fields on TeamConfig, remove this for cfg
-pub fn add_team(client: &mut Client, team: &TeamId, cfg: &TeamConfig) -> Result<(), imp::Error> {
+pub fn add_team(client: &mut Client, cfg: &AddTeamConfig) -> Result<(), imp::Error> {
     let client = client.imp();
-    let cfg: &imp::TeamConfig = cfg.deref();
-    client
-        .rt
-        .block_on(client.inner.add_team(team.into(), cfg.into()))?;
+    let cfg: &imp::AddTeamConfig = cfg.deref();
+    client.rt.block_on(client.inner.add_team(cfg.into()))?;
     Ok(())
 }
 
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -767,8 +767,8 @@ pub type CreateTeamConfigBuilder = Safe<imp::CreateTeamConfigBuilder>;
 ///
 /// By default, the QUIC syncer config is not set.
 ///
-/// @param cfg a pointer to the builder for a team config [`AddTeamConfigBuilder`]
-/// @param quic set the QUIC syncer config [`QuicSyncConfig`]
+/// @param[in,out] cfg a pointer to the builder for a team config [`AddTeamConfigBuilder`]
+/// @param[in] quic set the QUIC syncer config [`QuicSyncConfig`]
 ///
 /// @relates AranyaAddTeamConfigBuilder.
 pub fn add_team_config_builder_set_quic_syncer(
@@ -784,8 +784,8 @@ pub fn add_team_config_builder_set_quic_syncer(
 ///
 /// By default, the ID is not set.
 ///
-/// @param cfg a pointer to the builder for a team config [`AddTeamConfigBuilder`]
-/// @param id a pointer to a [`TeamId`]
+/// @param[in,out] cfg a pointer to the builder for a team config [`AddTeamConfigBuilder`]
+/// @param[in] id a pointer to a [`TeamId`]
 ///
 /// @relates AranyaAddTeamConfigBuilder.
 pub fn add_team_config_builder_set_id(cfg: &mut AddTeamConfigBuilder, id: &TeamId) {
@@ -797,8 +797,8 @@ pub fn add_team_config_builder_set_id(cfg: &mut AddTeamConfigBuilder, id: &TeamI
 /// This function consumes and releases any resources associated
 /// with the memory pointed to by `cfg`.
 ///
-/// @param cfg a pointer to the team config builder [`AddTeamConfigBuilder`]
-/// @param out a pointer to write the team config to [`AddTeamConfig`]
+/// @param[in] cfg a pointer to the team config builder [`AddTeamConfigBuilder`]
+/// @param[out] out a pointer to write the team config to [`AddTeamConfig`]
 ///
 /// @relates AranyaAddTeamConfigBuilder.
 pub fn add_team_config_build(
@@ -814,8 +814,8 @@ pub fn add_team_config_build(
 ///
 /// By default, the QUIC syncer config is not set.
 ///
-/// @param cfg a pointer to the builder for a team config [`CreateTeamConfigBuilder`]
-/// @param quic set the QUIC syncer config [`QuicSyncConfig`]
+/// @param[in,out] cfg a pointer to the builder for a team config [`CreateTeamConfigBuilder`]
+/// @param[in] quic set the QUIC syncer config [`QuicSyncConfig`]
 ///
 /// @relates AranyaCreateTeamConfigBuilder.
 pub fn create_team_config_builder_set_quic_syncer(
@@ -832,8 +832,8 @@ pub fn create_team_config_builder_set_quic_syncer(
 /// This function consumes and releases any resources associated
 /// with the memory pointed to by `cfg`.
 ///
-/// @param cfg a pointer to the team config builder [`CreateTeamConfigBuilder`]
-/// @param out a pointer to write the team config to [`CreateTeamConfig`]
+/// @param[in] cfg a pointer to the team config builder [`CreateTeamConfigBuilder`]
+/// @param[out] out a pointer to write the team config to [`CreateTeamConfig`]
 ///
 /// @relates AranyaCreateTeamConfigBuilder.
 pub fn create_team_config_build(
--- a/crates/aranya-client-capi/src/imp/config.rs
+++ b/crates/aranya-client-capi/src/imp/config.rs
@@ -317,7 +317,7 @@ pub struct TeamConfig {
     quic_sync: Option<QuicSyncConfig>,
 }
 
-impl From<TeamConfig> for aranya_client::TeamConfig {
+impl From<TeamConfig> for aranya_client::AddTeamConfig {
     fn from(value: TeamConfig) -> Self {
         let mut builder = Self::builder();
         if let Some(cfg) = value.quic_sync {
@@ -328,7 +328,7 @@ impl From<TeamConfig> for aranya_client::TeamConfig {
     }
 }
 
-impl From<&TeamConfig> for aranya_client::TeamConfig {
+impl From<&TeamConfig> for aranya_client::AddTeamConfig {
     fn from(value: &TeamConfig) -> Self {
         Self::from(value.to_owned())
     }
--- a/crates/aranya-client-capi/src/imp/config.rs
+++ b/crates/aranya-client-capi/src/imp/config.rs
@@ -8,7 +8,7 @@ use aranya_daemon_api::{SeedMode, SEED_IKM_SIZE};
 use tracing::error;
 
 use super::Error;
-use crate::api::defs::{self, Duration};
+use crate::api::defs::{self, Duration, TeamId};
 
 /// Configuration info for Aranya
 #[derive(Clone, Debug)]
@@ -311,14 +311,14 @@ impl Builder for QuicSyncConfigBuilder {
     }
 }
 
-/// Configuration info when creating or adding a team in Aranya
+/// Configuration info when adding a team in Aranya
 #[derive(Clone, Debug)]
-pub struct TeamConfig {
+pub struct CreateTeamConfig {
     quic_sync: Option<QuicSyncConfig>,
 }
 
-impl From<TeamConfig> for aranya_client::AddTeamConfig {
-    fn from(value: TeamConfig) -> Self {
+impl From<CreateTeamConfig> for aranya_client::CreateTeamConfig {
+    fn from(value: CreateTeamConfig) -> Self {
         let mut builder = Self::builder();
         if let Some(cfg) = value.quic_sync {
             builder = builder.quic_sync(cfg.into());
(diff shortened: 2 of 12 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -673,7 +673,7 @@ pub type QuicSyncConfig = Safe<imp::QuicSyncConfig>;
 
 /// A builder for initializing a [`QuicSyncConfig`].
 ///
-/// The [`QuicSyncConfig`] is an optional part of initializing a [`TeamConfig`].
+/// The [`QuicSyncConfig`] is an optional part of initializing an [`AddTeamConfig`] or [`CreateTeamConfig`].
```
ssweetney-spok:
```suggestion
/// The [`QuicSyncConfig`] is an optional part of initializing an [`AddTeamConfig`] or a [`CreateTeamConfig`].
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -741,48 +741,104 @@ pub fn quic_sync_config_build(
...
+pub type AddTeamConfig = Safe<imp::AddTeamConfig>;
+
+/// A builder for initializing an [`AddTeamConfig`].
+#[aranya_capi_core::derive(Init, Cleanup)]
+#[aranya_capi_core::opaque(size = 328, align = 8)]
+pub type AddTeamConfigBuilder = Safe<imp::AddTeamConfigBuilder>;
+
+/// Team configuration used when creating a team.
+///
+/// Use an [`CreateTeamConfigBuilder`] to construct this object.
```
```suggestion
/// Use a [`CreateTeamConfigBuilder`] to construct this object.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -741,48 +741,104 @@ pub fn quic_sync_config_build(
...
+    quic: OwnedPtr<QuicSyncConfig>,
+) {
+    // SAFETY: the user is responsible for passing in a valid QuicSyncConfig pointer.
+    let quic = unsafe { quic.read() };
+    cfg.quic(quic.imp());
+}
+
+/// Configures team ID field for [`AddTeamConfigBuilder`].
+///
+/// By default, the ID is not set.
```
```suggestion
/// By default, the team ID is not set.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -741,48 +741,104 @@ pub fn quic_sync_config_build(
...
+///
+/// @param[in,out] cfg a pointer to the builder for a team config [`AddTeamConfigBuilder`]
+/// @param[in] id a pointer to a [`TeamId`]
+///
+/// @relates AranyaAddTeamConfigBuilder.
+pub fn add_team_config_builder_set_id(cfg: &mut AddTeamConfigBuilder, id: &TeamId) {
```
```suggestion
pub fn add_team_config_builder_set_id(cfg: &mut AddTeamConfigBuilder, team_id: &TeamId) {
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1095,17 +1150,13 @@ pub unsafe fn encrypt_psk_seed_for_peer(
...
 ///
 /// @param client the Aranya Client [`Client`].
-/// @param team the team's ID [`TeamId`].
-/// @param cfg the Team Configuration [`TeamConfig`].
+/// @param cfg the Team Configuration [`AddTeamConfig`].
 ///
 /// @relates AranyaClient.
-#[allow(unused_variables)] // TODO(nikki): once we have fields on TeamConfig, remove this for cfg
-pub fn add_team(client: &mut Client, team: &TeamId, cfg: &TeamConfig) -> Result<(), imp::Error> {
+pub fn add_team(client: &mut Client, cfg: &AddTeamConfig) -> Result<(), imp::Error> {
```
gknopf-aranya: Note that we'll want to update the MVP spec API to reflect this change

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -311,14 +311,14 @@ impl Builder for QuicSyncConfigBuilder {
     }
 }
 
-/// Configuration info when creating or adding a team in Aranya
+/// Configuration info when adding a team in Aranya
```
gknopf-aranya:
```suggestion
/// Configuration info when creating a new team in Aranya.
```

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -328,44 +328,118 @@ impl From<TeamConfig> for aranya_client::TeamConfig {
...
+            out,
+            CreateTeamConfig {
+                quic_sync: self.quic_sync,
+            },
+        );
+        Ok(())
+    }
+}
+
+/// Configuration info when adding a team in Aranya
```
gknopf-aranya:
```suggestion
/// Configuration info when adding a team in Aranya.
```

crates/aranya-daemon-api/src/service.rs
```diff
@@ -134,9 +134,17 @@ pub enum Role {
...
+/// A configuration for adding a team to a daemon.
 #[derive(Debug, Serialize, Deserialize)]
-pub struct TeamConfig {
+pub struct AddTeamConfig {
+    pub id: TeamId,
+    pub quic_sync: Option<QuicSyncConfig>,
+}
+
+// Note: any fields added to this type should be public
+/// A configuration for creating a team to a daemon.
```
```suggestion
/// A configuration for creating a team in the daemon.
```

gknopf-aranya:
There are two primary reasons for splitting the create_team, add_team configs:
1. Creating a type that can be easily serialized, sent from the owner to a peer device, and used as input to `AddTeam()` that includes the team ID and QUIC syncer PSK.
2. Limiting the possible PSK modes in `SeedMode` to only what is valid when invoking either `CreateTeam()` or `AddTeam()`. This will require defining a new seed mode type for `CreateTeam()` and `AddTeam()`.

Currently this PR is handling 1. but not 2.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: https://github.com/aranya-project/aranya-docs/issues/109

ssweetney-spok:
> There are two primary reasons for splitting the create_team, add_team configs:
> 
>     1. Creating a type that can be easily serialized, sent from the owner to a peer device, and used as input to `AddTeam()` that includes the team ID and QUIC syncer PSK.
> 
>     2. Limiting the possible PSK modes in `SeedMode` to only what is valid when invoking either `CreateTeam()` or `AddTeam()`. This will require defining a new seed mode type for `CreateTeam()` and `AddTeam()`.
> 
> 
> Currently this PR is handling 1. but not 2.

Are you suggesting to spit the `SeedMode` in this PR or is there another issue?

examples/c/example.c
```diff
@@ -478,31 +478,48 @@ AranyaError init_team(Team *t) {
...
+        err = aranya_add_team_config_builder_set_quic_syncer(&build, &quic_cfg);
         if (err != ARANYA_ERROR_SUCCESS) {
             fprintf(stderr,
                     "unable to set `QuicSyncConfig` for "
-                    "`AranyaTeamConfigBuilder`\n");
+                    "`AranyaAddTeamConfigBuilder`\n");
             return err;
         }
+
+        err = aranya_add_team_config_builder_set_quic_syncer(&build, &quic_cfg);
```
djohnson-spideroak: is this being called twice?

ssweetney-spok: Yea, good catch

ssweetney-spok:
I can split out SeedMode on this PR. Should we also create new types for the QuicSyncConfigBuilder, etc?

I'm asking because I was thinking of doing something like:

```Rust
pub struct QuicSyncConfigBuilder {
    create_seed_mode: CreateSeedMode,
    add_seed_mode: Option<AddSeedMode> // optional because there's no sensible default between IKM and wrapped
}
```

So the same `QuicSyncConfigBuilder` can be used in `CreateTeamConfigBuilder` and `AddTeamConfigBuilder` but then the APIs would go from `QuicSyncConfigBuilder::seed_ikm` -> `QuicSyncConfigBuilder::add_seed_ikm`,` QuicSyncConfigBuilder::create_seed_ikm`. That feels more error prone.

Separate types seems more tedious to maintain but it may be easier for users

gknopf-aranya:
`AddSeedMode` could default to the wrapped mode since that's more secure. We want to force the user to be intentional about using the IKM.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -276,49 +348,186 @@ impl QuicSyncConfigBuilder {
...
+impl Builder for AddQuicSyncConfigBuilder {
+    type Output = defs::AddQuicSyncConfig;
+    type Error = Error;
+
+    /// # Safety
+    ///
+    /// No special considerations.
+    unsafe fn build(self, out: &mut MaybeUninit<Self::Output>) -> Result<(), Self::Error> {
+        let Some(mode) = self.data.mode else {
+            todo!("Don't panic. Return invalid arg error.")
```
ssweetney-spok: Fix

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -651,35 +651,50 @@ pub fn client_config_builder_set_aqc_config(cfg: &mut ClientConfigBuilder, aqc_c
...
     Ok(())
 }
 
-/// Attempts to set wrapped PSK seed value on [`QuicSyncConfigBuilder`].
+/// Attempts to set wrapped PSK seed value on [`AddQuicSyncConfigBuilder`].
 ///
 /// @param[in,out] cfg a pointer to the quic sync config builder
 /// @param[in] encap_seed a pointer the encapsulated PSK seed
 ///
 /// @relates AranyaQuicSyncConfigBuilder.
```
ssweetney-spok:
```suggestion
/// @relates AranyaAddQuicSyncConfigBuilder.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -651,35 +651,50 @@ pub fn client_config_builder_set_aqc_config(cfg: &mut ClientConfigBuilder, aqc_c
...
 #[aranya_capi_core::opaque(size = 288, align = 8)]
-pub type QuicSyncConfigBuilder = Safe<imp::QuicSyncConfigBuilder>;
+pub type CreateQuicSyncConfigBuilder = Safe<imp::CreateQuicSyncConfigBuilder>;
 
-/// Attempts to set PSK seed generation mode value on [`QuicSyncConfigBuilder`].
+/// Attempts to set PSK seed generation mode value on [`CreateQuicSyncConfigBuilder`].
 ///
 /// @param[in,out] cfg a pointer to the quic sync config builder
 ///
 /// @relates AranyaQuicSyncConfigBuilder.
```
ssweetney-spok:
```suggestion
/// @relates AranyaCreateQuicSyncConfigBuilder.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -693,80 +708,168 @@ pub struct SeedIkm {
     bytes: [u8; ARANYA_SEED_IKM_LEN],
 }
 
-/// Attempts to set raw PSK seed IKM value [`SeedIkm`] on [`QuicSyncConfigBuilder`].
+/// Attempts to set raw PSK seed IKM value [`SeedIkm`] on [`CreateQuicSyncConfigBuilder`].
+///
+/// @param[in,out] cfg a pointer to the quic sync config builder [`CreateQuicSyncConfigBuilder`]
+/// @param[in] ikm a pointer the raw PSK seed IKM [`SeedIkm`]
+///
+/// @relates AranyaQuicSyncConfigBuilder.
```
ssweetney-spok:
```suggestion
/// @relates AranyaCreateQuicSyncConfigBuilder.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -693,80 +708,168 @@ pub struct SeedIkm {
...
+
+/// Attempts to construct a [`CreateQuicSyncConfig`].
+///
+/// This function consumes and releases any resources associated
+/// with the memory pointed to by `cfg`.
+///
+/// @param[in] cfg a pointer to the QUIC sync config builder [`CreateQuicSyncConfigBuilder`]
+/// @param[out] out a pointer to write the QUIC sync config to [`CreateQuicSyncConfig`]
+///
+/// @relates AranyaQuicSyncConfigBuilder.
```
ssweetney-spok:
```suggestion
/// @relates AranyaCreateQuicSyncConfigBuilder.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -693,80 +708,168 @@ pub struct SeedIkm {
...
 ///
 /// This function consumes and releases any resources associated
 /// with the memory pointed to by `cfg`.
 ///
-/// @param[in] cfg a pointer to the QUIC sync config builder [`QuicSyncConfigBuilder`]
-/// @param[out] out a pointer to write the QUIC sync config to [`QuicSyncConfig`]
+/// @param[in] cfg a pointer to the QUIC sync config builder [`AddQuicSyncConfigBuilder`]
+/// @param[out] out a pointer to write the QUIC sync config to [`AddQuicSyncConfig`]
 ///
 /// @relates AranyaQuicSyncConfigBuilder.
```
ssweetney-spok:
```suggestion
/// @relates AranyaAddQuicSyncConfigBuilder.
```

crates/aranya-daemon-api/src/service.rs
ssweetney-spok, line 1: I didn't make the config types in this file generic but once there are more overlapping fields between the Add\*Config and Create\*Config types then that change can be made

examples/rust/src/main.rs
```diff
@@ -213,12 +214,11 @@ async fn main() -> Result<()> {
         owner.client.rand(&mut buf).await;
         buf
     };
-    let cfg = {
-        let qs_cfg = QuicSyncConfig::builder().seed_ikm(seed_ikm).build()?;
-        TeamConfig::builder().quic_sync(qs_cfg).build()?
+    let create_team_cfg = {
```
gknopf-aranya: nit: could name the variable `owner_cfg` since it's only used by the owner device when creating the team.
```suggestion
    let owner_cfg = {
```
The `add_team_cfg` could become `cfg` again. The type of config is clear based on the types used to initialize each one.

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
```diff
@@ -0,0 +1,199 @@
...
+pub struct AddBuild {
+    pub(super) mode: Option<AddSeedMode>,
+}
+
+#[derive(Clone, Default)]
+pub struct CreateBuild {
+    pub(super) mode: CreateSeedMode,
+}
+#[derive(Clone)]
+pub struct QuicSyncConfig<T> {
```
gknopf-aranya: I'm not sure I like using generics for the `QuicSyncConfig`. We want to restrict the configuration to either `CreateSeedMode` or `AddSeedMode` which can be accomplished using an enum. E.g.
```
pub enum QuicSyncConfig {
    Create(CreateSeedMode),
    Add(AddSeedMode)
}
```

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
```diff
@@ -0,0 +1,199 @@
...
+
+impl<T: Clone> QuicSyncConfig<T> {
+    /// Useful for deref coercion.
+    pub(crate) fn imp(&self) -> Self {
+        self.clone()
+    }
+}
+
+pub type CreateQuicSyncConfig = QuicSyncConfig<Create>;
+pub type AddQuicSyncConfig = QuicSyncConfig<Add>;
```
gknopf-aranya:
```suggestion
/// QUIC syncer configuration for CreateTeam() operation.
pub type CreateTeamQuicSyncConfig = QuicSyncConfig<Create>;
/// QUIC syncer configuration for AddTeam() operation.
pub type AddTeamQuicSyncConfig = QuicSyncConfig<Add>;
```

crates/aranya-client-capi/src/imp/config/team.rs
```diff
@@ -0,0 +1,199 @@
...
+
+#[derive(Clone)]
+/// Configuration info for creating or adding teams.
+pub struct TeamConfig<T, U> {
+    data: T,
+    quic_sync: Option<QuicSyncConfig<U>>,
+}
+
+pub type CreateTeamConfig = TeamConfig<Create, quic_sync::Create>;
+pub type AddTeamConfig = TeamConfig<Add, quic_sync::Add>;
```
gknopf-aranya:
```suggestion
/// A team config required to create a new Aranya team.
pub type CreateTeamConfig = TeamConfig<Create, quic_sync::Create>;
/// A team config required to add an existing Aranya team to a device.
pub type AddTeamConfig = TeamConfig<Add, quic_sync::Add>;
```

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
ssweetney-spok: Yea, an enum is probably best if there's only gonna be a single field in the `QuicSyncConfig`. We may want to add more in the future though. 

I could use an enum which would make things simpler but defining it as a struct in a later version would be a breaking change. (That doesn't matter much for `aranya-client-capi` but it's something to consider for the `aranya-client` crate which has similar type)

crates/aranya-client-capi/src/imp/config/team.rs
gknopf-aranya, line 1: Organization of type/struct definitions and impls could be improved in this file.
For example, all the `CreateTeamConfig` code could be grouped together, while all the `CreateTeamConfigBuilder` code is grouped together in another spot. Right now, the types and impls are interleaved with each other in a way that is difficult to follow.

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
ssweetney-spok: I'm also realizing that we lose some type safety with an enum since the enum variants are all the same type which means that an `AddTeamConfig` could hold a `QuicSyncConfig::Create` variant so we'd have to rely on runtime checks.

gknopf-aranya: Enums support putting structs/types in each variant. You could define a struct with a single field in each where more fields can be added later.
```
pub enum QuicSyncConfig {
    Create(CreateTeamQuicSyncConfig),
    Add(AddTeamQuicSyncConfig)
}

pub struct CreateTeamQuicSyncConfig{...}
pub struct AddTeamQuicSyncConfig{...}
```

gknopf-aranya: Type-safety would be ideal. This could be achieved by limiting the quic sync config to a specific type rather than using an enum or generic:
```
pub type CreateTeamConfig {
    quic CreateTeamQuicSyncConfig,
}
```

ssweetney-spok: Ah, yea that works also. Should we go with an enum if it means still having to use a match statement at runtime?

gknopf-aranya: I prefer a compile time type check over a runtime check. Enum is better than generics because the compiler will at least force the match to check each case exhaustively.

ssweetney-spok: > Type-safety would be ideal. This could be achieved by limiting the quic sync config to a specific type rather than using an enum or generic:
> 
> ```
> pub type CreateTeamConfig {
>     quic CreateTeamQuicSyncConfig,
> }
> ```

Concrete types also work but then there may be an additional type that's needed to hold common fields between the Add\*Config and Create\*Config types. I actually had [something similar](https://github.com/aranya-project/aranya/pull/368/commits/49840ef27d6db09bd1f7dd91cb7c405c67faf76e) before the current implementation

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -693,80 +708,168 @@ pub struct SeedIkm {
...
-/// @param[out] out a pointer to write the QUIC sync config to [`QuicSyncConfig`]
+/// @param[in] cfg a pointer to the QUIC sync config builder [`CreateQuicSyncConfigBuilder`]
+/// @param[out] out a pointer to write the QUIC sync config to [`CreateQuicSyncConfig`]
 ///
-/// @relates AranyaQuicSyncConfigBuilder.
-pub fn quic_sync_config_build(
-    cfg: OwnedPtr<QuicSyncConfigBuilder>,
-    out: &mut MaybeUninit<QuicSyncConfig>,
+/// @relates AranyaCreateQuicSyncConfigBuilder.
+pub fn create_quic_sync_config_build(
```
gknopf-aranya: Let's rename these methods to be more clear that create/add are referring to the team operation and not the QUIC sync config.
```suggestion
pub fn create_team_quic_sync_config_build(
```

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
gknopf-aranya: What was the reason for changing from that original implementation?

crates/aranya-client/src/config/team.rs
```diff
@@ -0,0 +1,169 @@
...
+impl CreateTeamConfig {
+    /// Creates a default [`CreateTeamConfigBuilder`].
+    pub fn builder() -> CreateTeamConfigBuilder {
+        CreateTeamConfigBuilder::default()
+    }
+}
+
+impl AddTeamConfigBuilder {
+    /// Sets the ID of the team to add.
+    pub fn id(mut self, id: TeamId) -> Self {
```
gknopf-aranya: nit: slightly more clear
```suggestion
    pub fn team_id(mut self, team_id: TeamId) -> Self {
```

crates/aranya-client/src/config/team/quic_sync.rs
```diff
@@ -0,0 +1,185 @@
+//! Configuration for team synchronization over QUIC connections.
+//!
+//!
```
gknopf-aranya: Extra line?

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
ssweetney-spok: I wanted to avoid having to duplicate common methods between the builders so I wrote that macro but Rust Analyzer doesn't work on code in macros so no autocomplete, etc. You could also get around that by defining the common methods on the `CommonBuilderFields` types but it's a bit less flexible.

I also saw https://github.com/aranya-project/aranya-core/pull/327 and thought generics might be a good fit if we need additional team configs for other operations.

crates/aranya-client/src/config/team/quic_sync.rs
```diff
@@ -0,0 +1,185 @@
...
+//! - [`AddQuicSyncConfig`] - For adding members to existing teams
+
+use aranya_daemon_api::{AddSeedMode, CreateSeedMode, SEED_IKM_SIZE};
+use tracing::error;
+
+use crate::{error::InvalidArg, ConfigError, Result};
+
+#[derive(Clone)]
+/// Configuration data for adding members to an existing team.
+/// See [`QuicSyncConfig`].
```
gknopf-aranya: @jdygert-spok, @ssweetney-spok I believe we have a convention of putting the derives after the docs in most of this codebase:
```suggestion
/// Configuration data for adding members to an existing team.
/// See [`QuicSyncConfig`].
#[derive(Clone)]
```

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: What do you mean "by referring to the team operation"?

This function is how the QUIC sync config is built

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
gknopf-aranya: While I'm generally a fan of consolidating duplicate code, in this case I believe type-safety is more important because it decreases the chance that the API could be misused or that a team config could be misconfigured.
As long as we ensure type safety in the types that are publicly exposed by the API, I don't see a reason not to use generics or enums under the hood though.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: In this context, `create_` refers to the fact that `create_quic_sync_config_build` should be used to construct a `CreateQuicSyncConfig` which is only supposed to be used by `CreateTeam()`.
The create is referring to whether this is related to `CreateTeam()` or `AddTeam()` and not to creating a QUIC syncer config.
I'm suggesting adding a bit more context to the name to make this more clear.

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
ssweetney-spok: We use generics internally but the generic team config type is not exposed. I realized over the weekend that adding a new type parameter would always be a breaking change. So going from:

```Rust
pub fn create_team(cfg: TeamConfig<Create, QuicCreate>) { .. }
```

to something like:

```Rust
pub fn create_team(cfg: TeamConfig<Create, QuicCreate, NewSyncTypeCreate>) { .. }
```

would require a major version bump.

gknopf-aranya: If the generics aren't exposed in the API, the current implementation sounds good to me.

ssweetney-spok: Moving back towards concrete types for the configs since "[obake](https://github.com/cofibrant/obake)", a crate I plan to use for another PR, doesn't support generic types.

gknopf-aranya: Looks like this thread can be resolved

crates/aranya-client-capi/src/imp/config/team.rs
gknopf-aranya, line 1: `pub` types/structs in this file should have docs

crates/aranya-daemon-api/src/service.rs
```diff
@@ -134,10 +134,18 @@ pub enum Role {
 }
 
 // Note: any fields added to this type should be public
-/// A configuration for creating or adding a team to a daemon.
+/// A configuration for adding a team in the daemon.
 #[derive(Debug, Serialize, Deserialize)]
-pub struct TeamConfig {
-    pub quic_sync: Option<QuicSyncConfig>,
+pub struct AddTeamConfig {
+    pub id: TeamId,
```
```suggestion
    pub team_id: TeamId,
```

gknopf-aranya (approved):

gknopf-aranya (dismissed):

gknopf-aranya:
> @ssweetney-spok, @djohnson-spideroak, @jdygert-spok how do we want to handle documentation for unstable types/method introduced on this PR? The PR I have open now doesn't have the new types introduced on this PR. So I think we should document the unstable things here (or open a follow-up PR after this is merged). If we're releasing on Thursday, we may not have time to open a follow-up PR though, so my preference is to handle that here.

Not relevant anymore since we are only documenting the methods which will be removed for the PSK -> certificate change.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -651,35 +651,50 @@ pub fn client_config_builder_set_aqc_config(cfg: &mut ClientConfigBuilder, aqc_c
 
 /// QUIC syncer configuration.
 ///
-/// Use a [`QuicSyncConfigBuilder`] to construct this object.
+/// Use a [`CreateTeamQuicSyncConfigBuilder`] to construct this object.
 #[aranya_capi_core::opaque(size = 288, align = 8)]
```
djohnson-spideroak: is this size still correct?

djohnson-spideroak (dismissed):

crates/aranya-daemon/src/api.rs
```diff
@@ -374,13 +374,9 @@ impl DaemonApi for Api {
...
-    #[instrument(skip(self), err)]
-    async fn add_team(
-        mut self,
-        _: context::Context,
-        team: api::TeamId,
-        cfg: api::TeamConfig,
-    ) -> api::Result<()> {
+    #[instrument(skip(self))]
+    async fn add_team(mut self, _: context::Context, cfg: api::AddTeamConfig) -> api::Result<()> {
+        let team = cfg.team_id;
```
jdygert-spok, line 379: nit: remove `team` param from `add_team_quic_sync`. since it is in the cfg now.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: Yes and no. 288 works but it adds excess padding. Reduced this to 56 for this and the other related types

crates/aranya-daemon/src/api.rs
ssweetney-spok: `add_team_quic_sync` doesn't take the team config that has the id in it as a param. It takes the `AddTeamQuicSyncConfig`

Edit: I see how the `cfg` variable being shadowed could make that confusing. I could rename it

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -651,35 +651,50 @@ pub fn client_config_builder_set_aqc_config(cfg: &mut ClientConfigBuilder, aqc_c
...
-pub type QuicSyncConfigBuilder = Safe<imp::QuicSyncConfigBuilder>;
+pub type AddTeamQuicSyncConfigBuilder = Safe<imp::AddTeamQuicSyncConfigBuilder>;
 
-/// Attempts to set PSK seed generation mode value on [`QuicSyncConfigBuilder`].
+/// A builder for initializing a [`CreateQuicSyncConfig`].
+///
+/// The [`QuicSyncConfig`] is an optional part of initializing an [`AddTeamConfig`] or a [`CreateTeamConfig`].
+#[aranya_capi_core::derive(Init, Cleanup)]
+#[aranya_capi_core::opaque(size = 56, align = 8)]
+pub type CreateTeamQuicSyncConfigBuilder = Safe<imp::CreateTeamQuicSyncConfigBuilder>;
```
```suggestion
/// A builder for initializing a [`CreateTeamQuicSyncConfig`].
///
/// The [`CreateTeamQuicSyncConfig`] is an optional part of initializing a [`CreateTeamConfig`].
#[aranya_capi_core::derive(Init, Cleanup)]
#[aranya_capi_core::opaque(size = 56, align = 8)]
pub type CreateTeamQuicSyncConfigBuilder = Safe<imp::CreateTeamQuicSyncConfigBuilder>;
```

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
```diff
@@ -0,0 +1,177 @@
...
+            error!(?err);
+            InvalidArg::new("wrapped_seed", "could not deserialize")
+        })?;
+        self.mode = Some(AddSeedMode::Wrapped(wrapped));
+
+        Ok(())
+    }
+}
+
+impl Typed for AddTeamQuicSyncConfig {
```
gknopf-aranya: I would group all the `impl` for a type together, rather than grouping all the `Typed` impls together.
The impls for a type are essentially a class/module that could be moved into their own file so it makes sense to keep them together even when there are multiple types implemented in the same file.

crates/aranya-client/src/config/team.rs
```diff
@@ -0,0 +1,128 @@
+//! Team configuration for creating new teams or joining existing ones.
+//!
+//! This module provides builders for configuring team operations with support
+//! for multiple transport mechanisms.
+//!
+//! # Overview
+//!
+//! There are two primary operations:
+//! - **Create Team**: Establish a new team with [`CreateTeamConfig`]
+//! - **Join Team**: Join an existing team with [`AddTeamConfig`]
```
gknopf-aranya: Let's be consistent with wording of "add team" vs. "join team". It seems the code/specs have mostly used "add team".
```suggestion
//! - **Add Team**: Add an existing team with [`AddTeamConfig`]
```

crates/aranya-client/src/config/team.rs
gknopf-aranya, line 1: Ditto on grouping impls for types together

crates/aranya-client/tests/common/mod.rs
```diff
@@ -135,26 +138,38 @@ impl TeamCtx {
...
+                .quic_sync(qs_cfg)
+                .build()?
+        };
+
         // Owner has the team added due to calling `create_team`, now we assign it to all other peers
-        self.admin.client.add_team(team_id, cfg.clone()).await?;
-        self.operator.client.add_team(team_id, cfg.clone()).await?;
-        self.membera.client.add_team(team_id, cfg.clone()).await?;
-        self.memberb.client.add_team(team_id, cfg).await?;
+        self.admin.client.add_team(cfg.clone()).await?;
```
gknopf-aranya, line 169: `Copy` can easily be implemented for the config type which removes the need for `.clone()` here:
https://github.com/aranya-project/aranya/pull/386/files#diff-d8c814f3642ce3699f179c12468986f3fece96206cd8b0ebe0e979d3c379d6ab

jdygert-spok: No, then we can't add any non-copy types to the config.

crates/aranya-client/src/config/team.rs
jdygert-spok: Yeah, IMO "join" implies a modification to the team, so "add" is right for this operation.

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
ssweetney-spok: Could you give an example? They're already grouped together by type aside for the `From` impls that are left at the bottom

ssweetney-spok: Oh, nvm this isn't team.rs

ssweetney-spok: I changed it. I find the updated version harder to read though since I have to scroll more to see which fields/methods the types have in common. Also knowing that those types implement `Typed` and `Builder` aren't that *important* which is why I wanted to leave them towards the bottom of the file.

gknopf-aranya: I see your point. It's a small file so it's not too bad. Usually impls for types would be different, and if they are mostly the same, that's an indication some code could be refactored into a common field or generic types.

PR #369, opened 2025-07-02
WIP: capi: guard against concurrent use

gknopf-aranya:
Safeguard C API against concurrent use.

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -77,6 +77,10 @@ pub enum Error {
     #[capi(msg = "serialization")]
     Serialization,
 
+    /// Locking error.
+    #[capi(msg = "unable to aquire mutex")]
+    TryLock,
+
     /// Some other error occurred.
     #[capi(msg = "other")]
     Other,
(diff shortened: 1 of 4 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1693,11 +2027,22 @@ pub fn aqc_delete_uni_channel(
...
+        client
+            .as_mut()
+            .ok_or(InvalidArg::new(
+                "client",
+                "unable to get mutable reference to client",
+            ))?
+            .imp()
+            .get_arcs()
+    };
+    let chan = client.try_lock()?.aqc().try_receive_channel()?;
```
gknopf-aranya, line 2045: I wanted to avoid deadlock by always obtaining the locks in a specific order: `rt` first, then `client`.
But, some methods only need to use `rt` or `client` so this could deadlock.

gknopf-aranya: Then again, since `try_lock()` is used everywhere, the worst that could happen is returning an error. And that would only happen if functions are called concurrently which is what this implementation is guarding against. So this shouldn't be an issue.

jdygert-spok:
I took a simpler approach here: #371.

gknopf-aranya:
Closing in favor of:
https://github.com/aranya-project/aranya/pull/371

PR #371, opened 2025-07-02
capi: add lock to client

### LINKED ISSUE #355: capi: prevent concurrent client usage
It would be fairly easy to accidentally use the client concurrently in the C api. Wrap the C client in a mutex-like type. Unsure if we want to block or return an error if locked.

jdygert-spok:
Closes #355.

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -12,7 +12,10 @@ use aranya_daemon_api::Text;
 use bytes::Bytes;
 use tracing::error;
 
-use crate::imp::{self, aqc::consume_bytes};
+use crate::{
+    imp::{self, aqc::consume_bytes},
+    lock::Lock,
+};
 
 /// An error code.
 ///
@@ -221,7 +224,12 @@ pub unsafe fn client_init(
             .connect()
     })?;
 
-    Client::init(client, imp::Client { rt, inner });
+    Client::init(
+        client,
+        imp::Client {
+            lock: Lock::new(imp::ClientInner { rt, inner }),
+        },
+    );
     Ok(())
 }
 
@@ -495,10 +503,11 @@ pub fn init_logging() -> Result<(), imp::Error> {
 ///
 /// @relates AranyaClient.
 pub unsafe fn get_key_bundle(
-    client: &mut Client,
+    client: *mut Client,
     keybundle: *mut MaybeUninit<u8>,
     keybundle_len: &mut usize,
 ) -> Result<(), imp::Error> {
+    let mut client = aranya_capi_core::try_as_ref!(client).try_lock()?;
     let client = client.imp();
     let keys = client.rt.block_on(client.inner.get_key_bundle())?;
     // SAFETY: Must trust caller provides valid ptr/len for keybundle buffer.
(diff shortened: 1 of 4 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -855,12 +865,13 @@ pub fn sync_peer_config_builder_set_sync_later(cfg: &mut SyncPeerConfigBuilder)
...
 /// @relates AranyaClient.
-pub fn assign_role(
-    client: &mut Client,
+pub unsafe fn assign_role(
+    client: *mut Client,
     team: &TeamId,
     device: &DeviceId,
     role: Role,
 ) -> Result<(), imp::Error> {
+    let mut client = aranya_capi_core::try_as_ref!(client).try_lock()?;
```
gknopf-aranya, line 874: No safety comment needed for these?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1493,11 +1532,12 @@ pub unsafe fn aqc_assign_net_identifier(
...
 /// @relates AranyaClient.
 pub unsafe fn aqc_remove_net_identifier(
-    client: &mut Client,
+    client: *mut Client,
     team: &TeamId,
     device: &DeviceId,
     net_identifier: NetIdentifier,
 ) -> Result<(), imp::Error> {
+    let mut client = aranya_capi_core::try_as_ref!(client).try_lock()?;
     let client = client.imp();
```
gknopf-aranya, line 1541: Two `let client` in a row is a bit awkward.
```suggestion
    let mut client = aranya_capi_core::try_as_ref!(client).try_lock()?.imp();
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -495,10 +503,11 @@ pub fn init_logging() -> Result<(), imp::Error> {
 ///
 /// @relates AranyaClient.
 pub unsafe fn get_key_bundle(
-    client: &mut Client,
+    client: *mut Client,
     keybundle: *mut MaybeUninit<u8>,
     keybundle_len: &mut usize,
 ) -> Result<(), imp::Error> {
+    let mut client = aranya_capi_core::try_as_ref!(client).try_lock()?;
```
gknopf-aranya, line 510: Should a new error be defined for the API if the `try_lock()` fails?

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: It needs to be separate or else the guard gets dropped since it is just a temporary.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Yeah, we should probably remove the internal unsafe block from the macro so the caller must have one.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: We could. One reason not to is that if we later make it usable concurrently, we'll be stuck with an unused variant.

gknopf-aranya: Is it using one of the existing error variants currently?

jdygert-spok: Yes, `InvalidArgument`

crates/aranya-client-capi/src/lock.rs
ssweetney-spok, line 1: Seems sound. Could you add an unit test that tests concurrent access/updates?

Also, one day it'd be nice to add Miri the CI pipeline.

ssweetney-spok: Nvm, the test may not give deterministic results

PR #373, opened 2025-07-02
Initialize AddTeamConfigBuilder from TeamInfo

### LINKED ISSUE #341: Initialize an AddTeamConfigBuilder from existing data

ssweetney-spok:
Resolves #341

- [x] Use team info in C example
- [x] Audit [Obake](https://github.com/cofibrant/obake)


crates/aranya-client-capi/src/imp/config/team.rs
```diff
@@ -168,3 +192,43 @@ impl From<&CreateTeamConfig> for aranya_client::CreateTeamConfig {
...
+            let qs_cfg_builder = value.quic_sync();
+
+            if let Some(mode) = qs_cfg_builder.get_mode() {
+                builder.quic_sync().mode(mode.clone());
+            }
+        }
+
+        builder
+    }
+}
```
ssweetney-spok: Not a fan of how I currently implemented this. The acessor methods don't seem like a good idea especially when more fields are added.

I think the easier path forward would be to move some code in this crate into `aranya-client` (behind a feature flag) instead.

There's also the option of duplicating the code for the versioned types so that there's less of a reliance of `aranya-client` 🤷

examples/rust/src/main.rs
```diff
@@ -246,17 +246,28 @@ async fn main() -> Result<()> {
...
-            .build()?
+    let (add_team_cfg, owner_team_cfg_builder) = {
+        let mut team_cfg_builder = AddTeamConfig::builder();
+
+        team_cfg_builder.quic_sync().seed_ikm(seed_ikm);
+        team_cfg_builder = team_cfg_builder.team_id(team_id);
+
+        let builder_clone = team_cfg_builder.clone();
+
+        (team_cfg_builder.build()?, builder_clone)
```
gknopf-aranya, line 257:
```suggestion
        (team_cfg_builder.clone().build()?, team_cfg_builder)
```

examples/rust/src/main.rs
```diff
@@ -246,17 +246,28 @@ async fn main() -> Result<()> {
...
+    };
+
+    let admin_team = {
+        // This is where the owner could create a TeamInfo struct, serialize it,
+        // and transmit it to the admin through an out-of-band process.
+        let team_info = owner_team_cfg_builder.to_team_info();
```
gknopf-aranya: Let's show an example or serializing/deserializing `team_info`

examples/c/example.c
```diff
@@ -471,40 +480,38 @@ AranyaError init_team(Team *t) {
...
         }
 
-        AranyaAddTeamConfigBuilder build;
-        err = aranya_add_team_config_builder_init(&build);
-        if (err != ARANYA_ERROR_SUCCESS) {
-            fprintf(stderr, "unable to init `AranyaAddTeamConfigBuilder`\n");
-            return err;
+        size_t team_info_len = BUFFER_LEN;
+        uint8_t *team_info   = calloc(team_info_len, 1);
+        err = aranya_add_team_config_builder_to_team_info(&build, team_info, &team_info_len);
```
gknopf-aranya: Confirming that the serialized team info data is written to `team_info` when this is called?

crates/aranya-client/tests/tests.rs
```diff
@@ -209,19 +207,15 @@ async fn test_add_team() -> Result<()> {
...
-                .build()?
-        })
-        .await?;
+
+    let admin_cfg = {
+        let mut team_cfg_builder = AddTeamConfig::builder().team_id(team_id);
+        team_cfg_builder.quic_sync().wrapped_seed(&admin_seed)?;
+
+        team_cfg_builder.build()?
+    };
```
gknopf-aranya, line 225: Mixing builder types returned by methods on `AddTeamConfigBuilder` means we can't simplify the code to this:
```suggestion
    let admin_cfg = AddTeamConfig::builder()
        .team_id(team_id)
        .quic_syncer_wrapped_seed(&admin_seed)?
        .build()?;
```
I'd prefer to implement the builder in a way where any setter methods return `Self` as the type.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -849,6 +836,63 @@ pub fn add_team_config_build(
...
+///
+/// This function consumes and releases any resources associated
+/// with the memory pointed to by `cfg`.
+///
+/// @param[in] cfg a pointer to the team config builder [`AddTeamConfigBuilder`]
+/// @param[out] buffer the serialized team info data.
+/// @param[in,out] buffer_len the number of bytes written to the buffer.
+///
+/// @relates AranyaAddTeamConfigBuilder.
+pub unsafe fn add_team_config_builder_to_team_info(
```
gknopf-aranya: The name of this function suggests that the this is adding the team config builder to team info.
But, this is converting add_team_config_builder into a serialized team_info.
```suggestion
pub unsafe fn add_team_config_builder_into_serialized_team_info(
```

crates/aranya-client-capi/src/imp/config/team.rs
gknopf-aranya: You could make the fields of the builder `pub(crate)` then return:
```
Self {
    field: value.field,
    ...
}
```

crates/aranya-client/src/config/team.rs
```diff
@@ -13,14 +13,43 @@
...
+    #![allow(missing_docs)] // TODO: Document the type alias by generated by _obake_
+    #![allow(unused_macros)]
+    //! This module contains the [`TeamInfo`] type that is used to
+    //! initialize an [`super::AddTeamConfigBuilder`].
+
+    use serde::{Deserialize, Serialize};
+
+    use super::{quic_sync, TeamId};
+
+    // The doc comment below does not get applied to the generated type alias :(
```
gknopf-aranya: What if you wrap it in another type alias that is documented?

crates/aranya-client/src/config/team.rs
```diff
@@ -13,14 +13,43 @@
...
+    /// Team data for initializing an [`super::AddTeamConfigBuilder`].
+    #[obake::versioned]
+    #[obake(version("0.1.0"))]
+    #[obake(derive(Clone, Debug, Serialize, Deserialize,))]
```
gknopf-aranya: Is there an extra comma at the end?
```suggestion
    #[obake(derive(Clone, Debug, Serialize, Deserialize))]
```

crates/aranya-client/src/config/team.rs
```diff
@@ -13,14 +13,43 @@
...
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub struct TeamInfo {
```
gknopf-aranya, line 48: Is there a way to determine the version of a `TeamInfo` object received from a peer?

crates/aranya-client/src/config/team.rs
ssweetney-spok: Yea, that seems like a good workaround for now

crates/aranya-client-capi/src/imp/config/team.rs
ssweetney-spok: How would making the fields `pub(crate)` help here?

crates/aranya-client/tests/tests.rs
ssweetney-spok: Do you have suggestions on how to do that?

It seems like based on the code above we'd have to have a corresponding method in `AddTeamConfigBuilder` for each method in `AddTeamQuicSyncConfigBuilder`.

crates/aranya-client/src/config/team.rs
ssweetney-spok: Yea, it's `VersionedTeamInfo` is an enum so users can use `match`

examples/c/example.c
```diff
@@ -471,40 +480,38 @@ AranyaError init_team(Team *t) {
...
         }
 
-        AranyaAddTeamConfigBuilder build;
-        err = aranya_add_team_config_builder_init(&build);
-        if (err != ARANYA_ERROR_SUCCESS) {
-            fprintf(stderr, "unable to init `AranyaAddTeamConfigBuilder`\n");
-            return err;
+        size_t team_info_len = BUFFER_LEN;
+        uint8_t *team_info   = calloc(team_info_len, 1);
+        err = aranya_add_team_config_builder_into_serialized_team_info(&build, team_info, &team_info_len);
```
djohnson-spideroak, line 493: unfortunate this name is so long, dont know if there is a better name though.

crates/aranya-client/src/config/team/quic_sync.rs
```diff
@@ -136,3 +157,59 @@ impl From<AddTeamQuicSyncConfig> for aranya_daemon_api::AddTeamQuicSyncConfig {
...
+    //! Versioned types for use in [`super::super::TeamInfo`]
+
+    use serde::{Deserialize, Serialize};
+
+    use super::AddSeedMode;
+
+    // Values added here should be set
+    // in super::AddTeamQuicSyncConfigBuilder::set_from_team_info
+    #[obake::versioned]
+    #[obake(version("0.1.0"))]
```
djohnson-spideroak, line 180: did we want the versioning of the structs to be independent of the aranya version?

crates/aranya-client-capi/src/imp/config/team.rs
gknopf-aranya: It would allow the type conversion to directly access fields rather than using getter/setting methods during the type conversion.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Yeah, I would use the builder syntax for setting up the `AddTeamQuicSyncConfigBuilder` first, build it, then set the `AddTeamQuicSyncConfig` on `AddTeamConfigBuilder` one time.

gknopf-aranya: It would look something like this:
```
    let admin_cfg = AddTeamConfig::builder()
        .team_id(team_id)
        .quic_syncer(QuicSyncConfig::builder().wrapped_seed(&admin_seed).build()?)?
        .build()?;
```

crates/aranya-client-capi/src/imp/config/team.rs
jdygert-spok: They are in separate crates, `aranya-client` and `aranya-client-capi`

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1377,27 +1334,31 @@ AranyaError aranya_create_team_config_builder_cleanup_ext(struct AranyaCreateTea
...
  * @param[in,out] cfg a pointer to the builder for a team config [`AranyaAddTeamConfigBuilder`](@ref AranyaAddTeamConfigBuilder)
- * @param[in] quic set the QUIC syncer config [`AranyaAddTeamQuicSyncConfig`](@ref AranyaAddTeamQuicSyncConfig)
+ * @param[out] __output a ptr to a builder for a quic sync config [`AranyaAddTeamQuicSyncConfigBuilder`](@ref AranyaAddTeamQuicSyncConfigBuilder)
  *
  * @relates AranyaAddTeamConfigBuilder.
  */
-AranyaError aranya_add_team_config_builder_set_quic_syncer(struct AranyaAddTeamConfigBuilder *cfg,
-                                                           struct AranyaAddTeamQuicSyncConfig *quic);
+AranyaError aranya_add_team_config_builder_get_quic_syncer(struct AranyaAddTeamConfigBuilder *cfg,
+                                                           struct AranyaAddTeamQuicSyncConfigBuilder **__output);
```
djohnson-spideroak: I think this doc needs updating, its not clear to me how this would be used to get the quic syncer set up.

crates/aranya-client/tests/tests.rs
ssweetney-spok: That's how it worked before this PR. I changed it because it seemed valuable to allow users to modify the `AddTeamQuicSyncConfig` values when they initialized an `AddTeamConfigBuilder` from some team info data.

crates/aranya-client/src/config/team/quic_sync.rs
ssweetney-spok: They can be tied together. I thought it would make sense to keep them independent since we may not add new fields every time the Aranya version is bumped.

Edit: We could just declare a new version whenever a new field is added and have it match the current/next Aranya version.

It should be OK if versions are skipped. (TeamInfov1_0_0 -> TeamInfov1_3_0)

crates/aranya-client/tests/tests.rs
gknopf-aranya: If that's needed, we could add a `.get_quic_syncer()` method that returns a `AddTeamQuicSyncConfigBuilder` that can be modified. I don't think that should be the normal way of building the object though.

ssweetney-spok: We should probably have a single way to build the object rather than multiple. That seems less prone to user error. I can revert the changes if we don't want the `AddTeamQuicSyncConfig` values to be modifiable after it's already set in the `AddTeamConfigBuilder`.

ssweetney-spok: Another reason I don't want to combine approaches is that it makes the code a bit awkward.

`AddTeamConfigBuilder` would hold an `AddTeamQuicSyncConfigBuilder` internally so it can return a mutable reference from `AddTeamConfigBuilder::get_quic_syncer`. So we'd end up with having to do this conversion

`AddTeamQuicSyncConfigBuilder` -> `AddTeamQuicSyncConfig` -> `AddTeamQuicSyncConfigBuilder`, when `AddTeamQuicSyncConfigBuilder::quic_sync` is called

gknopf-aranya: My preference is to implement the builder pattern in as idiomatic a way as possible unless there's a good reason not to.

examples/c/example.c
gknopf-aranya: Yeah, it isn't ideal. I don't see a way to shorten it without breaking the serialization step out into another function.

crates/aranya-client-capi/src/imp/config/team.rs
gknopf-aranya: @jdygert-spok Steve and I came up with 2 ideas so far:
1. Wrap the C API builder types around the client builder types directly (so no type conversion is needed)
2. Define `WrappedBuilder` types in the client and capi crates that have a private `Builder` field with its own `pub` fields. Type conversions between `Builder` types with `pub` fields would be simpler than conversion via getter/setter methods. `Builder` types with `pub` fields will need to be public so they can be shared between the crates and could be hidden with `#[doc(hidden)]`

(2) sounds overly complex so I'm leaning toward (1)

jdygert-spok: We haven't done (1) because the rust builder methods are `Self -> Self` but the C API builder methods take `&mut self`

ssweetney-spok: > We haven't done (1) because the rust builder methods are Self -> Self but the C API builder methods take &mut self

Yea, I remember you mentioned that [here](https://github.com/aranya-project/aranya/pull/357#discussion_r2175953752)

Also, I'd have to change from having nested builders and methods that return mutable references to them like [this](https://github.com/aranya-project/aranya/pull/373/files#diff-50e1de8616d8eeba4d987fddc6dcf64eeaad5727a8e5b3dbea8c9457ad5d2c06R96-R121).

It'd be easier if the *QuicSyncConfigBuilders weren't exposed at all and we just moved all those methods into the *TeamConfigBuilders. That would make the function/method names even longer...

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: Update doc includes:

> By default, the QUIC syncer config is not set. This function must be called
 > in order to initialize the builder with default values. Subsequent calls will
 > return a pointer to the existing config builder.

examples/rust/src/main.rs
```diff
@@ -246,17 +246,32 @@ async fn main() -> Result<()> {
...
-            .quic_sync(qs_cfg)
-            .team_id(team_id)
-            .build()?
+    let (add_team_cfg, owner_team_cfg_builder) = {
+        let mut team_cfg_builder = AddTeamConfig::builder();
+
+        team_cfg_builder.quic_sync().seed_ikm(seed_ikm);
+        team_cfg_builder = team_cfg_builder.team_id(team_id);
+
+        let builder_clone = team_cfg_builder.clone();
```
gknopf-aranya, line 255: The way this builder currently works is a bit awkward. It returns a quic sync config builder which has a method invoked on it, then reassigns a new builder to the original builder variable `team_cfg_builder`. It would be more idiomatic to construct the builder by setting methods which return the builder type. If there is a complex field to set on the builder that needs its own builder, then construct that field first using its own builder pattern before setting it on the original builder.

Then the code simplifies to this:
```suggestion
        let quic_sync_builder = AddTeamQuicSyncConfig:builder().seed_ikm(seed_ikm);
        let team_cfg_builder = AddTeamConfig::builder().team_id(team_id).quic_sync(quic_sync_builder);

        let builder_clone = team_cfg_builder.clone();
```
or this:
```suggestion
        let team_cfg_builder = AddTeamConfig::builder()
            .team_id(team_id)
            .quic_sync(AddTeamQuicSyncConfig:builder().seed_ikm(seed_ikm));

        let builder_clone = team_cfg_builder.clone();
```

examples/rust/src/main.rs
```diff
@@ -246,17 +246,32 @@ async fn main() -> Result<()> {
...
+    };
+
+    let admin_team = {
+        // This is where the owner could create a TeamInfo struct, serialize it,
+        // and transmit it to the admin through an out-of-band process.
+        let team_info = {
+            let input = owner_team_cfg_builder.to_team_info();
+            let serialized = postcard::to_allocvec(&input)?;
+            postcard::from_bytes(&serialized)?
+        };
```
gknopf-aranya:
```suggestion
        let team_info = {
            let team_info_for_peer = owner_team_cfg_builder.to_team_info();
            let serialized_team_info = postcard::to_allocvec(&team_info_for_peer)?;
            // This is where the serialized `TeamInfo` bytes would be transmitted to the peer.
            postcard::from_bytes(&serialized_team_info)?
        };
```

examples/c/example.c
```diff
@@ -421,16 +421,25 @@ AranyaError init_team(Team *t) {
...
+
+        AranyaAddTeamConfigBuilder build;
+        err = aranya_add_team_config_builder_init(&build);
+        if (err != ARANYA_ERROR_SUCCESS) {
+            fprintf(stderr, "unable to init `AranyaAddTeamConfigBuilder`\n");
+            return err;
+        }
+
+        AranyaAddTeamQuicSyncConfigBuilder *quic_build;
+        err = aranya_add_team_config_builder_get_quic_syncer(&build, &quic_build);
```
gknopf-aranya, line 433: Same feedback as for the Rust example. Build the quic sync config builder first, then pass it into the add team config builder.

examples/c/example.c
gknopf-aranya: We could shorten `config_builder` to `cfg_bldr`

examples/rust/src/main.rs
```diff
@@ -246,17 +246,32 @@ async fn main() -> Result<()> {
...
+
+        (team_cfg_builder.build()?, builder_clone)
+    };
+
+    let admin_team = {
+        // This is where the owner could create a TeamInfo struct, serialize it,
+        // and transmit it to the admin through an out-of-band process.
+        let team_info = {
+            let input = owner_team_cfg_builder.to_team_info();
+            let serialized = postcard::to_allocvec(&input)?;
```
gknopf-aranya, line 264: Providing a serialization method on `TeamInfo` would be more convenient for the user:
```suggestion
            let team_info = owner_team_cfg_builder.to_team_info();
            let serialized_team_info = team_info.serialize();
```

examples/rust/src/main.rs
ssweetney-spok: This seems similar to the convo we had above? I have some thoughts that I expressed in this thread [here](https://github.com/aranya-project/aranya/pull/373#discussion_r2208851385).

I'm curious to here other people's thoughts on this as well

examples/rust/src/main.rs
ssweetney-spok: Do Rust projects typically add serialization/de-serialization methods for types that already implement `serde::Serialize` and `serde::Deserialize`?

examples/rust/src/main.rs
ssweetney-spok: This also works. I think I didn't write this initially because I wanted to separate the team id field being set from the creation of the builder.

```suggestion
        let mut team_cfg_builder = AddTeamConfig::builder().team_id(team_id);

        team_cfg_builder.quic_sync().seed_ikm(seed_ikm);

        let builder_clone = team_cfg_builder.clone();
```

gknopf-aranya: "We should probably have a single way to build the object rather than multiple. That seems less prone to user error. I can revert the changes if we don't want the AddTeamQuicSyncConfig values to be modifiable after it's already set in the AddTeamConfigBuilder."

That complexity isn't needed for this specific Rust example so I'd prefer to simplify it here.
If there is a case where that extra complexity is needed, we can show an example of it or hide it internally so users don't see it.

gknopf-aranya: While the above may work, it requires the ` team_cfg_builder.quic_sync().seed_ikm(seed_ikm);` to be performed last.
In general, builder methods should be able to be invoked in any order.

gknopf-aranya: @jdygert-spok do you have thoughts on this?

examples/rust/src/main.rs
gknopf-aranya: I haven't seen it before. Usually serialization is done before putting data for an internal protocol onto the wire so users wouldn't need to perform the serialization themselves.
It's probably not very common to allow a user to serialize data as part of a project's public crate API.

examples/rust/src/main.rs
ssweetney-spok: > While the above may work, it requires the ` team_cfg_builder.quic_sync().seed_ikm(seed_ikm);` to be performed last. In general, builder methods should be able to be invoked in any order.

Yea, that is an unfortunate downside.

> That complexity isn't needed for this specific Rust example so I'd prefer to simplify it here.
If there is a case where that extra complexity is needed, we can show an example of it or hide it internally so users don't see it.

I'm not sure I understand. By "complexity" do you mean being able to modify the values on the inner `AddTeamQuicSyncConfigBuilder` from an `AddTeamConfigBuilder`?

gknopf-aranya: "I'm not sure I understand. By "complexity" do you mean being able to modify the values on the inner AddTeamQuicSyncConfigBuilder from an AddTeamConfigBuilder?"

Yes, in this particular Rust example the `AddTeamQuicSyncConfigBuilder` doesn't need to be modified after it is set on the `AddTeamConfigBuilder` so it's simpler to follow an idiomatic builder pattern where both builders are initialized up front.

ssweetney-spok: Oh, I see. That's what I getting at when I said that there should be a "single way to build the object rather than multiple", i.e. not both `AddTeamConfigBuilder::quic_sync` and `AddTeamConfigBuilder::get_quic_sync`. I'm OK with going with the former, users just would have to override the entire `AddTeamQuicSyncConfigBuilder`when wanting to modify values.

gknopf-aranya: That sounds like a good approach

examples/rust/src/main.rs
gknopf-aranya: If we want users to call this themselves: `postcard::to_allocvec(&team_info)?;` we'll probably want to document it somewhere in the `TeamInfo` docs and/or `rustdoc` examples. I wouldn't expect everyone to figure that out by looking at the `Serialize` impls in the source code.

ssweetney-spok: Added that in the docs.

Normally, it'd show up in the "Trait implementations" section for the item in the docs like [this](https://docs.rs/aranya-daemon-api/latest/aranya_daemon_api/enum.AddSeedMode.html#trait-implementations) but it's not appearing for the new type alias. I guess because it's a type alias of a type alias?

PR #374, opened 2025-07-02
codespace: rust example

gknopf-aranya:
Add codespace with Rust installed to the repo.

References:
https://github.com/codespaces-examples/base
https://hub.docker.com/_/rust

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
         "rust-lang.rust-analyzer"
       ],
       "settings": {
-        "terminal.integrated.shell.linux": "/usr/bin/zsh"
+        "terminal.integrated.defaultProfile.linux": "zsh"
       }
     }
   },


gknopf-aranya:
Latest commit allowed the example to run inside the codespace:
Screenshot 2025-07-02 at 3 44 54 PM

gknopf-aranya:
We may want to copy over the license, although the original has been substantially modified and could be modified more for the current use-case.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+## update and install some things we should probably have
+apt-get update
```
gknopf-aranya: These could be included in the container rather than installed at runtime.

jdygert-spok (approved):

ycarmy (approved):

PR #378, opened 2025-07-03
quic sync: cleanup daemon quic sync cfg

gknopf-aranya:
Enable the QUIC syncer by default.
Replace `"quic_sync": {}` with;
`"quic_sync": { "enabled": true }`
or
`"quic_sync": { "enabled": false }`

Add tests for various possible JSON configurations to ensure the default is set properly and the QUIC syncer must be explicitly disabled for it not to initialize.

### DIFF
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -533,7 +533,6 @@ mod tests {
             logs_dir: work_dir.join("logs"),
             config_dir: work_dir.join("config"),
             sync_addr: any,
-            quic_sync: None,
             afc: Some(AfcConfig {
                 shm_path: "/test_daemon1".to_owned(),
                 unlink_on_startup: true,
@@ -542,6 +541,7 @@ mod tests {
                 max_chans: 100,
             }),
             aqc: None,
+            ..Config::default()
         };
         for dir in [
             &cfg.runtime_dir,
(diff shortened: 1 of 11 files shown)


crates/aranya-daemon/src/daemon.rs
```diff
@@ -481,7 +487,53 @@ mod tests {
...
+        let any = Addr::new("localhost", 0).expect("should be able to create new Addr");
+        let cfg = Config {
+            name: "name".to_string(),
+            runtime_dir: work_dir.join("run"),
+            state_dir: work_dir.join("state"),
+            cache_dir: work_dir.join("cache"),
+            logs_dir: work_dir.join("logs"),
+            config_dir: work_dir.join("config"),
+            sync_addr: any,
+            quic_sync: None,
```
gknopf-aranya: This technically works based on the daemon's startup logic, but it would be nice to have a way to set this to `quic_sync: Some(QuicSyncConfig::default())` more explicitly.
I explored adding a `Default` impl for the rest of the `Config` struct, but the `quic_sync` field seems to be the only field that needs it.

gknopf-aranya: The only thing preventing deriving `Default` on the `Config` type is `Addr`.
We could make a custom `Default` impl for `Config` that initializes `sync_addr` to a default if it's not set.
Or, impl `Default` for `Addr` which I don't think is a good idea if we want to make sure network addresses are intentionally set to certain values.

gknopf-aranya: Pushed a commit I'm relatively happy with to solve this.

djohnson-spideroak:
should probably wait until after #380

gknopf-aranya:
Closing for now. May possibly reopen after https://github.com/aranya-project/aranya/pull/380, although it will likely be a new PR then.

PR #380, opened 2025-07-03
daemon: use toml for config and reorganize config

### LINKED ISSUE #338: aranya-daemon: revisit empty config objects
For aqc and quic syncer we have empty config objects. For quic-syncer it is used as enable/disable, and currently required. IMO `"quic_syncer": {}` is pretty awkward in the config. Should we have some kind of enabled flag, in or out of the object? Should we enable it by default?

jdygert-spok:
Switches the config to use toml and closes #338.


crates/aranya-daemon/src/config/enable.rs
```diff
@@ -0,0 +1,126 @@
+use serde::{de, ser, Deserialize, Serialize};
+
+/// Serde wrapper for an `enable` field on an optional.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct Enable<T>(pub Option<T>);
```
jdygert-spok: Looking for a better name. I think `Enableable` is more accurate but too much of a mouthful.

I think also I will switch to an explicit enum instead of holding option.

crates/aranya-daemon/example.toml
```diff
@@ -0,0 +1,87 @@
...
+# For example this could be `/etc/aranya`.
+#
+# See also = systemd `ConfigDirectory=` and
+# `$XDG_CONFIG_HOME`.
+config_dir = "/etc/aranya"
+
+[sync.quic]
+enable = true
+# Aranya sync server address.
+addr = "0.0.0.0:4321"
```
jdygert-spok, line 90: Moved `addr` to under the `quic` syncer since presumably each syncer would have its own address. Also moved the `quic` sync config under a `sync` heirarchy.

crates/aranya-daemon/example.toml
```diff
@@ -0,0 +1,87 @@
...
+#
+# This directory should be unique for each instance of the
+# daemon.
+#
+# # Example
+#
+# For example this could be `/var/cache/aranya`.
+#
+# See also = systemd `CacheDirectory=` and
+# `$XDG_CACHE_HOME`.
```
djohnson-spideroak, line 51: we should specify if the cache is temporary or needs to be preserved.

jdygert-spok:
`serde_json` doesn't support json5 comments and maybe trailing commas, which I think is why we ended up using hjson with `deser_hjson`, but that is a pretty unstandard format. Our config is not very nested and is human written so I think toml is a much better fit.

crates/aranya-client/tests/common/mod.rs
```diff
@@ -175,19 +175,17 @@ impl DeviceCtx {
...
             runtime_dir: work_dir.join("run"),
             state_dir: work_dir.join("state"),
             cache_dir: work_dir.join("cache"),
             logs_dir: work_dir.join("log"),
             config_dir: work_dir.join("config"),
-            sync_addr: addr_any,
-            afc: None,
-            aqc: None,
-            quic_sync,
+            aqc: daemon_cfg::AqcConfig {},
```
jdygert-spok: `aqc` is still empty and optional, not sure what we want to do with it.

crates/aranya-daemon/src/config.rs
```diff
@@ -181,46 +178,33 @@ impl Config {
...
+    let buf = fs::read_to_string(path.as_ref())?;
+    Ok(toml::from_str(&buf)?)
 }
 
-/// AFC configuration.
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+/// Sync configuration
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
-pub struct AfcConfig {
```
jdygert-spok, line 193: Removed `AfcConfig` since we completely tore AFC out a while ago.

crates/aranya-daemon/src/config/enable.rs
djohnson-spideroak: maybe Toggleable? also a bit hard to say. I think its probably file, or Enableable. no strong preference.

crates/aranya-client/tests/common/mod.rs
djohnson-spideroak: we are probably going to end up with cert-related fields post-MVP correct? I think its fine to leave blank. We should probably be erroring or warning if there are no data planes enabled.

crates/aranya-daemon/src/config/enable.rs
jdygert-spok: I switched to `Toggle`, simple enough and removes the repetition of `Enabled::Enable`.

crates/aranya-daemon/example.toml
jdygert-spok: "non-essential" implies it and presumably the "see also" references are more thorough. But we could expand on it.

elagergren-spideroak: It's non-essential cache data, so it's transient.

crates/aranya-client/tests/common/mod.rs
elagergren-spideroak: > we are probably going to end up with cert-related fields post-MVP correct?

Maaaybe? I'm not exactly sure right now. From the little bit I've thought about it so far, I think it'll work very similarly to how PSKs currently work.

crates/aranya-daemon/src/config/toggle.rs
```diff
@@ -0,0 +1,122 @@
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Toggle<T> {
+    /// An enabled option.
+    ///
+    /// This will serialize as `{ enable: true, ...T }`.
+    Enabled(T),
+    /// A disabled option.
+    ///
+    /// This will serialize as `{ enable: false }`.
+    Disabled,
```
elagergren-spideroak, line 14: nit: #[default] + derive(Default)

jdygert-spok: macro expansion tells me that requires `T: Default` but it doesn't error out when I change it to that even though `QuicSyncConfig: !Default` :thinking:

jdygert-spok: Turns out rust-analyzer's expansion is wrong. :shrug:

crates/aranya-client/tests/common/mod.rs
gknopf-aranya: I'm also under the impression that QUIC certs will be configured similarly to how QUIC PSKs are currently configured via the Aranya client Rust/C APIs and not via the daemon config.

crates/aranya-daemon/example.toml
```diff
@@ -0,0 +1,87 @@
+# The name of the daemon used for logging and debugging
+# purposes.
+name = "name"
```
gknopf-aranya: If someone sees the default value of this field output in a tracing log, defaulting to "deamon" would be more descriptive than "name".
```suggestion
name = "daemon"
```

crates/aranya-daemon/src/config/toggle.rs
gknopf-aranya, line 1: Adding this toggle type seems like more complexity than what is needed when an optional value is enough.
If there is no `[sync.quic]` section in the TOML file, or the optional value is set to `None`, we can assume the QUIC syncer is disabled.
If there is a `[sync.quic]` section in the TOML file, or the optional value is set to `Some`, we can assume the QUIC syncer is enabled.

jdygert-spok: I thought we agreed on `enable = true` for optional components.

gknopf-aranya: That was when we were using a JSON config format. I'm not sure it still makes sense if we're using the TOML format.

jdygert-spok: Discussed in call, agreed to keep `enable` for explicitness.

crates/aranya-client/tests/common/mod.rs
jdygert-spok: Should we have `aqc.enable` as well?

djohnson-spideroak: > Should we have `aqc.enable` as well?

I think so? It would be good to configure the syncers in the client configs.

jdygert-spok: Added `aqc.enable` and made it required for now. (It should be backward compatible to later allow disabling it)

Issue #327, opened 2025-06-24
example: send file via AQC


gknopf-aranya:
Update the Rust example to send a file via AQC.

Requirements:
1. Load the file in parts
2. Send the file via AQC in smaller parts
3. Receive the data via AQC on another peer and save it to a file
4. Close the file on both peers after it has been transmitted
5. Check that the file received matches the file sent (e.g. by checking hash). First iteration of this could check that size is correct.

cc: @recluse-so, @djohnson-spideroak, @ycarmy

gknopf-aranya:
Issue is obsolete since we no longer have AQC

PR #381, opened 2025-07-07
example: transmit file in multiple chunks and AQC streams

recluse-so:
# Overview
📦 Add chunked file transfer example demonstrating Aranya's secure communication

⚡ The example showcases team creation, role-based access control, label assignment, and secure peer-to-peer file transfer with progress tracking and error handling.

# Implementation
🔐 Implements a complete file transfer demo with chunked data transmission, metadata handling, and file reassembly using Aranya's AQC bidirectional channels. Includes optimal chunk sizing based on file size and CPU cores, CBOR serialization for chunk metadata, and proper file reconstruction on the receiver side.

--- Edit (Tue July 8th) --- 

### Overview
Implements a streaming file transfer system using Aranya's AQC protocol with file integrity verification.

### Key Features
- **Streaming transfer**: Files sent in chunks without loading entire file into memory
- **File integrity**: Hash verification ensures transferred files are complete and uncorrupted
- **Multi-stream**: Parallel streams for better performance
- **File ID system**: Efficient protocol for multi-file support

### Technical Details
- Uses Aranya's authenticated channels for secure transfer
- Buffered I/O for optimal disk performance
- File metadata sent once, chunks use compact IDs
- Memory usage: O(chunk_size) regardless of file size

### Testing
✅ File transfer with integrity verification  
✅ Multi-chunk streaming
✅ Hash verification working correctly  
✅ AQC channel establishment and cleanup  

# [Issue](https://github.com/aranya-project/aranya/issues/327#issuecomment-3001328904)


examples/rust/files/chunk_sender/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "sender"
+description = "Example Rust program which sends a file to a receiver using the Aranya client library"
```
gknopf-aranya:
```suggestion
description = "Example Rust program which sends a file to a receiver using AQC"
```

examples/rust/files/sender/src/main.rs
gknopf-aranya, line 1: I don't see a reason to keep this file transmission example around when there is a better example that splits files into chunks.

examples/rust/files/chunk_sender/src/main.rs
gknopf-aranya, line 1: Could you add a `README.md` for this example so the reader knows how to run it and what the example a doing?

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    operator_team.assign_label(sender.id, file_transfer_label, op).await?;
+    info!("assigning label to receiver");
+    operator_team.assign_label(receiver.id, file_transfer_label, op).await?;
+
+    // wait for syncing.
+    sleep(sleep_interval).await;
+
+    // Read the file to transfer
+    info!("reading file to transfer: {:?}", file_path);
+    let file_content = fs::read(&file_path).await
```
gknopf-aranya: This line is still reading the entire file into memory in one shot:
https://doc.rust-lang.org/std/fs/fn.read.html

`io::read()` is a way to read a specified number of bytes from a source:
https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read

A buffered reader is more efficient because previous data read from disk will be buffered in memory:
https://doc.rust-lang.org/std/io/struct.BufReader.html

gknopf-aranya: I recommend calling the buffered file reader to read a chunk of bytes from the file into a buffer right before sending the file chunk via AQC.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+        .with_context(|| format!("failed to read file: {:?}", file_path))?;
+    let file_bytes = Bytes::from(file_content);
```
gknopf-aranya, line 422: Use file metadata to get the length of the file without needing to read the entire file into memory:
https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    let file_size = file_bytes.len();
```
gknopf-aranya: Use file metadata to get the length of the file without needing to read the entire file into memory:
https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    // receiver receives channel stream created by sender.
+    info!("receiver receiving aqc bidi stream");
+    let mut receiver_stream = received_aqc_chan
+        .receive_stream()
+        .await
+        .assume("stream not received")?;
+
+    // receiver receives data from stream in chunks
+    info!("receiver receiving file data in chunks");
+    let mut received_chunks: Vec<Option<Vec<u8>>> = vec![None; total_chunks as usize];
```
gknopf-aranya: This code is storing all the received chunks in memory.
It would be better to write each chunk to the file and then receive the next chunk.
The buffer for receiving each chunk can be reused.
https://doc.rust-lang.org/std/io/struct.BufWriter.html

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+        // Receive chunk data
+        let chunk_data = stream.receive().await?.assume("no chunk data received")?;
+        total_received += chunk_data.len();
+        received_chunks.push(chunk_data.clone());
+        info!("received chunk {}: {} bytes", chunk_index, chunk_data.len());
+        
+        chunk_index += 1;
+        
+        // If we received a small chunk, it's likely the last one
+        if chunk_data.len() < chunk_size {
```
gknopf-aranya: What if all the chunks are the same size? This loop needs a way to terminate based on receiving the expected number of chunks or total bytes.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    // Keep receiving streams until we get all chunks
+    loop {
+        info!("receiving stream {}", chunk_index);
+        
+        // Receive stream
+        let mut stream = received_channel.receive_stream().await
+            .assume("stream not received")?;
+        
+        // Receive chunk data
+        let chunk_data = stream.receive().await?.assume("no chunk data received")?;
```
gknopf-aranya: `read_to_end()` is a more reliable way to read all the data from a stream, but our API doesn't currently expose that method.
https://docs.rs/futures-util/0.3.31/futures_util/io/trait.AsyncReadExt.html#method.read_to_end

Without that, the correct way to read all the data from the stream would be to loop over `stream.receive()` until it returns `None` indicating the stream has been closed:
```
    /// Receive the next available data from a stream. If the stream has
    /// been closed, return None.
    ///
    /// This method will block until data is available to return.
    /// The data is not guaranteed to be complete, and may need to be called
    /// multiple times to receive all data from a message.
    pub async fn receive(&mut self) -> Result<Option<Bytes>, AqcError> {
        match self {
            AqcPeerStream::Bidi(s) => s.receive().await,
            AqcPeerStream::Receive(s) => s.receive().await,
        }
    }
```

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    layer::{Context, Filter},
+    prelude::*,
+    EnvFilter,
+};
+use serde::{Serialize, Deserialize};
+use serde_cbor;
+
+#[derive(Serialize, Deserialize, Debug)]
+struct FileChunk {
+    filename: String,
```
gknopf-aranya: The file name only needs to be sent with the first chunk. After that, an index/ID could be used to associate all the file chunks with each other.
If multiple files/chunks are sent via the AQC channel in parallel, the file index/ID could be used to keep track of which file each transmitted file chunk belongs to.
```suggestion
    filename: Option<String>,
```

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    chunk_index: u64,
+    chunk_size: usize,
```
gknopf-aranya: The size of a received file chunk can be computed from `data.len()` so this doesn't need to be transmitted.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    total_chunks: u64,
```
gknopf-aranya: This only needs to be sent with the first chunk.
```suggestion
    total_chunks: Option<u64>,
```
Or instead of defining optional fields, different types for the first file chunk and the rest of the file chunks could be defined. E.g. `FirstFileChunk` and `FileChunk`.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    file_size: u64,
```
gknopf-aranya: Ditto

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+use tracing_subscriber::{
+    layer::{Context, Filter},
+    prelude::*,
+    EnvFilter,
+};
+use serde::{Serialize, Deserialize};
+use serde_cbor;
+
+#[derive(Serialize, Deserialize, Debug)]
+struct FileChunk {
```
gknopf-aranya: I'd recommend sending a hash of the file with the first file chunk that can be checked after the file has been received and written to disk.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+    let file_content = fs::read(&file_path).await
+        .with_context(|| format!("failed to read file: {:?}", file_path))?;
+    let file_bytes = Bytes::from(file_content);
+    let file_size = file_bytes.len();
+    info!("file size: {} bytes", file_size);
+
+    info!("starting file transfer");
+    info!("sender creating aqc bidi channel");
+    
+    // Prepare arguments that need to be captured by the async move block
```
gknopf-aranya: Let's move the code for establishing the AQC channel (both creating and receiving the channel) above any code related to sending the file via the AQC channel.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+
+
+fn optimal_chunk_size(file_size: usize) -> usize {
+    let min_chunk = 8 * 1024; // 8 KiB
+    let max_chunks = 128.max(num_cpus::get() * 4);
+    let chunk = (file_size + max_chunks - 1) / max_chunks;
+    chunk.max(min_chunk)
+}
+
+async fn send_file_over_multiple_streams(
```
gknopf-aranya: This method is currently unused

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,654 @@
...
+        chunk_index += 1;
+        
+        info!("Sent chunk {}: {}/{} bytes", chunk_index - 1, total_sent, file_size);
+    }
+
+    info!("File transfer completed! Sent {} chunks, {} total bytes", chunk_index, total_sent);
+    Ok(())
+}
+
+async fn receive_file_from_multiple_streams(
```
gknopf-aranya: This method is currently unused

examples/rust/files/chunk_sender/src/main.rs
gknopf-aranya: I like having the logic for sending and receiving files in their own methods.
Just need to make sure those methods are invoked by the example.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,638 @@
...
+    } else {
+        chunk_size
+    }
+}
+
+async fn send_file_over_multiple_streams(
+    file_path: &Path,
+    channel: &mut AqcBidiChannel,
+    chunk_size: usize,
+) -> Result<()> {
```
gknopf-aranya: The chunk size can be computed within this function:
```suggestion
async fn send_file_over_multiple_streams(
    file_path: &Path,
    channel: &mut AqcBidiChannel,
) -> Result<()> {
```

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,638 @@
...
+
+    info!("File transfer completed! Sent {} chunks, {} total bytes", chunk_index, total_sent);
+    Ok(())
+}
+
+async fn receive_file_from_multiple_streams(
+    output_path: &Path,
+    chunk_size: usize,
+    received_channel: &mut AqcBidiChannel,
+) -> Result<()> {
```
gknopf-aranya: The chunk size does not need to be passed into this function. Each chunk received will have a known size.
The file path also does not need to be passed in. There should be a directory configured where received files are written. The filename will be passed in with the first received file chunk.
```suggestion
async fn receive_file_from_multiple_streams(
    received_channel: &mut AqcBidiChannel,
) -> Result<()> {
```

examples/rust/files/chunk_sender/src/main.rs
gknopf-aranya, line 1: nit: the file transfer code is complex enough that it could be moved to its own module

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+    info!("Starting multi-threaded file transfer: {} chunks, {} bytes", total_chunks, file_size);
+    
+    // Read all chunks into memory first
+    let mut file = File::open(file_path).await?;
+    let mut chunks = Vec::new();
+    let mut chunk_index = 0;
+    
+    loop {
+        let mut buffer = vec![0u8; chunk_size];
+        let bytes_read = file.read(&mut buffer).await?;
```
gknopf-aranya: This file `read()` is not currently buffered, which means it will need to read directly from disk each time.
Better to use a buffered reader.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+        
+        if bytes_read == 0 {
+            break;
+        }
+        
+        chunks.push((chunk_index, buffer[..bytes_read].to_vec()));
```
gknopf-aranya: This is pushing each file chunk onto a buffer which means the entire file will be in memory (even though each chunk is read from the file separately). Better to send each chunk first before reading the next one:
```
loop {
   file.read()
   ...
   stream.send()
}
```

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+        let mut stream = channel.create_bidi_stream().await?;
+        stream.send(Bytes::copy_from_slice(&chunk_data)).await?;
+        stream.close().await?;
+        
+        total_sent += chunk_data.len();
+        info!("Sent chunk {}: {}/{} bytes", chunk_idx, total_sent, file_size);
+    }
+    
+    // Send EOF marker
+    let mut eof_stream = channel.create_bidi_stream().await?;
```
gknopf-aranya: No need to send EOF in another stream.
On the receiving side, check if `chunk index == total chunks - 1`

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+            break;
+        }
+        
+        if stream_data.is_empty() {
+            info!("stream {}: no data received, assuming end of transfer", chunk_index);
+            break;
+        }
+        
+        total_received += stream_data.len();
+        received_chunks.push(stream_data);
```
gknopf-aranya: This is pushing each chunk onto a vector which means the receiver needs to store the entire file in memory. Better to write each chunk to the file as it's received:
```
loop {
   stream.receive(...)
   ...
   file.write(...)
}
```
Rust will automatically drop objects that fall out of scope or reuse the buffer used to receive chunks so that only a single chunk of data is stored in memory at a given time.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+        info!("received chunk {}: {} bytes", chunk_index, received_chunks.last().unwrap().len());
+        
+        chunk_index += 1;
```
gknopf-aranya: The chunk index should be read from the corresponding field of the `Chunk` that was received.
We cannot assume all the file chunks were received in a particular order if they are sent in parallel via multiple streams.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+    info!("completed Aranya file transfer application");
+
+    Ok(())
+}
+
+
+
+fn optimal_chunk_size(file_size: usize) -> usize {
+    // Maximum available memory is 1GB, so max chunk size is 1/15 of 1GB to leave room for system and other services
+    let max_chunk_size = (1024 * 1024 * 1024) / 15; // ~68 MB
```
gknopf-aranya: https://docs.rs/byte-unit/latest/byte_unit/index.html

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+    let min_chunk = 8 * 2; // 16 KiB
+    
+    // Ensure at least 4 chunks by limiting chunk size to a quarter of the file size
+    let max_chunk_for_min_4 = file_size / 4;
```
gknopf-aranya: If the file is small enough, it should be okay to send the whole thing in a single chunk

examples/rust/files/chunk_sender/src/main.rs
recluse-so: Will do.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,673 @@
...
+    operator_team.assign_label(receiver.id, file_transfer_label, op).await?;
+
+    // wait for syncing.
+    sleep(sleep_interval).await;
+
+    // Read the file to transfer
+    info!("reading file to transfer: {:?}", file_path);
+    let file_size = fs::metadata(&file_path).await?.len() as usize;
+    let chunk_size = optimal_chunk_size(file_size);
+    let total_chunks = ((file_size + chunk_size - 1) / chunk_size) as u64;
```
gknopf-aranya: If `file_size` or `total_chunks` is zero, there's no need to attempt to send the file.

examples/rust/files/chunk_sender/src/main.rs
```diff
@@ -0,0 +1,741 @@
...
+    let _filename_str = file_path.file_name().unwrap().to_string_lossy().to_string();
+
+    info!("starting file transfer");
+    // Use the new multi-stream approach
+    info!("sender sending file over multiple streams");
+    send_file_over_multiple_streams(&file_path, &mut created_aqc_chan, chunk_size).await?;
+
+    // receiver receives data from multiple streams
+    info!("receiver receiving file data from multiple streams");
+    let output_filename = file_path.file_name()
```
gknopf-aranya: Get this filename from the first file chunk sent. The receiver won't have access to this variable when this example is split into multiple nodes.

examples/rust/files/chunk_sender/send_file.sh
```diff
@@ -0,0 +1,33 @@
...
+  FILE_PATH="$SCRIPT_DIR/data/test.yaml"
+fi
+
+cd "$WORKSPACE_ROOT" || exit 1
+
+echo "Building aranya-daemon..."
+cargo build --release --bin aranya-daemon
+
+echo "Building sender example..."
+cargo build --release --manifest-path examples/rust/files/sender/Cargo.toml --bin sender
```
gknopf-aranya: This is building the other example:
```suggestion
cargo build --release --manifest-path examples/rust/files/chunk_sender/Cargo.toml --bin sender
```

examples/rust/files/chunk_sender/send_file.sh
```diff
@@ -0,0 +1,33 @@
...
+set -xeuo pipefail
+
+# Usage: ./send_file.sh [path/to/file]
+# If no file is provided, defaults to data/test.yaml
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+WORKSPACE_ROOT="$SCRIPT_DIR/../../../../"
+
+DAEMON_PATH="$WORKSPACE_ROOT/target/release/aranya-daemon"
+SENDER_PATH="$WORKSPACE_ROOT/examples/rust/files/sender/target/release/sender"
```
gknopf-aranya: This is the path for the other example:
```suggestion
SENDER_PATH="$WORKSPACE_ROOT/examples/rust/files/chunk_sender/target/release/sender"
```

examples/rust/files/sender/src/main.rs
```diff
@@ -0,0 +1,491 @@
...
+    let received_bytes = receiver_stream.receive().await?.assume("no data received")?;
+    assert_eq!(received_bytes, file_bytes);
+    info!("file transfer completed successfully!");
+    
+    // Write the received file to disk in target/tmp directory
+    let output_filename = file_path.file_name()
+        .context("input file has no filename")?
+        .to_str()
+        .context("filename is not valid UTF-8")?;
+    let output_path = Path::new("target/tmp").join(output_filename);
```
gknopf-aranya: `tempdir` and `tempfile` can be used to create temporary files/directories:
https://docs.rs/tempfile/latest/tempfile/
The `target` dir is for build artifacts.

examples/rust/files/chunk_sender/README.md
```diff
@@ -0,0 +1,174 @@
+# Aranya Chunk Sender Example
+
+A demonstration of secure file transfer using Aranya's peer-to-peer communication framework. This example showcases chunked file transmission with metadata handling, optimal chunk sizing, and secure team-based access control.
```
gknopf-aranya:
```suggestion
This example demonstrates how to securely transmit a file via AQC with team-based access control.
Each file is split into multiple chunks which are streamed via AQC streams so that memory usage is low.
```

examples/rust/files/chunk_sender/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+For detailed debugging, run with debug logging:
+```bash
+ARANYA_EXAMPLE=debug ./target/release/sender <daemon_path> <file_path>
+```
+
+## Related Examples
+
+- [Basic Sender](../sender/): Simple file transfer example
+- [C Examples](../../c/): C language bindings and examples
```
gknopf-aranya: Don't need to mention the C example in the Rust example folder

jdygert-spok:
This seems over-complicated. QUIC already gives us multiple in-order byte streams so for each file we want to send, we can open a stream, write the header (probably just filename), write all the contents of the file in chunks directly, then close the stream.

elagergren-spideroak:
Claude?

PR #385, opened 2025-07-08
Multi-node rust provisioning example

jdygert-spok:


gknopf-aranya:
@jdygert-spok this PR is stale and is partially replaced by the following PR (minus containerization):
https://github.com/aranya-project/aranya/pull/443
Should we close it?

jdygert-spok:
I think it's worth updating this.

PR #386, opened 2025-07-09
client: eagerly implement Debug trait in public API

gknopf-aranya:
Eagerly implement `Debug` trait in the client's public Rust API:
https://rust-lang.github.io/api-guidelines/checklist.html

This PR adds whatever `Debug` traits can be easily added.
There are some cases where implementing traits would be more involved because the underlying types in `aranya-core` or a 3rd party lib would also need to implement those traits. This PR does not attempt to handle those cases.

### DIFF
--- a/crates/aranya-client/src/aqc/net.rs
+++ b/crates/aranya-client/src/aqc/net.rs
@@ -65,7 +65,7 @@ pub(crate) struct AqcClient {
     daemon: DaemonApiClient,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 struct ClientState {
     /// Quic client used to create channels with peers.
     quic_client: Client,
@@ -463,14 +463,14 @@ pub enum TryReceiveError<E = AqcError> {
     Closed,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone)]
 struct AqcChannelInfo {
     label_id: LabelId,
     channel_id: AqcChannelId,
 }
 
 /// An AQC Channel ID.
-#[derive(Copy, Clone, Debug)]
+#[derive(Debug, Copy, Clone)]
 enum AqcChannelId {
     Bidi(BidiChannelId),
     Uni(UniChannelId),
--- a/crates/aranya-client/src/aqc/net.rs
+++ b/crates/aranya-client/src/aqc/net.rs
@@ -463,14 +463,14 @@ pub enum TryReceiveError<E = AqcError> {
     Closed,
 }
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug)]
 struct AqcChannelInfo {
     label_id: LabelId,
     channel_id: AqcChannelId,
 }
 
 /// An AQC Channel ID.
-#[derive(Debug, Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 enum AqcChannelId {
     Bidi(BidiChannelId),
     Uni(UniChannelId),
--- a/crates/aranya-client/src/aqc/net.rs
+++ b/crates/aranya-client/src/aqc/net.rs
@@ -65,7 +65,7 @@ pub(crate) struct AqcClient {
     daemon: DaemonApiClient,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 struct ClientState {
     /// Quic client used to create channels with peers.
     quic_client: Client,
--- a/crates/aranya-client/src/client.rs
+++ b/crates/aranya-client/src/client.rs
@@ -14,7 +14,6 @@ use aranya_daemon_api::{
 };
 use aranya_util::Addr;
 use buggy::BugExt as _;
-use serde::{Deserialize, Serialize};
 use tarpc::context;
 use tokio::{fs, net::UnixStream};
 use tracing::{debug, error, info, instrument};
@@ -26,7 +25,7 @@ use crate::{
 };
 
 /// List of device IDs.
-#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone)]
 pub struct Devices {
     data: Vec<DeviceId>,
 }
@@ -44,7 +43,7 @@ impl Devices {
 }
 
 /// List of labels.
-#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone)]
 pub struct Labels {
     data: Vec<Label>,
 }
@@ -62,7 +61,7 @@ impl Labels {
 }
 
 /// Builds a [`Client`].
-#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Default)]
+#[derive(Debug, Default)]
 pub struct ClientBuilder<'a> {
     /// The UDS that the daemon is listening on.
     #[cfg(unix)]
--- a/crates/aranya-client/src/client.rs
+++ b/crates/aranya-client/src/client.rs
@@ -25,7 +25,7 @@ use crate::{
 };
 
 /// List of device IDs.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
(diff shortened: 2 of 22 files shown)


jdygert-spok (commented):
Most of these are unnecessary IMO and could lead to needing breaking changes.

gknopf-aranya:
> Most of these are unnecessary IMO and could lead to needing breaking changes.

Are there any that we want to add? Possibly `Debug`?

crates/aranya-util/src/addr.rs
```diff
@@ -45,6 +45,7 @@ macro_rules! const_assert {
 /// `Addr` ensures that the host part is a syntactically valid domain name or IP address.
 /// It provides methods for DNS lookup, conversion to socket addresses, and serde
 /// serialization/deserialization.
+// TODO: implement `Serialize`, `Deserialize`
```
jdygert-spok: `Addr` implements ser/de manually, see below.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -413,6 +415,15 @@ impl Daemon {
     }
 }
 
+impl Debug for Daemon {
```
gknopf-aranya: Wasn't able to derive `Debug` on `Daemon` because underlying types don't implement it.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -65,7 +65,7 @@ pub(crate) struct AqcClient {
     daemon: DaemonApiClient,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
jdygert-spok: Let's leave this off, there's no reason to clone this, and it could accidentally hide that we are not actually locking the client to protect against concurrent psk store usage.

crates/aranya-client/src/client.rs
```diff
@@ -25,6 +26,7 @@ use crate::{
 };
 
 /// List of device IDs.
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Default, Serialize, Deserialize)]
```
jdygert-spok: These list types shouldn't have much, we don't want to commit to them being a certain format.

crates/aranya-client/src/client.rs
```diff
@@ -59,6 +62,7 @@ impl Labels {
 }
 
 /// Builds a [`Client`].
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Default)]
```
jdygert-spok: Just debug and default.

crates/aranya-client/src/error.rs
```diff
@@ -74,7 +74,7 @@ pub enum ConfigError {
 }
 
 /// An invalid argument.
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, Clone, thiserror::Error)]
```
jdygert-spok: Probably fine here, just be wary of putting clone on more general error types since some sub errors aren't cloneable.

crates/aranya-daemon/src/config/toggle.rs
```diff
@@ -77,7 +78,7 @@ mod imp {
         }
     }
 
-    #[derive(Serialize, Deserialize)]
+    #[derive(Debug, Clone, Serialize, Deserialize)]
```
jdygert-spok: These 3 are an implementation detail, these derives not needed.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -413,6 +415,15 @@ impl Daemon {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Daemon")
+            .field("api", &self.api)
+            .field("span", &self.span)
+            .finish_non_exhaustive()
+    }
+}
```
jdygert-spok: Can probably make those other fields `Debug` and then derive this instead.

crates/aranya-util/src/task.rs
```diff
@@ -69,10 +69,12 @@ where
 
 type Panic = Box<dyn Any + Send>;
 
+#[derive(Debug, Clone)]
```
jdygert-spok: not clone

crates/aranya-daemon/src/daemon.rs
gknopf-aranya: Looks like we'll need some changes in `aranya-core` for that:
```
124 | #[derive(Debug)]
    |          ----- in this derive macro expansion
125 | pub struct Daemon {
126 |     sync_server: SyncServer,
    |     ^^^^^^^^^^^^^^^^^^^^^^^ `aranya_runtime::linear::LinearStorageProvider<aranya_runtime::linear::libc::FileManager>` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
    |
```
```
124 | #[derive(Debug)]
    |          ----- in this derive macro expansion
125 | pub struct Daemon {
126 |     sync_server: SyncServer,
    |     ^^^^^^^^^^^^^^^^^^^^^^^ `aranya_crypto::default::DefaultEngine` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
    |
```

elagergren-spideroak (changes requested):
please, let's only add `Debug` for now. Nothing else.

crates/aranya-daemon/src/config/toggle.rs
gknopf-aranya: The linter will complain if they don't implement `Debug`. I can remove `Clone`.

crates/aranya-daemon/src/daemon.rs
jdygert-spok: The sync server's debug impl shouldn't depend on its generics.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -463,14 +463,14 @@ pub enum TryReceiveError<E = AqcError> {
...
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone)]
 struct AqcChannelInfo {
     label_id: LabelId,
     channel_id: AqcChannelId,
 }
 
 /// An AQC Channel ID.
-#[derive(Copy, Clone, Debug)]
+#[derive(Debug, Copy, Clone)]
```
jdygert-spok: nit: I try to keep a consistent order, this is the order I use:

```rust
Copy, Clone, Debug, Default, Hash, Eq, PartialEq, Ord, PartialOrd, <third party>
```

crates/aranya-daemon/src/daemon.rs
gknopf-aranya: The code does not compile if I attempt to derive `Debug` on `Daemon`. `SyncServer` is defined based on generics:
`pub(crate) type SyncServer = crate::sync::task::quic::Server<EN, SP>;`

jdygert-spok: Yes, I mean the debug impl for `crate::sync::task::quic::Server` should be changed to not depend on its generics.

gknopf-aranya: How would we do this? Even a custom `Debug` impl on `Server` requires the generics:
```
error[E0107]: missing generics for struct `sync::task::quic::Server`
   --> crates/aranya-daemon/src/sync/task/quic.rs:551:16
    |
551 | impl Debug for Server {}
    |                ^^^^^^ expected 2 generic arguments
    |
note: struct defined here, with 2 generic parameters: `EN`, `SP`
   --> crates/aranya-daemon/src/sync/task/quic.rs:334:12
    |
334 | pub struct Server<EN, SP> {
    |            ^^^^^^ --  --
help: add missing generic arguments
    |
551 | impl Debug for Server<EN, SP> {}
```

gknopf-aranya: Dygert pushed a commit that uses `derive_where` for this.

gknopf-aranya:
> please, let's only add `Debug` for now. Nothing else.

@elagergren-spideroak PR has been updated to only add `Debug` impls

elagergren-spideroak (approved):

Issue #320, opened 2025-06-17
QUIC: Investigate how to get the PSK identity from the connection


ssweetney-spok:
Currently AQC and the QUIC syncer rely on `aranya_internal_rustls::server::SelectsPresharedKeys::chosen` to get the PSK identity used for the connection but this may not be reliable.

See this [comment](https://github.com/aranya-project/aranya/pull/258#discussion_r2151013984)

gknopf-aranya:
Another discussion about this:
https://github.com/aranya-project/aranya/pull/387#discussion_r2195694648

We may want to add a test with multiple simultaneous connections to see if there's a way to cause an incorrect PSK to be used.
This will make it easier to verify a potential fix.

If the client cancels a QUIC connection before the server has a chance to accept the connection, could that cause the PSK queue to be received in a different order than the server accepts the connections in?

PR #387, opened 2025-07-09
daemon: ensure spawns inherit span and refactor quic sync server

jdygert-spok:


crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -390,80 +389,64 @@ where
...
-            let active_team = {
-                let Ok(guard) = self.active_team.lock().inspect_err(|e| error!(%e)) else {
+        info!("QUIC sync server listening for incoming connections");
+
+        scope(async |s| {
+            // Accept incoming QUIC connections
+            while let Some(mut conn) = self.server.accept().await {
+                debug!("received incoming QUIC connection");
+
+                let Ok(active_team) = self.active_team_rx.try_recv() else {
```
gknopf-aranya, line 412: Does rustls guarantee `active_team_rx()` has been set before the connection is accepted?
What if there are 2 incoming connections, will the `active_team_rx` be received in the same order as the connections are accepted?
What if a connection was dropped by the client before the server has a chance to accept the connection, could this cause an incorrect active team to be received for an accepted connection?

jdygert-spok: Those are the issues I was referring to in my post.

Note that this was an issue before the refactor, although the behavior is different since it now uses the next untaken ID instead of using the latest set ID.

gknopf-aranya: If that means this approach is more reliable than the previous one and the other issues are documented as a future enhancement, then this refactor looks good to me.

gknopf-aranya (approved):

crates/aranya-daemon/src/sync/task/quic.rs
jdygert-spok: I wouldn't say _more_ reliable, just fails in a different situation.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -390,80 +389,64 @@ where
...
-            let Ok(peer) = conn.remote_addr() else {
-                error!("unable to get peer address from connection");
-                continue;
-            };
-            let client = self.aranya.clone();
-            let active_team = {
-                let Ok(guard) = self.active_team.lock().inspect_err(|e| error!(%e)) else {
+        info!("QUIC sync server listening for incoming connections");
+
+        scope(async |s| {
```
ssweetney-spok, line 407: Your `scope` function works now since JoinSet isn't being used now 👍

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: Yea, they're both "not good" options but this may be better since it keeps a queue of IDs. So it may work better in some scenarios 🤷

gknopf-aranya: As long as it's not a regression. Sounds like it improves in some cases, but not in others. So it's hard to say whether this adversely affects security overall. @djohnson-spideroak, @elagergren-spideroak are you okay with this change?

djohnson-spideroak: Does this have the potential to cause a connection to use a wrong PSK or communicate with a peer that it isnt allowed to? I assume it could cause some race conditions using the wrong team ID.

jdygert-spok: Yes, before and after this change it might be possible receive the wrong info used for the connection. I'm not sure how abusable that'd be with the current API.

gknopf-aranya: There is an open issue for this here:
https://github.com/aranya-project/aranya/issues/320

gknopf-aranya (approved):

PR #403, opened 2025-07-11
queries: fix flake

gknopf-aranya:
Queries in unit tests were sometimes executing before commands had been synced and committed to the graph.
Retry queries that don't return expected value until expected value is returned.
Timeout if query never returns expected value within a certain timeframe.

Adds more test coverage of queries while we're updating the test.
Removes unnecessary `&mut self` from `acq_net_id()`.

### DIFF
--- a/crates/aranya-client/tests/common/mod.rs
+++ b/crates/aranya-client/tests/common/mod.rs
@@ -21,8 +21,7 @@ use tokio::{fs, time};
 use tracing::{info, instrument, trace};
 
 // Default sync interval for testing.
-pub const SYNC_INTERVAL_MS: u64 = 100;
-pub const SYNC_INTERVAL: Duration = Duration::from_millis(SYNC_INTERVAL_MS);
+pub const SYNC_INTERVAL: Duration = Duration::from_millis(100);
 // Allow for one missed sync and a misaligned sync rate, while keeping run times low.
 pub const SLEEP_INTERVAL: Duration = Duration::from_millis(250);
 
--- a/crates/aranya-client/tests/common/mod.rs
+++ b/crates/aranya-client/tests/common/mod.rs
@@ -47,6 +47,9 @@ impl TeamCtx {
         let membera = DeviceCtx::new(name, "membera", work_dir.join("membera")).await?;
         let memberb = DeviceCtx::new(name, "memberb", work_dir.join("memberb")).await?;
 
+        // Wait for daemons to start up.
+        sleep(Duration::from_millis(20)).await;
+
         Ok(Self {
             owner,
             admin,
--- a/crates/aranya-client/tests/common/mod.rs
+++ b/crates/aranya-client/tests/common/mod.rs
@@ -251,7 +251,7 @@ impl DeviceCtx {
     }
 
     #[allow(unused, reason = "module compiled for each test file")]
-    pub fn aqc_net_id(&mut self) -> NetIdentifier {
+    pub fn aqc_net_id(&self) -> NetIdentifier {
         NetIdentifier(
             self.client
                 .aqc()
--- a/crates/aranya-client/tests/common/mod.rs
+++ b/crates/aranya-client/tests/common/mod.rs
@@ -21,7 +21,8 @@ use tokio::{fs, time};
 use tracing::{info, instrument, trace};
 
 // Default sync interval for testing.
-pub const SYNC_INTERVAL: Duration = Duration::from_millis(100);
+pub const SYNC_INTERVAL_MS: u64 = 100;
+pub const SYNC_INTERVAL: Duration = Duration::from_millis(SYNC_INTERVAL_MS);
 // Allow for one missed sync and a misaligned sync rate, while keeping run times low.
 pub const SLEEP_INTERVAL: Duration = Duration::from_millis(250);
 
--- a/crates/aranya-client/tests/common/mod.rs
+++ b/crates/aranya-client/tests/common/mod.rs
@@ -20,7 +20,8 @@ use backon::{ExponentialBuilder, Retryable as _};
 use tokio::{fs, time};
 use tracing::{info, instrument, trace};
 
-const SYNC_INTERVAL: Duration = Duration::from_millis(100);
+// Default sync interval for testing.
+pub const SYNC_INTERVAL: Duration = Duration::from_millis(100);
 // Allow for one missed sync and a misaligned sync rate, while keeping run times low.
 pub const SLEEP_INTERVAL: Duration = Duration::from_millis(250);
 
--- a/crates/aranya-client/tests/tests.rs
+++ b/crates/aranya-client/tests/tests.rs
@@ -149,13 +149,10 @@ async fn test_query_functions() -> Result<()> {
     // First, let's check how many devices are on the team.
     time::timeout(Duration::from_secs(1), async {
         loop {
-            let devices = queries
-                .devices_on_team()
-                .await
-                .expect("expected to query devices on team");
-            let count = devices.iter().count();
-            if count == 5 {
-                break;
+            if let Ok(devices) = queries.devices_on_team().await {
+                if devices.iter().count() == 5 {
+                    break;
+                }
             }
             sleep(Duration::from_millis(10)).await;
         }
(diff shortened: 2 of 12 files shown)


gknopf-aranya:
There is probably a similar change needed in the Rust example queries. Although I don't recall seeing the Rust example flake in CI.

djohnson-spideroak (approved):

crates/aranya-client/tests/tests.rs
```diff
@@ -128,27 +129,122 @@ async fn test_query_functions() -> Result<()> {
...
-    assert_eq!(devices.iter().count(), 5);
-    debug!("membera devices on team: {:?}", devices.iter().count());
+    time::timeout(Duration::from_secs(1), async {
+        loop {
+            if let Ok(devices) = queries.devices_on_team().await {
+                if devices.iter().count() == 5 {
+                    break;
+                }
+            }
+            sleep(Duration::from_millis(10)).await;
```
jdygert-spok: These should be at least the sync interval.

crates/aranya-client/tests/tests.rs
```diff
@@ -128,27 +129,122 @@ async fn test_query_functions() -> Result<()> {
...
     // Test all our fact database queries.
-    let memberb = team.membera.client.team(team_id);
+    let memberb = team.memberb.client.team(team_id);
     let queries = memberb.queries();
 
     // First, let's check how many devices are on the team.
-    let devices = queries.devices_on_team().await?;
-    assert_eq!(devices.iter().count(), 5);
-    debug!("membera devices on team: {:?}", devices.iter().count());
+    time::timeout(Duration::from_secs(1), async {
```
jdygert-spok: Using the extension trait is a bit nicer:
```rust
use tokio_util::time::FutureExt as _;

async {
   // ...
}
.timeout(...)
.await
```

crates/aranya-client/tests/tests.rs
gknopf-aranya: True. We alternatively call a `sync_now()` right before running a query which could remove the need for waiting.

jdygert-spok: `sync_now` doesn't wait until the sync happens, and also a sync is currently only one transfer of up to 100 commands, although here the diff _should_ be small enough.

gknopf-aranya: Opened an issue to allow polling `sync_now()`:
https://github.com/aranya-project/aranya/issues/404
That was part of the original intent of adding `sync_now()` functionality.

gknopf-aranya:
> There is probably a similar change needed in the Rust example queries. Although I don't recall seeing the Rust example flake in CI.

This flake wasn't happening in the Rust example because it sleeps for the sync interval before calling the queries.

djohnson-spideroak (approved):

crates/aranya-client/tests/tests.rs
```diff
@@ -117,6 +120,8 @@ async fn test_remove_devices() -> Result<()> {
 /// Tests functionality to make sure that we can query the fact database for various things.
 #[test(tokio::test(flavor = "multi_thread"))]
 async fn test_query_functions() -> Result<()> {
+    const TIMEOUT: Duration = Duration::from_millis(SYNC_INTERVAL_MS * 2);
```
jdygert-spok:
```suggestion
    const TIMEOUT: Duration = SYNC_INTERVAL * 2;
```

gknopf-aranya: That's what I tried originally but the compiler complained:
```
error[E0015]: cannot call non-const operator in constants
    --> crates/aranya-client/tests/tests.rs:123:31
     |
123  |     const TIMEOUT: Duration = SYNC_INTERVAL * 2;
```

jdygert-spok: Oh right, would have to either use `const fn checked_mul` or do in `let` instead. Oh well :shrug:

gknopf-aranya: `let` should work since it's only used in this method

djohnson-spideroak (dismissed):

gknopf-aranya:
@jdygert-spok ready for another look

benz-spideroak (dismissed):

crates/aranya-client/tests/tests.rs
```diff
@@ -128,27 +133,130 @@ async fn test_query_functions() -> Result<()> {
...
 
+    // TODO: #404 invoke sync_now() before queries when long-polling is supported
+
     // First, let's check how many devices are on the team.
-    let devices = queries.devices_on_team().await?;
-    assert_eq!(devices.iter().count(), 5);
-    debug!("membera devices on team: {:?}", devices.iter().count());
+    async {
+        loop {
+            if let Ok(devices) = queries.devices_on_team().await {
```
jdygert-spok: nit: the queries themselves should never fail so these should use `?` or `expect`

gknopf-aranya: The queries can fail if the fact does not exist yet. E.g. if we attempt to query the number of devices on the team and no devices have been added to the team (or the device hasn't synced those commands yet), the query will fail.

jdygert-spok: I believe it will just return an empty list.

crates/aranya-client/tests/tests.rs
```diff
@@ -128,27 +133,130 @@ async fn test_query_functions() -> Result<()> {
...
+            }
+            sleep(SYNC_INTERVAL).await;
+        }
+    }
+    .timeout(timeout)
+    .await
+    .expect("expected label to exist");
+
+    // Query labels.
+    async {
```
elagergren-spideroak: don't we already have a dependency that does retries w/ exponential backoff?

ssweetney-spok: Yea, `backon` is already in our dependency tree. I was wondering about that but I'm not sure if it'd make the code simpler

gknopf-aranya: If the `sync_now()` PR is merged before this PR, we may end up calling `sync_now()` right before performing the queries instead of polling each query.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Closing discussion as this is no longer relevant now that `sync_now()` is used in the tests.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Closing discussion as this is no longer relevant now that `sync_now()` is used in the tests.

crates/aranya-client/tests/tests.rs
```diff
@@ -126,27 +128,65 @@ async fn test_query_functions() -> Result<()> {
...
     // Test all our fact database queries.
-    let memberb = team.membera.client.team(team_id);
+    let memberb = team.memberb.client.team(team_id);
     let queries = memberb.queries();
 
+    // Wait sync interval before having membera sync with owner.
+    sleep(SYNC_INTERVAL).await;
+    memberb
+        .sync_now(team.owner.aranya_local_addr().await?.into(), None)
+        .await?;
```
jdygert-spok:
```suggestion
    memberb
        .sync_now(team.operator.aranya_local_addr().await?.into(), None)
        .await?;
```

gknopf-aranya: Tried this but it is flakey.

gknopf-aranya: Is it possible the owner needs some time to commit commands to the graph before membera syncs with it?

jdygert-spok: Did you try with syncing from the operator? The operator is the one that does all those commands above.

gknopf-aranya: That is more reliable. I must have been syncing with the operator indirectly via the owner which caused this delay to be needed before.

gknopf-aranya: There are a few other places in the tests where sleeps could be traded for `sync_now()`. I'll leave that for another PR.

jdygert-spok (dismissed):

nikki-aranya (dismissed):

crates/aranya-client/tests/common/mod.rs
```diff
@@ -46,6 +47,9 @@ impl TeamCtx {
         let membera = DeviceCtx::new(name, "membera", work_dir.join("membera")).await?;
         let memberb = DeviceCtx::new(name, "memberb", work_dir.join("memberb")).await?;
 
+        // Wait for daemons to start up.
+        sleep(Duration::from_millis(20)).await;
```
gknopf-aranya, line 51: This sleep resolved the flake I was seeing in CI. But we need a better way to confirm when the daemon sync servers have completed startup.

ssweetney-spok (approved):

crates/aranya-client/tests/common/mod.rs
gknopf-aranya: https://github.com/aranya-project/aranya/issues/411

gknopf-aranya:
Closing in favor of:
https://github.com/aranya-project/aranya/pull/412

PR #405, opened 2025-07-11
long poll on sync-now

### LINKED ISSUE #404: sync_now: support long poll
Currently, the application can call `sync_now()` but there is no guarantee the sync completes before the method returns and no way for the application to be notified when the sync has completed. The syncer will do its best to sync as soon as possible.

There are certain cases where it would be useful to know exactly when the sync has completed if there are certain operations that are expected to complete before proceeding to the next step.

Some examples of this:
1. Calling `sync_now()` to get a command needed before the next team onboarding step can continue. E.g. getting a command indicating a device was added to a team before assigning a label to that device.
2. Calling `sync_now()` to get a command before running a query where the expected output depends on receiving the command.

It would be useful for the application to have a way to wait for the sync to complete with a `sync_now_poll().await?` or some other way of notifying the application.

cc: @djohnson-spideroak, @jdygert-spok, @ycarmy, @ericlagergren

jdygert-spok:
Closes #404.

### DIFF
--- a/crates/aranya-daemon/src/sync/task.rs
+++ b/crates/aranya-daemon/src/sync/task.rs
@@ -8,16 +8,16 @@
 
 use std::{collections::HashMap, future::Future};
 
-use anyhow::{Context, Result};
+use anyhow::Context;
 use aranya_daemon_api::SyncPeerConfig;
 use aranya_runtime::{storage::GraphId, ClientError, Engine, Sink};
 use aranya_util::Addr;
-use buggy::{Bug, BugExt};
+use buggy::BugExt;
 use futures_util::StreamExt;
 use serde::{Deserialize, Serialize};
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, oneshot};
 use tokio_util::time::{delay_queue::Key, DelayQueue};
-use tracing::{error, instrument, trace};
+use tracing::{error, instrument, trace, warn};
 
 use super::Result as SyncResult;
 use crate::{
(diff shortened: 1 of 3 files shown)


ssweetney-spok (approved):
These changes also add long poll on the `add_peer` call when the `sync_now` value in the config is true. That seems like an improvement as well 👍

gknopf-aranya:
I started seeing new CI flakes after this PR was merged:
https://github.com/aranya-project/aranya/actions/runs/16278659698/job/45964564242?pr=403

crates/aranya-daemon/src/sync/task.rs
```diff
@@ -73,15 +75,17 @@ pub(crate) enum SyncResponse {
 
 impl SyncPeers {
     /// Create a new peer manager.
-    fn new(sender: mpsc::Sender<Msg>) -> Self {
+    fn new(sender: mpsc::Sender<Request>) -> Self {
         Self { sender }
     }
 
-    async fn send(&self, msg: Msg) -> Result<(), Bug> {
+    async fn send(&self, msg: Msg) -> Reply {
```
gknopf-aranya, line 82: Since `self.sender.send()` and `rx.await` are not called atomically if `send()` is called by multiple threads at the same time, this could result in one thread receiving a `Reply` meant for another thread.

gknopf-aranya: Nevermind, `tx` and `rx` are local to `send()` not global so this should be fine.

PR #406, opened 2025-07-11
Clean up errors

jdygert-spok:
Adds `ReportExt` for displaying error chains and cleans up error impls and usage.

### DIFF
--- a/crates/aranya-client/src/aqc/net.rs
+++ b/crates/aranya-client/src/aqc/net.rs
@@ -13,7 +13,10 @@ use aranya_crypto::aqc::{BidiChannelId, UniChannelId};
 use aranya_daemon_api::{
     AqcBidiPsks, AqcCtrl, AqcPsks, AqcUniPsks, DaemonApiClient, LabelId, TeamId,
 };
-use aranya_util::rustls::{NoCertResolver, SkipServerVerification};
+use aranya_util::{
+    error::ReportExt as _,
+    rustls::{NoCertResolver, SkipServerVerification},
+};
 use buggy::BugExt as _;
 use bytes::{Bytes, BytesMut};
 use channels::AqcPeerChannel;
@@ -313,10 +316,7 @@ impl AqcClient {
                     // The original function logged an error and returned ControlFlow::Break
                     // which implies the loop should terminate or an error state.
                     // For try_receive_channel, this might mean the connection is unusable for ctrl messages.
-                    warn!(
-                        "Receiving control message failed: {}, potential issue with connection.",
-                        e
-                    );
+                    warn!(error = %e.report(), "Receiving control message failed, potential issue with connection.");
                     // Depending on desired behavior, you might return an error or continue.
                     // For now, let's assume it's an error if control message processing fails critically.
                     return Err(TryReceiveError::Error(e));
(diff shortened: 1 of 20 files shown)


crates/aranya-client/src/aqc/net.rs
```diff
@@ -392,17 +393,18 @@ impl AqcClient {
                 }
             }
             Err(e) => {
-                error!("Failed to deserialize AqcCtrlMessage: {}", e);
-                let ack_msg =
-                    AqcAckMessage::Failure(format!("Failed to deserialize AqcCtrlMessage: {e}"));
+                let ack_msg = AqcAckMessage::Failure(format!(
+                    "Failed to deserialize AqcCtrlMessage: {}",
```
gknopf-aranya: If an error is returned via an ack, it seems like we'd want the peer to already know what that error is rather than sending it a formatted string that isn't shared between both peers as a known error type.

crates/aranya-client/src/error.rs
```diff
@@ -115,43 +115,43 @@ pub(crate) enum IpcRepr {
...
-    /// Serde serialization/deserialization error.
-    #[error("serialization/deserialization error: {0}")]
-    Serde(postcard::Error),
+    /// Error parsing control message.
+    #[error("failed to parse control message")]
+    InvalidCtrlMessage(postcard::Error),
 
     /// Peer failed to process control message.
-    #[error("error from peer processing control message: {0}")]
+    #[error("peer could not processing control message: {0}")]
```
gknopf-aranya:
```suggestion
    #[error("peer could not process control message: {0}")]
```

crates/aranya-util/src/error.rs
```diff
@@ -0,0 +1,81 @@
...
+    inner: &'a mut D,
+    number: Option<usize>,
+    started: bool,
+}
+
+impl<T> Write for Indented<'_, T>
+where
+    T: Write,
+{
+    fn write_str(&mut self, s: &str) -> fmt::Result {
```
gknopf-aranya, line 59: A bit difficult to follow what the format will look like. Assuming since this was copied from `anyhow` that we don't care about documenting that here.

gknopf-aranya (approved):
lgtm aside from a few comments

crates/aranya-util/src/error.rs
jdygert-spok: Yeah, it's just copied from anyhow.

crates/aranya-client/src/aqc/net.rs
jdygert-spok: Actually, I realized we should avoid sending the whole error to avoid potentially leaking information. It currently just sends the postcard error but really it should also signify an error occurs when the control processing fails. I simplified this to just use a quic stream reset with no error message. The peer will still log the error on their side.

gknopf-aranya (approved):

crates/aranya-daemon/src/sync/task/quic/psk.rs
```diff
@@ -137,32 +137,22 @@ impl PskStore {
...
-                    .or_default()
-                    .push(Arc::clone(&psk));
-                inner.identity_team.insert(PskIdAsKey(psk), team_id);
-                Ok(())
-            }
-            Err(e) => bail!(e.to_string()),
-        }
+    pub(crate) fn insert(&self, team_id: TeamId, psk: Arc<PresharedKey>) {
+        #[allow(clippy::expect_used, reason = "poison")]
+        let mut inner = self.inner.lock().expect("poisoned");
```
ssweetney-spok, line 142: Is it OK to panic in the daemon? (I see why we wouldn't want to do so in the client libraries)

If the mutex is poisoned that would prevent future syncs so this is pretty much an unrecoverable error. ~~Not sure if we should implement some type of graceful shutdown so that the other async tasks on the thread with the panic don't get stopped abruptly.~~

Edit: Oh does the tokio runtime catch the panic instead? Yea, ignoring is the default

ssweetney-spok (approved):

PR #412, opened 2025-07-15
Wait for daemon startup and use sync now for tests

jdygert-spok:
Reduces test flakes and improves behavior.

### DIFF
--- a/crates/aranya-client/tests/aqc.rs
+++ b/crates/aranya-client/tests/aqc.rs
@@ -3,6 +3,7 @@
 use std::time::Duration;
 
 mod common;
+
 use anyhow::{Context as _, Result};
 use aranya_client::aqc::AqcPeerChannel;
 use aranya_crypto::dangerous::spideroak_crypto::csprng::rand;
@@ -10,11 +11,10 @@ use aranya_daemon_api::{text, ChanOp};
 use backon::{ConstantBuilder, Retryable as _};
 use buggy::BugExt;
 use bytes::{Bytes, BytesMut};
-use common::TeamCtx;
 use futures_util::{future::try_join, FutureExt};
 use tempfile::tempdir;
 
-use crate::common::sleep;
+use crate::common::{sleep, TeamCtx};
 
 /// Demonstrate nominal usage of AQC channels.
 /// 1. Create bidirectional and unidirectional AQC channels.
--- a/crates/aranya-client/tests/aqc.rs
+++ b/crates/aranya-client/tests/aqc.rs
@@ -3,15 +3,18 @@
 use std::time::Duration;
 
 mod common;
-use anyhow::Result;
+use anyhow::{Context as _, Result};
 use aranya_client::aqc::AqcPeerChannel;
 use aranya_crypto::dangerous::spideroak_crypto::csprng::rand;
 use aranya_daemon_api::{text, ChanOp};
+use backon::{ConstantBuilder, Retryable as _};
 use buggy::BugExt;
 use bytes::{Bytes, BytesMut};
-use common::{sleep, TeamCtx};
-use futures_util::future::try_join;
+use common::TeamCtx;
+use futures_util::{future::try_join, FutureExt};
 use tempfile::tempdir;
+
+use crate::common::sleep;
 
 /// Demonstrate nominal usage of AQC channels.
 /// 1. Create bidirectional and unidirectional AQC channels.
@@ -19,9 +22,6 @@ use tempfile::tempdir;
 /// 3. Delete AQC channels.
 #[test_log::test(tokio::test(flavor = "multi_thread"))]
 async fn test_aqc_chans() -> Result<()> {
-    let interval = Duration::from_millis(100);
-    let sleep_interval = interval * 6;
-
     let tmp = tempdir()?;
     let work_dir = tmp.path().to_path_buf();
 
@@ -30,14 +30,9 @@ async fn test_aqc_chans() -> Result<()> {
     // create team.
     let team_id = team.create_and_add_team().await?;
 
-    sleep(sleep_interval).await;
-
     // Tell all peers to sync with one another, and assign their roles.
     team.add_all_sync_peers(team_id).await?;
     team.add_all_device_roles(team_id).await?;
-
-    // wait for syncing.
-    sleep(sleep_interval).await;
 
     let operator_team = team.operator.client.team(team_id);
     operator_team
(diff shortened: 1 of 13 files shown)


crates/aranya-client/tests/common/mod.rs
```diff
@@ -66,6 +67,9 @@ impl TeamCtx {
     }
 
     pub async fn add_all_sync_peers(&mut self, team_id: TeamId) -> Result<()> {
+        if true {
+            return Ok(());
+        }
```
jdygert-spok: The tests no longer actually need to use `add_all_sync_peers`, but removing it entirely will mean we are no longer testing the interval syncer :/ I guess we should add one test that relies on the interval?

jdygert-spok: I guess we could leave that to the examples?

gknopf-aranya: Agree that we want to keep test coverage for that by having a separate test for it.
I don't like that manually calling `sync_now()` everywhere makes tests more difficult to maintain.
But the performance improvement to unit tests is probably worth it.

Removing `add_all_sync_peers()` from tests could potentially hide race conditions that may be introduced later.
This is because `sync_now()` ensures commands arrive before performing operations. Code that relies on having certain commands before running but has a sync interval could race in a way that isn't caught by the unit tests.
Do we plan to wait until race conditions are encountered before fixing them and adding regression tests for them?
Or do we want a more proactive way of catching these types of races?

crates/aranya-client/tests/aqc.rs
```diff
@@ -30,15 +30,10 @@ async fn test_aqc_chans() -> Result<()> {
     // create team.
     let team_id = team.create_and_add_team().await?;
 
-    sleep(sleep_interval).await;
-
     // Tell all peers to sync with one another, and assign their roles.
     team.add_all_sync_peers(team_id).await?;
```
gknopf-aranya: No longer needed?

crates/aranya-client/tests/aqc.rs
```diff
@@ -223,10 +226,14 @@ async fn test_aqc_chans() -> Result<()> {
         // Send from 1 to 2 with a unidirectional stream
         let msg1 = Bytes::from_static(b"hello");
         send1_1.send(msg1.clone()).await?;
-        tokio::time::sleep(Duration::from_millis(100)).await;
         // Receive a unidirectional stream from peer 1
-        let mut recv2_1 = bidi_chan2
-            .try_receive_stream()
+        let mut recv2_1 = (|| std::future::ready(bidi_chan2.try_receive_stream()))
```
gknopf-aranya, line 229: Wonder if we should provide the application with a long poll future they can await?
Easier for the user than implementing their own retry logic.

jdygert-spok: This is testing `try_receive`. `receive` will let you poll it.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -259,26 +259,36 @@ impl Daemon {
...
         );
         set.spawn(
-            async move {
-                loop {
-                    if let Err(err) = self.syncer.next().await {
-                        error!(?err, "unable to sync with peer");
+            {
+                let notifier = waiter.notifier();
+                async move {
+                    notifier.notify();
```
gknopf-aranya: Does the `async move` introduce enough of a delay to be significant?
```suggestion
                waiter.notifier().notify();
                async move {
```

crates/aranya-util/src/ready.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Debug)]
+pub struct Waiter {
+    rx: tokio::sync::mpsc::Receiver<()>,
+    notifier: Notifier,
+}
+
+impl Waiter {
+    /// Create a waiter that will wait for `count` ready notifications.
+    pub fn new(count: usize) -> Self {
+        let (tx, rx) = tokio::sync::mpsc::channel(1);
```
gknopf-aranya, line 21: What if there are multiple notifications at the same time? Shouldn't the size of the `mpsc` channel be `count`?
```suggestion
        let (tx, rx) = tokio::sync::mpsc::channel(count);
```

crates/aranya-daemon/src/daemon.rs
jdygert-spok: The async block is the task, if I move it above then it will just immediately notify. The current version notifies once the task has been initially run once, which admittedly doesn't do much for us.

jdygert-spok: I added a `Syncer::run` method to make this the same as the other tasks so it is more clear.

crates/aranya-util/src/ready.rs
jdygert-spok: No, only the last notifier sends a message.

NB I think this could be implemented with roughly just `*mut (AtomicUsize, UnsafeCell<Waker>)`, but I didn't want to get bogged down in that so I did the simplest implemenation that didn't use unsafe.

crates/aranya-daemon/src/daemon.rs
gknopf-aranya: Do we really need the notification for this particular task? I thought the notification is for tasks that have a delay before they are ready. I suppose there could be a delay before the task is scheduled to run for the first time.
It's mostly the QUIC server and API startup we are interested in. Even if the syncer task took a while to startup, the worst thing that happens is it takes a little longer to initiate a sync.

crates/aranya-util/src/ready.rs
gknopf-aranya: It's a one-time operation at startup so it doesn't seem like it's worth optimizing yet.

crates/aranya-daemon/src/daemon.rs
jdygert-spok: Not really, I just figured it was nicer to do for all and then we wouldn't forget to add it if we add some setup for the syncer.

crates/aranya-client/tests/aqc.rs
```diff
@@ -47,12 +41,6 @@ async fn test_aqc_chans() -> Result<()> {
         .assign_aqc_net_identifier(team.memberb.id, team.memberb.aqc_net_id())
         .await?;
 
-    // wait for syncing.
-    sleep(sleep_interval).await;
-
-    // wait for ctrl message to be sent.
-    sleep(Duration::from_millis(100)).await;
-
```
djohnson-spideroak, line 55: Is there syncing happening at some point before this? I dont see any syncing happening between creating the team and assigning the label.

djohnson-spideroak (dismissed):

crates/aranya-client/tests/aqc.rs
jdygert-spok: `add_all_device_roles` syncs everyone.

crates/aranya-client/tests/common/mod.rs
jdygert-spok: > Removing `add_all_sync_peers()` from tests could potentially hide race conditions that may be introduced later.

Without interval syncing, there should be no way to have races in our tests. If we want to see what happens when we try things without syncing first, then we can test that directly.

djohnson-spideroak (approved):

PR #417, opened 2025-07-17
ci: don't cancel concurrent workflows on main branch

gknopf-aranya:
Concurrency groups are used by our CI pipeline to cancel older workflows in favor of running new ones for new commits.
This reduces CI usage by cancelling workflows that are no longer needed.

The problem with this is that while this strategy works well for development branches/PRs, it is not a good approach for `main`, `master`, `release`, `production`, etc. branches. These types of branches must guarantee each commit builds, passes tests, completes upload of release artifacts, or completes deployment of software to a production server. Cancelling such jobs could have unintended consequences such as releasing software with partial artifacts or partially deploying software to a server which results in a broken deployment.

The solution to this is to skip concurrency group cancellation if a branch matches a certain pattern. The only branch we're concerned about at this time is `main`, although this could extend to other branches in the future.

Relevant docs:
https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-when-your-workflow-runs/control-the-concurrency-of-workflows-and-jobs#example-only-cancel-in-progress-jobs-on-specific-branches

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}
-  cancel-in-progress: ${{ !contains(github.ref, 'main/') }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
   build-release:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}
-  cancel-in-progress: ${{ !contains(github.ref, 'main/')}}
+  cancel-in-progress: ${{ !contains(github.ref, 'main/') }}
 
 jobs:
   build-release:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ !contains(github.ref, 'main/')}}
 
 jobs:
   build-release:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}-${{ github.ref != 'refs/heads/main' || github.run_id }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+  cancel-in-progress: true
 
 jobs:
   build-release:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ env:
   REV: ${{ github.run_id }}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}-${{ github.ref != 'refs/heads/main' || github.run_id }}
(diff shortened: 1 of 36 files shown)


gknopf-aranya:
Confirmed that workflows were cancelled as expected after the `fmt` commit.

.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ !contains(github.ref, 'main/') }}
```
djohnson-spideroak: will this match branches with the name `something-main`?

jdygert-spok: I think it won't match anything actually. Looks like the format for `github.ref` is `refs/heads/main`. Should check for just that explicitly.

jdygert-spok: (Or rather it would match branches named `main/foo`)

.github/workflows/correctness.yml
```diff
@@ -17,7 +17,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ !contains(github.ref, 'main/') }}
```
jdygert-spok: Note that since a group is still set, it will wait until the previous commit's action finishes, which isn't necessary. I don't think it's a big deal in the circumstances though.

jdygert-spok:
I confirmed publish.yml and release.yml are already set up correctly.

.github/workflows/build.yml
gknopf-aranya: It would match to `refs/heads/main/<something>`. I'm going to change this to:
`cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`

.github/workflows/correctness.yml
gknopf-aranya: It should be possible to conditionally append the unique job ID to the group name if running on the main branch.
Without that, the CI pipeline works correctly but causes us to wait an extra ~20min before the next set of jobs start.

djohnson-spideroak (dismissed):

.github/workflows/correctness.yml
jdygert-spok: I believe you'd want to just leave the group off. (Set to null to do it conditionally?)

It shouldn't usually cause us to wait very long, this would only trigger if we immediately merge another PR which took shorter to run than the actions on main. And it'd just delay the checks for that new main commit which isn't a huge deal either. It'd back up if we kept merging a bunch of PRs quickly but we don't really do that :shrug:

jdygert-spok: Seems like we do need this though. If you merge A>B>C, if B hasn't started because A is still running, then it will get cancelled and only C will be queued. Because B isn't "in-progress", so it gets cancelled anyways.

gknopf-aranya: It isn't clear from the docs if there's a way to define a concurrency group for some branches but not others.
https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-when-your-workflow-runs/control-the-concurrency-of-workflows-and-jobs
I tried setting the group to an empty string:
`group: ''`
But see this error:
`Unexpected value ''`
Our best bet is probably putting each workflow/job ID for a certain branch into a unique concurrency group.

.github/workflows/build.yml
```diff
@@ -16,8 +16,8 @@ env:
   REV: ${{ github.run_id }}
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.workflow }}-${{ github.ref != 'refs/heads/main' || github.run_id }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
gknopf-aranya: If each workflow on `main` is now in its own concurrency group, we don't need this conditional `cancel-in-progress` anymore.

PR #421, opened 2025-07-18
Share connection map between sync client and sync server

### LINKED ISSUE #420: Share connection map between sync client and sync server
We want to allow for one peer to initiate a connection and then both peers can use the established connection to send sync requests.

[Passive Sync Spec](https://github.com/aranya-project/aranya-docs/pull/113#discussion_r2216393543)

ssweetney-spok:
Resolves #420

### DIFF
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -34,7 +34,7 @@ use crate::{
     keystore::{AranyaStore, LocalStore},
     policy,
     sync::task::{
-        quic::{PskStore, SharedConnectionMap, State as QuicSyncState},
+        quic::{Notification, PskStore, SharedConnectionMap, State as QuicSyncState},
         Syncer,
     },
     util::{load_team_psk_pairs, SeedDir},
@@ -63,7 +63,8 @@ struct SyncParams {
     psk_store: Arc<PskStore>,
     active_team_rx: Receiver<TeamId>,
     external_sync_addr: Addr,
-    conns: SharedConnectionMap,
+    conns: Arc<SharedConnectionMap>,
+    conn_rx: Receiver<Notification>,
 }
 
 mod invalid_graphs {
@@ -178,7 +179,8 @@ impl Daemon {
             let psk_store = Arc::new(psk_store);
 
             // Initialize the connection map used by the syncer and sync server
-            let conn_map: SharedConnectionMap = Default::default();
+            let (conn_map, conn_rx) = SharedConnectionMap::new();
+            let conn_map = Arc::new(conn_map);
 
             // Initialize Aranya client.
             let (client, sync_server) = Self::setup_aranya(
@@ -192,6 +194,7 @@ impl Daemon {
                     psk_store: Arc::clone(&psk_store),
                     conns: Arc::clone(&conn_map),
                     external_sync_addr: qs_config.addr,
+                    conn_rx,
                     active_team_rx,
                 },
             )
@@ -349,6 +352,7 @@ impl Daemon {
             active_team_rx,
             external_sync_addr,
             conns,
+            conn_rx,
         }: SyncParams,
     ) -> Result<(Client, SyncServer)> {
         let device_id = pk.ident_pk.id()?;
@@ -368,6 +372,7 @@ impl Daemon {
             &external_sync_addr,
             psk_store,
             conns,
+            conn_rx,
             active_team_rx,
         )
         .await
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -58,6 +58,14 @@ pub(crate) type EF = policy::Effect;
 pub(crate) type Client = aranya::Client<EN, SP>;
 pub(crate) type SyncServer = crate::sync::task::quic::Server<EN, SP>;
 
+/// Sync configuration for setting up Aranya.
+struct SyncParams {
+    psk_store: Arc<PskStore>,
+    active_team_rx: Receiver<TeamId>,
+    external_sync_addr: Addr,
+    conns: SharedConnectionMap,
+}
+
 mod invalid_graphs {
     use std::{
         collections::HashSet,
@@ -180,10 +188,12 @@ impl Daemon {
                     .try_clone()
                     .context("unable to clone keystore")?,
                 &pks,
-                qs_config.addr,
-                Arc::clone(&psk_store),
-                Arc::clone(&conn_map),
-                active_team_rx,
+                SyncParams {
+                    psk_store: Arc::clone(&psk_store),
+                    conns: Arc::clone(&conn_map),
+                    external_sync_addr: qs_config.addr,
+                    active_team_rx,
+                },
             )
             .await?;
             let local_addr = sync_server.local_addr()?;
@@ -334,10 +344,12 @@ impl Daemon {
         eng: CE,
         store: AranyaStore<KS>,
         pk: &PublicKeys<CS>,
-        external_sync_addr: Addr,
-        psk_store: Arc<PskStore>,
-        conns: SharedConnectionMap,
-        active_team_rx: Receiver<TeamId>,
+        SyncParams {
+            psk_store,
+            active_team_rx,
+            external_sync_addr,
+            conns,
+        }: SyncParams,
     ) -> Result<(Client, SyncServer)> {
         let device_id = pk.ident_pk.id()?;
 
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -22,15 +22,15 @@ use tracing::{error, info, info_span, Instrument as _};
 
 use crate::{
     actions::Actions,
-    api::{ApiKey, DaemonApiServer, QSData},
+    api::{ApiKey, DaemonApiServer, EffectReceiver, QSData},
     aqc::Aqc,
     aranya,
     config::{Config, Toggle},
     keystore::{AranyaStore, LocalStore},
     policy,
     sync::task::{
-        quic::{PskStore, SharedConnectionMap, State as QuicSyncClientState, SyncParams},
-        Syncer,
+        quic::{PskStore, State as QuicSyncClientState, SyncParams},
+        SyncPeers, Syncer,
     },
     util::{load_team_psk_pairs, SeedDir},
     vm_policy::{PolicyEngine, TEST_POLICY_1},
@@ -164,12 +164,10 @@ impl Daemon {
             let (psk_store, active_team_rx) = PskStore::new(initial_keys);
             let psk_store = Arc::new(psk_store);
 
-            // Initialize the connection map used by the syncer and sync server
-            let (conn_map, conn_rx) = SharedConnectionMap::new();
-            let conn_map = Arc::new(conn_map);
+            let invalid_graphs = InvalidGraphs::default();
 
-            // Initialize Aranya client.
-            let (client, sync_server) = Self::setup_aranya(
+            // Initialize Aranya client, sync client,and sync server.
+            let (client, sync_server, syncer, peers, recv_effects) = Self::setup_aranya(
                 &cfg,
                 eng.clone(),
                 aranya_store
@@ -178,22 +176,13 @@ impl Daemon {
                 &pks,
                 SyncParams {
                     psk_store: Arc::clone(&psk_store),
-                    conns: Arc::clone(&conn_map),
-                    external_sync_addr: qs_config.addr,
-                    conn_rx,
+                    server_addr: qs_config.addr,
                     active_team_rx,
                 },
+                invalid_graphs.clone(),
             )
             .await?;
             let local_addr = sync_server.local_addr()?;
-
-            // Sync in the background at some specified interval.
-            let (send_effects, recv_effects) = tokio::sync::mpsc::channel(256);
-
-            let invalid_graphs = InvalidGraphs::default();
-            let state = QuicSyncClientState::new(psk_store.clone(), conn_map)?;
-            let (syncer, peers) =
-                Syncer::new(client.clone(), send_effects, invalid_graphs.clone(), state);
 
             let graph_ids = client
                 .aranya
@@ -327,7 +316,7 @@ impl Daemon {
         Ok(())
     }
 
-    /// Creates the Aranya client and sync server.
+    /// Creates the Aranya client, sync client, and sync server.
     async fn setup_aranya(
         cfg: &Config,
         eng: CE,
@@ -336,11 +325,16 @@ impl Daemon {
         SyncParams {
             psk_store,
             active_team_rx,
-            external_sync_addr,
-            conns,
-            conn_rx,
+            server_addr,
         }: SyncParams,
-    ) -> Result<(Client, SyncServer)> {
+        invalid_graphs: InvalidGraphs,
+    ) -> Result<(
+        Client,
+        SyncServer,
+        Syncer<QuicSyncClientState>,
+        SyncPeers,
+        EffectReceiver,
+    )> {
         let device_id = pk.ident_pk.id()?;
 
         let aranya = Arc::new(Mutex::new(ClientState::new(
@@ -352,10 +346,21 @@ impl Daemon {
 
         let client = Client::new(Arc::clone(&aranya));
 
-        info!(addr = %external_sync_addr, "starting QUIC sync server");
+        // Sync in the background at some specified interval.
+        let (send_effects, recv_effects) = tokio::sync::mpsc::channel(256);
+
+        // Initialize the syncer
+        let (syncer, peers, conns, conn_rx) = Syncer::new(
+            client.clone(),
+            send_effects,
+            invalid_graphs,
+            psk_store.clone(),
+        )?;
+
+        info!(addr = %server_addr, "starting QUIC sync server");
         let server = SyncServer::new(
             client.clone(),
-            &external_sync_addr,
+            &server_addr,
             psk_store,
             conns,
             conn_rx,
@@ -366,7 +371,7 @@ impl Daemon {
 
         info!(device_id = %device_id, "set up Aranya");
 
-        Ok((client, server))
+        Ok((client, server, syncer, peers, recv_effects))
     }
 
     /// Loads the crypto engine.
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -7,22 +7,17 @@ use aranya_crypto::{
     keystore::{fs_keystore::Store, KeyStore},
     Engine, Rng,
 };
-use aranya_daemon_api::TeamId;
 use aranya_keygen::{KeyBundle, PublicKeys};
 use aranya_runtime::{
     storage::linear::{libc::FileManager, LinearStorageProvider},
     ClientState, StorageProvider,
 };
-use aranya_util::{ready, Addr};
+use aranya_util::ready;
 use bimap::BiBTreeMap;
 use buggy::{bug, Bug, BugExt};
 use ciborium as cbor;
 use serde::{de::DeserializeOwned, Serialize};
-use tokio::{
-    fs,
-    sync::{mpsc::Receiver, Mutex},
-    task::JoinSet,
-};
+use tokio::{fs, sync::Mutex, task::JoinSet};
 use tracing::{error, info, info_span, Instrument as _};
 
 use crate::{
@@ -34,7 +29,7 @@ use crate::{
     keystore::{AranyaStore, LocalStore},
     policy,
     sync::task::{
-        quic::{ConnectionUpdate, PskStore, SharedConnectionMap, State as QuicSyncState},
+        quic::{PskStore, SharedConnectionMap, State as QuicSyncClientState, SyncParams},
         Syncer,
     },
     util::{load_team_psk_pairs, SeedDir},
@@ -57,15 +52,6 @@ pub(crate) type EF = policy::Effect;
 
 pub(crate) type Client = aranya::Client<EN, SP>;
 pub(crate) type SyncServer = crate::sync::task::quic::Server<EN, SP>;
-
-/// Sync configuration for setting up Aranya.
-struct SyncParams {
-    psk_store: Arc<PskStore>,
-    active_team_rx: Receiver<TeamId>,
-    external_sync_addr: Addr,
-    conns: Arc<SharedConnectionMap>,
-    conn_rx: Receiver<ConnectionUpdate>,
-}
 
 mod invalid_graphs {
     use std::{
@@ -132,7 +118,7 @@ impl DaemonHandle {
 #[derive(Debug)]
 pub struct Daemon {
     sync_server: SyncServer,
-    syncer: Syncer<QuicSyncState>,
+    syncer: Syncer<QuicSyncClientState>,
     api: DaemonApiServer,
     span: tracing::Span,
 }
@@ -205,7 +191,7 @@ impl Daemon {
             let (send_effects, recv_effects) = tokio::sync::mpsc::channel(256);
 
             let invalid_graphs = InvalidGraphs::default();
-            let state = QuicSyncState::new(psk_store.clone(), conn_map)?;
+            let state = QuicSyncClientState::new(psk_store.clone(), conn_map)?;
             let (syncer, peers) =
                 Syncer::new(client.clone(), send_effects, invalid_graphs.clone(), state);
 
@@ -492,6 +478,7 @@ mod tests {
 
     use std::time::Duration;
 
+    use aranya_util::Addr;
     use tempfile::tempdir;
     use test_log::test;
     use tokio::time;
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -34,7 +34,7 @@ use crate::{
     keystore::{AranyaStore, LocalStore},
     policy,
     sync::task::{
-        quic::{Notification, PskStore, SharedConnectionMap, State as QuicSyncState},
+        quic::{ConnectionUpdate, PskStore, SharedConnectionMap, State as QuicSyncState},
         Syncer,
     },
     util::{load_team_psk_pairs, SeedDir},
@@ -64,7 +64,7 @@ struct SyncParams {
     active_team_rx: Receiver<TeamId>,
     external_sync_addr: Addr,
     conns: Arc<SharedConnectionMap>,
-    conn_rx: Receiver<Notification>,
+    conn_rx: Receiver<ConnectionUpdate>,
 }
 
 mod invalid_graphs {
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -34,7 +34,7 @@ use crate::{
     keystore::{AranyaStore, LocalStore},
     policy,
     sync::task::{
-        quic::{PskStore, State as QuicSyncState},
+        quic::{PskStore, SharedConnectionMap, State as QuicSyncState},
         Syncer,
     },
     util::{load_team_psk_pairs, SeedDir},
@@ -169,6 +169,9 @@ impl Daemon {
             let (psk_store, active_team_rx) = PskStore::new(initial_keys);
             let psk_store = Arc::new(psk_store);
 
+            // Initialize the connection map used by the syncer and sync server
+            let conn_map: SharedConnectionMap = Default::default();
+
             // Initialize Aranya client.
             let (client, sync_server) = Self::setup_aranya(
                 &cfg,
@@ -179,6 +182,7 @@ impl Daemon {
                 &pks,
                 qs_config.addr,
                 Arc::clone(&psk_store),
+                Arc::clone(&conn_map),
                 active_team_rx,
             )
             .await?;
@@ -188,7 +192,7 @@ impl Daemon {
             let (send_effects, recv_effects) = tokio::sync::mpsc::channel(256);
 
             let invalid_graphs = InvalidGraphs::default();
-            let state = QuicSyncState::new(psk_store.clone())?;
+            let state = QuicSyncState::new(psk_store.clone(), conn_map)?;
             let (syncer, peers) =
                 Syncer::new(client.clone(), send_effects, invalid_graphs.clone(), state);
 
@@ -332,6 +336,7 @@ impl Daemon {
         pk: &PublicKeys<CS>,
         external_sync_addr: Addr,
         psk_store: Arc<PskStore>,
+        conns: SharedConnectionMap,
         active_team_rx: Receiver<TeamId>,
     ) -> Result<(Client, SyncServer)> {
         let device_id = pk.ident_pk.id()?;
@@ -350,6 +355,7 @@ impl Daemon {
             client.clone(),
             &external_sync_addr,
             psk_store,
+            conns,
             active_team_rx,
         )
         .await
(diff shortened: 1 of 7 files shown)


crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -409,65 +436,123 @@ where
...
-                                error!(error = %err.report(), "server unable to sync with peer");
-                                return;
+                            match conn_map_guard.entry(key) {
+                                Entry::Occupied(_) => {
+                                    // Close the new connection because one already exists
+
+                                    // TODO: Use appropriate error code
+                                    conn.close(AppError::UNKNOWN);
+
+                                    continue;
```
ssweetney-spok: Should probably send a ping frame here to see if the connection that exists in the map is still open

crates/aranya-daemon/src/sync/task/quic/map.rs
gknopf-aranya, line 1: Rename this file to `connections.rs` then add a doc top the top of the file like this:
```
//! Map for storing persistent QUIC connections between pairs of sync peers.
//! Once a connection has been opened, it is shared between the QUIC syncer client and server.
```

crates/aranya-daemon/src/sync/task/quic/map.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl<T> DerefMut for RwLockWriteGuard<'_, T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.guard
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct SharedConnectionMap {
```
gknopf-aranya: The name of the struct should describe the purpose. The fact that it is implemented using a map is an implementation detail.
```suggestion
/// Stores active QUIC syncer connections with other peers.
pub(crate) struct Connections {
```

crates/aranya-daemon/src/daemon.rs
```diff
@@ -169,6 +178,10 @@ impl Daemon {
             let (psk_store, active_team_rx) = PskStore::new(initial_keys);
             let psk_store = Arc::new(psk_store);
 
+            // Initialize the connection map used by the syncer and sync server
+            let (conn_map, conn_rx) = SharedConnectionMap::new();
+            let conn_map = Arc::new(conn_map);
```
gknopf-aranya:
```suggestion
            let (conns, conn_rx) = Connections::new();
            let conns = Arc::new(conns);
```

crates/aranya-daemon/src/daemon.rs
```diff
@@ -177,9 +190,13 @@ impl Daemon {
                     .try_clone()
                     .context("unable to clone keystore")?,
                 &pks,
-                qs_config.addr,
-                Arc::clone(&psk_store),
-                active_team_rx,
+                SyncParams {
+                    psk_store: Arc::clone(&psk_store),
+                    conns: Arc::clone(&conn_map),
```
gknopf-aranya: Isn't this the same as calling `.clone()`?
```suggestion
                    conns: conns.clone(),
```

crates/aranya-daemon/src/daemon.rs
```diff
@@ -188,7 +205,7 @@ impl Daemon {
             let (send_effects, recv_effects) = tokio::sync::mpsc::channel(256);
 
             let invalid_graphs = InvalidGraphs::default();
-            let state = QuicSyncState::new(psk_store.clone())?;
+            let state = QuicSyncState::new(psk_store.clone(), conn_map)?;
```
gknopf-aranya: We're calling the QUIC sync server `SyncServer` while the client is called `QuicSyncState`. I think we should call the client `QuicSyncClient` or `SyncClient` to avoid confusion.
```suggestion
            let state = SyncClient::new(psk_store.clone(), conn_map)?;
```

crates/aranya-daemon/src/sync/task/quic/map.rs
```diff
@@ -0,0 +1,94 @@
...
+use s2n_quic::{
+    application::Error as AppError,
+    connection::{Handle, StreamAcceptor},
+    Connection,
+};
+use tokio::sync::{self, mpsc, Mutex, RwLock};
+
+use crate::sync::task::quic::ConnectionKey;
+
+pub(crate) type ConnectionUpdate = (ConnectionKey, StreamAcceptor);
```
gknopf-aranya: Needs doc

crates/aranya-daemon/src/daemon.rs
```diff
@@ -58,6 +58,15 @@ pub(crate) type EF = policy::Effect;
 pub(crate) type Client = aranya::Client<EN, SP>;
 pub(crate) type SyncServer = crate::sync::task::quic::Server<EN, SP>;
 
+/// Sync configuration for setting up Aranya.
+struct SyncParams {
```
gknopf-aranya: This type should be defined inside one of the QUIC syncer implementation modules. E.g. in `aranya-daemon/src/sync/task/quic.rs`.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -195,18 +195,25 @@ impl Syncer<State> {
         debug!("client connecting to QUIC sync server");
         // Check if there is an existing connection with the peer.
         // If not, create a new connection.
-        let conns = &mut self.state.conns;
+        let conn_map_guard = self.state.conns.read().await;
```
gknopf-aranya: Is this approach better than using a `Arc<Mutex<Connections>>` to access the connections map?

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -195,18 +195,25 @@ impl Syncer<State> {
...
         let client = &self.state.client;
 
         let key = ConnectionKey { addr: *peer, id };
-        let conn = match conns.entry(key) {
-            Entry::Occupied(entry) => {
+
+        let conn_ref = match conn_map_guard.get(&key) {
```
gknopf-aranya: Now that we have another module for sharing QUIC connections, we should move the logic for managing read/write to the object, checking if a connection exists, and returning a connection into that module.
The QUIC client/server only need a simple interface to the connections object like this that hides the complexity behind the abstraction that was created:
```
let conn = conns.get(&key).await?;
conns.remove(key).await?;
```
In the near future, we'll add `connect()` so a peer can setup a connection without syncing:
```
conns.connect(&key)
```

crates/aranya-daemon/src/sync/task/quic/map.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub(super) async fn insert(
+        &mut self,
+        key: ConnectionKey,
+        conn: Connection,
+    ) -> (Option<Arc<Mutex<Handle>>>, Arc<Mutex<Handle>>) {
+        let (handle, acceptor) = conn.split();
+
+        let handle = Arc::new(Mutex::new(handle));
+        let existing = self.deref_mut().insert(key, Arc::clone(&handle));
+        self.tx.send((key, acceptor)).await.expect("channel closed");
```
gknopf-aranya: I think this should only be called if a new connection that didn't already exist has been inserted.

crates/aranya-daemon/src/sync/task/quic/map.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[derive(Debug)]
+pub(crate) struct SharedConnectionMap {
+    data: RwLock<BTreeMap<ConnectionKey, Arc<Mutex<Handle>>>>,
+    tx: mpsc::Sender<ConnectionUpdate>,
+}
+
+impl SharedConnectionMap {
+    pub(crate) fn new() -> (Self, mpsc::Receiver<ConnectionUpdate>) {
+        let (tx, rx) = mpsc::channel(10);
```
gknopf-aranya: Seems too easy to accidentally fill up this buffer:
```suggestion
        let (tx, rx) = mpsc::channel(32);
```

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -409,65 +436,123 @@ where
...
+                    else => break,
+                }
             }
         })
         .await;
 
         error!("server terminated");
     }
 
+    async fn handle_stream(
```
gknopf-aranya:
```suggestion
    async fn handle_streams(
```

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: I thought it would be but with the way that syncing works, we only allow for 1 outgoing sync request at a time so there's no benefit to having multiple readers so I'll refactor this

crates/aranya-daemon/src/daemon.rs
ssweetney-spok: Yea, the naming is hard. This type isn't actually the sync client. Calling the `Syncer` the sync client would be more appropriate. This is a type that implements `aranya_daemon::sync::task::SyncState`.

`QuicSyncClientState` would probably make the most sense

crates/aranya-daemon/src/daemon.rs
ssweetney-spok: This is the idiomatic way to write it. It shows that it's a shared clone

crates/aranya-daemon/src/sync/task/quic/map.rs
ssweetney-spok: That's a good point. Isn't the purpose to map (Addr, TeamID) to Connection objects?

djohnson-spideroak: I think its probably fine, it does use a map, but the purpose is to be a specific map implementation.

gknopf-aranya: I'm thinking of it as an object that stores connections in memory. The fact that the object is shared and contains a map are implementation details.

ssweetney-spok: Oh, I see. I think what you're describing is a connection pool which is a bit different. It's pretty much the same except instead of grabbing any connection, we get connections by a specific key

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -354,6 +368,10 @@ pub struct Server<EN, SP> {
     /// Receives updates for the "active team".
     /// Used to ensure that the chosen PSK corresponds to an incoming sync request.
     active_team_rx: mpsc::Receiver<TeamId>,
+    /// Connection map shared with [`super::Syncer`]
+    conns: Arc<SharedConnectionMap>,
+    /// Receives updates for connections created during outgoing sync requests.
```
ssweetney-spok:
```suggestion
    /// Receives updates for connections inserted into [`Self::conns`].
```

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: I did some refactoring which made things a bit simpler.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -409,65 +431,93 @@ where
...
+            let client = self.aranya.clone();
+
+            loop {
+                tokio::select! {
+                    // Accept incoming QUIC connections.
+                    Some(conn) = self.server.accept() => {
+                        debug!("received incoming QUIC connection");
+                        let Ok(active_team) = self.active_team_rx.try_recv() else {
+                            warn!("no active team for accepted connection");
+                            continue;
```
djohnson-spideroak: should this be a warn or error?

crates/aranya-daemon/src/sync/task/quic/connections.rs
```diff
@@ -0,0 +1,130 @@
...
+                } else {
+                    let _ = occupied_handle_entry.insert(new_handle);
+                    (occupied_handle_entry.into_mut(), true)
+                }
+            }
+            Entry::Vacant(entry) => (entry.insert(new_handle), true),
+        };
+
+        if inserted {
+            self.tx.send((key, acceptor)).await.expect("channel closed");
```
gknopf-aranya: I understand that the reason for panicking here is that this channel being closed indicates there is a bug in the syncer. But, I think it would be better to log and return an error here rather than panicking. Panicking would crash the daemon and prevent the syncer from syncing over existing connections.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -169,6 +164,10 @@ impl Daemon {
             let (psk_store, active_team_rx) = PskStore::new(initial_keys);
             let psk_store = Arc::new(psk_store);
 
+            // Initialize the connection map used by the syncer and sync server
+            let (conn_map, conn_rx) = SharedConnectionMap::new();
```
gknopf-aranya: The daemon doesn't need to know about how to initialize a connections map, only the server and client.
There are a couple ways to handle this:
1. Return `conn_map` and `conn_rx` from `SyncServer::new()`
2. Return a handle to the client when initializing the server `let (client_state, server) = SyncServer::new()`. This would work because `QuicSyncClientState` needs the `psk_store` and `conn_map` which are already available to the server at initialization.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -97,9 +94,18 @@ impl From<Infallible> for Error {
 /// Unique key for a connection with a peer.
 /// Each team/graph is synced over a different QUIC connection so a team-specific PSK can be used.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
-struct ConnectionKey {
-    addr: Addr,
-    id: GraphId,
+pub(crate) struct ConnectionKey {
```
gknopf-aranya: Since the connections map has been moved to `connections.rs`, this type can be moved there as well since it is the key for the map.

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: I think either one would be appropriate. This scenario means that we ran into some unexpected behavior but it's not something that would prevent the daemon from continuing normally.

Also I wasn't the original author, just moved some code around

crates/aranya-daemon/src/daemon.rs
```diff
@@ -330,9 +333,13 @@ impl Daemon {
         eng: CE,
         store: AranyaStore<KS>,
         pk: &PublicKeys<CS>,
-        external_sync_addr: Addr,
-        psk_store: Arc<PskStore>,
-        active_team_rx: Receiver<TeamId>,
+        SyncParams {
+            psk_store,
+            active_team_rx,
+            external_sync_addr,
```
gknopf-aranya: nit: since this is the address the server binds to:
```suggestion
            server_addr,
```
`_sync_` isn't needed in the name since that's implied by the field being inside the `SyncParams` struct.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -195,16 +203,17 @@ impl Syncer<State> {
         debug!("client connecting to QUIC sync server");
         // Check if there is an existing connection with the peer.
         // If not, create a new connection.
-        let conns = &mut self.state.conns;
+        let mut conn_map_guard = self.state.conns.lock().await;
```
gknopf-aranya:
```suggestion
        let mut conns = self.state.conns.lock().await;
```

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -195,16 +203,17 @@ impl Syncer<State> {
...
         let client = &self.state.client;
 
         let key = ConnectionKey { addr: *peer, id };
-        let conn = match conns.entry(key) {
-            Entry::Occupied(entry) => {
+
+        let conn_ref = match conn_map_guard.get_mut(&key) {
```
gknopf-aranya:
```suggestion
        let conn = match conns.get_mut(&key) {
```

crates/aranya-daemon/src/sync/task/quic/connections.rs
ssweetney-spok: I had a [similar concern](https://github.com/aranya-project/aranya/pull/406/files#r2208660783) while reviewing another PR. The daemon wouldn't crash but it would terminate the tokio task that the panic happened in. 

If we return an error here then syncing over existing connections would be stopped but then we can't accept sync requests for new connections. Not sure what the ideal behavior should be 🤔 

I think I'd prefer to return an error here but as a general question: when do we want to panic in the daemon code?

crates/aranya-daemon/src/sync/task/quic/connections.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Inserts a QUIC connection into the map.
+    ///
+    /// Splits the connection into a handle and stream acceptor. If a connection already exists
+    /// for the key, checks for open connections via ping - reuses open connections and replaces
+    /// closed ones. Sends a [`ConnectionUpdate`] when a new connection is inserted.
+    ///
+    /// # Parameters
+    ///
+    /// * `key` - The [`ConnectionKey`] that uniquely identifies the connection pair
+    /// * `conn` - The [`Connection`] to insert
```
gknopf-aranya:
```suggestion
    /// * `key` - The [`ConnectionKey`] that uniquely identifies the connection pair based on team ID and the peer's network address.
    /// * `conn` - The QUIC [`Connection`] to insert
```

crates/aranya-daemon/src/sync/task/quic/connections.rs
gknopf-aranya: Usually, it's best to error on the side of returning an error rather than panicking. Panicking is non-recoverable so should be reserved for conditions that can't be recovered from or could compromise security/functionality of the daemon. Some cases I can think of where it could be okay to panic:
- Unable to load daemon config file
- Unable to initialize or bind sync server to address (daemon isn't very useful without ability to sync)
- Unable to initialize or bind AQC server to address
- Unable to initialize the Aranya graph/policy
- Internal crypto library error

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -409,65 +431,93 @@ where
...
+                        };
+
+                        let Ok(peer) = conn.remote_addr().inspect_err(|err| {
+                            error!(error = %err.report(), "unable to get peer address from connection");
+                        }) else {
+                            continue;
+                        };
+
+                        let mut conn_map_guard = self.conns.lock().await;
+                        let key = ConnectionKey {
```
gknopf-aranya: Since the connection map and key are being moved to their own module, let's provide a well-defined constructor for the key with `ConnectionKey::new(...)` and make the fields private.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -409,65 +431,93 @@ where
...
+                            addr: peer.into(),
+                            id: active_team.into_id().into(),
+                        };
+                        let _ = conn_map_guard.insert(key, conn).await;
```
gknopf-aranya: Why is result ignored?

ssweetney-spok: The type is `(&mut Handle, bool)` which isn't needed for the server. 

There's some indirection going on. The call to `MutexGuard::insert` sends an update over the channel which is then handled a few lines below

`Some((key, acceptor)) = self.conn_rx.recv() => { ..`

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -409,65 +431,93 @@ where
...
     }
 
+    async fn handle_streams(
+        client: AranyaClient<EN, SP>,
+        mut acceptor: StreamAcceptor,
+        peer: SocketAddr,
+        active_team: TeamId,
+    ) {
+        // Accept incoming streams.
+        while let Some(res) = acceptor.accept_bidirectional_stream().await.transpose() {
```
gknopf-aranya: I imagine it would be a bug if a unidirectional stream was received since the QUIC syncer only operates on bidirectional streams for the request/response. Do we check for unidirectional streams and return/log an error if that occurs?

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -409,65 +431,93 @@ where
...
+            let stream = match res {
+                Ok(stream) => stream,
+                Err(err) => {
+                    error!(error = %err.report(), "error receiving QUIC stream");
+                    return;
+                }
+            };
+            debug!("received incoming QUIC stream");
+            if let Err(err) = Self::sync(client.clone(), peer, stream, &active_team).await {
+                error!(error = %err.report(), "server unable to sync with peer");
```
gknopf-aranya: Is this sync initiated by the client or server? Based on the error message, it indicates the sync was initiated by the server. But, "received incoming QUIC stream" make it seem as though the server is receiving a stream created by another client.

crates/aranya-daemon/src/sync/task/quic/connections.rs
```diff
@@ -0,0 +1,130 @@
+//! This modules contains a map for storing persistent QUIC connections between pairs of sync peers.
```
gknopf-aranya, line 1: Could mention that it stores a connection for each team ID that is synced between a pair of peers. So the correct PSK is used to sync each team.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -222,14 +231,21 @@ impl Syncer<State> {
 
                 conn.keep_alive(true).map_err(Error::from)?;
                 debug!("created new quic connection");
-                entry.insert(conn)
+
+                let (conn, inserted) = conn_map_guard.insert(key, conn).await;
```
gknopf-aranya: Returning a boolean isn't very idiomatic for a Rust map insertion. Usually insert would return an `Option<value>`. I understand that the reason for doing that here is because the handle to the connection should always be returned.
It should be possible to check if the key/value already exists in the map first before inserting using some combination of these methods:
https://doc.rust-lang.org/beta/std/collections/btree_map/struct.BTreeMap.html#method.try_insert
https://doc.rust-lang.org/beta/std/collections/btree_map/struct.BTreeMap.html#method.insert
https://doc.rust-lang.org/beta/std/collections/btree_map/struct.BTreeMap.html#method.contains_key
https://doc.rust-lang.org/beta/std/collections/btree_map/struct.BTreeMap.html#method.get

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -222,14 +231,21 @@ impl Syncer<State> {
...
+                if !inserted {
+                    // Note(Steve): This shouldn't be reached because the map
+                    // is still locked until `conn_map_guard` is dropped
+                    debug!("New connection wasn't inserted. A healthy connection was found")
```
gknopf-aranya: No need to add logic for unreachable states.

crates/aranya-daemon/src/sync/task/quic/connections.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// # Panics
+    ///
+    /// Panics if the internal connection update channel is closed.
+    #[allow(clippy::expect_used, reason = "channel closed")]
+    pub(super) async fn insert(
+        &mut self,
+        key: ConnectionKey,
+        conn: Connection,
+    ) -> (&mut Handle, bool) {
+        let (new_handle, acceptor) = conn.split();
```
gknopf-aranya:
```suggestion
        let (conn_handle, acceptor) = conn.split();
```

crates/aranya-daemon/src/sync/task/quic/connections.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        if inserted {
+            self.tx.send((key, acceptor)).await.expect("channel closed");
+        }
+
+        (handle, inserted)
+    }
+
+    pub(super) async fn remove(&mut self, key: ConnectionKey) {
+        if let Some(existing_conn) = self.guard.remove(&key) {
```
gknopf-aranya: This is only removing the connection handles used by the client from the map.
How are the server's `StreamAcceptor` cleaned up? I'm guessing the server cleans them up automatically if the stream operations return an error?

crates/aranya-daemon/src/sync/task/quic/connections.rs
gknopf-aranya, line 1: The rest of the types/methods in this file could use more docs

crates/aranya-util/src/addr.rs
```diff
@@ -200,6 +200,13 @@ impl<'de> Deserialize<'de> for Addr {
     }
 }
 
+impl TryFrom<&Addr> for SocketAddr {
+    type Error = AddrParseError;
+    fn try_from(value: &Addr) -> std::result::Result<Self, Self::Error> {
+        Self::from_str(&value.to_string())
+    }
+}
```
jdygert-spok: This doesn't seem right, an `Addr` can have a hostname, so converting to a socket addr is a matter of dns resolving.

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: We only attempt to accept bidirectional streams. There are no calls to `StreamAcceptor::accept` or `StreamAcceptor::accept_receive_stream`, etc.

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: It functions the way it did before this PR. The server accept bidi streams and treats them as incoming sync requests

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: How would using these methods help?

> Returning a boolean isn't very idiomatic for a Rust map insertion. Usually insert would return an Option<value>.

Yea, I stole the idea from `Hashset::insert` in the standard library. The docs for `MutexGuard::insert` explain better but the idea is that flag just tells you if you're using an existing connection or the one that was just inserted.

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: It's unreachable now but it may not be if during a refactor something like `drop(conn_map_guard)` was added a few lines above

crates/aranya-daemon/src/sync/task/quic/connections.rs
ssweetney-spok: Yea, the connection is closed just below this. If the task spawned by the sync server that accepts stream can't accept a stream for some reason then the task exits

crates/aranya-util/src/addr.rs
ssweetney-spok: Oh, I see. Good catch

crates/aranya-daemon/src/sync/task/quic/connections.rs
ssweetney-spok: Are there any ones in particular you'd like to see docs for? The rest are mostly wrappers around existing methods/types that don't change behavior

crates/aranya-daemon/src/sync/task/quic.rs
gknopf-aranya:
```suggestion
                error!(error = %err.report(), "server unable to respond to sync request from peer");
```

crates/aranya-daemon/src/sync/task/quic.rs
gknopf-aranya: It's preventing us from writing code like this:
```
if let Some(conn) = conns.insert(conn) { ... }
```
But it works so I can't complain too much.

crates/aranya-daemon/src/sync/task/quic.rs
gknopf-aranya: Then we should handle that case if/when it's refactored.

ssweetney-spok: In that case, I'll remove the bool being returned by `insert`

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -183,28 +180,64 @@ where {
...
     #[instrument(skip_all)]
     async fn connect(&mut self, peer: &Addr, id: GraphId) -> SyncResult<BidirectionalStream> {
         debug!("client connecting to QUIC sync server");
         // Check if there is an existing connection with the peer.
         // If not, create a new connection.
-        let conns = &mut self.state.conns;
+        let mut conn_map_guard = self.state.conns.lock().await;
         let client = &self.state.client;
 
         let key = ConnectionKey { addr: *peer, id };
```
jdygert-spok: The address should be resolved before this point and `ConnectionKey` should store `SocketAddr`.

ssweetney-spok: Yea, I just realized [that](https://github.com/aranya-project/aranya/pull/421/commits/d5b5965f9377054a482708cb12870d3ddf987bdd)

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -183,35 +176,72 @@ where {
...
-            Entry::Occupied(entry) => {
+        let addr = tokio::net::lookup_host(peer.to_socket_addrs())
+            .await
+            .context("DNS lookup on for peer address")?
+            .next()
+            .context("could not resolve peer address")?;
+
+        let key = ConnectionKey { addr, id };
+
+        let conn_ref = match conn_map_guard.get_mut(&key) {
```
jdygert-spok: Could have something like this on the map guard to be used here:
```rust
    async fn get_or_try_insert_with(
        &mut self,
        key: ConnectionKey,
        make_conn: impl F: AsyncFn() -> Result<Connection, Error>,
    ) -> Result<&mut Handle, Error>
```

Not sure it's all that much more clear :shrug:

crates/aranya-daemon/src/sync/task/quic/connections.rs
ssweetney-spok: Added some docs

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: Oh, passing the closure to the function is nice. Idk if makes things more clear but it protects against something like
`drop(conn_map_guard)` being added at the top of the `None` branch.

Edit: ehh, we'd have to obtain the lock again anyway 🤷

ssweetney-spok: This works better now since I wasn't checking that if `Some(conn)` was returned by `MutexGuard::get_mut` that it was still open.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -183,53 +184,80 @@ where {
...
-                conn.keep_alive(true).map_err(Error::from)?;
-                debug!("created new quic connection");
-                entry.insert(conn)
-            }
-        };
+                    .and_then(|mut conn| match conn.keep_alive(true) {
+                        Ok(_) => Ok(conn),
+                        Err(e) => Err(e),
+                    })
+                    .map_err(Error::from)
```
jdygert-spok: nit: this chain is awkward, I'd just use `?`.

```rust
                let mut conn = client
                    .await?;
                conn.keep_alive(true)?;
                Ok(conn)
```

djohnson-spideroak (commented):
LGTM pending open convos. I do think its important to mention the makeup of the composite key, and Dygert has a good conciseness suggestion.

crates/aranya-daemon/src/sync/task/quic.rs
ssweetney-spok: I'm using the fields in the parent module so I'm not really sure if making the fields private makes sense but I did reduce the visibility

crates/aranya-daemon/src/sync/task/quic/connections.rs
ssweetney-spok: Mentioned in the docs for `ConnectionKey`

jdygert-spok (approved):

PR #422, opened 2025-07-22
rustdoc: connect example

gknopf-aranya:
Add a `rustdoc` example showing how the Aranya client can connect to the Aranya daemon.

### DIFF
--- a/crates/aranya-client/src/client.rs
+++ b/crates/aranya-client/src/client.rs
@@ -137,6 +137,20 @@ impl Client {
     }
 
     /// Creates a client connection to the daemon.
+    ///
+    /// # Example
+    /// ```rust,no_run
+    /// use std::net::Ipv4Addr;
+    /// # use aranya_client::Client;
+    /// # #[tokio::main]
+    /// # async fn main() -> anyhow::Result<()> {
+    /// let client = Client::builder()
+    ///     .daemon_uds_path("/var/run/aranya/uds.sock".as_ref())
+    ///     .aqc_server_addr(&(Ipv4Addr::UNSPECIFIED, 1234).into())
+    ///     .connect()
+    ///     .await?;
+    /// #    Ok(())
+    /// # }
     #[instrument(skip_all)]
     async fn connect(uds_path: &Path, aqc_addr: &Addr) -> Result<Self> {
         info!(path = ?uds_path, "connecting to daemon");
--- a/crates/aranya-util/src/task.rs
+++ b/crates/aranya-util/src/task.rs
@@ -40,7 +40,8 @@ use tracing::Instrument;
 ///         sleep(Duration::from_secs(1)).await;
 ///         println!("{msg}");
 ///     });
-/// }).await;
+/// })
+/// .await;
 /// # }
 /// ```
 pub async fn scope<F>(f: F)
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,3 +1,4 @@
 unstable_features = true
 group_imports = "StdExternalCrate"
 imports_granularity = "Crate"
+format_code_in_doc_comments = true


crates/aranya-client/src/client.rs
```diff
@@ -137,6 +137,81 @@ impl Client {
...
+    /// use std::path::Path;
+    /// use tokio::process::Command;
+    /// use tokio::process::Child;
+    /// use tokio::time::sleep;
+    /// use tracing::debug;
+    /// use aranya_daemon::Daemon;
+    /// use aranya_daemon::config::Config;
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> Result<(), anyhow::Error> {
```
gknopf-aranya: Most of the code in this example is for spawning the daemon. We should use a utility method to spawn an example daemon instead. That will allow the example to be much shorter and focus more on the method the example is for.

crates/aranya-client/src/client.rs
```diff
@@ -137,6 +137,81 @@ impl Client {
     }
 
     /// Creates a client connection to the daemon.
+    ///
+    /// # Example
+    /// ```
+    /// use anyhow::{bail, Context as _, Result};
```
gknopf-aranya: @jdygert-spok is there a way to run `rustfmt` on `rustdocs` examples? Formatting this example manually would be too much effort.

crates/aranya-client/src/client.rs
```diff
@@ -137,6 +137,41 @@ impl Client {
...
+    ///     let work_dir = TempDir::with_prefix("daemon")?;
+    ///
+    ///     let daemon = DaemonProcess::spawn(work_dir.path()).await?;
+    ///     let uds_sock = daemon.uds_sock();
+    ///
+    ///     // Give the daemon time to start up and write its public key.
+    ///     sleep(Duration::from_millis(1000)).await;
+    ///
+    ///     let addr = Addr::from((Ipv4Addr::LOCALHOST, 0));
+    ///     let client = Client::builder()
```
gknopf-aranya: Use `ClientCtx`

crates/aranya-client/src/client.rs
```diff
@@ -137,6 +137,41 @@ impl Client {
...
+    ///
+    /// use anyhow::Result;
+    /// use aranya_client::{Client, DaemonProcess};
+    /// use aranya_util::Addr;
+    /// use tempfile::TempDir;
+    /// use tokio::time::sleep;
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> Result<(), anyhow::Error> {
+    ///     let work_dir = TempDir::with_prefix("daemon")?;
```
gknopf-aranya: Hide temp dir management in util

crates/aranya-client/src/client.rs
```diff
@@ -137,6 +137,41 @@ impl Client {
...
+    ///
+    ///     let daemon = DaemonProcess::spawn(work_dir.path()).await?;
+    ///     let uds_sock = daemon.uds_sock();
+    ///
+    ///     // Give the daemon time to start up and write its public key.
+    ///     sleep(Duration::from_millis(1000)).await;
```
gknopf-aranya: Hide this in `spawn()`?

crates/aranya-client/src/client.rs
gknopf-aranya, line 1: Once the daemon/client util code is in a good spot, the Rust example can be refactored to use it.

crates/aranya-client/src/client.rs
gknopf-aranya: Nevermind. Using `ClientCtx` would hide the `connect()` call. The purpose of this example is to show an example of invoking `connect()`.

crates/aranya-client/src/client.rs
gknopf-aranya: Delay should no longer be needed after https://github.com/aranya-project/aranya/pull/412

crates/aranya-client/src/daemon.rs
jdygert-spok, line 1: If we want to have this, we should move it to an unpublished `aranya-test-utils` crate.

gknopf-aranya: There is a lot of common code between the Rust example and `rustdocs` examples that could be shared via a util crate. If the crate is unpublished, would we be able to use it from both places? Would users need to copy the example code from the util crate before using it in their projects?

jdygert-spok: Yes, we will be able to use it from the doctests and rust example.

Users should not be using this code, it is not a realistic way to run things, only useful for our testing.

gknopf-aranya: I started moving this code to another crate but this caused a cyclical package dependency.
`aranya-test-util` depends on the `Client` in `aranya-client`.
`aranya-client` depends on `use aranya-test-util::daemon::DaemonProcess` in `aranya-test-util`.

gknopf-aranya: Made the `aranya-test-util` crate a dev dep of `aranya-client` rather than a regular dep.

crates/aranya-client/src/client.rs
gknopf-aranya: Using common util code for both `examples/rust` and `rustdoc` examples.

ssweetney-spok (dismissed):

crates/aranya-client/src/client.rs
```diff
@@ -137,6 +151,27 @@ impl Client {
...
+    /// # Example
+    /// ```
+    /// use anyhow::Result;
+    /// use aranya_client::Client;
+    /// use aranya_test_util::TestDaemon;
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> Result<(), anyhow::Error> {
+    ///     // Spawn the daemon.
+    ///     let daemon = TestDaemon::spawn("daemon").await?;
```
djohnson-spideroak: I think it would be good to add a bit of info on what this does, especially because connect is going to be very commonly seen by engineers exploring aranya for this first time.

crates/aranya-client/src/client.rs
```diff
@@ -137,6 +151,27 @@ impl Client {
...
+    ///
+    ///     // Connect the client to the daemon.
+    ///     let client = Client::builder()
+    ///         .daemon_uds_path(&daemon.uds_sock())
+    ///         .connect()
+    ///         .await?;
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
```
jdygert-spok: IMO we don't really need to run these examples and can make them simpler, something like this:

```suggestion
    /// ```no_run
    /// use std::net::Ipv4Addr;
    /// # use aranya_client::Client;
    /// # #[tokio::main]
    /// # async fn main() -> anyhow::Result<()> {
    /// let client = Client::builder()
    ///     .daemon_uds_path("/var/run/aranya/uds.sock".as_ref())
    ///     .aqc_server_addr(&(Ipv4Addr::UNSPECIFIED, 1234).into())
    ///     .connect()
    ///     .await?;
    /// #    Ok(())
    /// # }
```

This will show in the rustdoc as

```rust
use std::net::Ipv4Addr;
let client = Client::builder()
    .daemon_uds_path("/var/run/aranya/uds.sock".as_ref())
    .aqc_server_addr(&(Ipv4Addr::UNSPECIFIED, 1234).into())
    .connect()
    .await?;
```

`cargo test` will still check that it compiles but will not run it.

gknopf-aranya: Running the example is beneficial since it guarantees we will know when the example breaks.
These examples can also serve as a basic integration test suite for the client API. We've been wanting to add more integration tests.

gknopf-aranya: If we don't offer runnable examples, users may have difficulty translating the example into runnable code.

crates/aranya-client/src/client.rs
gknopf-aranya: Added some more docs to describe the test daemon. It's a daemon process with:
- a temporary working directory
- sample daemon config file with some default values
- sync server configured to bind to localhost address

crates/aranya-client/src/client.rs
jdygert-spok: I'm wary of all our examples looking like test code and not... examples. We don't expect users to use `TestDaemon` or something like it to spawn their daemon.

gknopf-aranya: Looking through some examples from `s2n_quic`. 
https://docs.rs/s2n-quic/latest/s2n_quic/connection/struct.Connection.html#examples
They don't appear to be compilable or runnable. They assume the object was already instantiated.

jdygert-spok: Look at the source, they do a similar thing with `no_run` and hidden code:
https://docs.rs/s2n-quic/latest/src/s2n_quic/connection.rs.html#43

gknopf-aranya: Pushed commit with `no_run`

PR #425, opened 2025-07-23
policy: add priorities to commands

gknopf-aranya:
Add priorities to commands.
Basing PR on custom roles since it has reworked the policy significantly.

Rules for determining command priorities:
- Revocation and deletion have highest priority
- Creation command has higher priority than commands after it
- Command that modifies role should have higher priority than command authorized by role

### DIFF
--- a/crates/aranya-daemon/src/policy.md
+++ b/crates/aranya-daemon/src/policy.md
@@ -119,8 +119,8 @@ CreateTeam is automatically the highest priority based on ancestry.
 In general, deletion and revocation commands should be higher priority than create/modify/use commands. E.g.:
 TerminateTeam(500) -> DeleteFoo(400) -> RevokeFoo(300) -> CreateFoo(200) -> UseFoo(100)
 
-Commands that must be called first are higher priority than commands that depend on them.
-For example, a command to assign a label to a device would be higher priority than a command that uses the label based on ancestry in the weave.
+Commands that must be called first should have a higher priority than commands that depend on them.
+For example, a command to create/assign a label to a device should be higher priority than a command that uses the label based.
 Delete*, Revoke*, Terminate*, Remove*, etc. commands must have a higher priority assigned to them since they occur later in the weave, but must take precedence over other commands that modify the state of the object.
 For example, deleting a label should be higher priority than assigning a label to a device because if the label doesn't exist, operations with the label are invalid.
 


crates/aranya-daemon/src/policy.md
```diff
@@ -3194,7 +3230,11 @@ effect LabelRevokedFromRole {
     author_id id,
 }
 
-command RevokeLabelFromDevice {
```
gknopf-aranya: There were 2 `RevokeLabelFromDevice` commands in the policy, so I changed this to `RevokeLabelFromRole` since it appears to be a typo

gknopf-aranya:
CI is not expected to pass on this PR since it is not passing on the base branch. That should not prevent merging this in as long as the policy compiles.

gknopf-aranya:
@elagergren-spideroak I'll let you merge this in since the custom roles base PR is yours.

gknopf-aranya:
Need to update this PR so it explicitly defines zero priorities

jdygert-spok:
Would it make sense to space out the values so we can add other commands between without changing them?

gknopf-aranya:
> Would it make sense to space out the values so we can add other commands between without changing them?

What do you mean by this?

crates/aranya-daemon/src/policy.md
gknopf-aranya: This has been addressed in a more recent version of the custom roles policy

jdygert-spok:
> > Would it make sense to space out the values so we can add other commands between without changing them?
> 
> What do you mean by this?

If you have priorities 0,1,2,3 and want to add a new one in the middle, you have to change 2 to 3 and 3 to 4. If you have them spaced out like 0,10,20,30 then you can just use 15.

crates/aranya-daemon/src/policy.md
```diff
@@ -438,6 +438,10 @@ effect QueryDevicesOnTeamResult {
 
 // A trampoline that forwards `device_id` to the effect.
 ephemeral command QueryDevicesOnTeam {
+    attributes {
+        priority: 0
+    }
```
jdygert-spok: Ephemeral commands shouldn't have priority.

crates/aranya-daemon/src/policy.md
```diff
@@ -2431,6 +2491,10 @@ effect TeamCreated {
 }
 
 command CreateTeam {
+    attributes {
+        priority: 0
+    }
```
jdygert-spok, line 2507: Init commands should have `init: true` instead.

crates/aranya-daemon/src/policy.md
djohnson-spideroak, line 1: Can you add docs on the reasoning behind why these priority values were chosen?

chip-so (approved):

crates/aranya-daemon/src/policy.md
gknopf-aranya: Added a section on priorities to the policy

crates/aranya-daemon/src/policy.md
```diff
@@ -101,6 +101,38 @@ publish to the graph. Commands are signed with the device's
 Signing Key. The policy controls which commands each device is
 authorized to publish.
 
+## Command Priorities
+
+Priorities can defined for commands by specifying the `priority` attribute in the command's `attribute` block:
```
```suggestion
Priorities can defined for commands by specifying the `priority` attribute in the command's `attributes` block:
```

chip-so (approved):

crates/aranya-daemon/src/policy.md
```diff
@@ -101,6 +101,38 @@ publish to the graph. Commands are signed with the device's
...
+TerminateTeam(500) -> DeleteFoo(400) -> RevokeFoo(300) -> CreateFoo(200) -> UseFoo(100)
+
+Commands that must be called first are higher priority than commands that depend on them.
+For example, a command to assign a label to a device would be higher priority than a command that uses the label based on ancestry in the weave.
+Delete*, Revoke*, Terminate*, Remove*, etc. commands must have a higher priority assigned to them since they occur later in the weave, but must take precedence over other commands that modify the state of the object.
+For example, deleting a label should be higher priority than assigning a label to a device because if the label doesn't exist, operations with the label are invalid.
+
+Command priorities will be required to be defined for each command in the policy. Even the zero priority should be defined. That way, the policy can be audited and does not have any default priorities that cannot be audited.
+
+Ephemeral commands do not need to specify a priority since they are automatically lower priority than all commands on the graph before the ephemeral command is evaluated.
```
jdygert-spok:
```suggestion
Ephemeral commands do not need to specify a priority since they do not participate in braiding.
```

crates/aranya-daemon/src/policy.md
```diff
@@ -101,6 +101,38 @@ publish to the graph. Commands are signed with the device's
...
+Commands are ordered first by weave ancestry, then priority, then by comparing their IDs.
+The weave ancestry is the deterministic order that commands are added to the graph when weaving branches together.
+CreateTeam is the first command (a.k.a. the init command), so it doesn't need an assigned priority.
+CreateTeam is automatically the highest priority based on ancestry.
+
+In general, deletion and revocation commands should be higher priority than create/modify/use commands. E.g.:
+TerminateTeam(500) -> DeleteFoo(400) -> RevokeFoo(300) -> CreateFoo(200) -> UseFoo(100)
+
+Commands that must be called first are higher priority than commands that depend on them.
+For example, a command to assign a label to a device would be higher priority than a command that uses the label based on ancestry in the weave.
```
jdygert-spok: If assigning should be higher than using, should UseFoo be higher than CreateFoo?

Not really sure what "ancestry in the weave" means here.

gknopf-aranya: `CreateFoo` is automatically higher priority than `UseFoo` because `CreateFoo` must be added to the graph before `UseFoo` is invoked for `UseFoo` to be a valid command. `CreateFoo` is higher priority based on ancestry in the weave, so it doesn't need a priority defined that's higher than `UseFoo` for the prioritization to be correct.

crates/aranya-daemon/src/policy.md
```diff
@@ -101,6 +101,38 @@ publish to the graph. Commands are signed with the device's
...
+```
+command CommandFoo {
+    attributes {
+        priority: 100
+    }
+}
+```
+
+Commands are ordered first by weave ancestry, then priority, then by comparing their IDs.
+The weave ancestry is the deterministic order that commands are added to the graph when weaving branches together.
```
jdygert-spok: This seems odd. How about this:

```suggestion
Branches in the graph are deterministically ordered by the braid algorithm. Higher priority commands are generally evaluated before lower priority commands.
```

(If we want to go in more detail, just "ancestry then priority than ID" isn't exactly correct)

crates/aranya-daemon/src/policy.md
gknopf-aranya:
```suggestion
TerminateTeam(500) -> DeleteFoo(400) -> RevokeFoo(300) -> CreateFoo(200) -> UseFoo(100)

Commands that must be called first should have a higher priority than commands that depend on them.
For example, a command to create/assign a label to a device should be higher priority than a command that uses the label based.
```

jdygert-spok: > `CreateFoo` is higher priority based on ancestry in the weave, so it doesn't need a priority defined that's higher than `UseFoo` for the prioritization to be correct.

I suppose that's true for our current policy, e.g. you can't use a label before it is assigned to you. I don't think it's immediately obvious that that should be true for all resources. (You might add to the graph that you are "attempting" to do something, and then later get permission. And it might be important to commit your attempt early on.)

But in short I don't think there's any downside to making creation a lower priority, so we could just recommend that?

jdygert-spok: I think I am confusing myself :thinking:

jdygert-spok: CreateFoo is higher than UseFoo as it "should" be 👍

jdygert-spok (approved):

PR #426, opened 2025-07-24
Shared connection refactor

jdygert-spok:
I got drawn into doing some refactoring. I moved the server select arms into methods, and made the shared connection map not expose the lock as part of the API.

### DIFF
--- a/crates/aranya-daemon/src/sync/task/quic.rs
+++ b/crates/aranya-daemon/src/sync/task/quic.rs
@@ -499,7 +499,7 @@ where
             anyhow::Ok(())
         }
         .unwrap_or_else(move |err| {
-            error!(error = ?err, "TODO");
+            error!(error = ?err, "server unable to accept connection");
             handle.close(AppError::UNKNOWN);
         })
     }
--- a/crates/aranya-daemon/src/sync/task/quic.rs
+++ b/crates/aranya-daemon/src/sync/task/quic.rs
@@ -108,7 +108,7 @@ pub struct State {
     /// QUIC client to make sync requests to another peer's sync server and handle sync responses.
     client: QuicClient,
     /// Address -> Connection map to lookup existing connections before creating a new connection.
-    conns: Arc<SharedConnectionMap>,
+    conns: SharedConnectionMap,
     /// PSK store shared between the daemon API server and QUIC syncer client and server.
     /// This store is modified by [`crate::api::DaemonApiServer`].
     store: Arc<PskStore>,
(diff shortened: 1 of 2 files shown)


crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -523,27 +475,61 @@ where
...
             };
-            debug!("received incoming QUIC stream");
-            if let Err(err) = Self::sync(client.clone(), peer, stream, &active_team).await {
-                error!(error = %err.report(), "server unable to respond to sync request from peer");
-                return;
+            self.conns.insert(key, conn).await;
+            anyhow::Ok(())
+        }
+        .unwrap_or_else(move |err| {
+            error!(error = ?err, "TODO");
```
gknopf-aranya: TODO

gknopf-aranya (approved):

crates/aranya-daemon/src/sync/task/quic/connections.rs
```diff
@@ -59,55 +81,50 @@ impl MutexGuard<'_> {
...
     }
 
-    /// Attempts to insert a QUIC connection into the map, failing if an open connection exists.
+    /// Inserts a QUIC connection into the map, unless an open connection exists.
     ///
     /// Checks if a connection already exists for the key. If found and still alive via ping,
-    /// returns an error with the original connection. If found but closed, replaces it with the
-    /// new connection. If no connection exists, inserts the new one. Sends a [`ConnectionUpdate`]
-    /// when a connection is successfully inserted. If the update cannot be sent, the newly
-    /// inserted connection is closed.
```
ssweetney-spok, line 110: I was thinking that users may encounter some weird states if we the connection isn't closed when sending the update fails. A server could accept an incoming connection but never be able to receive incoming sync requests if it never gets the update from the channel.

ssweetney-spok: Relatedly, I tried something like this yesterday

```
// Inside the insert methods

let send_result = self.tx.send((key, acceptor)).await;
match send_result {
    Ok(_) => Ok(handle),
    Err(e) => {
        self.remove(handle);
        return Err(e.into);
    }
}
```

but that doesn't compile with the current version of the borrow checker. Have to use nightly with the Polonious flag

ssweetney-spok (commented):
Good call on moving the code in the select branches into methods. `cargo fmt` will actually work on that code now.

LGTM pending the convo about closing the connection when sending an update fails

crates/aranya-daemon/src/sync/task/quic/connections.rs
jdygert-spok: That should never happen on the server side since the accepting and receiving happen at the same level. And on the client side if the receiver was dropped, it's still valid to use the connections.

PR #427, opened 2025-07-25
aqc: zeroize psks when deleting a channel

gknopf-aranya:
Delete PSKs when an AQC channel is deleted. More details in AQC channel deletion spec:
https://github.com/aranya-project/aranya-docs/pull/117

These changes are applicable to the future PSK->cert switch because certs will be deleted under similar conditions that PSKs are deleted under.

TODOs:
- [x] Update Rust example
- [x] Update C example
- [x] Update unit tests
- [x] Update C API in `defs.rs`
- [x] Implement bidi and uni channel delete APIs
- [x] Send signal from steam objects to `AqcClient` to delete channel when there is a connection error during stream operations.
- [x] Switch client keys data structure from `Vec` to map for faster lookup/delete based on PSK identity.
- [x] Zeroize PSKs instead of deleting them.
- [x] Close AQC channel when QUIC connection times out.
- [x] `Drop` impls for deleting channels, streams, and `AqcClient`. More idiomatic Rust API and prevents accidentally forgetting to clean up objects or PSKs when a channel falls out of scope.
- [x] Revert channel/stream to `AqcClient` signals.
- [x] Delete channel PSKs when channel is deleted.
- [x] Delete channel PSKs when stream operation encounters closed AQC channel error in connection error.
- [x] Send channel close error via connection close() when deleting an AQC channel.
- [x] Finalize error handling
- [x] Confirm that after a channel is closed, new streams cannot be opened on the channel and existing streams can no longer send/receive data
- [x] Delete PSKs from key stores when they are selected by rustls

Future work to handle on other PRs:
- [ ] Query daemon to see which active AQC channels are valid, delete invalid channels (may do this on another PR).
- [x] Provide API for closing streams in C API

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1621,7 +1621,7 @@ pub unsafe fn aqc_remove_net_identifier(
 /// `try_receive_channel`, and is invalidated after calling
 /// `get_bidi_channel`/`get_receive_channel`.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 152, align = 8)]
+#[aranya_capi_core::opaque(size = 160, align = 8)]
 pub type AqcPeerChannel = Safe<imp::AqcPeerChannel>;
 
 /// An enum containing all [`AqcPeerChannel`] variants.
@@ -1634,17 +1634,17 @@ pub enum AqcChannelType {
 
 /// An AQC Bidirectional Channel Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 144, align = 8)]
+#[aranya_capi_core::opaque(size = 152, align = 8)]
 pub type AqcBidiChannel = Safe<imp::AqcBidiChannel>;
 
 /// An AQC Sender Channel Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 144, align = 8)]
+#[aranya_capi_core::opaque(size = 152, align = 8)]
 pub type AqcSendChannel = Safe<imp::AqcSendChannel>;
 
 /// An AQC Receiver Channel Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 144, align = 8)]
+#[aranya_capi_core::opaque(size = 152, align = 8)]
 pub type AqcReceiveChannel = Safe<imp::AqcReceiveChannel>;
 
 /// An AQC Bidirectional Stream Object.
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -68,11 +68,7 @@ pub enum Error {
 
     /// A connection got unexpectedly closed.
     #[capi(msg = "connection got closed")]
-    ConnectionClosed,
-
-    /// A stream got unexpectedly closed.
-    #[capi(msg = "stream got closed")]
-    StreamClosed,
+    Closed,
 
     /// Unable to create configuration info.
     #[capi(msg = "invalid config")]
@@ -110,8 +106,7 @@ impl From<&imp::Error> for Error {
                 }
             },
             imp::Error::WouldBlock => Self::WouldBlock,
-            imp::Error::ConnectionClosed => Self::ConnectionClosed,
-            imp::Error::StreamClosed => Self::StreamClosed,
+            imp::Error::Closed => Self::Closed,
             imp::Error::Config(_) => Self::Config,
             imp::Error::Serialization(_) => Self::Serialization,
             imp::Error::Other(_) => Self::Other,
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -68,7 +68,15 @@ pub enum Error {
 
     /// A connection got unexpectedly closed.
     #[capi(msg = "connection got closed")]
-    Closed,
+    ConnectionClosed,
+
+    /// A stream got unexpectedly closed.
+    #[capi(msg = "stream got closed")]
+    StreamClosed,
+
+    /// A channel got unexpectedly closed.
+    #[capi(msg = "channel got closed")]
+    ChannelClosed,
 
     /// Unable to create configuration info.
     #[capi(msg = "invalid config")]
@@ -106,7 +114,9 @@ impl From<&imp::Error> for Error {
                 }
             },
             imp::Error::WouldBlock => Self::WouldBlock,
-            imp::Error::Closed => Self::Closed,
+            imp::Error::ConnectionClosed => Self::ConnectionClosed,
+            imp::Error::StreamClosed => Self::StreamClosed,
+            imp::Error::ChannelClosed => Self::ChannelClosed,
             imp::Error::Config(_) => Self::Config,
             imp::Error::Serialization(_) => Self::Serialization,
             imp::Error::Other(_) => Self::Other,
@@ -1649,17 +1659,17 @@ pub type AqcReceiveChannel = Safe<imp::AqcReceiveChannel>;
 
 /// An AQC Bidirectional Stream Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 184, align = 8)]
+#[aranya_capi_core::opaque(size = 208, align = 8)]
 pub type AqcBidiStream = Safe<imp::AqcBidiStream>;
 
 /// An AQC Sender Stream Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 152, align = 8)]
+#[aranya_capi_core::opaque(size = 176, align = 8)]
 pub type AqcSendStream = Safe<imp::AqcSendStream>;
 
 /// An AQC Receiver Stream Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 184, align = 8)]
+#[aranya_capi_core::opaque(size = 208, align = 8)]
 pub type AqcReceiveStream = Safe<imp::AqcReceiveStream>;
 
 /// Create a bidirectional AQC channel between this device and a peer.
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1732,8 +1732,8 @@ pub unsafe fn aqc_create_uni_channel(
 }
 
 /// Delete a bidirectional AQC channel.
-///
-/// Note that this function takes ownership of the [`AqcBidiChannel`] and invalidates any further use.
+/// Zeroizes PSKs associated with the channel.
+/// Closes all associated QUIC connections and streams.
 ///
 /// @param[in] client the Aranya Client [`Client`].
 /// @param[in] channel the AQC Channel [`AqcBidiChannel`] to delete.
@@ -1741,21 +1741,18 @@ pub unsafe fn aqc_create_uni_channel(
 /// @relates AranyaClient.
 pub fn aqc_delete_bidi_channel(
     client: &Client,
-    channel: OwnedPtr<AqcBidiChannel>,
+    channel: &mut AqcBidiChannel,
 ) -> Result<(), imp::Error> {
-    // SAFETY: the user is responsible for passing in a valid AqcBidiChannel pointer.
-    let channel = unsafe { Opaque::into_inner(channel.read()).into_inner().inner };
-
     let client = client.imp();
     client
         .rt
-        .block_on(client.inner.aqc().delete_bidi_channel(channel))?;
+        .block_on(client.inner.aqc().delete_bidi_channel(&mut channel.inner))?;
     Ok(())
 }
 
 /// Delete a send unidirectional AQC channel.
-///
-/// Note that this function takes ownership of the [`AqcSendChannel`] and invalidates any further use.
+/// Zeroizes PSKs associated with the channel.
+/// Closes all associated QUIC connections and streams.
 ///
 /// @param[in] client the Aranya Client [`Client`].
 /// @param[in] channel the AQC Channel [`AqcSendChannel`] to delete.
@@ -1763,21 +1760,21 @@ pub fn aqc_delete_bidi_channel(
 /// @relates AranyaClient.
 pub fn aqc_delete_send_uni_channel(
     client: &Client,
-    channel: OwnedPtr<AqcSendChannel>,
+    channel: &mut AqcSendChannel,
 ) -> Result<(), imp::Error> {
-    // SAFETY: the user is responsible for passing in a valid AqcSendChannel pointer.
-    let channel = unsafe { Opaque::into_inner(channel.read()).into_inner().inner };
-
     let client = client.imp();
-    client
-        .rt
-        .block_on(client.inner.aqc().delete_send_uni_channel(channel))?;
+    client.rt.block_on(
+        client
+            .inner
+            .aqc()
+            .delete_send_uni_channel(&mut channel.inner),
+    )?;
     Ok(())
 }
 
 /// Delete a receive unidirectional AQC channel.
-///
-/// Note that this function takes ownership of the [`AqcReceiveChannel`] and invalidates any further use.
+/// Zeroizes PSKs associated with the channel.
+/// Closes all associated QUIC connections and streams.
 ///
 /// @param[in] client the Aranya Client [`Client`].
 /// @param[in] channel the AQC Channel [`AqcReceiveChannel`] to delete.
@@ -1785,15 +1782,15 @@ pub fn aqc_delete_send_uni_channel(
 /// @relates AranyaClient.
 pub fn aqc_delete_receive_uni_channel(
     client: &Client,
-    channel: OwnedPtr<AqcReceiveChannel>,
+    channel: &mut AqcReceiveChannel,
 ) -> Result<(), imp::Error> {
-    // SAFETY: the user is responsible for passing in a valid AqcSendChannel pointer.
-    let channel = unsafe { Opaque::into_inner(channel.read()).into_inner().inner };
-
     let client = client.imp();
-    client
-        .rt
-        .block_on(client.inner.aqc().delete_receive_uni_channel(channel))?;
+    client.rt.block_on(
+        client
+            .inner
+            .aqc()
+            .delete_receive_uni_channel(&mut channel.inner),
+    )?;
     Ok(())
 }
 
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1753,7 +1753,7 @@ pub fn aqc_delete_bidi_channel(
     Ok(())
 }
 
-/// Delete a unidirectional AQC channel.
+/// Delete a send unidirectional AQC channel.
 ///
 /// Note that this function takes ownership of the [`AqcSendChannel`] and invalidates any further use.
 ///
@@ -1761,7 +1761,7 @@ pub fn aqc_delete_bidi_channel(
 /// @param[in] channel the AQC Channel [`AqcSendChannel`] to delete.
 ///
 /// @relates AranyaClient.
-pub fn aqc_delete_uni_channel(
+pub fn aqc_delete_send_uni_channel(
     client: &Client,
     channel: OwnedPtr<AqcSendChannel>,
 ) -> Result<(), imp::Error> {
@@ -1771,7 +1771,29 @@ pub fn aqc_delete_uni_channel(
     let client = client.imp();
     client
         .rt
-        .block_on(client.inner.aqc().delete_uni_channel(channel))?;
+        .block_on(client.inner.aqc().delete_send_uni_channel(channel))?;
+    Ok(())
+}
+
+/// Delete a receive unidirectional AQC channel.
+///
+/// Note that this function takes ownership of the [`AqcReceiveChannel`] and invalidates any further use.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] channel the AQC Channel [`AqcReceiveChannel`] to delete.
+///
+/// @relates AranyaClient.
+pub fn aqc_delete_receive_uni_channel(
+    client: &Client,
+    channel: OwnedPtr<AqcReceiveChannel>,
+) -> Result<(), imp::Error> {
+    // SAFETY: the user is responsible for passing in a valid AqcSendChannel pointer.
+    let channel = unsafe { Opaque::into_inner(channel.read()).into_inner().inner };
+
+    let client = client.imp();
+    client
+        .rt
+        .block_on(client.inner.aqc().delete_receive_uni_channel(channel))?;
     Ok(())
 }
 
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1920,23 +1920,6 @@ pub fn aqc_bidi_create_bidi_stream(
     Ok(())
 }
 
-/// Close a bidirectional stream from an [`AqcBidiChannel`].
-///
-/// @param[in]  client the Aranya Client [`Client`].
-/// @param[in]  stream the AQC send stream object [`AqcBidiStream`].
-///
-/// @relates AranyaClient.
-pub fn aqc_bidi_stream_close(
-    client: &Client,
-    stream: OwnedPtr<AqcBidiStream>,
-) -> Result<(), imp::Error> {
-    // SAFETY: the user is responsible for passing in a valid AqcBidiStream pointer.
-    let mut stream = unsafe { Opaque::into_inner(stream.read()).into_inner().inner };
-
-    client.rt.block_on(stream.close())?;
-    Ok(())
-}
-
 /// Send some data to a peer using an [`AqcBidiStream`].
 ///
 /// @param[in] client the Aranya Client [`Client`].
@@ -2068,42 +2051,6 @@ pub fn aqc_send_create_uni_stream(
     let send = client.rt.block_on(channel.inner.create_uni_stream())?;
 
     AqcSendStream::init(stream, imp::AqcSendStream::new(send));
-    Ok(())
-}
-
-/// Close a unidirectional stream from an [`AqcSendChannel`].
-///
-/// @param[in]  client the Aranya Client [`Client`].
-/// @param[in]  stream the AQC send stream object [`AqcSendStream`].
-///
-/// @relates AranyaClient.
-pub fn aqc_send_uni_stream_close(
-    client: &Client,
-    stream: OwnedPtr<AqcSendStream>,
-) -> Result<(), imp::Error> {
-    // SAFETY: the user is responsible for passing in a valid AqcSendStream pointer.
-    let mut stream = unsafe { Opaque::into_inner(stream.read()).into_inner().inner };
-
-    client.rt.block_on(stream.close())?;
-    Ok(())
-}
-
-/// Close a unidirectional stream from an [`AqcReceiveChannel`].
-///
-/// The client paramater is currently unused but is required for backward compitibility in case it is needed in the future.
-///
-/// @param[in]  _client the Aranya Client [`Client`].
-/// @param[in]  stream the AQC send stream object [`AqcReceiveStream`].
-///
-/// @relates AranyaClient.
-pub fn aqc_receive_uni_stream_close(
-    _client: &Client,
-    stream: OwnedPtr<AqcReceiveStream>,
-) -> Result<(), imp::Error> {
-    // SAFETY: the user is responsible for passing in a valid AqcReceiveStream pointer.
-    let mut stream = unsafe { Opaque::into_inner(stream.read()).into_inner().inner };
-
-    stream.close();
     Ok(())
 }
 
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1621,7 +1621,7 @@ pub unsafe fn aqc_remove_net_identifier(
 /// `try_receive_channel`, and is invalidated after calling
 /// `get_bidi_channel`/`get_receive_channel`.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 120, align = 8)]
+#[aranya_capi_core::opaque(size = 144, align = 8)]
 pub type AqcPeerChannel = Safe<imp::AqcPeerChannel>;
 
 /// An enum containing all [`AqcPeerChannel`] variants.
@@ -1634,17 +1634,17 @@ pub enum AqcChannelType {
 
 /// An AQC Bidirectional Channel Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 112, align = 8)]
+#[aranya_capi_core::opaque(size = 136, align = 8)]
 pub type AqcBidiChannel = Safe<imp::AqcBidiChannel>;
 
 /// An AQC Sender Channel Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 112, align = 8)]
+#[aranya_capi_core::opaque(size = 136, align = 8)]
 pub type AqcSendChannel = Safe<imp::AqcSendChannel>;
 
 /// An AQC Receiver Channel Object.
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 112, align = 8)]
+#[aranya_capi_core::opaque(size = 136, align = 8)]
 pub type AqcReceiveChannel = Safe<imp::AqcReceiveChannel>;
 
 /// An AQC Bidirectional Stream Object.
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1920,6 +1920,23 @@ pub fn aqc_bidi_create_bidi_stream(
     Ok(())
 }
 
+/// Close a bidirectional stream from an [`AqcBidiChannel`].
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  stream the AQC send stream object [`AqcBidiStream`].
+///
+/// @relates AranyaClient.
+pub fn aqc_bidi_stream_close(
+    client: &Client,
+    stream: OwnedPtr<AqcBidiStream>,
+) -> Result<(), imp::Error> {
+    // SAFETY: the user is responsible for passing in a valid AqcBidiStream pointer.
+    let mut stream = unsafe { Opaque::into_inner(stream.read()).into_inner().inner };
+
+    client.rt.block_on(stream.close())?;
+    Ok(())
+}
+
 /// Send some data to a peer using an [`AqcBidiStream`].
 ///
 /// @param[in] client the Aranya Client [`Client`].
@@ -2051,6 +2068,42 @@ pub fn aqc_send_create_uni_stream(
     let send = client.rt.block_on(channel.inner.create_uni_stream())?;
 
     AqcSendStream::init(stream, imp::AqcSendStream::new(send));
+    Ok(())
+}
+
+/// Close a unidirectional stream from an [`AqcSendChannel`].
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  stream the AQC send stream object [`AqcSendStream`].
+///
+/// @relates AranyaClient.
+pub fn aqc_send_uni_stream_close(
+    client: &Client,
+    stream: OwnedPtr<AqcSendStream>,
+) -> Result<(), imp::Error> {
+    // SAFETY: the user is responsible for passing in a valid AqcSendStream pointer.
+    let mut stream = unsafe { Opaque::into_inner(stream.read()).into_inner().inner };
+
+    client.rt.block_on(stream.close())?;
+    Ok(())
+}
+
+/// Close a unidirectional stream from an [`AqcReceiveChannel`].
+///
+/// The client paramater is currently unused but is required for backward compitibility in case it is needed in the future.
+///
+/// @param[in]  _client the Aranya Client [`Client`].
+/// @param[in]  stream the AQC send stream object [`AqcReceiveStream`].
+///
+/// @relates AranyaClient.
+pub fn aqc_receive_uni_stream_close(
+    _client: &Client,
+    stream: OwnedPtr<AqcReceiveStream>,
+) -> Result<(), imp::Error> {
+    // SAFETY: the user is responsible for passing in a valid AqcReceiveStream pointer.
+    let mut stream = unsafe { Opaque::into_inner(stream.read()).into_inner().inner };
+
+    stream.close();
     Ok(())
 }
 
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -66,10 +66,6 @@ pub enum Error {
     #[capi(msg = "no response ready yet")]
     WouldBlock,
 
-    /// A channel got unexpectedly closed.
-    #[capi(msg = "channel got closed")]
-    ChannelClosed,
-
     /// A connection got unexpectedly closed.
     #[capi(msg = "connection got closed")]
     ConnectionClosed,
@@ -114,7 +110,6 @@ impl From<&imp::Error> for Error {
                 }
             },
             imp::Error::WouldBlock => Self::WouldBlock,
-            imp::Error::ChannelClosed => Self::ChannelClosed,
             imp::Error::ConnectionClosed => Self::ConnectionClosed,
             imp::Error::StreamClosed => Self::StreamClosed,
             imp::Error::Config(_) => Self::Config,
@@ -1755,7 +1750,6 @@ pub fn aqc_delete_bidi_channel(
     client
         .rt
         .block_on(client.inner.aqc().delete_bidi_channel(channel))?;
-
     Ok(())
 }
 
@@ -1778,7 +1772,6 @@ pub fn aqc_delete_uni_channel(
     client
         .rt
         .block_on(client.inner.aqc().delete_uni_channel(channel))?;
-
     Ok(())
 }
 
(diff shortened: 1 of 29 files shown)


gknopf-aranya:
Noticed that PSK secret field is already wrapped in `Zeroizing` for the `PresharedKey` type. So, zeroization should be happening automatically when the PSKs are removed from the map.

gknopf-aranya:
QUIC connection will automatically close when the connection times out. `keep_alive()` is configured for connections to keep them open longer than the timeout, so the connection should only close due to loss of connectivity rather than inactivity.

crates/aranya-client/src/aqc/net/channels.rs
```diff
@@ -215,49 +355,96 @@ impl AqcBidiChannel {
...
 impl Drop for AqcBidiChannel {
     fn drop(&mut self) {
-        self.close();
+        debug!("dropping aqc bidi channel");
+        let _ = futures_lite::future::block_on(self.close());
     }
 }
 
 /// Used to send and receive data with a peer.
+// TODO: implement `Drop`
```
gknopf-aranya: Unable to implement `Drop` due to `split()` method moving ownership of fields out of object.

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -85,35 +98,53 @@ impl SelectsPresharedKeys for ServerPresharedKeys {
 
 #[derive(Debug)]
 pub struct ClientPresharedKeys {
-    keys: Mutex<Vec<Arc<PresharedKey>>>,
+    keys: Mutex<BTreeMap<PskIdentity, Arc<PresharedKey>>>,
```
ssweetney-spok: Check out [this type](https://github.com/aranya-project/aranya/blob/main/crates/aranya-daemon/src/sync/task/quic/psk.rs#L213-L229).

You can avoid having to store the identity and the key separately. That'll let you use a hashset/btreeset

gknopf-aranya: In this case, I'm storing a list of PSK identities for a specific channel and references to the keystores for all channels. When a channel is deleted, I only want to delete the PSKs for that specific channel from the keystores. We aren't managing a different keystore for each channel.

ssweetney-spok: I think you're describing the `ChannelKeys` type? I was referring to the type that's in the diff above: `ClientPresharedKeys`

ssweetney-spok: Oh, I think see where the confusion is.

The `PskIdAsKey` has custom implementations for `Borrow`, `Hash`, etc. so you can lookup and remove from the set by the PSK identity.

gknopf-aranya: What I mean is the set of PSK identities and the set of PSKs are different here. So it doesn't make sense to use the PSK as the key to the map and the entry in the map.

ssweetney-spok: The entire PSK itself wouldn't be the key, just the identity. That's what the custom trait impls are doing

See [my commit here](https://github.com/aranya-project/aranya/commit/8c93c194cb3d4447019a0f042a9e32015eb85417).

It should function the same. We just save 34 bytes (I think that's the size of the PSK identity?) per PSK

gknopf-aranya: Cool, I tried using a `HashMap` instead of a `BTreeMap` at some point but it didn't work since the order seems to matter.

ssweetney-spok: Ah, maybe use a `BtreeSet` just in case then

gknopf-aranya: `PresharedKey` doesn't have `Ord` implemented on it which is needed by `BTreeSet`.

gknopf-aranya: Opened an issue to add more impls to the `PresharedKey` type:
https://github.com/aranya-project/aranya-internal-rustls/issues/3
Not planning to block this PR on that.

crates/aranya-client/src/aqc/net/channels.rs
```diff
@@ -215,49 +338,72 @@ impl AqcBidiChannel {
...
-    /// Close the stream.
-    pub async fn close(&mut self) -> Result<(), AqcError> {
-        self.0.close().await?;
+        if let Err(e) = self.stream.send(data).await {
+            let _ = self.close().await;
+            if channel_closed(e) {
+                self.keys.zeroize();
+            }
+            return Err(AqcError::StreamError(e));
+        }
```
djohnson-spideroak: is it correct to always close if we get an error from the stream? I would imagine there are recoverable errors.

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -60,6 +62,13 @@ impl ServerPresharedKeys {
         }
     }
 
+    pub fn zeroize_psks(&self, identities: &[PskIdentity]) {
+        let mut keys = self.keys.lock().expect("poisoned");
+        identities.iter().for_each(|i| {
+            keys.remove(i);
+        });
+    }
```
djohnson-spideroak: is this actually zeroing?

gknopf-aranya: Yes, the underlying `PresharedKey` is wrapped in `Zeroizing` which ensures the PSK will be zeroized when dropped.

crates/aranya-client/src/aqc/net/channels.rs
gknopf-aranya: These are the possible stream and connection errors:
https://docs.rs/s2n-quic/latest/s2n_quic/stream/enum.Error.html
https://docs.rs/s2n-quic/latest/s2n_quic/connection/enum.Error.html
There are a few that could be recoverable like `SendingBlocked`.
In general the errors would indicate loss of network connectivity, a channel/connection was closed, or a bug in AQC.

gknopf-aranya: My reasoning for always closing the stream is that a stream error could indicate something nefarious is going on. Since AQC channels/streams are ephemeral, it doesn't hurt to be overly cautious and close the stream if an error is encountered.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: This provides a better guarantee than setting the values to zero manually which could get optimized out if the value is never used again. It also protects against the case where more copies of the data were made somewhere in the implementation or data structure, because each copy will be zeroized when it is dropped.

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -85,35 +98,53 @@ impl SelectsPresharedKeys for ServerPresharedKeys {
...
-            .collect::<Option<Vec<_>>>()
-            .expect("can create psks");
-        *self.keys.lock().expect("poisoned") = keys;
+        let mut keys = self.keys.lock().expect("poisoned");
+        for (suite, psk) in psks {
+            let identity = psk.identity().as_bytes().to_vec();
+            let key = make_preshared_key(suite, psk).expect("can make psk");
+            keys.insert(identity, key);
+        }
+    }
```
jdygert-spok: This isn't right, this means we can connect with any channel's psks. This method needs to set the "active" psks.

gknopf-aranya: Was that a bug already or did this change introduce it?

jdygert-spok: Just introduced, the existing code overwrites the vec.

ssweetney-spok: I think this might be the cause of [this issue](https://github.com/aranya-project/aranya/pull/427#discussion_r2243546842). Adding `keys.clear()` seems to have fixed the issue with the example hanging more than half the time when trying to receive a channel if you use a `HashMap` instead of a `BtreeMap`.

gknopf-aranya: Added a `keys.clear()` before inserting so it resets the psks contained in the map.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: Switched over to `HashSet` for client and server PSK maps.

crates/aranya-client/src/aqc/net/channels.rs
gknopf-aranya: I read through the errors in more detail and confirmed that most of them are not recoverable or are the result of a bug/misconfiguration.
It's possible to recover from `SendingBlocked`, but it indicates we tried to send data before the stream was ready:
"The stream is blocked on writing data. This is caused by trying to send data before polling readiness"

gknopf-aranya: It's not possible for the `SendingBlocked` error to occur with our current implementation because we are calling `send()` and waiting on the future. If we were calling `start_send()` without checking `poll_ready()` first, it's possible to get a `SendingBlocked` error.
https://docs.rs/s2n-quic/latest/s2n_quic/stream/struct.SendStream.html
https://docs.rs/futures-sink/0.3.31/futures_sink/trait.Sink.html#tymethod.poll_ready
https://docs.rs/futures-sink/0.3.31/futures_sink/trait.Sink.html#tymethod.start_send

examples/rust/src/main.rs
```diff
@@ -450,6 +450,10 @@ async fn main() -> Result<()> {
     let bytes = peer2.receive().await?.assume("no data received")?;
     assert_eq!(bytes, msg);
 
+    // membera deletes the AQC channel.
+    info!("membera deleting aqc channel");
+    membera.client.aqc().delete_bidi_channel(created_aqc_chan).await?;
+
```
djohnson-spideroak: do we have any other tests for deleting the channel?

gknopf-aranya: There are more tests in `crates/aranya-client/tests/aqc.rs`

gknopf-aranya: Adding `delete_uni_channel` to the unit tests. Noticed we currently only have a way to delete a uni channel from the send side. Do we want to support deleting a channel from the receive side as well?

gknopf-aranya:
Added methods for closing streams to the C API

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -40,63 +42,90 @@ impl ServerPresharedKeys {
...
+            error!("Duplicate PSK identity inserted: {:?}", identity);
         }
     }
 
+    pub fn zeroize_psks(&self, identities: &[PskIdentity]) {
+        let mut keys = self.keys.lock().expect("poisoned");
+        identities.iter().for_each(|i| {
+            keys.remove(i.as_slice());
+        });
+    }
```
ssweetney-spok: nit: The name is kind of misleading because it may not zeroize the PSKs if there are other references.

```Rust
fn foo(key: Arc<PresharedKey>) {
    let identity = key.identity();

    let keys = ServerPresharedKeys::new();
    keys.insert(Arc::clone(&key)); // Increase the ref count by 1
    keys.zeroize_psks(&[identity.to_vec()]); // decrease the ref count by 1

    // can still access data with `key` variable
}
```

crates/aranya-client/src/aqc/net.rs
```diff
@@ -253,10 +323,12 @@ impl AqcClient {
                 );
                 AqcError::NoChannelInfoFound
             })?;
+            let keys = ChannelKeys::new(vec![identity], client_keys, self.server_keys.clone());
```
ssweetney-spok: This is grabbing the identity from the connection but there should be multiple PSK identities associated with a channel. It seems like there would still be some data left in `AqcClient::channels` that should be removed when a channel is deleted.

Edit: Some PSKs left in the server key store possibly?

gknopf-aranya: There is only a single `identity` returned from `get_conn_identity()` above. When an AQC channel is deleted, only the PSKs for that channel should be deleted from the key store, the rest can remain. The server may have PSKs for multiple channels stored in the in-memory key store.

djohnson-spideroak: I think the distinction is that a channel could have multiple connections, each with their own PSK.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: The method is zeroing PSKs that correspond to the list of identities, not all the PSKs in the key store.

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: In the AQC spec, it's possible to have multiple connections per AQC channel.
However, in the current implementation of AQC there is only a single connection per channel.
Since the goal of this PR is to delete channel PSKs when a channel is deleted, I'd prefer not to complicate things by handling multiple connection corner cases on this PR.
We could open another PR to add support for multiple connections if that is desired.

gknopf-aranya: I thought the multiple PSKs per AQC channel was for different crypto engines.

gknopf-aranya: As far as I can tell, created channels may have multiple PSKs, while received channels only have a single PSK from the identity hint.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: Added some docs to clarify what this method does.

crates/aranya-client/src/aqc/net.rs
ssweetney-spok: I could be misinterpreting it but the `AqcClient::process_ctrl_message` method creates a new entry for each PSK

crates/aranya-client/src/aqc/crypto.rs
ssweetney-spok: > The method is zeroing PSKs that correspond to the list of identities, not all the PSKs in the key store.

This method isn't guaranteed to actually zeroize the PSKs if there are other references. I put a contrived example above just to demonstrate that

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: If we add the PSK identities to `ChannelInfo` they could be looked up there. Not sure if I like making multiple copies of the identity data to make it easier to look up though.

gknopf-aranya: I can put the identities in an `Arc` since they are ready-only. That avoids extra copies of the data.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: Because the `PresharedKey` secret is wrapped in `Zeroizing` the PSK will be zeroized once all references have been dropped. This is only removing the entry from the key store which will cause the `PresharedKey` to be dropped as long as there are no other references. If there are other references to the PSK somewhere else, then those references wouldn't zeroized until they are all dropped.

It would be a bug if the AQC implementation held references to the PSKs outside the keystore for anything other than temporarily:
1. Initializing the channel
2. Loading the PSK into `rustls` for a connection
3. Forming the ctrl msg to send to a peer

There should not be any objects holding a `PresharedKey` long-term other than the keystores.

gknopf-aranya: I could call the method `remove_psks()` which may be more accurate since PSKs are removed from the map. But then the name of the method would not communicate the intent to zeroize the PSKs after they are dropped from the key store.

gknopf-aranya: Documented the assumptions made about references and the drop impl.

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok it looks like we can't call `delete` and `cleanup` on channels in the C API without it causing a segfault. Why are both methods available to the application?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1766,7 +1771,29 @@ pub fn aqc_delete_uni_channel(
...
+}
+
+/// Delete a receive unidirectional AQC channel.
+///
+/// Note that this function takes ownership of the [`AqcReceiveChannel`] and invalidates any further use.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] channel the AQC Channel [`AqcReceiveChannel`] to delete.
+///
+/// @relates AranyaClient.
```
djohnson-spideroak: Lets make it more clear what these delete functions are doing, especially around what this is actually deleting.

crates/aranya-client/tests/aqc.rs
```diff
@@ -245,6 +259,19 @@ async fn test_aqc_chans() -> Result<()> {
...
+            .client
+            .aqc()
+            .delete_bidi_channel(bidi_chan1)
+            .await?;
+        devices
+            .memberb
+            .client
+            .aqc()
+            .delete_bidi_channel(bidi_chan2)
+            .await?;
```
djohnson-spideroak: do we have tests to verify the channel is actually deleted locally?

gknopf-aranya: Do you mean tests that verify the AQC PSKs have been deleted?
We could add test only methods that get the PSKs from the key stores before and after deleting a channel to confirm the channels PSKs have been deleted after the channel is deleted.

djohnson-spideroak: I think it would be fine to try and send/recv on the channel after deletion.

gknopf-aranya: That case is tested at the end of `test_aqc_chans_delete_chan_send_recv`

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I'll mention the zeroized PSKs here

gknopf-aranya:
I have the channel delete methods in the C API taking a `chan: &mut <ChannelType>` instead of an `OwnedPtr` so `cleanup()` methods are the only ones that free the objects now.

djohnson-spideroak (approved):

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -38,65 +40,101 @@ impl ServerPresharedKeys {
...
 
     pub fn load_psks(&self, psks: AqcPsks) {
-        let keys = psks
-            .into_iter()
-            .map(|(suite, psk)| make_preshared_key(suite, psk))
-            .collect::<Option<Vec<_>>>()
-            .expect("can create psks");
-        *self.keys.lock().expect("poisoned") = keys;
+        let mut keys = self.keys.lock().expect("poisoned");
+        keys.clear();
```
jdygert-spok: If we are just going to clear the previous keys whenever we open a new channel, we might as well clear them when we finish opening the channel and not need to stash them in the channel object.

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -38,65 +40,101 @@ impl ServerPresharedKeys {
...
+        let mut keys = self.keys.lock().expect("poisoned");
+        if !keys.insert(PskIdAsKey(psk)) {
+            error!("Duplicate PSK identity inserted: {:?}", identity);
         }
     }
 
+    /// Zeroize PSKs with the provided identities.
+    /// Removes the PSKs from the map so the secret can be zeroized via `Zeroizing` when dropped.
+    /// Assumes there are no other references to the zeroized PSKs held long-term outside of this keystore.
+    pub fn zeroize_psks(&self, identities: &[PskIdentity]) {
```
jdygert-spok: These zeroize methods should just be called remove.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -415,6 +521,20 @@ impl AqcClient {
...
+        self.channels.write().expect("poisoned").clear();
+        self.client_state.lock().await.clear();
+        self.server_keys.clear();
+    }
+}
+
+impl Drop for AqcClient {
+    fn drop(&mut self) {
+        futures_lite::future::block_on(self.close());
+    }
```
jdygert-spok: This close method and drop impl are redundant, all of the clearing that it does will happen just the same when those fields are dropped.

crates/aranya-client/src/aqc/net/channels.rs
```diff
@@ -215,49 +319,61 @@ impl AqcBidiChannel {
...
+                let _ = self.close().await;
+                Err(AqcError::ConnectionError(e))
+            }
+        }
     }
 
     /// Close the channel if it's open. If the channel is already closed, do nothing.
-    pub fn close(&mut self) {
+    #[instrument(skip_all)]
+    pub async fn close(&mut self) -> Result<(), AqcError> {
```
jdygert-spok: These dont' need to be async

crates/aranya-client/src/aqc/net.rs
```diff
@@ -209,32 +249,65 @@ impl AqcClient {
...
+        ))
+    }
+
+    /// Zeroize PSKs with provided identities from client and server key stores.
+    #[instrument(skip(self))]
+    async fn zeroize_psks(&self, identities: Vec<PskIdentity>) {
+        self.client_state.lock().await.zeroize_psks(&identities);
+        self.server_keys.zeroize_psks(&identities);
+        for identity in identities {
+            self.channels.write().expect("poisoned").remove(&identity);
```
jdygert-spok: Looks like this channel info map doesn't get updated when the channels are closed directly.

crates/aranya-client/src/aqc/net/channels.rs
```diff
@@ -15,6 +22,39 @@ mod s2n {
...
 }
 
+/// AQC channel Keys.
+///
+/// Each AQC channel will have a unique set of PSK identities and keys.
+/// This structure provides access to a copy of client and server keys.
+/// It enables the channel PSKs to easily be zeroized when an AQC channel is deleted.
+#[derive(Debug, Clone)]
+pub struct ChannelKeys {
+    identities: Arc<Vec<PskIdentity>>,
```
jdygert-spok: We should probably just pass the channel ID and have the key managers store mappings as needed.

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: That was also the case before this PR.
The channel info map doesn't contain any PSKs, so I'm less concerned about removing entries from the map in this PR.
Removing them would be beneficial from a garbage collection standpoint, but it isn't much data.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: There was a similar discussion about this here:
https://github.com/aranya-project/aranya/pull/427#discussion_r2252438786
I want to communicate that the intent is to zeroize the PSKs, not just remove them from the map.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: Where would the PSKs be cleared? Right after the connection is established in `connect_data()`?

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: Removing this

crates/aranya-client/src/aqc/crypto.rs
jdygert-spok: Yes, or `PresharedKeyStore::psks` could `mem::take` them.

gknopf-aranya: This approach did not work:
```
    /// Establish connection for sending a data message.
    #[instrument(skip(self, psks))]
    fn connect_data(
        &mut self,
        addr: SocketAddr,
        psks: AqcPsks,
    ) -> s2n_quic::client::ConnectionAttempt {
        self.client_keys.load_psks(psks);
        let conn = self
            .quic_client
            .connect(Connect::new(addr).with_server_name(addr.ip().to_string()));
        self.client_keys.clear();
        conn
    }
```
Could try the `mem::take` approach, although that means there could only ever be a single connection made with the channel PSKs.

gknopf-aranya: I don't think we should add multi-connection support in this PR anyways so it should be okay.

crates/aranya-client/src/aqc/api.rs
```diff
@@ -126,43 +127,50 @@ impl<'a> AqcChannels<'a> {
...
-        //     .daemon
-        //     .delete_aqc_bidi_channel(context::current(), chan.aqc_id().into_id().into())
-        //     .await
-        //     .map_err(IpcError::new)?
-        //     .map_err(aranya_error)?;
-        chan.close();
+    /// Zeroizes PSKs associated with the channel.
+    /// Closes all associated QUIC connections and streams.
+    #[instrument(skip_all, fields(aqc_id = %chan.aqc_id(), label = %chan.label_id()))]
+    pub async fn delete_bidi_channel(&mut self, chan: &mut AqcBidiChannel) -> crate::Result<()> {
```
ssweetney-spok, line 134: Now that these methods no longer consume the channel, do we plan on adding a way to "restore" a channel from an existing channel object?

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: Ended up being simpler to `drain()` the hash map to return an iterator to the removed elements.

gknopf-aranya: Should we use a similar approach for getting rid of the server keys after a connection is established?

crates/aranya-client/src/aqc/net/channels.rs
```diff
@@ -215,49 +319,61 @@ impl AqcBidiChannel {
...
-pub struct AqcBidiStream(s2n::BidirectionalStream);
+pub struct AqcBidiStream {
+    stream: s2n::BidirectionalStream,
+}
 
 impl AqcBidiStream {
+    /// Creates a new AQC bidirectional stream.
+    pub fn new(stream: s2n::BidirectionalStream) -> Self {
+        Self { stream }
+    }
```
ssweetney-spok: Do users need this?

crates/aranya-client/src/aqc/crypto.rs
ssweetney-spok: Could note your intentions in the comments. The method/function names should align with the most accurate behavior IMO

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: I tried removing the server psks from the key store when they are loaded into `rustls`. However, it appears that the same PSK identities are loaded more than once:
```
2025-08-05T21:12:06.985013Z  INFO aranya_client::aqc::crypto: loading server psk: [97, 114, 97, 110, 121, 97, 45, 99, 116, 114, 108, 45, 112, 115, 107, 33]
2025-08-05T21:12:06.990062Z  INFO aranya_client::aqc::crypto: loading server psk: [19, 4, 14, 114, 173, 255, 24, 170, 206, 86, 20, 237, 249, 180, 218, 64, 58, 184, 236, 122, 84, 184, 202, 242, 183, 80, 154, 150, 140, 21, 136, 53, 12, 175]
2025-08-05T21:12:06.990064Z  INFO aranya_client::aqc::crypto: loading server psk: [19, 3, 14, 114, 173, 255, 24, 170, 206, 86, 20, 237, 249, 180, 218, 64, 58, 184, 236, 122, 84, 184, 202, 242, 183, 80, 154, 150, 140, 21, 136, 53, 12, 175]
2025-08-05T21:12:06.990066Z  INFO aranya_client::aqc::crypto: loading server psk: [19, 2, 14, 114, 173, 255, 24, 170, 206, 86, 20, 237, 249, 180, 218, 64, 58, 184, 236, 122, 84, 184, 202, 242, 183, 80, 154, 150, 140, 21, 136, 53, 12, 175]
2025-08-05T21:12:07.040568Z ERROR aranya_client::aqc::crypto:  unable to load server psk [97, 114, 97, 110, 121, 97, 45, 99, 116, 114, 108, 45, 112, 115, 107, 33]
2025-08-05T21:12:08.043421Z ERROR aranya_client::aqc::crypto:  unable to load server psk [97, 114, 97, 110, 121, 97, 45, 99, 116, 114, 108, 45, 112, 115, 107, 33]
...
```

gknopf-aranya: Strange that some of the PSK identities only have a single byte difference.

crates/aranya-client/src/aqc/api.rs
gknopf-aranya: How would a channel be restored if the PSKs have been deleted?

gknopf-aranya: This change was made so there is a single way to free the channel objects in the C API.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: Renaming this method

crates/aranya-client/src/aqc/net/channels.rs
gknopf-aranya: This can be reverted. It happened because I was going to add more fields to the object at some point.

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: Looks like they can be dropped after the first read instead of keeping them around.

crates/aranya-client/src/aqc/net/channels.rs
gknopf-aranya: What is the benefit of this approach? Passing a single ID around instead of a vector of PSK identities?

gknopf-aranya: One of the benefits is that it gives the key manager more control over which channel keys are deleted.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -385,28 +412,40 @@ impl AqcClient {
             .map_err(IpcError::new)?
             .map_err(aranya_error)?;
 
-        self.server_keys.load_psks(psks.clone());
-
         let mut channels = self.channels.write().expect("poisoned");
-        match psks {
-            AqcPsks::Bidi(psks) => {
-                for (_suite, psk) in psks {
+        match psks.clone() {
```
gknopf-aranya: Would prefer not to clone `psks` so many times in this function. But the match takes ownership of `psks` and they need to be inserted into the server key store based on the channel ID.

crates/aranya-client/src/aqc/api.rs
ssweetney-spok: Yea, I saw your other comment about why the change was made. I was asking because now it makes the Rust API slightly worse since these methods should really take the channel objects by value. I wonder if it'd be better to just remove the cleanup methods in the C API and just rely on the delete functions

crates/aranya-client/src/aqc/crypto.rs
djohnson-spideroak: That seems worth investigating right?

jdygert-spok: That is because they are just `(ChannelId, SuiteId)`.

jdygert-spok: (Or rather `(SuiteId, ChannelId)` because they got reordered by the compiler. I opened https://github.com/aranya-project/aranya-core/pull/353 to make the order consistent).

jdygert-spok: You could remove the set of IDs after `fn chosen` is called.

gknopf-aranya: `chosen`? I'm not seeing this in the codebase.

jdygert-spok: It is a method on `trait SelectsPresharedKeys`

gknopf-aranya: Would we remove all the PSKs for that channel here or only the selected one?
I tried both approaches and it causes the AQC tests to no longer pass.
First one causes some tests to hang, second one resulted in a:
```
thread 'test_aqc_chans' panicked at crates/aranya-client/tests/aqc.rs:170:10:
can create uni channel: Aqc(ConnectionError(MaxHandshakeDurationExceeded { max_handshake_duration: 10s, source: Location { file: "/Users/gknopf/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/s2n-quic-transport-0.58.0/src/connection/connection_impl.rs", line: 1078, col: 24 } }))
```

gknopf-aranya: removing chosen psk:
```
    fn chosen(&self, identity: &[u8]) {
        self.keys.lock().expect("poisoned").take(identity);
    }
```
removing psks for the channel:
```
    fn chosen(&self, identity: &[u8]) {
        if let Some(channel_id) =
            self.identities
                .lock()
                .expect("poisoned")
                .iter()
                .find_map(|(k, v)| {
                    if v.iter().any(|i| *i == identity) {
                        Some(k)
                    } else {
                        None
                    }
                })
        {
            info!("removing chosen channel PSKs");
            self.remove(channel_id);
        }
    }
```

gknopf-aranya: The hang looks like it was due to deadlock, one sec

jdygert-spok: I'd guess the issue is that you remove the ctrl psk

gknopf-aranya: Yep, that appears to be the reason

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -28,75 +30,114 @@ pub(super) static CTRL_PSK: LazyLock<Arc<PresharedKey>> = LazyLock::new(|| {
...
+
+    fn chosen(&self, identity: &[u8]) {
+        // Don't remove ctrl PSK.
+        if identity == CTRL_PSK.identity() {
+            return;
+        }
+
+        let mut keys = self.keys.lock().expect("poisoned");
+        let mut identities = self.identities.lock().expect("poisoned");
+        let channel_id = {
```
gknopf-aranya: To simplify this code, we'd need an `identity -> channel_id` mapping

gknopf-aranya:
@jdygert-spok server key store channel PSKs are now deleted when a PSK is chosen

djohnson-spideroak (dismissed):

crates/aranya-client/src/aqc/net/channels.rs
jdygert-spok, line 1: I think we can revert all of this.

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -28,75 +30,106 @@ pub(super) static CTRL_PSK: LazyLock<Arc<PresharedKey>> = LazyLock::new(|| {
 
 #[derive(Debug)]
 pub(crate) struct ServerPresharedKeys {
-    keys: Mutex<HashMap<Vec<u8>, Arc<PresharedKey>>>,
+    keys: Mutex<HashSet<PskIdAsKey>>,
+    identities: Mutex<HashMap<AqcChannelId, Vec<PskIdentity>>>,
+    channels: Mutex<HashMap<PskIdentity, AqcChannelId>>,
```
jdygert-spok: Could just store the related psks:

```suggestion
    keys: Mutex<HashMap<PskIdAsKey, Arc<[Arc<PresharedKey>]>>>,
```

then:
```rust
if let Some(psks) = keys.remove(chosen) {
    for psk in psks.iter() {
        keys.remove(psk.identity());
    }
}
```

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -28,75 +30,106 @@ pub(super) static CTRL_PSK: LazyLock<Arc<PresharedKey>> = LazyLock::new(|| {
...
+                });
+            }
+        }
     }
 }
 
 #[derive(Debug)]
 pub struct ClientPresharedKeys {
-    keys: Mutex<Vec<Arc<PresharedKey>>>,
+    keys: Mutex<HashSet<PskIdAsKey>>,
```
jdygert-spok: This can be reverted, then `fn psks` can `mem::take` the vec.

crates/aranya-client/src/aqc/net.rs
jdygert-spok: The match doesn't need to consume, match on `&psks`.

crates/aranya-client/src/aqc/net.rs
```diff
@@ -426,9 +446,12 @@ pub enum TryReceiveError<E = AqcError> {
...
     Error(E),
-    /// The channel or stream is closed.
-    #[error("channel or stream is closed")]
-    Closed,
+    /// The  stream is closed.
+    #[error("stream is closed")]
+    StreamClosed,
+    /// The connection is closed.
+    #[error("connection is closed")]
+    ConnectionClosed,
```
jdygert-spok: This is never created now.

crates/aranya-client/src/aqc/net/channels.rs
gknopf-aranya: Except for the typo fixes

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: If I do this, the inner iterators want to be cloned:
```
help: consider cloning the value if the performance cost is acceptable
    |
400 -                 for (_suite, psk) in *psks {
400 +                 for (_suite, psk) in psks.clone() {
```

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -30,106 +29,98 @@ pub(super) static CTRL_PSK: LazyLock<Arc<PresharedKey>> = LazyLock::new(|| {
...
-        }
-        self.identities
-            .lock()
-            .expect("poisoned")
-            .insert(*channel_id, identities);
+
+        let psks: Vec<Arc<PresharedKey>> = psks
+            .into_iter()
+            .map(|(suite, psk)| make_preshared_key(suite, psk).expect("can make psk"))
+            .collect();
```
jdygert-spok: You can `collect` straight into an `Arc<[T]>`

crates/aranya-client/src/aqc/net.rs
```diff
@@ -191,7 +196,7 @@ impl AqcClient {
             .client_state
             .lock()
             .await
-            .connect_data(addr, AqcPsks::Uni(psks))
+            .connect_data(addr, AqcPsks::Uni(psks.clone()))
```
jdygert-spok: revert clone

crates/aranya-client/src/aqc/net.rs
```diff
@@ -213,28 +218,29 @@ impl AqcClient {
             .client_state
             .lock()
             .await
-            .connect_data(addr, AqcPsks::Bidi(psks))
+            .connect_data(addr, AqcPsks::Bidi(psks.clone()))
```
jdygert-spok: revert clone

crates/aranya-client/src/aqc/net.rs
```diff
@@ -313,14 +319,14 @@ impl AqcClient {
             // If the PSK identity hint is not the control PSK, check if it's in the channel map.
             // If it is, create a channel of the appropriate type. We should have already received
             // the control message for this PSK, if we don't we can't create a channel.
-            let channels = self.channels.read().expect("poisoned");
-            let channel_info = channels.get(&identity).ok_or_else(|| {
+            let Some(channel_info) = self.channels.write().expect("poisoned").remove(&identity)
```
jdygert-spok: I think `channels` can just be a mutex now.

crates/aranya-client/src/aqc/api.rs
```diff
@@ -126,43 +127,50 @@ impl<'a> AqcChannels<'a> {
...
+        &mut self,
+        chan: &mut AqcSendChannel,
+    ) -> crate::Result<()> {
+        chan.close();
+        Ok(())
+    }
+
+    /// Deletes a receive AQC uni channel.
+    /// Zeroizes PSKs associated with the channel.
+    /// Closes all associated QUIC connections and streams.
```
jdygert-spok: Blank line after first line on each of these 3 docs.

crates/aranya-client/src/aqc/crypto.rs
```diff
@@ -38,31 +39,54 @@ impl ServerPresharedKeys {
...
-            Entry::Vacant(v) => {
-                v.insert(psk);
-            }
-            Entry::Occupied(_) => {
-                error!("Duplicate PSK identity inserted: {:?}", identity);
-            }
+        let mut keys = self.keys.lock().expect("poisoned");
+        if keys.insert(PskIdAsKey(psk), [].into()).is_some() {
+            error!("Duplicate PSK identity inserted: {:?}", identity);
         }
```
ssweetney-spok: Don't need to create a copy of the identity if it's not gonna be used every time

```Rust
        let mut keys = self.keys.lock().expect("poisoned");
        match keys.entry(PskIdAsKey(psk)) {
            Entry::Vacant(v) => {
                v.insert(Arc::new([]));
            }
            Entry::Occupied(v) => {
                let identity = v.key().0.identity();
                error!("Duplicate PSK identity inserted: {:?}", identity);
            }
        }
```

crates/aranya-client/src/aqc/net.rs
```diff
@@ -102,7 +107,7 @@ struct ServerState {
 }
 
 /// Identity of a preshared key.
-type PskIdentity = Vec<u8>;
+pub type PskIdentity = Vec<u8>;
```
ssweetney-spok: can be reverted. Don't think we need this to be pub

crates/aranya-client/src/aqc/net.rs
```diff
@@ -438,8 +449,8 @@ struct AqcChannelInfo {
 }
 
 /// An AQC Channel ID.
-#[derive(Copy, Clone, Debug)]
-enum AqcChannelId {
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+pub enum AqcChannelId {
```
ssweetney-spok: Why pub here?

crates/aranya-client/src/aqc/crypto.rs
jdygert-spok: Also use `Arc::default()` so you get the static default.

crates/aranya-client/src/aqc/crypto.rs
gknopf-aranya: Like this? `let psks: Vec<_>`

jdygert-spok:
```rust
let psks: Arc<[_]> = iter.collect();
```

crates/aranya-client/src/aqc/net.rs
gknopf-aranya: `Eq, Hash, PartialEq` can be reverted too. This is from when this was a key in a map.

jdygert-spok (approved):

PR #430, opened 2025-07-30
client: config experiment

jdygert-spok:
This PR basically removes the builders by moving the "build" step up into the methods which take configs, and also lets those configs be used as the "info" structs which can be sent to other users.

### DIFF
--- a/crates/aranya-client/src/client.rs
+++ b/crates/aranya-client/src/client.rs
@@ -20,7 +20,7 @@ use tracing::{debug, error, info, instrument};
 
 use crate::{
     aqc::{AqcChannels, AqcClient},
-    config::{AddTeamConfig, CreateTeamConfig, SyncPeerConfig},
+    config::{private::ConfigFor, SyncPeerConfig},
     error::{self, aranya_error, InvalidArg, IpcError, Result},
 };
 
@@ -246,10 +246,13 @@ impl Client {
     }
 
     /// Create a new graph/team with the current device as the owner.
-    pub async fn create_team(&self, cfg: CreateTeamConfig) -> Result<Team<'_>> {
+    pub async fn create_team(
+        &self,
+        cfg: impl ConfigFor<aranya_daemon_api::CreateTeamConfig>,
+    ) -> Result<Team<'_>> {
         let team_id = self
             .daemon
-            .create_team(context::current(), cfg.into())
+            .create_team(context::current(), cfg.resolve()?)
             .await
             .map_err(IpcError::new)?
             .map_err(aranya_error)?;
@@ -274,8 +277,11 @@ impl Client {
     }
 
     /// Add a team to local device storage.
-    pub async fn add_team(&self, cfg: AddTeamConfig) -> Result<Team<'_>> {
-        let cfg = aranya_daemon_api::AddTeamConfig::from(cfg);
+    pub async fn add_team(
+        &self,
+        cfg: impl ConfigFor<aranya_daemon_api::AddTeamConfig>,
+    ) -> Result<Team<'_>> {
+        let cfg = cfg.resolve()?;
         let team_id = cfg.team_id;
 
         self.daemon
--- a/crates/aranya-client/src/config/team.rs
+++ b/crates/aranya-client/src/config/team.rs
@@ -11,118 +11,102 @@
 //!
 //! Both operations support optional transport configuration.
 
+use aranya_daemon_api as api;
 use aranya_daemon_api::TeamId;
 
-use crate::{error::InvalidArg, ConfigError, Result};
-
 pub mod quic_sync;
-pub use quic_sync::{
-    AddTeamQuicSyncConfig, CreateTeamQuicSyncConfig, CreateTeamQuicSyncConfigBuilder,
+use serde::{Deserialize, Serialize};
+use serde_with::skip_serializing_none;
+
+use crate::{
+    config::{
+        private::{ConfigFor, UseDefaultSpread},
+        ConfigResult,
+    },
+    error::InvalidArg,
 };
 
-/// Builder for [`CreateTeamConfig`].
-#[derive(Debug, Default)]
-pub struct CreateTeamConfigBuilder {
-    quic_sync: Option<CreateTeamQuicSyncConfig>,
+#[skip_serializing_none]
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct CreateTeamConfigV1 {
+    pub quic_sync: Option<quic_sync::CreateTeamQuicSyncConfigV1>,
+    #[doc(hidden)]
+    #[serde(skip)]
+    pub __use_default_spread: UseDefaultSpread,
 }
 
-impl CreateTeamConfigBuilder {
-    /// Configures the quic_sync config..
-    ///
-    /// This is an optional field that configures how the team
-    /// synchronizes data over QUIC connections.
-    pub fn quic_sync(mut self, cfg: CreateTeamQuicSyncConfig) -> Self {
-        self.quic_sync = Some(cfg);
-        self
-    }
-
-    /// Builds the configuration for creating a new team.
-    pub fn build(self) -> Result<CreateTeamConfig> {
-        Ok(CreateTeamConfig {
-            quic_sync: self.quic_sync,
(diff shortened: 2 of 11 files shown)


crates/aranya-client/src/client.rs
```diff
@@ -246,10 +246,13 @@ impl Client {
     }
 
     /// Create a new graph/team with the current device as the owner.
-    pub async fn create_team(&self, cfg: CreateTeamConfig) -> Result<Team<'_>> {
+    pub async fn create_team(
+        &self,
+        cfg: impl ConfigFor<aranya_daemon_api::CreateTeamConfig>,
+    ) -> Result<Team<'_>> {
```
jdygert-spok, line 252: Taking `impl SomePrivateTrait` means we can easily expand the types which could be passed to this method. (e.g. V2, V3, later).

crates/aranya-client/src/config/team.rs
```diff
@@ -11,118 +11,102 @@
...
-#[derive(Debug, Default)]
-pub struct CreateTeamConfigBuilder {
-    quic_sync: Option<CreateTeamQuicSyncConfig>,
+#[skip_serializing_none]
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct CreateTeamConfigV1 {
+    pub quic_sync: Option<quic_sync::CreateTeamQuicSyncConfigV1>,
+    #[doc(hidden)]
+    #[serde(skip)]
+    pub __use_default_spread: UseDefaultSpread,
```
jdygert-spok, line 35: All fields are optional. This is _basically_ equivalent to the builder, except we've moved the build step "inside" the methods that take `impl ConfigFor`.

All fields are public, but there is a hidden field. If we assume that users won't directly reference this field, the only two ways to construct this are:
```rust
let mut cfg = CreateTeamConfigV1::default();
cfg.quic_sync = ...;
// or
let cfg = CreateTeamConfigV1 {
    quic_sync = ...,
    ..Default::default()
};
```

jdygert-spok: Since the serde impl skips `None` fields, if we use cbor, if we add a new optional field, then you can deserialize new as old or old as new. With this we don't need to make V2 if we are only adding an optional field.

crates/aranya-client/src/config/team.rs
```diff
@@ -11,118 +11,102 @@
...
-/// Builder for [`AddTeamConfig`].
-#[derive(Debug, Default)]
-pub struct AddTeamConfigBuilder {
-    id: Option<TeamId>,
-    quic_sync: Option<AddTeamQuicSyncConfig>,
+#[skip_serializing_none]
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct AddTeamConfigV1 {
+    pub team_id: Option<TeamId>,
+    pub quic_sync: Option<quic_sync::AddTeamQuicSyncConfigV1>,
```
jdygert-spok, line 50: I could take or leave the pub fields with default. This is pretty much equivalent to an opaque config type with getters and setters. The pub fields are less boilerplate. But setters would let you change the type of the argument, e.g. taking `T` instead of the `Some(T)` field, or even `impl Into<T>` etc.

It's also a bit hacky, you _could_ directly use `__use_default_spread: Default::default()` and then you'd stop compiling if we added a field. Although this is kinda nice for the C API since then we know we aren't forgetting a field.

crates/aranya-client/tests/common/mod.rs
```diff
@@ -132,14 +136,13 @@ impl DevicesCtx {
...
-                .team_id(team_id)
-                .quic_sync(qs_cfg)
-                .build()?
+        let cfg = AddTeamConfigV1 {
+            team_id: Some(team_id),
+            quic_sync: Some(AddTeamQuicSyncConfigV1 {
+                seed_mode: Some(seed_ikm.into()),
+                ..Default::default()
+            }),
+            ..Default::default()
```
jdygert-spok, line 145: compare to config with setters:

```rust
let cfg = AddTeamConfigV1::new()
    .with_team_id(team_id)
    .with_quic_sync(AddTeamQuicSyncConfigV1::new().with_seed_ikm(seed_ikm));
```

or maybe:

```rust
let cfg = AddTeamConfigV1::new(team_id)
    .with_quic_sync(AddTeamQuicSyncConfigV1::new().with_seed_ikm(seed_ikm));
```

gknopf-aranya: I can't imagine this struct not having a `team_id` field in the future, but it's safer to require a setter method for each field just in case.

gknopf-aranya: I'd use default instead of new to indicate that all the fields are populated with a default value initially.
```
let cfg = AddTeamConfigV1::new()
    .with_team_id(team_id)
    .with_quic_sync(AddTeamQuicSyncConfigV1::new().with_seed_ikm(seed_ikm));
```
->
```
let cfg = AddTeamConfigV1::default()
    .with_team_id(team_id)
    .with_quic_sync(AddTeamQuicSyncConfigV1::new().with_seed_ikm(seed_ikm));
```

jdygert-spok: > I can't imagine this struct not having a `team_id` field in the future, but it's safer to require a setter method for each field just in case.

If we removed the team_id field we'd have to make it a new version.

gknopf-aranya: Wouldn't that be the case if any other fields were removed? What's the rationale for treating the `team_id` field differently than the rest of the fields?

jdygert-spok: > I'd use default instead of new to indicate that all the fields are populated with a default value initially.

I think `new` is fine, It is generally accepted that a no-arg `new` and `default` are the same. If anything I might make it `empty()` to be explicit that nothing is set. But `new` is much more common/expected.

jdygert-spok: > What's the rationale for treating the `team_id` field differently than the rest of the fields?

Same as the reason we had it as a separate argument to `add_team` for a while. It's _the team_ you are adding, not some additional configuration.

crates/aranya-client/src/client.rs
ssweetney-spok: Not sure how we'd replicate this for the C API. Does that mean we'd have to have a different `create_team` function for each version?

crates/aranya-client/src/config/team.rs
ssweetney-spok: > if we use cbor

Can you explain the importance of cbor here?

crates/aranya-client/src/config/team/quic_sync.rs
```diff
@@ -6,133 +6,128 @@
...
+};
 
 /// Configuration for creating a new team with QUIC synchronization.
-#[derive(Clone, Debug)]
-pub struct CreateTeamQuicSyncConfig {
-    mode: CreateSeedMode,
+#[skip_serializing_none]
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct CreateTeamQuicSyncConfigV1 {
+    pub seed_mode: Option<CreateSeedModeV1>,
```
ssweetney-spok, line 25: Couldn't the `Option` be removed?

ssweetney-spok: Oh, but then I guess we'd need a V2 if we removed it

crates/aranya-client/src/config/team.rs
ssweetney-spok, line 1: One advantage over Obake is that this approach works with generics so I could bring back this type I wanted

```Rust
#[derive(serde::Serialize, serde::Deserialize)]
struct Frozen<T> {
    data: T,
    frozen: bool,
}

impl<T> Frozen<T> {
    fn freeze(&mut self) {
        self.frozen = true;
    }

    fn get_mut(&mut self) -> &mut T {
        if self.frozen {
            panic!("unable to mutate")
        }

        &mut self.data
    }
}

impl<T> std::ops::Deref for Frozen<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.data
    }
}
```

jdygert-spok:
We decided no change to the builders is needed and that we will not address the "info" passing ATM.

PR #431, opened 2025-07-31
WIP: aqc: query daemon for valid channels

gknopf-aranya:
Implement daemon query to determine which active AQC channels are still valid according to the policy.

TODO:
- [ ] Test with more channels

Out of scope for this PR:
- Have the Aranya client periodically run this query so it can delete invalid AQC channels
- Delete PSKs when an AQC channel is deleted


gknopf-aranya:
Closing in favor of:
https://github.com/aranya-project/aranya/issues/432

PR #435, opened 2025-08-01
policy: query whether aqc channel is valid

gknopf-aranya:
Query the policy to determine whether an AQC channel is valid.


gknopf-aranya:
Dygert mentioned today that we would likely want to delete PSKs/certs from the shared memory after processing effects for delete_label, revoke_label, terminate_team, remove_member, etc. This would remove the need for querying channel validity from the graph.

PR #436, opened 2025-08-01
Use stashed PSK identity for aqc and quic syncer

### LINKED ISSUE #320: QUIC: Investigate how to get the PSK identity from the connection

jdygert-spok:
Closes #320.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ futures-util = { version = "0.3.30" }
 heapless = { version = "0.8", default-features = false }
 postcard = { version = "1", default-features = false, features = ["use-std", "heapless", "experimental-derive"] }
 pretty_assertions = { version = "1.4" }
-s2n-quic = { version = "1.58.0-psk.1", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"], package = "aranya-internal-s2n-quic" }
+s2n-quic = { version = "1.58.0-psk.2", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"], package = "aranya-internal-s2n-quic" }
 semver = { version = "1", features = ["serde"] }
 serde = "1"
 serial_test = { version = "3" }
@@ -98,3 +98,7 @@ allow-branch = ["main"]
 shared-version = true
 tag-message = ""
 tag-prefix = ""
+
+
+[patch.crates-io]
+aranya-internal-s2n-quic = { git = "https://github.com/aranya-project/aranya-internal-s2n-quic.git", branch = "stash-chosen-identity" }
--- a/crates/aranya-client/src/aqc/net.rs
+++ b/crates/aranya-client/src/aqc/net.rs
@@ -18,7 +18,7 @@ use aranya_util::{
     rustls::{NoCertResolver, SkipServerVerification},
     s2n_quic::read_to_end,
 };
-use buggy::BugExt as _;
+use buggy::{Bug, BugExt as _};
 use bytes::Bytes;
 use channels::AqcPeerChannel;
 use s2n_quic::{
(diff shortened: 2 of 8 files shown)


crates/aranya-client/src/aqc/net.rs
```diff
@@ -234,12 +230,7 @@ impl AqcClient {
                 .await
                 .ok_or(AqcError::ServerConnectionTerminated)?;
             // Receive a PSK identity hint.
-            // TODO: Instead of receiving the PSK identity hint here, we should
-            // pull it directly from the connection.
-            let identity = server_state
-                .identity_rx
-                .try_recv()
-                .assume("identity received after accepting connection")?;
+            let identity = get_conn_identity(&mut conn)?;
```
gknopf-aranya, line 233: Would be neat if the method could be invoked on the object:
```suggestion
            let identity = conn.get_identity()?;
```
I suppose that would require making our own fork of `s2n-quic` or wrapping it in our own type which seems like too much effort.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -491,9 +484,15 @@ where
         let handle = conn.handle();
         async {
             debug!("received incoming QUIC connection");
+            let identity = *conn
```
gknopf-aranya: Since the code for getting the identity from the connection is shared by AQC and the QUIC syncer, it could be moved to `crates/aranya-util/src/s2n_quic.rs`

crates/aranya-client/src/aqc/net.rs
```diff
@@ -466,3 +452,13 @@ struct AqcCtrlMessage {
...
+
+// Extract the chosen PSK identity from the connection context.
+fn get_conn_identity(conn: &mut Connection) -> Result<Vec<u8>, Bug> {
+    Ok(*conn
+        .take_tls_context()
+        .assume("connection has tls context")?
+        .downcast::<Vec<u8>>()
+        .ok()
+        .assume("can downcast identity")?)
+}
```
djohnson-spideroak: should we be using panic here with assume or just dropping/giving an error?

gknopf-aranya: Would rather return an error instead so this doesn't crash the program. It will prevent the connection from being established.

jdygert-spok: This _should_ never fail. `assume` panics if `debug_assertions` is enabled or returns a `Bug` otherwise.

gknopf-aranya: What if someone tries to make a QUIC connection to our AQC or QUIC syncer servers with a client that doesn't implement our fork of `rustls`? Could that cause a panic?

jdygert-spok: No, this depends only on the local version. And again it will only panic in dev profile.

Cargo.toml
```diff
@@ -98,3 +98,7 @@ allow-branch = ["main"]
 shared-version = true
 tag-message = ""
 tag-prefix = ""
+
+
+[patch.crates-io]
+aranya-internal-s2n-quic = { git = "https://github.com/aranya-project/aranya-internal-s2n-quic.git", branch = "stash-chosen-identity" }
```
gknopf-aranya: Remove patch before merging

crates/aranya-client/src/aqc/net.rs
ssweetney-spok: It seems like this function can fail if it's called multiple times with the same connection. I don't think we'd ever need to do that though

ssweetney-spok (commented):
Looks good. I'm unable to run the rust example but the sync tests are passing

PR #441, opened 2025-08-05
Fix tempfile usage in client tests

### LINKED ISSUE #217: incorrect use of `tempfile`
For example https://github.com/aranya-project/aranya/blob/5d20c9ff4696b806cb6a11a06365216b7da8ba21/crates/aranya-client/tests/tests.rs#L389

The directory is deleted when the `TempDir` is dropped, which it is in/after this expression. I only think this is working b/c the daemon or somebody else recreates the directory.

ssweetney-spok:
Closes #217

Moved tempdir creation into `DevicesCtx::new`

### DIFF
--- a/crates/aranya-client/tests/aqc.rs
+++ b/crates/aranya-client/tests/aqc.rs
@@ -21,7 +21,7 @@ use crate::common::{sleep, DevicesCtx};
 /// 3. Delete AQC channels.
 #[test_log::test(tokio::test(flavor = "multi_thread"))]
 async fn test_aqc_chans() -> Result<()> {
-    let (mut devices, _dir) = DevicesCtx::new("test_aqc_chans").await?;
+    let mut devices = DevicesCtx::new("test_aqc_chans").await?;
 
     // create team.
     let team_id = devices.create_and_add_team().await?;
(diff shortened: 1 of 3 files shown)


crates/aranya-client/tests/common/mod.rs
```diff
@@ -35,22 +36,28 @@ pub struct DevicesCtx {
 }
 
 impl DevicesCtx {
-    pub async fn new(name: &str, work_dir: PathBuf) -> Result<Self> {
+    pub async fn new(name: &str) -> Result<(Self, TempDir)> {
```
jdygert-spok: Could just make it a field in `DevicesCtx`

gknopf-aranya: Yep, that ensures the object holds it so it doesn't get dropped

jdygert-spok (approved):

crates/aranya-client/tests/aqc.rs
```diff
@@ -22,10 +21,7 @@ use crate::common::{sleep, DevicesCtx};
 /// 3. Delete AQC channels.
 #[test_log::test(tokio::test(flavor = "multi_thread"))]
 async fn test_aqc_chans() -> Result<()> {
-    let tmp = tempdir()?;
-    let work_dir = tmp.path().to_path_buf();
-
-    let mut devices = DevicesCtx::new("test_aqc_chans", work_dir).await?;
+    let mut devices = DevicesCtx::new("test_aqc_chans").await?;
```
gknopf-aranya, line 24: The downside of initializing the temp dirs in `DevicesCtx` is that there is a different temp dir per device.
It is easier to navigate the directory structure when all the files for the test/demo are grouped into the same folder.

gknopf-aranya: Oh I see, it's held by `DevicesCtx` not `DeviceCtx`. This looks good.

gknopf-aranya (approved):

Issue #437, opened 2025-08-01
Unify Example Code and Metrics


nikki-aranya:
The `aranya-metrics` crate introduces a second copy of all the code from the Rust example with multiple architectural changes. We need to unify them and backport the changes so we can continue to easily maintain them.

PR #443, opened 2025-08-05
WIP: examples

gknopf-aranya:


Makefile.toml
```diff
@@ -255,7 +255,13 @@ dependencies = ["build-capi-header", "build-capi-lib"]
 [tasks.run-rust-example]
 category = "examples"
 description = "Run Rust example application"
-cwd = "examples/rust"
+cwd = "examples/rust/aranya-example"
+script = "./run.bash"
+
+[tasks.run-rust-example-multi-node]
+category = "examples"
+description = "Run Rust example applicationon multiple nodes"
```
gknopf-aranya:
```suggestion
description = "Run Rust example application on multiple nodes"
```

examples/rust/aranya-example-multi-node/Cargo.toml
```diff
@@ -0,0 +1,69 @@
+[package]
+name = "aranya-example-multi-node"
+description = "Example of how run each Aranya device on a different node"
```
gknopf-aranya:
```suggestion
description = "Example of how to run each Aranya device on a different machine in a network"
```

examples/rust/aranya-example-multi-node/README.md
```diff
@@ -0,0 +1,27 @@
...
+- `example.env` - an environment file for loading shared configuration info such as IP addresses into executables
+- `aranya-daemon` - the Aranya daemon executable
+- `aranya-example-multi-node-<node variant>` - the team member's Aranya client executable
+
+For example, the `owner` device would copy the `aranya-example-multi-node-owner` executable from the `target/release/` folder onto the corresponding machine acting as the owner on the team.
+
+Once the artifacts have been copied onto each machine, source the environment file into the current environment on each machine: `. example.env`
+
+Start the example Aranya client executable on each machine on the network at the same time. E.g. on the owner machine run:
+`aranya-example-multi-node-owner --daemon-path <path to Aranya daemon executable>/aranya-daemon`
```
gknopf-aranya:
```suggestion
Start the Aranya daemon `aranya-daemon` executable and Aranya client executable on each machine on the network at the same time. E.g. on the owner machine run:
`aranya-daemon --config <path to daemon config file>`
`aranya-example-multi-node-owner --uds-sock <path to daemon's unix domain socket API>`
```

examples/rust/aranya-example-multi-node/src/config.rs
gknopf-aranya, line 1: Should we include an example config in this example directory?
Or refer to the example daemon config provided in the daemon crate?
I wouldn't expect a user to know how to make a daemon config file from scratch easily.

Cargo.toml
```diff
@@ -60,9 +61,10 @@ aranya-daemon-api = { version = "1.0.0", path = "crates/aranya-daemon-api" }
 aranya-keygen = { version = "1.0.0", path = "crates/aranya-keygen" }
 aranya-util = { version = "1.0.0", path = "crates/aranya-util" }
 
+age = { version = "0.11.1" }
```
gknopf-aranya: Move this to example's `Cargo.toml` so someone doesn't accidentally use it in non-example code.

examples/rust/aranya-example-multi-node/src/admin/main.rs
```diff
@@ -0,0 +1,105 @@
...
+    info!("admin: sent public keys to owner");
+
+    // Setup sync peers.
+    info!("admin: adding owner sync peer");
+    let sync_interval = Duration::from_millis(100);
+    let sync_cfg = SyncPeerConfig::builder().interval(sync_interval).build()?;
+    team.add_sync_peer(env.owner.sync_addr, sync_cfg.clone())
+        .await?;
+
+    info!("admin: complete");
```
gknopf-aranya, line 109: The admin isn't currently performing any operations on the team other than onboarding itself. Do we want to move some of the operator operations over to the admin?

gknopf-aranya: We could have the admin create the label instead of the operator. Here's a summary of role operations from the policy:
```
* Admin:
  * Assign/revoke Operator role.
  * Define/undefine channel label.
  * Revoke channel label.
  * Unset AQC network identifier.

* Operator:
  * Add (new) / remove Member.
  * Define channel label.
  * Assign/revoke channel label.
  * Set/unset AQC address&name.
```

gknopf-aranya: Updated PR to have admin create the label instead of the operator

deny.toml
```diff
@@ -235,12 +238,13 @@ deny = [
 # this is set there is no point setting `deny`
 #exact = true
 
-# Certain crates/versions that will be skipped when doing duplicate detection.
 skip = [
+    "base64@0.21.7",
```
nikki-aranya, line 244: I don't love pulling in more outdated deps but I don't see another alternative for age encryption, maybe I'll open a PR upstream, or I guess we could just fork it but bleh

examples/rust/README.md
```diff
@@ -1,56 +1,6 @@
...
 
-This crate contains example code showing how to use Aranya in a Rust application.
+Specific examples of how to use the Aranya client API.
 
-This example shows how to use the [`aranya-client`](../../crates/aranya-client) library to:
-- Setup a team
-- Sync Aranya graphs
-- Create an AQC (Aranya QUIC) channel
-- Send encrypted data between peers
+Each example can be built and run as a stand-alone executable.
```
nikki-aranya:
```suggestion
Each example can be built and ran as a stand-alone executable.
```
Also, we really should update the workspace-level README now that we're adding more examples. I'd love something more like "Each example can be built and ran as a stand-alone executable. See `[the top-level README](../../README.md#cargo-make)` on how to set that up." and refactor top-level to have a dedicated section for cargo make since we lean so heavily on it compared to native "cargo run --example multiple-nodes"

nikki-aranya: It would also be nice to write a short blurb about each example's purpose kinda like bevy does: <https://github.com/bevyengine/bevy/tree/main/examples#table-of-contents>

examples/rust/aranya-example-multi-node/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+
+[features]
+default = []
+
+# Enable debugging.
+debug = []
+
+[dependencies]
+aranya-client = { workspace = true }
+aranya-daemon-api = { workspace = true }
```
nikki-aranya, line 24: at what point do we just re-export daemon-api under aranya-client or rename it to aranya-ipc, it's always been named poorly and you shouldn't expect to see "daemon" anywhere in a user facing application

examples/rust/aranya-example-multi-node/src/admin/main.rs
```diff
@@ -0,0 +1,127 @@
...
+    .retry(ExponentialBuilder::default())
+    .await
+    .expect("expected to initialize client");
+    info!("admin: initialized client");
+
+    // Initialize `age` encryptor.
+    let encryptor = AgeEncryptor::new(env.passphrase);
+
+    // Get team ID from owner.
+    let team_id: TeamId = postcard::from_bytes(&encryptor.decrypt(&server.recv().await?)?)?;
```
nikki-aranya: should probably make a helper util/couple functions that wraps the encrypter and does this from_bytes/to_allocvec nonsense in a cleaner call

deny.toml
jdygert-spok: It doesn't seem to take all that much to do ourself, and then we know it is using our crypto instead of whatever age is doing. https://git.spideroak-inc.com/spideroak-inc/secure-kex/blob/init/crates/onboarding/src/diceware/mod.rs @elagergren-spideroak

But also the reason this is getting pulled in now is because this changed the examples to be part of the workspace.

examples/rust/aranya-example-multi-node/src/admin/main.rs
```diff
@@ -0,0 +1,127 @@
...
+            .team_id(team_id)
+            .build()?
+    };
+    let team = client.add_team(add_team_cfg.clone()).await?;
+    info!("admin: added team");
+
+    // Send device ID to owner.
+    info!("admin: sending device ID to owner");
+    let device_id = client.get_device_id().await?;
+    TcpClient::connect(env.owner.tcp_addr)
```
nikki-aranya: is there a reason we don't keep the connection alive, we're hammering the tcp server with a new handshake every single command since this is tcp

examples/rust/aranya-example-multi-node/Cargo.toml
jdygert-spok: Yes, we've said that the necessary types from aranya-daemon-api need to be re-exported and/or wrapped in aranya-client.

examples/rust/aranya-example-multi-node/src/admin/main.rs
```diff
@@ -0,0 +1,127 @@
...
+        .await?
+        .send(&encryptor.encrypt(&postcard::to_allocvec(&client.get_key_bundle().await?)?)?)
+        .await?;
+    info!("admin: sent public keys to owner");
+
+    // Setup sync peers.
+    info!("admin: adding owner sync peer");
+    let sync_interval = Duration::from_millis(100);
+    let sleep_interval = sync_interval * 6;
```
nikki-aranya: constants should be up top so they don't get lost in the sauce

examples/rust/aranya-example-multi-node/Cargo.toml
gknopf-aranya: We should keep types that are only needed for using the daemon API in the daemon API crate.
Types also used in the Rust API should be defined somewhere in the client crate or maybe a `-common-types` crate. But that's not an issue to solve on this PR.

examples/rust/aranya-example-multi-node/src/admin/main.rs
```diff
@@ -0,0 +1,127 @@
...
+    // Loop until this device has the `Admin` role assigned to it.
+    info!("admin: waiting for owner to assign admin role");
+    let queries = team.queries();
+    'outer: loop {
+        if let Ok(devices) = queries.devices_on_team().await {
+            for device in devices.iter() {
+                if let Ok(Role::Admin) = queries.device_role(*device).await {
+                    break 'outer;
+                }
+            }
```
nikki-aranya: you should be able to make an iterator that does a for_each and flattens this so we don't need an 'outer

examples/rust/aranya-example-multi-node/src/config.rs
```diff
@@ -0,0 +1,50 @@
...
+
+                aqc.enable = true
+
+                [sync.quic]
+                enable = true
+                addr = {sync_addr:?}
+                "#
+    );
+    fs::write(&cfg, buf).await?;
+    info!("generated config file: {:?}", cfg);
```
nikki-aranya, line 55: we use this pattern so much, should probably shove it in aranya-util or something, I can also shove it in demo-utils when I tackle #437 if that's easier/more fitting

gknopf-aranya: As a function, it would take too many parameters. Unless we do something like `fn generate_config_file(cfg: Config) -> PathBuf` that takes a config object.

examples/rust/aranya-example-multi-node/src/admin/main.rs
gknopf-aranya: There are a lot of things that could be improved with this TCP implementation. It's the simplest implementation that I could get working quickly. But it's just for example purposes so I don't want to spend too much time optimizing it.
Another thing that could be improved is sending more structured types across so it's not possible to race on sending or receiving data if transmission happens in an unexpected order.

examples/rust/aranya-example-multi-node/src/env.rs
```diff
@@ -0,0 +1,187 @@
...
+                sync_addr: env_var(&format!("ARANYA_SYNC_ADDR_{}", device.0.to_uppercase()))?,
+                role: device.1,
+            };
+            devices.push(device);
+        }
+        let owner = devices[0].clone();
+        let admin = devices[1].clone();
+        let operator = devices[2].clone();
+        let membera = devices[3].clone();
+        let memberb = devices[4].clone();
```
nikki-aranya: I feel like there should be a better way to do this than cloning every device, but I know interior mutability is annoying in this instance. `HashMap<&'static str, Device>` instead of devices, and then do `env.devices["membera"].tcp_addr`? it's not as clean, but meh

gknopf-aranya: Could put devices behind `Arc<>` and only clone if data is mutated. This is all configuration info so it shouldn't be mutated anyways.

examples/rust/aranya-example-multi-node/src/env.rs
```diff
@@ -0,0 +1,187 @@
...
+        let passphrase = SecretString::from(env_var::<String>("ARANYA_ONBOARDING_PASSPHRASE")?);
```
nikki-aranya: should make this a const up top. I'd love the whole thing to be const, but it doesn't look like SecretString has any const stuff yet

deny.toml
gknopf-aranya: I decided to add examples to the workspace so we can minimize the amount of different library versions we have across the codebase and vet everything with `cargo vet`. This does mean there will be more code to audit and more duplicate versions detected. We already had this issue before, it just wasn't being checked in CI.

examples/rust/aranya-example-multi-node/Cargo.toml
gknopf-aranya: It was just supposed to be IPC methods and types originally. Looks like it's drifted away from the original intent.

examples/rust/aranya-example-multi-node/src/admin/main.rs
gknopf-aranya: I'll move it to a module so it can be imported by all the devices.

examples/rust/aranya-example-multi-node/Cargo.toml
gknopf-aranya: Opened an issue to cleanup the IPC types:
https://github.com/aranya-project/aranya/issues/460
That is not something to handle on this PR.

examples/rust/aranya-example-multi-node/src/admin/main.rs
gknopf-aranya: I'm already making a new type elsewhere so might as well define one to group the fields together here and derive serialize/deserialize on it.

examples/rust/aranya-example-multi-node/src/admin/main.rs
gknopf-aranya: Created a module to simplify onboarding

examples/rust/aranya-example-multi-node/src/admin/main.rs
gknopf-aranya: Bundled onboarding data into new types such as `DeviceInfo` and `TeamInfo`.

examples/rust/aranya-example-multi-node/src/env.rs
gknopf-aranya: Made const definitions for device and env var types to define defaults ahead of time.

examples/rust/aranya-example-multi-node/src/env.rs
gknopf-aranya: Ended up adding a `fn devices() -> Vec<&Device>` method to borrow the devices as a vector.

examples/rust/aranya-example-multi-node/src/config.rs
gknopf-aranya: Closing since I'd prefer to handle this in a different issue.

examples/rust/aranya-example-multi-node/src/admin/main.rs
gknopf-aranya: This is tricky because it iterates on a list of async futures.

gknopf-aranya: Revised logic to no longer require outer loops

examples/rust/README.md
gknopf-aranya: Improved docs for `cargo make` commands in the repo's top-level README.md. These docs point to the example directories for additional information.

nikki-aranya (dismissed):

djohnson-spideroak (dismissed):

nikki-aranya (approved):

djohnson-spideroak:
Taking another look at the code, but I am getting this error when running the new example:

```
2025-09-09T15:27:36.454372Z ERROR aranya_runtime::vm_policy: Could not seal command: check
2025-09-09T15:27:36.454403Z ERROR aranya_daemon::api: error=unable to query labels

Caused by:
    engine error: panic
```

gknopf-aranya:
> Taking another look at the code, but I am getting this error when running the new example:

We've opened an issue to investigate ways to handle the engine panic in seal when running a query before device keys have been added to the team:
https://github.com/aranya-project/aranya/issues/478

It is out of scope for this PR so should not prevent us from merging this PR.

gknopf-aranya:
Planning to switch this example from AQC to AFC in a follow-up.

examples/rust/aranya-example-multi-node/src/membera/main.rs
```diff
@@ -0,0 +1,198 @@
...
+    // Create a bidi AQC channel.
+    info!("membera: creating bidi channel");
+    let mut chan = client
+        .aqc()
+        .context("AQC is enabled")?
+        .create_bidi_channel(
+            team_info.team_id,
+            NetIdentifier::from_str(env.memberb.aqc_addr.to_string().as_str())
+                .expect("expected net identifier"),
+            LabelId { __id: label1.id },
```
gknopf-aranya, line 168: @jdygert-spok it appears that we've inadvertently exposed some of the daemon API via `Labels`:
```
/// List of labels.
#[derive(Debug)]
pub struct Labels {
    data: Vec<Label>,
}

impl Labels {
    /// Return iterator for list of labels.
    pub fn iter(&self) -> impl Iterator<Item = &Label> {
        self.data.iter()
    }

    #[doc(hidden)]
    pub fn __data(&self) -> &[Label] {
        self.data.as_slice()
    }
}
```

gknopf-aranya: Including this on an existing issue to clean up client types:
https://github.com/aranya-project/aranya/issues/460

README.md
gknopf-aranya, line 1: @djohnson-spideroak there are recent additions to these `cargo make` targets based on feature flags, should we make those updates here?

gknopf-aranya: Went ahead and made some updates for that

examples/rust/aranya-example-multi-node/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+workspace = true
+
+[features]
+default = []
+
+# Enable debugging.
+debug = []
+
+[dependencies]
+aranya-client = { workspace = true, features = ["preview", "afc", "aqc", "experimental"] }
```
jdygert-spok: This will cause the same issue that #526 is fixing.

supply-chain/audits.toml
jdygert-spok, line 1: Did you actually review these? if not they should be exemptions.

gknopf-aranya: I reviewed the diffs. Not every line as they were huge diffs.

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+CARGO_WORKSPACE_DIR = { value = "", relative = true }
```
jdygert-spok: I assume this is a work-around. I'd prefer to find another way if possible.

examples/rust/aranya-example-multi-node/src/env.rs
```diff
@@ -0,0 +1,269 @@
...
+                device.tcp_addr
+            );
+        }
+        fs::write(path, buf).await?;
+        Ok(())
+    }
+
+    /// Set environment variables.
+    pub fn set(&self) {
+        env::set_var("ARANYA_EXAMPLE", self.level.clone());
```
jdygert-spok, line 117: Oh, I just realized `aranya` is still using rust edition 2021. In edition 2024 this is marked `unsafe` because setting environment variables is not thread safe. We'll want to be careful with this.

examples/rust/aranya-example-multi-node/Cargo.toml
gknopf-aranya: Pushed fixes for this

examples/rust/aranya-example-multi-node/run.bash
```diff
@@ -0,0 +1,47 @@
...
+        --features aqc \
+        --locked
+done
+
+echo "Building aranya-daemon..."
+cargo build \
+    --release \
+    --manifest-path Cargo.toml \
+    --bin aranya-daemon \
+    --features experimental,aqc,preview,afc
```
gknopf-aranya: This example doesn't need afc, so I tried removing that feature from the daemon build:
```suggestion
    --features experimental,aqc
```
But that resulted in a compilation error:
```
error[E0046]: not all trait items implemented, missing: `afc_shm_info`, `create_afc_uni_send_channel`, `delete_afc_channel`, `receive_afc_ctrl`
   --> crates/aranya-daemon/src/api.rs:354:1
    |
354 | impl DaemonApi for Api {
    | ^^^^^^^^^^^^^^^^^^^^^^ missing `afc_shm_info`, `create_afc_uni_send_channel`, `delete_afc_channel`, `receive_afc_ctrl` in implementation
```

examples/rust/aranya-example/Cargo.toml
```diff
@@ -12,13 +12,16 @@ rust-version.workspace = true
 workspace = true
 
 [features]
-default = []
+default = ["afc", "aqc"]
+
+afc = ["aranya-client/afc", "aranya-client/preview"]
+aqc = ["aranya-client/aqc", "aranya-client/experimental"]
```
jdygert-spok: Enabling them by default is no different, you'll still have the same issue.

examples/rust/aranya-example-multi-node/src/env.rs
gknopf-aranya: This code is only for an example so I'm not very concerned about it. `env.set()` is called from `bin/main.rs` in a single thread.

examples/rust/aranya-example/Cargo.toml
gknopf-aranya: What is the issue? I copied the feature changes from the PR. It wasn't clear to me what problem was being solved because there was no description in the PR or GitHub issue linked to the PR.

gknopf-aranya: Oh I see, `cargo check` doesn't pass

jdygert-spok: Yes, it's the feature mismatch between daemon and daemon-api since this sets the feature for the client for the whole workspace.

.cargo/config.toml
gknopf-aranya: Changed this so `run.bash` passes in the paths.

README.md
```diff
@@ -37,6 +37,62 @@ on this component.
...
+
+Rather than requiring feature flags to be manually specified with `cargo build --features ...`, `cargo make` commands are provided in [Makefile.toml](Makefile.toml) for each feature set.
+
+## Cargo Make
+
+We rely heavely on `cargo make` targets to build software, run integration tests, perform unit tests, and run CICD checks. Here's how to install `cargo make`:
+[cargo-make](https://github.com/sagiegurari/cargo-make?tab=readme-ov-file#installation)
+
+Building Aranya:
+- `cargo make build` - builds release versions of crates at [crates](crates/) and [examples/rust](examples/rust). This includes the Aranya client Rust library, Aranya daemon executable, and Rust example executables.
```
nikki-aranya: this is incorrect, it only builds the daemon release binary

README.md
```diff
@@ -37,6 +37,62 @@ on this component.
...
+- `cargo make fmt`
+
+We allow certain targets to be run for specific sets of feature flags by appending `*-preview` or `*-experimental`:
+- `cargo make build-preview`
+- `cargo make build-experimental`
+- `cargo make build-capi-lib-preview`
+- `cargo make build-capi-lib-experimental`
+
+
+A complete list of `cargo make` targets can be found in the [Makefile.toml](Makefile.toml)
```
nikki-aranya: , or by running `cargo make` without any arguments.

deny.toml
```diff
@@ -237,30 +240,35 @@ deny = [
...
+    "toml@0.5.11",
     "untrusted@0.7.1",
     "getrandom@0.1.16",
     "getrandom@0.2.15",
     "hash32@0.2.1",
     "heapless@0.7.17",
     "zerocopy@0.7.35",
     "zerocopy-derive@0.7.35",
-]
+] # Certain crates/versions that will be skipped when doing duplicate detection.
```
nikki-aranya, line 271: we really should look into a better solution for this, as well as updating and replacing all our deps that use old versions :\ need to make an issue

crates/aranya-client/src/client.rs
```diff
@@ -51,7 +51,7 @@ impl Display for DeviceId {
 }
 
 /// The Team ID (a.k.a Graph ID).
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
 pub struct TeamId {
```
jdygert-spok, line 57: Use `#[serde(transparent)]` for both of these

examples/rust/aranya-example/src/main.rs
```diff
@@ -77,7 +77,7 @@ impl ClientCtx {
     pub async fn new(team_name: &str, user_name: &str, daemon_path: &DaemonPath) -> Result<Self> {
         info!(team_name, user_name, "creating `ClientCtx`");
 
-        let work_dir = TempDir::with_prefix(user_name)?;
+        let work_dir = TempDir::new()?;
```
jdygert-spok: Why no prefix?

examples/rust/aranya-example-multi-node/README.md
```diff
@@ -0,0 +1,28 @@
...
+An example that runs each Aranya device as a stand-alone executable.
+
+# How To Run The Example
+
+In this workspace, run:
+`cargo make run-rust-example-multi-node`
+
+# How To Run The Example On Multiple Nodes
+
+Copy the following artifacts onto each node:
```
nikki-aranya, line 12: does the makefile/bashfile not do this? If it does, we should note that this is how you'd do it if you're copying the implementation for your own usage

examples/rust/aranya-example-multi-node/src/age.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// Create a new `age` encryptor with a passphrase.
+    pub fn new(passphrase: SecretString) -> Self {
+        Self { passphrase }
+    }
+
+    /// Encrypt data with a passhrase.
+    pub fn encrypt(&self, plaintext: &[u8]) -> Result<Vec<u8>> {
+        let encryptor = Encryptor::with_user_passphrase(self.passphrase.clone());
+
+        let mut ciphertext = vec![];
```
nikki-aranya, line 27: it'd be nice to do vec::with_capacity here if we can get even a rough estimate, reduce allocations, just do plaintext.len() + 32 (and ciphertext.len() below)?

examples/rust/aranya-example/README.md
```diff
@@ -0,0 +1,56 @@
...
+$ git clone git@github.com:aranya-project/aranya.git
+
+# Change into the directory
+$ cd aranya/examples/rust/aranya-example
+
+# Build the binary
+$ cargo build --release
+
+# Run the binary:
+$ <repo>/target/release/aranya-example
```
jdygert-spok: Needs feature flags and daemon as arg to the example

supply-chain/audits.toml
jdygert-spok: There are many new non-diff audits here. Did you review those?

examples/rust/aranya-example-multi-node/src/bin/main.rs
```diff
@@ -0,0 +1,110 @@
...
+            .join("aranya-example-multi-node")
+            .join("example.env"),
+    )
+    .await?;
+    // Set environment variables before spawning child processes.
+    env.set();
+
+    // Start a daemon for each device.
+    let tmp = tempdir()?;
+    let mut children = vec![];
```
nikki-aranya: same here, we should be trying to tell the vec what size it'll be in advance if at all possible

examples/rust/aranya-example/README.md
jdygert-spok: Should we just point to `./run.bash` instead?

examples/rust/aranya-example-multi-node/src/bin/main.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        let mut child =
+            client(device.name.clone(), &uds_sock, &release).expect("expected to spawn client");
+        if device.name == "membera" {
+            set.spawn(async move {
+                let _ = child.wait().await;
+            });
+        } else {
+            children.push(child);
+        }
```
nikki-aranya: I feel like this would look cleaner as a match since you can match on specific strings

examples/rust/aranya-example-multi-node/src/config.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let cfg = work_dir.join("config.toml");
+
+    let shm = format!("/shm_{}", device);
+    let _ = rustix::shm::unlink(&shm);
+    let runtime_dir = work_dir.join("run");
+    let state_dir = work_dir.join("state");
+    let cache_dir = work_dir.join("cache");
+    let logs_dir = work_dir.join("logs");
+    let config_dir = work_dir.join("config");
```
nikki-aranya, line 25: I've seen this pattern a bunch of times, would be nice to move into utils/onboard, I think that's outside of the scope of this PR but I'd take a TODO

nikki-aranya: (i.e. beyond just this example, other examples use it iirc)

examples/rust/aranya-example-multi-node/src/bin/main.rs
jdygert-spok: It's not really a big deal for the example IMO

examples/rust/aranya-example-multi-node/src/membera/main.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    // Setup sync peers.
+    let sync_cfg = SyncPeerConfig::builder().interval(SYNC_INTERVAL).build()?;
+    info!("membera: adding operator sync peer");
+    team.add_sync_peer(env.operator.sync_addr, sync_cfg.clone())
+        .await
+        .expect("expected to add sync peer");
+
+    // wait for syncing.
+    sleep(SLEEP_INTERVAL).await;
```
nikki-aranya, line 100: this whole chunk could be one function with setup(), everything below should be another function since it's the actual example code (see: how I structured aranya-metrics, we'll need to dedupe all this so I'd also take a TODO here)

examples/rust/aranya-example-multi-node/src/env.rs
jdygert-spok: You're calling it in a task on a tokio multithreaded runtime. It's not just about writing, there can be issues if another thread is reading the environment. It's _probably_ ok. It's not a huge deal since it's in an example but it should be noted so this isn't blindly copied elsewhere.

Alternatively we can pass in the environment to `std::process::Command`.

examples/rust/aranya-example-multi-node/README.md
gknopf-aranya: The script does not do this. Everything is currently run locally on my dev machine or in CI. We don't currently have a lab environment where this is tested on multiple nodes. So users are expected to copy `example.env`, executables, etc. to the various machines and reconfigure IP addresses before running the example on different machines.

README.md
gknopf-aranya: Resolving this

deny.toml
gknopf-aranya: Sometimes the old versions are coming from 3rd party deps that use different versions. This is out of scope for this PR so I'll mark it as resolved.

crates/aranya-client/src/client.rs
gknopf-aranya: Will do

gknopf-aranya: What is the purpose of adding this here?

examples/rust/aranya-example/src/main.rs
gknopf-aranya: I don't remember exactly why this change was made. I vaguely recall the daemon adding the username to the path of the files because of `name = {user_name:?}` being configured in the `config.toml`. So there may have been a path like `tmp/owner/owner/...`

gknopf-aranya: Ran the example both ways. The username wasn't required because each device generates a different tmp path:
```
2025-10-09T21:30:31.445591Z DEBUG spawning daemon cmd=Command { std: cd "/var/folders/rh/gcfwmfzn79795x4t7q15h4cc0000gp/T/ownerailTwl/daemon" && "/Users/gknopf/scm/aranya/target/release/aranya-daemon" "--config" "/var/folders/rh/gcfwmfzn79795x4t7q15h4cc0000gp/T/ownerailTwl/daemon/config.toml", kill_on_drop: true }
```
```
+ /Users/gknopf/scm/aranya/target/release/aranya-example /Users/gknopf/scm/aranya/target/release/aranya-daemon
2025-10-09T21:33:39.752476Z  INFO starting example Aranya application
2025-10-09T21:33:39.752549Z  INFO creating `ClientCtx` team_name="rust_example" user_name="owner"
2025-10-09T21:33:39.753432Z DEBUG spawning daemon cmd=Command { std: cd "/var/folders/rh/gcfwmfzn79795x4t7q15h4cc0000gp/T/.tmpKSlzTO/daemon" && "/Users/gknopf/scm/aranya/target/release/aranya-daemon" "--config" "/var/folders/rh/gcfwmfzn79795x4t7q15h4cc0000gp/T/.tmpKSlzTO/daemon/config.toml", kill_on_drop: true }
```

gknopf-aranya: I'll revert this change. Seems easier to debug with the usernames in the path.

examples/rust/aranya-example-multi-node/src/age.rs
gknopf-aranya: Where is the 32 bytes coming from?

crates/aranya-client/src/client.rs
jdygert-spok: It serializes as if it is the inner type rather than a struct containing the inner type.

examples/rust/aranya-example-multi-node/src/age.rs
gknopf-aranya: afaict, the maximum overhead is 16 bytes:
https://docs.rs/chacha20poly1305/latest/chacha20poly1305/trait.AeadCore.html#associatedtype.CiphertextOverhead

gknopf-aranya: @jdygert-spok, @mooreoak does this sound like the correct overhead to you?
I looked at the source code for `age` and it appears to be using `ChaChaPoly1305` for encryption.

gknopf-aranya: Hmmm, it appears that individual chunks in the stream are encrypted which each have an overhead of 16 bytes, so the overhead of the ciphertext could change depending on how many chunks there are in the plaintext.

gknopf-aranya: Here are the constants I was able to find:
```
        /// Constants copied from age-0.11.1/src/primitives/stream.rs
        const CHUNK_SIZE: usize = 64 * 1024;
        const TAG_SIZE: usize = 16;
        const ENCRYPTED_CHUNK_SIZE: usize = CHUNK_SIZE + TAG_SIZE;

        let encryptor = Encryptor::with_user_passphrase(self.passphrase.clone());

        let mut ciphertext =
            Vec::with_capacity(plaintext.len() * ENCRYPTED_CHUNK_SIZE / CHUNK_SIZE);
```
Current plan is to compute the number of chunks in the plaintext, then scale that by the size of each ciphertext chunk to figure out the total overhead.

examples/rust/aranya-example-multi-node/src/bin/main.rs
gknopf-aranya: or:
```
        if device.name == "membera" {
            set.spawn(async move {
                let _ = child.wait().await;
            });
            continue;
        }
        children.push(child);
```

examples/rust/aranya-example-multi-node/src/config.rs
gknopf-aranya: That's part of the reason I moved the code out of a `main.rs` into `config.rs`.
Agree we could take this a step further and reuse it across examples.

gknopf-aranya: Adding a TODO here

examples/rust/aranya-example-multi-node/src/membera/main.rs
gknopf-aranya: There are a bunch of variables initialized in the setup section that are used in the example.
Separating the setup into another function requires returning all these variables. I'm already returning all these values and there are still more to go:
`Ok((env.clone(), team_info, client, team))`

gknopf-aranya: I'm not sure I see the value in moving the setup code into a function in this case since the code will not be reused anywhere else. Each device has unique setup and demo code.

examples/rust/aranya-example-multi-node/src/bin/main.rs
gknopf-aranya: Went ahead and made this change since it was easy to make.

examples/rust/aranya-example/README.md
gknopf-aranya: Yeah, users should only be using `run.bash` or `cargo make`.
Trying to keep the `cargo build` instructions up-to-date is too much work since it seems to change every release.

gknopf-aranya: I'm tempted to say we should only document `cargo make` since that relies on `run.bash` and only requires the user to install one more dependency. Then we don't need to continue documenting multiple ways to build things.

gknopf-aranya: If someone wants to customize the build more, they can trace `cargo make` to `run.bash` and customize that themselves.

gknopf-aranya: Removing the `cargo build` docs from this README.md.

examples/rust/aranya-example-multi-node/src/env.rs
gknopf-aranya: Adding a TODO for this

supply-chain/audits.toml
gknopf-aranya: Not sure what you mean by non-diff audits. I ran `cargo vet diff` for each dep `cargo vet` complained about and then certified it afterwards.

examples/rust/aranya-example-multi-node/src/age.rs
gknopf-aranya: Resolving since worst case of incorrect guess is an extra reallocation.

examples/rust/aranya-example-multi-node/src/membera/main.rs
gknopf-aranya: Marking this as resolved since it doesn't seem worth delaying merging this PR on refactoring this code.

examples/rust/aranya-example-multi-node/src/age.rs
jdygert-spok: Just leave it starting empty rather than complicating this example onboarding code. We really don't need to minimize allocations here.

supply-chain/audits.toml
gknopf-aranya: Switched over to `exemptions` for newly added `age` deps.

supply-chain/audits.toml
```diff
@@ -210,8 +210,8 @@ delta = "1.0.38 -> 1.0.39"
 
 [[audits.redox_syscall]]
 who = "gknopf <gknopf@spideroak.com>"
-criteria = "safe-to-run"
-delta = "0.5.8 -> 0.5.10"
+criteria = "safe-to-deploy"
+version = "0.5.10"
```
jdygert-spok: exempt

supply-chain/audits.toml
```diff
@@ -273,6 +273,31 @@ who = "gknopf <gknopf@spideroak.com>"
...
+
+[[audits.rustix]]
+who = "gknopf <gknopf@spideroak.com>"
+criteria = "safe-to-deploy"
+delta = "0.37.13 -> 0.37.28"
+
+[[audits.rustix]]
+who = "gknopf <gknopf@spideroak.com>"
+criteria = "safe-to-deploy"
+delta = "0.37.13 -> 0.37.28"
```
jdygert-spok: duplicates

examples/rust/aranya-example-multi-node/src/bin/main.rs
```diff
@@ -0,0 +1,117 @@
...
+            .path()
+            .join(device.name.clone())
+            .join("daemon")
+            .join("run")
+            .join("uds.sock");
+
+        let mut child =
+            client(device.name.clone(), &uds_sock, &release).expect("expected to spawn client");
+        // Wait on the device executable expected to complete last.
+        if device.name == "membera" {
```
djohnson-spideroak: is this dependable?

gknopf-aranya: `membera` is expected to send and receive data via an AQC channel with `memberb`. So, waiting on `membera` to complete ensures data was received by `membera` from `memberb` and the entire example completed successfully.

gknopf-aranya: If this condition is not met, it means:
- One of the example executables returned an error which will cause the example to fail
- The membera application did not complete which would be the result of an error or hanging somewhere

It shouldn't be possible for membera to hang waiting for data from memberb, because in that case memberb should return an error when attempting to send the data to membera.

examples/rust/aranya-example-multi-node/src/memberb/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    info!("memberb: waiting for admin to create label");
+    let queries = team.queries();
+    loop {
+        if let Ok(labels) = queries.labels().await {
+            if labels.iter().next().is_some() {
+                break;
+            }
+        }
+        sleep(SLEEP_INTERVAL).await;
+    }
```
djohnson-spideroak, line 112: Is the member syncing without having been actually added to the team?

gknopf-aranya: `memberb` is added to the team above:
```
    let team = client
        .add_team(add_team_cfg.clone())
        .await
        .expect("expected to add team");
    info!("memberb: added team");
```

This is checking whether certain effects have been synced/processed by querying until a condition is met.

gknopf-aranya: In this case, we want to check that the label exists and all devices have been added to the team, before transmitting data via AQC.

examples/rust/aranya-example-multi-node/src/bin/main.rs
gknopf-aranya: When running locally or in CI I haven't seen any cases of this not working. However, if this example were to be deployed to remote nodes on an unreliable network, it's possible some of the executables could hang while waiting to sync certain effects in the query loops.

examples/rust/aranya-example-multi-node/src/memberb/main.rs
djohnson-spideroak: I misread the portion where memberb is sending device info to the operator as part of it being added to the team.

examples/rust/aranya-example-multi-node/src/env.rs
ssweetney-spok, line 1: Are the "const" wrapper types needed if you use static variables and `LazyLock`?

`LazyLock` does implement `DerefMut` but I don't think we need to worry about that for an example

examples/rust/aranya-example-multi-node/src/bin/main.rs
gknopf-aranya: We could add code to ensure that loops timeout eventually after enough retries. Not sure if this level of complexity is needed for a non-production example though.

examples/rust/aranya-example-multi-node/src/tracing.rs
```diff
@@ -0,0 +1,42 @@
...
+};
+
+/// Example tracing filter.
+struct DemoFilter {
+    env_filter: EnvFilter,
+}
+
+impl<S> Filter<S> for DemoFilter {
+    fn enabled(&self, metadata: &Metadata<'_>, context: &Context<'_, S>) -> bool {
+        if metadata.target().starts_with(module_path!()) {
```
jdygert-spok: I know this is copied from the other example, this doesn't make much sense done in another module here since what this does is force enable all logging that comes from `crate::tracing::*`.

examples/rust/aranya-example-multi-node/src/env.rs
jdygert-spok: It seems like they are just used to implement `Default for EnvVars`? They could just be inlined into that.

(BTW `DerefMut` for `LazyLock` is `&mut LazyLock<T> -> &mut T`, when you have a static one you only get `&LazyLock<T>` so you can't `DerefMut` it.)

ssweetney-spok: > It seems like they are just used to implement Default for EnvVars? They could just be inlined into that.

Oh yea, that'd be even simpler

gknopf-aranya: I figured initializing everything as const could save some memory. Agree that it's not worth optimizing that right now and the simpler inlined defaults will work.

examples/rust/aranya-example-multi-node/src/tracing.rs
gknopf-aranya: What do you propose that we do instead?

examples/rust/aranya-example-multi-node/src/env.rs
ssweetney-spok: I think we lose any potential memory savings since we have to allocate in the conversion from `ConstEnvVars` to `EnvVars`

examples/rust/aranya-example-multi-node/src/bin/main.rs
jdygert-spok: I don't think we're currently handling errors from any of the clients. Since all the clients should terminate, let's just wait on all of them and handle their errors there. Do this for each client:

```rust
        set.spawn(async move {
            let status = child.wait().await.unwrap();
            assert!(status.success(), "{status:?}");
        });
```
(without adding them to the children vec)

examples/rust/aranya-example-multi-node/src/tracing.rs
jdygert-spok: Take it in as an argument so each bin calls `init_logging(module_path!())`.

examples/rust/aranya-example-multi-node/src/env.rs
```diff
@@ -0,0 +1,213 @@
...
+
+/// Default environment variables.
+impl Default for EnvVars {
+    fn default() -> Self {
+        Self {
+            level: "info".into(),
+            passphrase: "passphrase".into(),
+            owner: Device {
+                name: "owner".into(),
+                aqc_addr: Addr::from_str("127.0.0.1:10000").expect("expected addr"),
```
jdygert-spok:
```suggestion
                aqc_addr: Addr::from((Ipv4Addr::LOCALHOST, 10000)),
```

Same for all the others.

examples/rust/aranya-example-multi-node/src/env.rs
```diff
@@ -0,0 +1,213 @@
...
+            operator,
+            membera,
+            memberb,
+        })
+    }
+
+    /// Generate environment file.
+    pub async fn generate(&self, path: &Path) -> Result<()> {
+        let mut buf = "".to_string();
+        buf += &format!("export {}={}\r\n", LOG_LEVEL_ENV_VAR, self.level);
```
jdygert-spok:
```rust
        use std::fmt::Write;
        writeln!(buf, "export {}={}", LOG_LEVEL_ENV_VAR, self.level)?;
```

etc

examples/rust/aranya-example-multi-node/src/bin/main.rs
gknopf-aranya: Pushed a commit that waits for all devices processes to complete, then kills all the daemons.

examples/rust/aranya-example-multi-node/src/tracing.rs
gknopf-aranya: Done

PR #446, opened 2025-08-07
(aranya-client) require 'default' feature to be enabled

### LINKED ISSUE #423: client: require 'default' feature to be enabled
Add tokio feature default enabled to client and make it required. This will let us have features for other runtimes without a breaking change.

ssweetney-spok:
Resolves #423

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -192,7 +192,7 @@ workspace = true
 script = '''
 cargo clippy
 cargo clippy --all-targets
-cargo clippy --no-default-features
+# cargo clippy --no-default-features (TODO: Uncomment when aranya-client no longer requires the 'default' feature to be enabled) 
 cargo clippy --all-targets --all-features
 '''
 install_crate = { rustup_component_name = "clippy" }
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -192,7 +192,7 @@ workspace = true
 script = '''
 cargo clippy
 cargo clippy --all-targets
-# cargo clippy --no-default-features (TODO: https://github.com/aranya-project/aranya/issues/448) 
+# cargo clippy --no-default-features (TODO: uncomment in #448) 
 cargo clippy --all-targets --all-features
 '''
 install_crate = { rustup_component_name = "clippy" }
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -192,7 +192,7 @@ workspace = true
 script = '''
 cargo clippy
 cargo clippy --all-targets
-# cargo clippy --no-default-features (TODO: Uncomment when aranya-client no longer requires the 'default' feature to be enabled) 
+# cargo clippy --no-default-features (TODO: https://github.com/aranya-project/aranya/issues/448) 
 cargo clippy --all-targets --all-features
 '''
 install_crate = { rustup_component_name = "clippy" }
--- a/crates/aranya-client/Cargo.toml
+++ b/crates/aranya-client/Cargo.toml
@@ -48,5 +48,5 @@ name = "tests"
 path = "tests/tests.rs"
 
 [package.metadata.cargo-all-features]
-# TODO: https://github.com/aranya-project/aranya/issues/448
+# TODO: remove in #448
 always_include_features = ["default"]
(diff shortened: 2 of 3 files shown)


crates/aranya-client/Cargo.toml
```diff
@@ -32,7 +33,7 @@ s2n-quic = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 tarpc = { workspace = true }
 thiserror = { workspace = true }
-tokio = { workspace = true, features = ["io-util", "macros", "net", "sync"] }
+tokio = { workspace = true, features = ["io-util", "macros", "net", "sync"], optional = true }
```
jdygert-spok: Let's leave off the tokio feature for now.

ssweetney-spok: Gotcha. I didn't fully understand the second comment you left on the issue. I'll update this

jdygert-spok (dismissed):

Makefile.toml
```diff
@@ -192,7 +192,7 @@ workspace = true
 script = '''
 cargo clippy
 cargo clippy --all-targets
-cargo clippy --no-default-features
+# cargo clippy --no-default-features (TODO: Uncomment when aranya-client no longer requires the 'default' feature to be enabled)
```
gknopf-aranya: We should open an issue for this and reference it in all these TODOs. That way, when someone is assigned the issue in the future, they know where to make changes.

crates/aranya-client/src/lib.rs
```diff
@@ -22,6 +22,9 @@
     missing_docs
 )]
 
+#[cfg(not(feature = "default"))]
```
gknopf-aranya, line 26: Need a TODO to remove this later

crates/aranya-client/Cargo.toml
```diff
@@ -46,3 +46,7 @@ test-log = { workspace = true }
 [[test]]
 name = "tests"
 path = "tests/tests.rs"
+
+[package.metadata.cargo-all-features]
+# TODO: Remove when enabling the 'default' feature is no longer required
+always_include_features = ["default"]
```
gknopf-aranya, line 52: Is this enabling default features in the `aranya-client` or in `tokio`?
What is the current set of default features this is enabling?

ssweetney-spok: This is the config for the 'aranya-client' crate so it's just always enabling the default feature for this crate.

> What is the current set of default features this is enabling?

None at the moment. See above

```Toml
[features]
default = []
```

gknopf-aranya: So the purpose of this change is prevent users from updating their `Cargo.toml` if we add features later?

Makefile.toml
```diff
@@ -192,7 +192,7 @@ workspace = true
 script = '''
 cargo clippy
 cargo clippy --all-targets
-cargo clippy --no-default-features
+# cargo clippy --no-default-features (TODO: https://github.com/aranya-project/aranya/issues/448)
```
gknopf-aranya: Shorthand:
```suggestion
# cargo clippy --no-default-features (TODO: uncomment in #448) 
```
If it was in the `aranya-core` repo it would be `aranya-core#<issue number>`

crates/aranya-client/Cargo.toml
```diff
@@ -46,3 +46,7 @@ test-log = { workspace = true }
 [[test]]
 name = "tests"
 path = "tests/tests.rs"
+
+[package.metadata.cargo-all-features]
+# TODO: https://github.com/aranya-project/aranya/issues/448
```
gknopf-aranya:
```suggestion
# TODO: remove in #448
```

crates/aranya-client/Cargo.toml
jdygert-spok: The purpose is to allow us to put existing functionality behind a new feature flag which is default enabled, without breaking compatibility.

crates/aranya-client/src/lib.rs
```diff
@@ -22,6 +22,10 @@
     missing_docs
 )]
 
+// TODO: https://github.com/aranya-project/aranya/issues/448
```
gknopf-aranya, line 25:
```suggestion
// TODO: remove in #448
```

PR #451, opened 2025-08-11
logging: remove error! from From impl

djohnson-spideroak:
This is one location where excessive errors are logged by aranya. This line would log even during non-fatal errors like WouldBlock, causing logs to fill up quickly. 

Another solution could be to not log on WouldBlock and other non-fatal errors.

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -9,7 +9,6 @@ use anyhow::Context as _;
 use aranya_capi_core::{opaque::Opaque, prelude::*, ErrorCode, InvalidArg};
 use aranya_client::aqc::{self, AqcPeerStream};
 use aranya_daemon_api::Text;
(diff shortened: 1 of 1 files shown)


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -85,7 +84,6 @@ pub enum Error {
 
 impl From<&imp::Error> for Error {
     fn from(err: &imp::Error) -> Self {
-        error!(error = %err.report(), "Aranya client C API error");
```
gknopf-aranya, line 88: I thought we were changing the log level from `error` to `debug`

djohnson-spideroak: I was talking about the aranya core change, but I can do it for this one too.

ssweetney-spok (approved):

gknopf-aranya (approved):

PR #452, opened 2025-08-11
Add a finalize command to the policy

### LINKED ISSUE #321: policy: introduce finalization commands
Introduce finalization commands to the default policy.

ssweetney-spok:
Resolves #321


gknopf-aranya:
@djohnson-spideroak, @ssweetney-spok are we planning to add a way to invoke this command to the Aranya API?

djohnson-spideroak:
> > @djohnson-spideroak, @ssweetney-spok are we planning to add a way to invoke this command to the Aranya API?
> 
> I believe so but there isn't another issue for that yet.

I think we can do that in this PR

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1225,6 +1225,20 @@ pub fn close_team(client: &Client, team: &TeamId) -> Result<(), imp::Error> {
     Ok(())
 }
 
+/// Finalizes the team's current graph state.
```
gknopf-aranya: We should mention that only the owner role can perform this operation.
Readers may not know what finalization means. Is it finalizing graph commands before a certain point, or preventing the graph from ever being appended too?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1225,6 +1225,20 @@ pub fn close_team(client: &Client, team: &TeamId) -> Result<(), imp::Error> {
...
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+///
+/// @relates AranyaClient.
+pub fn finalize_team(client: &Client, team: &TeamId) -> Result<(), imp::Error> {
+    let client = client.imp();
+    client
+        .rt
+        .block_on(client.inner.team(team.into()).finalize_team())?;
```
gknopf-aranya, line 1274: Should we check if the device is the owner role before performing this operation or let the policy have an internal error if this fails?
If we let the policy return an error directly if the wrong role invokes this, we should make sure the error returned by the policy clearly indicates the reason the operation failed was because the device had the wrong role.

crates/aranya-daemon/src/api.rs
```diff
@@ -187,6 +187,7 @@ impl EffectHandler {
             trace!(?effect, "handling effect");
             match effect {
                 TeamCreated(_team_created) => {}
+                TeamFinalized(_team_finalized) => {}
                 TeamTerminated(_team_terminated) => {}
```
gknopf-aranya, line 192: Should `TeamTerminated` also finalize the graph?

crates/aranya-daemon/src/policy.md
```diff
@@ -436,6 +437,51 @@ finish function add_new_device(key_bundle struct KeyBundle, key_ids struct KeyId
...
+
+Finalizes the current state of the team's graph.
+
+```policy
+action finalize_team() {
+    publish FinalizeTeam {}
+}
+
+effect TeamFinalized {
+    owner_id id,
```
gknopf-aranya, line 454: May be worth including the `team_id` here so it's easily accessible.
The owner and other peers that sync this command should have the team ID already available via context, but it may be more convenient to access via the command.

crates/aranya-daemon/src/policy.md
```diff
@@ -436,6 +437,51 @@ finish function add_new_device(key_bundle struct KeyBundle, key_ids struct KeyId
...
+                owner_id: author.device_id,
+            }
+        }
+    }
+}
+```
+
+**Invariants:**
+
+- Only an Owner will create this event.
```
gknopf-aranya: Event isn't in our graph terminology. We have actions, commands, and effects.
```suggestion
- Only an Owner will create this command.
```

examples/rust/src/main.rs
```diff
@@ -371,6 +371,18 @@ async fn main() -> Result<()> {
     // wait for syncing.
     sleep(sleep_interval).await;
 
+    // finalize operation
+
+    // Admin tries to publish a finalize command but fails
+    match admin_team.finalize_team().await {
+        Ok(()) => bail!("expected this operation to fail"),
+        Err(Error::Aranya(_)) => {}
```
gknopf-aranya: Would be better to return an error that indicates the finalize operation failed due to an invalid device role. `Error::Aranya(_)` is too vague.

examples/c/example.c
```diff
@@ -668,6 +668,11 @@ AranyaError run(Team *t) {
 
     sleep(1);
 
+    printf("finalizing graph state\n");
+    
+    err = aranya_finalize_team(&owner->client, &t->id);
```
gknopf-aranya: Let's make sure when the admin attempts to perform this operation it fails in the C example just like in the Rust example. Want to make sure error handling for this is correct in the C API.

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: Yeah, I think this doc should include a pretty detailed description of finalization as well, maybe linking to an external document. We might want to write one targeted to this context for the docs site.

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: It should be enforced by policy. I think the default "not authorized" plus the description here is enough.

gknopf-aranya: If a user sees a "not authorized" error, how will they know it's because the device role is not authorized to perform the operation? It could also be because the team does not exist or the device has not been added to the team.

gknopf-aranya:
Should we update the MVP spec with this new API method?

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: Good points. There should probably be an entry for that in the [glossary](https://aranya-project.github.io/glossary).

The most detailed description I could find was this [section in the aranya-docs repo](https://github.com/aranya-project/aranya-docs/blob/1ecf718ca179a431db724a4ada45129d96edcbf2/docs/graph.md#finalization)

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: That is a general issue, not something specific to this action. We need to figure out how we want to bubble up more error info from the vm.

gknopf-aranya: Good point. We don't want to add policy specific error handling to the Aranya client when those permissions and errors should be managed by the policy itself and the policy could change in the future.

examples/rust/src/main.rs
ssweetney-spok: Not currently possible at the moment. FWIW, this is how the other checks for daemon errors look in the tests

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I do think this puts more of a burden on our API docs to fill in the gap in the meantime.

crates/aranya-daemon/src/policy.md
ssweetney-spok: I saw ['event' here](https://github.com/aranya-project/aranya-docs/blob/1ecf718ca179a431db724a4ada45129d96edcbf2/docs/graph.md#finalization) so I think that's why it popped in my mind. I do think command fits better in this case

gknopf-aranya: Is an event different from a command in that doc? As far as I can tell they are used interchangeably. If they are the same thing, it would be ideal to use the same word for it everywhere.

ssweetney-spok: Also, it's used in the description for `TerminateTeam`'s invariant list

```
**Invariants:**

- This is the final command in the graph.
- Only an Owner can create this event.
- Once terminated, no further communication will occur over the team graph.
```

Should we keep things consistent for now?

examples/rust/src/main.rs
djohnson-spideroak, line 1: Lets also get some tests going, ideally we should have a successful finalization command, parallel finalization commands to validate failure, etc.

crates/aranya-daemon/src/policy.md
ssweetney-spok: > Is an event different from a command in that doc? As far as I can tell they are used interchangeably. If they are the same thing, it would be ideal to use the same word for it everywhere.

Yea, they seem to be the same thing

examples/rust/src/main.rs
ssweetney-spok: We can't validate that parallel finalization commands produce an error since only the owner is able to publish the command.

There is [a test in aranya-core](https://github.com/aranya-project/aranya-core/pull/172/files#diff-7e6866574f71023fcb57b03ff1f28b156e8e55340523b80a1ffdcb6774b6aafdR964-R976) for that though

ssweetney-spok: I'm also not sure what a test would look like since we can't really inspect the graph

crates/aranya-daemon/src/policy.md
ssweetney-spok: Could you explain the use case your thinking of? `aranya_daemon::api::EffectHandler::handle_effects` has the graph/team ID passed in as a param so we can use that

crates/aranya-daemon/src/policy.md
gknopf-aranya: Yeah, I think we should try to be consistent. I've heard the team use command more than event to describe graph commands.

crates/aranya-daemon/src/policy.md
gknopf-aranya: The effect handler may have the team ID but suppose it calls a method like `fn foo(team_finalized: TeamFinalized)` that doesn't need the team ID now but needs it later. If the team ID isn't part of the command, we need to refactor code to pass it alongside the command as a function parameter, if it is included in the command, that context is automatically available wherever the command is.

gknopf-aranya: I'm borrowing from the concept of Go contexts here:
https://pkg.go.dev/context
It's a way to pass less information via function parameters because it's lumped into the context.
In this case, team ID is relevant context to a command and could be lumped together with it.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: [Related](https://github.com/aranya-project/aranya-project.github.io/pull/77)

[Follow-up](https://github.com/aranya-project/aranya/issues/468)

elagergren-spideroak:
Or rebase on to the policy PR?

examples/rust/src/main.rs
ssweetney-spok: I was wrong about that. Owners can create other owners so we could've tested the parallel finalization error.

Maybe, the tests could make use of the new "Epoch" fact somehow 🤔.

Edit: Added a test [here](https://github.com/aranya-project/aranya/pull/452/files#diff-4522f1fd60038a030ad93bcdf14ed0ef84359f6aad457eaff125d8b257162c61R12-R64)

PR #463, opened 2025-08-19
WIP: add afc policy back in

gknopf-aranya:
Old AFC policy must be updated to use IDs instead of ints for labels.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ wildcard_imports = "warn"
 
 [workspace.dependencies]
 aranya-aqc-util = { version = "0.6.0", features = ["alloc"] }
+aranya-afc-util = { version = "0.11.0", features = ["alloc"] }
 aranya-capi-core = { version = "0.4.2" }
 aranya-capi-codegen = { version = "0.3.2" }
 aranya-crypto = { version = "0.8.0", features = ["alloc", "aqc", "fs-keystore", "clone-aead", "std", "tls"] }
--- a/crates/aranya-daemon/Cargo.toml
+++ b/crates/aranya-daemon/Cargo.toml
@@ -14,13 +14,17 @@ workspace = true
 
 
 [features]
-default = []
-
+default = ["afc", "unstable"]
+# Unstable features
+unstable = ["afc"]
+# Aranya Fast Channels
+afc = []
 
 [dependencies]
 aranya-daemon-api = { workspace = true }
 aranya-keygen = { workspace = true }
 
+aranya-afc-util = { workspace = true }
 aranya-aqc-util = { workspace = true }
 aranya-crypto = { workspace = true }
 aranya-crypto-ffi = { workspace = true }
--- a/crates/aranya-daemon/src/actions.rs
+++ b/crates/aranya-daemon/src/actions.rs
@@ -335,6 +335,34 @@ where
         .in_current_span()
     }
 
+    /// Queries all AFC network names off-graph.
+    #[instrument(skip(self))]
+    fn query_afc_network_names_off_graph(
+        &self,
+    ) -> impl Future<Output = Result<Vec<(NetIdentifier, DeviceId)>>> + Send {
+        self.session_action(move || VmAction {
+            name: ident!("query_afc_network_names"),
+            args: Cow::Owned(vec![]),
+        })
+        .and_then(|(_, effects)| {
+            std::future::ready(
+                effects
+                    .into_iter()
+                    .map(|eff| {
+                        let Effect::QueryAfcNetworkNamesOutput(eff) = eff else {
+                            anyhow::bail!("bad effect in query_network_names");
+                        };
+                        Ok((
+                            NetIdentifier(eff.net_identifier),
+                            DeviceId::from(eff.device_id),
+                        ))
+                    })
+                    .collect(),
+            )
+        })
+        .in_current_span()
+    }
+
     /// Creates a bidirectional AQC channel off graph.
     #[allow(clippy::type_complexity)]
     #[instrument(skip(self), fields(peer_id = %peer_id, label_id = %label_id))]
--- a/crates/aranya-daemon/src/api.rs
+++ b/crates/aranya-daemon/src/api.rs
@@ -87,14 +87,18 @@ impl DaemonApiServer {
         recv_effects: EffectReceiver,
         invalid: InvalidGraphs,
         aqc: Aqc<CE, KS>,
+        #[cfg(all(feature = "afc", feature = "unstable"))] afc: Aqc<CE, KS>,
         crypto: Crypto,
         seed_id_dir: SeedDir,
         quic: Option<quic_sync::Data>,
     ) -> anyhow::Result<Self> {
         let listener = UnixListener::bind(&uds_path)?;
         let aqc = Arc::new(aqc);
+        let afc = Arc::new(afc);
         let effect_handler = EffectHandler {
             aqc: Arc::clone(&aqc),
+            #[cfg(all(feature = "afc", feature = "unstable"))]
+            afc: Arc::clone(&afc),
         };
         let api = Api(Arc::new(ApiInner {
             client,
@@ -175,6 +179,8 @@ impl DaemonApiServer {
 #[derive(Clone, Debug)]
 struct EffectHandler {
     aqc: Arc<Aqc<CE, KS>>,
+    #[cfg(all(feature = "afc", feature = "unstable"))]
+    afc: Arc<Aqc<CE, KS>>,
 }
 
 impl EffectHandler {
@@ -184,6 +190,7 @@ impl EffectHandler {
         trace!("handling effects");
 
         use Effect::*;
+        // TODO: support feature flag in interface generator to compile out certain effects.
         for effect in effects {
             trace!(?effect, "handling effect");
             match effect {
@@ -210,19 +217,40 @@ impl EffectHandler {
                         )
                         .await;
                 }
-                AqcNetworkNameUnset(e) => self.aqc.remove_peer(graph, e.device_id.into()).await,
+                AqcNetworkNameUnset(e) => self.afc.remove_peer(graph, e.device_id.into()).await,
+                AfcNetworkNameSet(e) => {
+                    #[cfg(all(feature = "afc", feature = "unstable"))]
+                    self.aqc
+                        .add_peer(
+                            graph,
+                            api::NetIdentifier(e.net_identifier.clone()),
+                            e.device_id.into(),
+                        )
+                        .await;
+                }
+                AfcNetworkNameUnset(e) =>
+                {
+                    #[cfg(all(feature = "afc", feature = "unstable"))]
+                    self.afc.remove_peer(graph, e.device_id.into()).await
+                }
                 QueriedLabel(_) => {}
                 AqcBidiChannelCreated(_) => {}
                 AqcBidiChannelReceived(_) => {}
                 AqcUniChannelCreated(_) => {}
                 AqcUniChannelReceived(_) => {}
+                AfcBidiChannelCreated(_) => {}
+                AfcBidiChannelReceived(_) => {}
+                AfcUniChannelCreated(_) => {}
+                AfcUniChannelReceived(_) => {}
                 QueryDevicesOnTeamResult(_) => {}
                 QueryDeviceRoleResult(_) => {}
                 QueryDeviceKeyBundleResult(_) => {}
-                QueryAqcNetIdentifierResult(_) => {}
                 QueriedLabelAssignment(_) => {}
                 QueryLabelExistsResult(_) => {}
+                QueryAqcNetIdentifierResult(_) => {}
                 QueryAqcNetworkNamesOutput(_) => {}
+                QueryAfcNetIdentifierResult(_) => {}
+                QueryAfcNetworkNamesOutput(_) => {}
             }
         }
         Ok(())
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -219,6 +219,31 @@ impl Daemon {
                 )
             };
 
+            let afc = {
+                let peers = {
+                    let mut peers = BTreeMap::new();
+                    for graph_id in &graph_ids {
+                        let graph_peers = BiBTreeMap::from_iter(
+                            client
+                                .actions(graph_id)
+                                .query_afc_network_names_off_graph()
+                                .await?,
+                        );
+                        peers.insert(*graph_id, graph_peers);
+                    }
+                    peers
+                };
+                // TODO: use AFC shm key store
+                Aqc::new(
+                    eng.clone(),
+                    pks.ident_pk.id()?,
+                    aranya_store
+                        .try_clone()
+                        .context("unable to clone keystore")?,
+                    peers,
+                )
+            };
+
             let data = QSData { psk_store };
 
             let crypto = crate::api::Crypto {
@@ -237,6 +262,8 @@ impl Daemon {
                 recv_effects,
                 invalid_graphs,
                 aqc,
+                #[cfg(all(feature = "afc", feature = "unstable"))]
+                afc,
                 crypto,
                 seed_id_dir,
                 Some(data),
--- a/crates/aranya-daemon/src/policy.md
+++ b/crates/aranya-daemon/src/policy.md
@@ -29,7 +29,8 @@ higher role back down to `Member`.
   * Assign/revoke Operator role.
   * Define/undefine channel label.
   * Assign/revoke channel label.
-  * Set/unset AQC address&name.
+  * Set/unset AQC address & name.
+  * Set/unset AFC address & name.
 
 * Admin:
   * Assign/revoke Operator role.
@@ -41,10 +42,12 @@ higher role back down to `Member`.
   * Add (new) / remove Member.
   * Define channel label.
   * Assign/revoke channel label.
-  * Set/unset AQC address&name.
+  * Set/unset AQC address & name.
+  * Set/unset AFC address & name.
 
 * Member:
   * Create/delete AQC channel.
+  * Create/delete AFC channel.
 
 **Invariants**:
 
@@ -63,6 +66,7 @@ higher role back down to `Member`.
 ### Imports & Global Constants
 
 ```policy
+use afc
 use aqc
 use crypto
 use device
@@ -133,6 +137,9 @@ fact TeamEnd[]=>{}
 
 // Stores a Member's associated network identifier for AQC.
 fact AqcMemberNetworkId[device_id id]=>{net_identifier string}
+
+// Stores a Member's associated network identifier for AFC.
+fact AfcMemberNetworkId[device_id id]=>{net_identifier string}
 ```
 
 ### Functions
@@ -310,6 +317,62 @@ function can_create_aqc_uni_channel(sender_id id, receiver_id id, label_id id) b
     // Devices cannot create channels with themselves.
     //
     // This should have been caught by the AQC FFI, so check
+    // instead of just returning false.
+    check sender_id != receiver_id
+
+    // The writer must have permissions to write (send) data.
+    let writer_op = get_allowed_op(sender_id, label_id)
+    match writer_op {
+        ChanOp::RecvOnly => { return false }
+        ChanOp::SendOnly => {}
+        ChanOp::SendRecv => {}
+    }
+
+    // The reader must have permission to read (receive) data.
+    let reader_op = get_allowed_op(receiver_id, label_id)
+    match reader_op {
+        ChanOp::RecvOnly => {}
+        ChanOp::SendOnly => { return false }
+        ChanOp::SendRecv => {}
+    }
+
+    return true
+}
+```
+
+##### AFC Channel Functions
+
+```policy
+// Reports whether the devices have permission to create
+// a bidirectional AFC channel with each other.
+function can_create_afc_bidi_channel(device1 id, device2 id, label_id id) bool {
+    // Devices cannot create channels with themselves.
+    //
+    // This should have been caught by the AFC FFI, so check
+    // instead of just returning false.
+    check device1 != device2
+
+    // Both devices must have permissions to read (recv) and
+    // write (send) data.
+    let device1_op = get_allowed_op(device1, label_id)
+    if device1_op != ChanOp::SendRecv {
+        return false
+    }
+
+    let device2_op = get_allowed_op(device2, label_id)
+    if device2_op != ChanOp::SendRecv {
+        return false
+    }
+
+    return true
+}
+
+// Reports whether the devices have permission to create
+// a unidirectional AFC channel with each other.
+function can_create_afc_uni_channel(sender_id id, receiver_id id, label_id id) bool {
+    // Devices cannot create channels with themselves.
+    //
+    // This should have been caught by the AFC FFI, so check
     // instead of just returning false.
     check sender_id != receiver_id
 
@@ -1072,7 +1135,127 @@ command UnsetAqcNetworkName {
 
 **Invariants**:
 
-- Only Owners and Operators Operators can unset AQC network names from Members.
+- Only Owners and Operators can unset AQC network names from Members.
+
+## SetAfcNetworkName
+
+Associates a network name and address to a `Member` for use in AFC.
+
+```policy
+action set_afc_network_name (device_id id, net_identifier string) {
+    publish SetAfcNetworkName {
+        device_id: device_id,
+        net_identifier: net_identifier,
+    }
+}
+
+effect AfcNetworkNameSet {
+    device_id id,
+    net_identifier string,
+}
+
+command SetAfcNetworkName {
+    fields {
+        device_id id,
+        net_identifier string,
+    }
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        check team_exists()
+
+        let author = get_valid_device(envelope::author_id(envelope))
+        let device = get_valid_device(this.device_id)
+
+        // Only Owners and Operators can associate a network name.
+        check is_owner(author.role) || is_operator(author.role)
+        // Only Members can be associated a network name.
+        check is_member(device.role)
+
+        let net_id_exists = query AfcMemberNetworkId[device_id: this.device_id]
+
+        if net_id_exists is Some {
+            let net_id = unwrap net_id_exists
+            finish {
+                update AfcMemberNetworkId[device_id: this.device_id]=>{net_identifier: net_id.net_identifier} to {
+                    net_identifier: this.net_identifier
+                }
+
+                emit AfcNetworkNameSet {
+                    device_id: device.device_id,
+                    net_identifier: this.net_identifier,
+                }
+            }
+        }
+        else {
+            finish {
+                create AfcMemberNetworkId[device_id: this.device_id]=>{net_identifier: this.net_identifier}
+
+                emit AfcNetworkNameSet {
+                    device_id: device.device_id,
+                    net_identifier: this.net_identifier,
+                }
+            }
+        }
+    }
+}
+```
+
+**Invariants**:
+
+- Only Owners and Operators can assign AFC network names to Members.
+- Members can only be assigned to one AFC network name.
+
+## UnsetAfcNetworkName
+
+Dissociates an AFC network name and address from a `Member`.
+
+```policy
+action unset_afc_network_name (device_id id) {
+    publish UnsetAfcNetworkName {
+        device_id: device_id,
+    }
+}
+
+effect AfcNetworkNameUnset {
+    device_id id,
+}
+
+command UnsetAfcNetworkName {
+    fields {
+        device_id id,
+    }
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        check team_exists()
+
+        let author = get_valid_device(envelope::author_id(envelope))
+        let device = get_valid_device(this.device_id)
+
+        // Only Owners, Admins, and Operators can unset a Member's network name.
+        check is_owner(author.role) || is_admin(author.role) || is_operator(author.role)
+        check is_member(device.role)
+
+        check exists AfcMemberNetworkId[device_id: this.device_id]
+        finish {
+            delete AfcMemberNetworkId[device_id: this.device_id]
+
+            emit AfcNetworkNameUnset {
+                device_id: device.device_id,
+            }
+        }
+    }
+}
+```
+
+**Invariants**:
+
+- Only Owners and Operators can unset AFC network names from Members.
 
 ## CreateChannel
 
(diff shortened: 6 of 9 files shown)


gknopf-aranya:
Policy will not compile until https://github.com/aranya-project/aranya-core/issues/376 is completed:
```
Caused by:
    0: should be able to compile policy
    1: invalid type: Argument 6 (`label`) in FFI call to `afc::create_bidi_channel` found `id`, not `int` at line 1678 col 5:
       	let channel = afc::create_bidi_channel(
               parent_cmd_id,
               author.enc_key_id,
               author_id,
               peer_enc_pk,
               peer_id,
               label_id,
           )
```

crates/aranya-daemon/src/afc.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub(crate) async fn delete_channel(&self, channel_id: ChannelId) -> Result<()>
+    where
+        E: Engine<CS = C>,
+    {
+        self.shm
+            .lock()
+            .await
+            .write
+            .remove(channel_id)
+            .map_err(|err| anyhow!("unable to removfe AFC channel: {err}"))
```
gknopf-aranya:
```suggestion
            .map_err(|err| anyhow!("unable to remove AFC channel: {err}"))
```

crates/aranya-daemon/src/afc.rs
```diff
@@ -0,0 +1,302 @@
...
+    cfg: AfcConfig,
+    write: WriteState<C, Rng>,
+}
+
+impl<C> AfcShm<C>
+where
+    C: CipherSuite,
+{
+    fn new(cfg: AfcConfig) -> Result<Self> {
+        // TODO: issue stellar-tapestry#34
```
elagergren-spideroak: stellar-tapestry?

crates/aranya-daemon/src/policy.md
```diff
@@ -133,6 +137,9 @@ fact TeamEnd[]=>{}
 
 // Stores a Member's associated network identifier for AQC.
 fact AqcMemberNetworkId[device_id id]=>{net_identifier string}
+
+// Stores a Member's associated network identifier for AFC.
+fact AfcMemberNetworkId[device_id id]=>{net_identifier string}
```
elagergren-spideroak: not needed for AFC

crates/aranya-client/tests/common/mod.rs
```diff
@@ -167,9 +171,11 @@ pub struct DeviceCtx {
 }
 
 impl DeviceCtx {
-    async fn new(_team_name: &str, name: &str, work_dir: PathBuf) -> Result<Self> {
+    async fn new(team_name: &str, name: &str, work_dir: PathBuf) -> Result<Self> {
         let addr_any = Addr::from((Ipv4Addr::LOCALHOST, 0));
 
+        let afc_shm_path = get_shm_path(format!("/{team_name}_{name}"));
```
djohnson-spideroak: This might not work on every platform given shm requirements.

examples/rust/src/main.rs
```diff
@@ -76,6 +76,7 @@ impl ClientCtx {
         let work_dir = TempDir::with_prefix(user_name)?;
 
         let daemon = {
+            let shm = format!("/shm_{}", user_name);
```
djohnson-spideroak, line 79: we may need to apply that cleanup/fix function here.

crates/aranya-daemon/src/policy.md
```diff
@@ -2171,17 +2573,23 @@ action query_aqc_network_names() {
 }
 
 effect QueryAqcNetworkNamesOutput {
+    // AFC network name assigned to device.
```
djohnson-spideroak: not AFC

crates/aranya-daemon/src/actions.rs
```diff
@@ -385,16 +370,39 @@ where
         .in_current_span()
     }
 
-    /// Creates a bidirectional AQC channel.
+    /// Creates a bidirectional AFC channel off graph.
+    #[cfg(any(feature = "afc", feature = "preview"))]
+    #[allow(clippy::type_complexity)]
     #[instrument(skip(self), fields(peer_id = %peer_id, label_id = %label_id))]
-    fn create_aqc_bidi_channel(
```
djohnson-spideroak, line 390: I dont think we want to overwrite/remove this, AQC is getting put behind experimental

crates/aranya-daemon/src/afc.rs
gknopf-aranya: I'll remove this comment. We're tracking the issue as part of the 3.0 milestone.

crates/aranya-daemon/src/actions.rs
gknopf-aranya: We were only using the off-graph channel creation commands for AQC so I removed these since they are unused.

gknopf-aranya: We can revert this since it isn't relevant to the current PR but we'll probably want to clean it up somewhere else.

examples/rust/src/main.rs
gknopf-aranya: Which function are you referring to?

gknopf-aranya: Is it this one that was removed with afc a while back?
```
    fn get_shm_path(path: String) -> String {
        if cfg!(target_os = "macos") && path.len() > 31 {
            // Shrink the size of the team name down to 22 bytes to work within macOS's limits.
            let d = Sha256::hash(path.as_bytes());
            let t: [u8; 16] = d[..16].try_into().unwrap();
            return format!("/{}", t.to_base58());
        };
        path
    }
```

gknopf-aranya: I moved this code into the shm util crate so the code always executes when creating an `AfcShmBuf` from a string.

gknopf-aranya:
"Put shm util code behind a feature flag (move to another crate?)"
I have this as a todo, but it doesn't seem required for the initial version of afc open/seal.

crates/aranya-daemon/src/config.rs
```diff
@@ -198,6 +202,16 @@ pub struct SyncConfig {
 #[serde(deny_unknown_fields)]
 pub struct AqcConfig {}
 
+/// AFC configuration.
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct AfcConfig {
+    /// Shared memory path.
+    pub shm_path: String,
```
jdygert-spok: Use `Box<shm::Path>` once https://github.com/aranya-project/aranya-core/pull/406 is merged. Same for the afc shm info.

crates/aranya-daemon/src/afc.rs
```diff
@@ -0,0 +1,325 @@
...
+                )) {
+                Ok(w) => w,
+                Err(e) => {
+                    warn!(?e);
+                    WriteState::open(&path, Flag::OpenOnly, Mode::ReadWrite, cfg.max_chans, Rng)
+                        .context(format!(
+                            "unable to open existing `WriteState`: {:?}",
+                            cfg.shm_path
+                        ))?
+                }
```
ssweetney-spok: Seems like we need that logic for getting the current channel ID when a shared memory object already exists

crates/aranya-daemon/src/api.rs
```diff
@@ -661,31 +693,11 @@ impl DaemonApi for Api {
...
-    #[instrument(skip(self), err)]
-    async fn delete_aqc_uni_channel(
-        self,
-        _: context::Context,
-        chan: api::AqcUniChannelId,
-    ) -> api::Result<api::AqcCtrl> {
-        // TODO: remove AQC uni channel from Aranya.
-        todo!();
-    }
-
```
ssweetney-spok: Were these supposed to be removed? This may cause conflicts with https://github.com/aranya-project/aranya/pull/473

crates/aranya-client/tests/common/mod.rs
```diff
@@ -167,9 +167,11 @@ pub struct DeviceCtx {
 }
 
 impl DeviceCtx {
-    async fn new(_team_name: &str, name: &str, work_dir: PathBuf) -> Result<Self> {
+    async fn new(team_name: &str, name: &str, work_dir: PathBuf) -> Result<Self> {
         let addr_any = Addr::from((Ipv4Addr::LOCALHOST, 0));
 
+        let afc_shm_path = format!("/{team_name}_{name}");
```
djohnson-spideroak: as long as each test uses a different team name this is fine. otherwise we might consider adding a random number or something to reduce collision chances.

crates/aranya-daemon/Cargo.toml
```diff
@@ -14,18 +14,25 @@ workspace = true
...
 [features]
-default = []
-
+# TODO: don't enable preview features or afc in daemon by default.
+# Enabled them because unit tests in aranya-client aren't passing without this.
+default = ["afc", "preview"]
+# Preview features
+preview = ["afc"]
+# Aranya Fast Channels
+afc = []
```
ssweetney-spok:
```suggestion
afc = ["aranya-daemon-api/afc", "aranya-util/afc"]
```

examples/rust/Cargo.toml
```diff
@@ -40,6 +40,7 @@ wildcard_imports = "warn"
 [dependencies]
 aranya-client = { path = "../../crates/aranya-client" }
 aranya-daemon-api = { path = "../../crates/aranya-daemon-api" }
+aranya-fast-channels = { version = "0.11.0" }
```
djohnson-spideroak: we will need to update this after release correct?

crates/aranya-daemon/Cargo.toml
```diff
@@ -14,18 +14,25 @@ workspace = true
 
 
 [features]
-default = []
-
+# TODO: don't enable preview features or afc in daemon by default.
+# Enabled them because unit tests in aranya-client aren't passing without this.
+default = ["afc", "preview"]
+# Preview features
+preview = ["afc"]
```
ssweetney-spok: Why is "afc" in the feature list for "preview"?

crates/aranya-daemon/src/config.rs
gknopf-aranya: Will update this PR once that is merged

examples/rust/Cargo.toml
gknopf-aranya: Yes. btw, once the multi-node example is merged, this can be updated to use the workspace dep instead.

crates/aranya-client/tests/common/mod.rs
gknopf-aranya: They do currently. We're trying to keep the shm paths short because they have length limits. That's why we can't use randomly generated temp dirs for the shm paths.

gknopf-aranya: We're working with 32 characters on macos and 255 characters on other OS.
```
    /// The maximum number of bytes allowed in a `Path`.
    pub const NAME_MAX: usize = if cfg!(target_os = "macos") { 32 } else { 255 };
```

gknopf-aranya: Opening another issue for this here:
https://github.com/aranya-project/aranya/issues/487
This isn't something that's currently a problem we're encountering, although it would help to improve the robustness of our unit tests.

Cargo.toml
```diff
@@ -105,3 +108,24 @@ allow-branch = ["main"]
 shared-version = true
 tag-message = ""
 tag-prefix = ""
+
+[patch.crates-io]
```
gknopf-aranya: Revert after aranya-core release

crates/aranya-daemon/src/config.rs
gknopf-aranya: Updated to `Box<shm::Path>` but I'm seeing a bunch of compiler errors locally related to `Clone`, `Serialize`, `Deserialize`

gknopf-aranya: Using the `alloc` feature for afc crates solved this

crates/aranya-daemon/Cargo.toml
gknopf-aranya: Because afc is a preview feature.

crates/aranya-daemon/src/api.rs
gknopf-aranya: I removed a few unused methods on this PR. They don't need to be included on this PR but it seemed like a simple enough thing to clean up while we have open client/daemon PRs.

crates/aranya-daemon/Cargo.toml
ssweetney-spok: It should be separate like how it's done in the [AQC feature flag PR](https://github.com/aranya-project/aranya/pull/473/files#diff-af547c21969c096209de84573218a86ff82a672b9053b9c6ceb0f65e94b7242b).

We should avoid adding features to other features if we know we'll be removing it in the future since that may [cause a breaking change](https://doc.rust-lang.org/cargo/reference/semver.html#cargo-feature-remove-another)

examples/rust/src/main.rs
```diff
@@ -76,6 +78,9 @@ impl ClientCtx {
         let work_dir = TempDir::with_prefix(user_name)?;
 
         let daemon = {
+            let shm = format!("/shm_{}\0", user_name);
```
djohnson-spideroak: Do we still need the null char ending? I thought we had some changes in core to not require it.

jdygert-spok: Yes, the conversion methods still require it. We _could_ have the `-> Box<Path>` conversions add it if missing.

gknopf-aranya: It's required now that this PR has been merged:
https://github.com/aranya-project/aranya-core/pull/406

crates/aranya-daemon/src/policy.md
```diff
@@ -1470,6 +1530,343 @@ command AqcCreateUniChannel {
...
+Creates a bidirectional AFC channel for off-graph messaging. This is an ephemeral command, which
+means that it can only be emitted within an ephemeral session so that it is not added to the graph
+of commands. Furthermore, it cannot persist any changes to the factDB.
+
+The `create_afc_bidi_channel` action creates the `ChannelKeys`, encapsulates them for the peer and the
+author, and sends the encapsulations through the `AfcCreateBidiChannel` command. When processing the
+command, the device will decapsulate their keys and store them in the shared memory DB.
+
+```policy
+action create_afc_bidi_channel(peer_id id, label_id id) {
```
djohnson-spideroak:
```suggestion
ephemeral action create_afc_bidi_channel(peer_id id, label_id id) {
```

same for other ephemeral actions/commands

crates/aranya-daemon/src/api.rs
```diff
@@ -320,6 +334,24 @@ impl DaemonApi for Api {
...
+        Ok(self.afc.get_shm_info().await)
+    }
+
+    #[cfg(not(feature = "afc"))]
+    #[instrument(skip(self), err)]
+    async fn afc_shm_info(
+        self,
+        context: ::tarpc::context::Context,
+    ) -> api::Result<api::AfcShmInfo> {
+        todo!()
```
djohnson-spideroak: maybe bug! instead?

crates/aranya-daemon/src/api.rs
```diff
@@ -320,6 +334,24 @@ impl DaemonApi for Api {
         self.device_id().map(|id| id.into_id().into())
     }
 
+    #[cfg(feature = "afc")]
+    #[instrument(skip(self), err)]
+    async fn afc_shm_info(
+        self,
+        context: ::tarpc::context::Context,
```
elagergren-spideroak: add an import for Context

crates/aranya-daemon/src/api.rs
elagergren-spideroak: or just remove it and gate it everywhere else

crates/aranya-daemon/src/afc.rs
ssweetney-spok: I think we need to add a method to `aranya_fast_channels::state::AranyaState` to query the current number of channels

cc: @jdygert-spok

ssweetney-spok: Went with [Dygert's suggestion](https://github.com/aranya-project/aranya/pull/463/commits/3d3cb869f61fa2b5db396993185fd4c71c13bf6a) to use `AranyaState::remove_if`. This may need to be changed if we decide to store the max ID in the shared memory.

crates/aranya-daemon/src/api.rs
ssweetney-spok: Removed it

crates/aranya-daemon/src/afc.rs
```diff
@@ -0,0 +1,347 @@
...
+pub struct AfcShm<C> {
+    cfg: AfcConfig,
+    write: WriteState<C, Rng>,
+}
+
+impl<C> AfcShm<C>
+where
+    C: CipherSuite,
+{
+    fn new(cfg: AfcConfig) -> Result<Self> {
```
djohnson-spideroak, line 42: thoughts on an "open if not exists" option?

examples/rust/Cargo.toml
ssweetney-spok: Removed it. It doesn't seem like the example should be depending on aranya crates other than aranya-client

examples/rust/src/main.rs
ssweetney-spok: Doesn't seem like it's needed since the [null byte is added](https://github.com/aranya-project/aranya-core/pull/406/files#diff-e8a6c07ef7a64993804c44f8f1864aaea19204eea903c2783870bf9232a7797fR300-R309) when de-serializing.

Should the null byte not be included when serializing `shm::Path` then? 🤔

crates/aranya-client-capi/Cargo.toml
```diff
@@ -20,6 +20,10 @@ doctest = false
 
 [features]
 default = []
+# Preview features
+preview = ["aranya-client/preview"]
```
gknopf-aranya, line 24: If the client is compiled with `preview` or `afc` we should pass those features to the aranya-client-api crate.
If the daemon is compiled with `preview` or `afc` we should also pass those features to the aranya-client-api crate.
I don't think we should let the aranya-client-api crate determine which features are in the client or daemon.
Because the client is a lib and the daemon is an executable, they are the release artifacts which should flow features to other crates they depend on.

crates/aranya-client/tests/common/mod.rs
gknopf-aranya, line 1: Let me know if there are any other changes required to the client crate. I'm trying to keep those in the client PR as much as possible:
https://github.com/aranya-project/aranya/pull/476

crates/aranya-client-capi/Cargo.toml
gknopf-aranya: Ignore this. I thought this was the `aranya-daemon-api` crate for a second.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -630,6 +635,14 @@ pub struct Label {
     pub name: Text,
 }
 
+/// AFC shared-memory info.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+
+pub struct AfcShmInfo {
```
gknopf-aranya: This contains the same info as the afc config in `config.rs`.
Having a new type here is good because we can update the types independently.

crates/aranya-daemon/src/afc.rs
djohnson-spideroak: maybe better as a followup when we are refining the lifecycle of the shm

crates/aranya-daemon/src/api.rs
```diff
@@ -216,12 +224,16 @@ impl EffectHandler {
                 AqcBidiChannelReceived(_) => {}
                 AqcUniChannelCreated(_) => {}
                 AqcUniChannelReceived(_) => {}
+                AfcBidiChannelCreated(_) => {}
+                AfcBidiChannelReceived(_) => {}
+                AfcUniChannelCreated(_) => {}
+                AfcUniChannelReceived(_) => {}
```
djohnson-spideroak, line 230: Dont we need to hook this up to call the channel received functions?

djohnson-spideroak: ahh I see, we manually handle them when creating or receiving the channel via the API

crates/aranya-daemon/src/afc.rs
```diff
@@ -0,0 +1,347 @@
...
+        debug!("setting up afc shm write side: {:?}", cfg.shm_path);
+        let write = {
+            #[cfg(any(test, feature = "example"))]
```
gknopf-aranya: We probably don't want to do this for the example. The example is supposed to show users how to use the public API.
We are currently expecting users to manage deleting the shm themselves before restarting the daemon.
At least until we have better signal handling to know when the daemon is being restarted vs. stopped.

crates/aranya-daemon/src/afc.rs
gknopf-aranya: The first `WriteState::open()` in this function would open the shm if it doesn't exist.
The second `WriteState::open()` will open an existing shm.

crates/aranya-daemon/src/afc.rs
```diff
@@ -0,0 +1,347 @@
...
+                Err(e) => {
+                    warn!(?e);
+                    WriteState::open(
+                        cfg.shm_path.clone(),
+                        Flag::OpenOnly,
+                        Mode::ReadWrite,
+                        cfg.max_chans,
+                        Rng,
+                    )
+                    .context(format!(
```
gknopf-aranya: Should we have better defined errors in this file rather than `.context()`?

crates/aranya-daemon/src/afc.rs
ssweetney-spok: > maybe better as a followup when we are refining the lifecycle of the shm

Yea, I think so

crates/aranya-daemon/src/afc.rs
```diff
@@ -0,0 +1,347 @@
...
+                        cfg.shm_path
+                    ))?
+                }
+            }
+        };
+
+        Ok(Self { cfg, write })
+    }
+
+    fn next_unused_id(&self) -> Result<u32> {
```
gknopf-aranya: We should document that this should only be done once at startup, rather than to obtain each channel ID.

crates/aranya-daemon/src/api.rs
gknopf-aranya: They are handled like this:
```
        let Some(Effect::AfcBidiChannelCreated(e)) =
            find_effect!(&effects, Effect::AfcBidiChannelCreated(e) if e.author_id == id.into())
        else {
            return Err(anyhow!("unable to find `AfcBidiChannelCreated` effect").into());
        };
```

crates/aranya-daemon/src/api.rs
```diff
@@ -216,12 +224,16 @@ impl EffectHandler {
                 AqcBidiChannelReceived(_) => {}
                 AqcUniChannelCreated(_) => {}
                 AqcUniChannelReceived(_) => {}
+                AfcBidiChannelCreated(_) => {}
```
gknopf-aranya, line 227: I wish there was a way to put some of these cases behind a feature flag so they don't appear to be handled for all feature sets.

crates/aranya-daemon/src/afc.rs
ssweetney-spok: Will remove this once https://github.com/aranya-project/aranya-core/pull/410 is merged

crates/aranya-daemon-api/src/service.rs
```diff
@@ -725,6 +737,46 @@ pub trait DaemonApi {
     /// Receive AQC ctrl message.
     async fn receive_aqc_ctrl(team: TeamId, ctrl: AqcCtrl) -> Result<(LabelId, AqcPsks)>;
 
+    /// Gets AFC shared-memory configuration info.
+    #[cfg(feature = "afc")]
```
gknopf-aranya, line 745: Confirming that you're able to only check for `afc` and not `preview` here because of:
```
#[cfg(all(feature = "afc", not(feature = "preview")))]
compile_error!(
    "AFC is currently a preview feature. Enable the 'preview' feature to opt into preview APIs."
);
```

crates/aranya-daemon/src/afc.rs
ssweetney-spok: I [added this](https://github.com/aranya-project/aranya/pull/463/commits/54db03865ef25ab910f44804e53983ca7c5cf1fd) for developer convenience since we don't want the example to rely on "aranya-fast-channels". I could remove it and add note in the README about having to unlink it manually

crates/aranya-daemon-api/src/service.rs
ssweetney-spok: Yea

crates/aranya-daemon/src/afc.rs
gknopf-aranya: I believe we can delete the shm file without importing the `aranya-fast-channels` create.
We wouldn't have access to the convenient methods from that crate.
cc: @jdygert-spok

elagergren-spideroak: Yes, you can. This whole method to unlink if we're testing or running an example is _super_ hacky and we should not add it.

elagergren-spideroak: Let's just get https://github.com/aranya-project/aranya-core/pull/410 merged and then keep the shared memory open indefinitely.

ssweetney-spok: Removed it

crates/aranya-daemon/src/afc.rs
ssweetney-spok: [Removed](https://github.com/aranya-project/aranya/pull/463/commits/3cc9ae08ee639cb5c103f2e04a0dfc20b0700875)

crates/aranya-daemon-api/src/service.rs
ssweetney-spok: What is the benefit of being able to update them independently? They are defined in internal crates so we don't have to worry about API stability

gknopf-aranya: That's a good point that we're not guaranteeing stability of the daemon IPC. In that case, we should use the same type for this so they don't get out of sync. It's possible the daemon config may need extra configuration than what needs to be sent over the IPC in the future, but that's not the case right now.

crates/aranya-daemon-api/Cargo.toml
```diff
@@ -8,13 +8,20 @@ license.workspace = true
...
+# Aranya Fast Channels
+afc = []
 
 [lints]
 workspace = true
 
 
 [dependencies]
 aranya-crypto = { workspace = true }
+aranya-fast-channels = { workspace = true, features = ["alloc", "posix"] }
```
jdygert-spok: Should be optional dep

crates/aranya-daemon/src/config.rs
ssweetney-spok: Considering going back to `String` since I'm making the "aranya-fast-channel" crate an optional dependency.

I could gate this type behind the "afc" feature instead but it feels kind of weird for the daemon config to change based on features 🤔

ssweetney-spok: Or maybe "aranya-daemon" always depends on "aranya-fast-channel"

jdygert-spok: I think gating it is fine.

Cargo.toml
```diff
@@ -105,3 +107,7 @@ allow-branch = ["main"]
...
```
djohnson-spideroak, line 111: PR LGTM, we will just need to update this after release and make sure everything is passing

crates/aranya-daemon/Cargo.toml
```diff
@@ -14,18 +14,25 @@ workspace = true
...
+# Preview features
+preview = ["aranya-daemon-api/preview"]
+# Aranya Fast Channels
+afc = ["aranya-daemon-api/afc", "dep:aranya-fast-channels"]
 
 [dependencies]
 aranya-daemon-api = { workspace = true }
 aranya-keygen = { workspace = true }
 
+aranya-afc-util = { workspace = true }
```
jdygert-spok, line 29: This should also be optional

jdygert-spok: Although we'd need to figure out the conditional policy for that.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -219,6 +221,23 @@ impl Daemon {
                 )
             };
 
+            #[cfg(feature = "afc")]
+            let afc = {
+                let Toggle::Enabled(afc_cfg) = &cfg.afc else {
+                    anyhow::bail!(
+                        "AFC is currently required, set `afc.enable = true` in daemon config."
+                    )
+                };
```
jdygert-spok, line 230: I think we should go ahead and make AFC optional, it can be on another PR though.

ssweetney-spok: Yea, I was wondering if it should be similar to how you implemented it for AQC on the other PR

Cargo.toml
ssweetney-spok: [Upgraded just the afc crates](https://github.com/aranya-project/aranya/pull/463/commits/b45db8650d21cf226f074c6b1642ba345db225d9). Unsure if there was gonna be another "chore PR" to upgrade the rest of the core crates

deny.toml
ssweetney-spok, line 1: Not sure why exactly this was necessary now since the rest of the aranay-core crates we were depending on should have the same license as the 2 crates, `aranya-fast-channels` and `aranya-afc-util`, that caused the cargo deny check fail

Edit: https://github.com/aranya-project/aranya/commit/9f3810c1b2ad2e4ad816d8c7dd25174ca5585c95#r166146333

Cargo.toml
jdygert-spok: Since it's just two and no changes needed, lets do it here.

crates/aranya-daemon/src/daemon.rs
ssweetney-spok: https://github.com/aranya-project/aranya/issues/490

jdygert-spok (approved):

crates/aranya-daemon/Cargo.toml
djohnson-spideroak: Does this need to be resolved on this PR?

examples/rust/src/main.rs
```diff
@@ -77,6 +77,7 @@ impl ClientCtx {
 
         let daemon = {
             let shm = format!("/shm_{}", user_name);
+            let _ = rustix::shm::unlink(&shm);
```
djohnson-spideroak, line 80: Do we not have another way to unlink? Or does this match how we do the other shm link/unlink operations?

ssweetney-spok: I believe this, `rustix`, crate and the unlink function in `aranya-fast-channels`crate  both call the same underlying `libc` function but this was brought in so the example doesn't have to use the `aranya-fast-channels` crate

crates/aranya-daemon/Cargo.toml
ssweetney-spok: If I'm understanding correctly, I think this would require much _larger_ changes like allowing for conditional compilation in the policy language

djohnson-spideroak (approved):

crates/aranya-daemon/src/afc.rs
ssweetney-spok: Should be OK for now but we can clean it up in the future. The fallible operations for AQC also use `anyhow::Result`.

crates/aranya-daemon/src/afc.rs
```diff
@@ -0,0 +1,307 @@
...
+    write: WriteState<C, Rng>,
+}
+
+impl<C> AfcShm<C>
+where
+    C: CipherSuite,
+{
+    fn new(cfg: AfcConfig) -> Result<Self> {
+        debug!("setting up afc shm write side: {:?}", cfg.shm_path);
+        let write = {
```
elagergren-spideroak, line 44: early return on the first success case instead of nesting so much

crates/aranya-daemon/src/api.rs
```diff
@@ -646,7 +666,7 @@ impl DaemonApi for Api {
             .context("did not find peer")?;
 
         let id = self.device_id()?;
-        let (ctrl, effects) = self
+        let (ctrl, effects): (_, _) = self
```
elagergren-spideroak, line 669: does the compiler really complain without `: (_, _)`? it's very obviously a binary tuple

crates/aranya-daemon/benches/metrics.rs
```diff
@@ -39,6 +39,8 @@ fn daemon_startup(bencher: divan::Bencher<'_, '_>) {
                     }),
                 },
                 aqc: Toggle::Enabled(AqcConfig {}),
+                #[cfg(feature = "afc")]
+                afc: Toggle::Disabled,
```
elagergren-spideroak, line 43: this breaks the benchmarks b/c Daemon::load will immediately exit

crates/aranya-daemon/src/config.rs
```diff
@@ -17,7 +19,6 @@ pub use toggle::Toggle;
 
 /// Options for configuring the daemon.
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-#[serde(deny_unknown_fields)]
```
elagergren-spideroak, line 20: intentional?

ssweetney-spok: Yea, because the `Config::afc` field is gated behind the "afc" feature. This makes it less annoying since you can re-use the same config without having to remove the field when the feature is disabled

crates/aranya-daemon/benches/metrics.rs
ssweetney-spok: Good catch. https://github.com/aranya-project/aranya/issues/490, should fix this. I'll enable it in the benchmarks instead then

crates/aranya-daemon/src/api.rs
ssweetney-spok: I had a similar question while reviewing. Removing it doesn't cause any compiler errors on my end so it may have been adding during debugging

crates/aranya-daemon/src/config.rs
jdygert-spok: I made the same change in my AQC PR but it's not ideal. Unsure how to handle it due to a few issues. I think we should address it on another PR soon.

ssweetney-spok: > I think we should address it on another PR soon.

Should the `Config::afc` and `Config::aqc` fields not be gated at all? That seems better but makes it tricky to use types that are gated behind features

crates/aranya-daemon/benches/metrics.rs
ssweetney-spok: https://github.com/aranya-project/aranya/pull/463/commits/4a958c4df8d0c1966909057ea8534a312ee1dd34

gknopf-aranya: @ssweetney-spok can this be marked as resolved?

crates/aranya-daemon/src/config.rs
gknopf-aranya: @ssweetney-spok, @jdygert-spok let's capture this on a backlog issue and leave the issue here so this thread can be resolved on the current PR.

gknopf-aranya (commented):
lgtm!

crates/aranya-daemon/src/config.rs
ssweetney-spok: https://github.com/aranya-project/aranya/issues/491

PR #464, opened 2025-08-19
Updates for compatibility with aranya-core/main

jdygert-spok:


gknopf-aranya:
@jdygert-spok `clippy` found a few things:
https://github.com/aranya-project/aranya/actions/runs/17080413178/job/48432635809?pr=464

gknopf-aranya:
I was expecting to see a change to `Cargo.toml` to pull in changes from `main` for `aranya-core` deps.

crates/aranya-daemon/src/policy.md
```diff
@@ -1587,8 +1587,8 @@ command DeleteLabel {
         let label = check_unwrap query Label[label_id: this.label_id]
 
         finish {
-            // Cascade deleting the label assignments.
-            delete AssignedLabel[label_id: label.label_id, device_id: ?]
+            // TODO: Cascade deleting the label assignments.
+            // delete AssignedLabel[label_id: label.label_id, device_id: ?]
```
ssweetney-spok, line 1591: Did the use of a bind marker here cause an error at runtime?

jdygert-spok: No, it was causing incorrect behavior since we don't have prefix deletion so the machine IO interface makes the storage think it is just deleting a single fact.

With https://github.com/aranya-project/aranya-core/pull/375 it will be a compile error.

ssweetney-spok: Oh, I expected there to be some mismatch in the stack values and instructions [since we don't put anything on the stack when we see a bind marker](https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-policy-compiler/src/compile.rs#L541-L562)

Edit: Nvm, I see we `continue` so the instruction below isn't added

jdygert-spok:
LGTM

crates/aranya-daemon/src/policy.md
djohnson-spideroak: Just want to confirm: this does remove the label, but not the assignments. Does this introduce any risk of a label-resurrection attack? I would imagine not due to the label ID being a command ID, but I want to be clear as to what this actually changes.

jdygert-spok: The deletion was already not working correctly, but I confirmed that we check for the label to exist before any uses of `AssignedLabel`. In theory if you recreated the label fact with the same label id, then those assignments would be live again. Maybe there's some way to do that with braiding? Not sure if you could get anything useful out of it. Either way, this PR doesn't introduce the issue.

Issue #442, opened 2025-08-05
example: split complex example into simple focused examples


djohnson-spideroak:
The goal of this is to take the example and split it into multiple examples that focus on specific common uses of aranya. Some possible examples:

- Onboarding
- AQC
- IDAM operations
- sync topologies

And included comments in the code explaining concepts, etc.

This ticket should focus on initial splitting of the example and ensuring that the examples continue to run in CI.

gknopf-aranya:
I wonder how we will be able to simplify the AQC example.
It depends on onboarding, IDAM, sync peers, labels, etc. to be setup in order to create and use channels.
We could have some test util code that sets up a default team in a single method, so the rest of the code is very AQC specific.

gknopf-aranya:
It would be convenient to have examples that are both automatically unit tested and compiled into stand-alone binaries.
If the examples are in a different `aranya-examples` crate, they are guaranteed to use the public `aranya-client` application API.
Each example could be part of a different module in that crate with its own `Cargo.toml`, `main.rs`, and `test.rs` so they can be unit tested in CI and compiled into stand-alone executables. Example directory structure:
`crates/aranya-examples`
    `aqc/Cargo.toml`
    `aqc/src/main.rs`
    `aqc/src/test.rs`

gknopf-aranya:
Reopening since only part of this issue has been resolved by the merged PR.

gknopf-aranya:
Unassigning this from myself since this is too large of a task to handle in a sprint. Will move smaller issues into sprints as we have time.

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok, @jnewell-spok we're thinking of adding an example of onboarding a single device next. The team owner and device added should be implemented in 2 different executables to show how onboarding info will be sent via the network. We can borrow the executable build structure and TCP networking code from the multi-node example for this.

PR #473, opened 2025-09-02
AQC: Add feature flag

### LINKED ISSUE #470: aqc: mark experimental
AQC should be marked as experimental in preparation for our move to transportless data plane.

cc @gknopf-aranya  @jdygert-spok @ycarmy

jdygert-spok:
Closes #470.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ wildcard_imports = "warn"
 aranya-aqc-util = { version = "0.7.0", features = ["alloc"] }
 aranya-capi-core = { version = "0.4.4" }
 aranya-capi-codegen = { version = "0.3.3" }
-aranya-crypto = { version = "0.9.0", features = ["alloc", "aqc", "fs-keystore", "clone-aead", "std", "tls"] }
+aranya-crypto = { version = "0.9.0", features = ["alloc", "fs-keystore", "clone-aead", "std", "tls"] }
 aranya-crypto-ffi = { version = "0.12.0" }
 aranya-device-ffi = { version = "0.12.0" }
 aranya-envelope-ffi = { version = "0.12.0" }
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -65,7 +65,7 @@ category = "build"
 description = "Build C API shared library"
 cwd = "crates/aranya-client-capi"
 command = "cargo"
-args = ["build", "--release"]
+args = ["build", "--release", "--features=aqc"]
 dependencies = ["install-nightly"]
 
 [tasks.tgz-capi]
@@ -192,7 +192,7 @@ workspace = true
 script = '''
 cargo clippy
 cargo clippy --all-targets
-# cargo clippy --no-default-features (TODO: uncomment in #448) 
+# cargo clippy --no-default-features (TODO: uncomment in #448)
 cargo clippy --all-targets --all-features
 '''
 install_crate = { rustup_component_name = "clippy" }
--- a/crates/aranya-client-capi/Cargo.toml
+++ b/crates/aranya-client-capi/Cargo.toml
@@ -21,6 +21,8 @@ doctest = false
 [features]
 default = []
 
+aqc = ["aranya-client/aqc"]
+
 
 [dependencies]
 aranya-client = { workspace = true }
--- a/crates/aranya-client-capi/cbindgen.toml
+++ b/crates/aranya-client-capi/cbindgen.toml
@@ -96,6 +96,7 @@ style = "both"
 "target_os = linux" = "__linux__"
 "target_os = vxworks" = "__VXWORKS__"
 "target_arch = x86_64" = "__x86_64__"
+"feature = aqc" = "ENABLE_AQC"
 
 [export]
 renaming_overrides_prefixing = true
@@ -124,7 +125,7 @@ clean = true
 
 [parse.expand]
 default_features = false
-features = []
+features = ["aqc"]
 crates = ["aranya-client-capi"]
 
 [ptr]
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -3,17 +3,23 @@ use core::{
     ffi::{c_char, CStr},
     ptr, slice,
 };
-use std::{ffi::OsStr, ops::Deref, os::unix::ffi::OsStrExt, str::FromStr};
+#[cfg(feature = "aqc")]
+use std::str::FromStr;
+use std::{ffi::OsStr, ops::Deref, os::unix::ffi::OsStrExt};
 
 use anyhow::Context as _;
-use aranya_capi_core::{opaque::Opaque, prelude::*, ErrorCode, InvalidArg};
+#[cfg(feature = "aqc")]
+use aranya_capi_core::opaque::Opaque;
+use aranya_capi_core::{prelude::*, ErrorCode, InvalidArg};
+#[cfg(feature = "aqc")]
 use aranya_client::aqc::{self, AqcPeerStream};
 use aranya_daemon_api::Text;
 use aranya_util::error::ReportExt as _;
+#[cfg(feature = "aqc")]
 use bytes::Bytes;
 use tracing::{debug, error};
 
-use crate::imp::{self, aqc::consume_bytes};
+use crate::imp;
 
 /// An error code.
 ///
@@ -59,6 +65,7 @@ pub enum Error {
     Aranya,
 
     /// AQC library error.
+    #[cfg(feature = "aqc")]
     #[capi(msg = "AQC library error")]
     Aqc,
 
@@ -96,6 +103,7 @@ impl From<&imp::Error> for Error {
             imp::Error::Client(err) => match err {
                 aranya_client::Error::Ipc(_) => Self::Ipc,
                 aranya_client::Error::Aranya(_) => Self::Aranya,
+                #[cfg(feature = "aqc")]
                 aranya_client::Error::Aqc(_) => Self::Aqc,
                 aranya_client::Error::Bug(_) => Self::Bug,
                 aranya_client::Error::Config(_) => Self::Config,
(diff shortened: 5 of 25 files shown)


jdygert-spok:
I think I'm hitting https://github.com/aranya-project/aranya-core/issues/179 now.

jdygert-spok:
Needs aranya-project/aranya-core#393 but ready for some review.

I only added an aqc feature and not an experimental flag. Unsure how we want to shape that (cargo feature vs bare cfg etc).

The rust/c examples and the model benchmarks all enable AQC unconditionally.

jdygert-spok:
Discussed on rust call, will add `cfg(aranya_unstable)`.

jdygert-spok:
I tried out the cfg but it was painful to manage. It also triggers a full rebuild. Going to stick with the feature flag.

Makefile.toml
gknopf-aranya, line 1: The `metrics` task requires the `aqc` feature

crates/aranya-client-capi/cbindgen.toml
```diff
@@ -96,6 +96,7 @@ style = "both"
 "target_os = linux" = "__linux__"
 "target_os = vxworks" = "__VXWORKS__"
 "target_arch = x86_64" = "__x86_64__"
+"feature = aqc" = "ENABLE_AQC"
```
gknopf-aranya, line 103: Do we want an `ENABLE_UNSTABLE` flag that gates all the unstable features for the C API?

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya, line 1: At some point we may want to consider adding support for multiple `defs.rs` modules so features can each go into a different files and be hidden with a single `#[cfg(feature = "aqc")]`

crates/aranya-client/src/client.rs
```diff
@@ -76,26 +79,113 @@ impl ClientBuilder<'_> {
         Self::default()
     }
 
-    /// Connects to the daemon.
+    /// Creates a client connection to the daemon.
```
gknopf-aranya, line 82: I thought this doc example was already added in by one of my previous PRs

gknopf-aranya: Nevermind, this was moved from another spot

crates/aranya-daemon-api/src/service.rs
```diff
@@ -630,6 +268,19 @@ pub struct Label {
     pub name: Text,
 }
 
+// TODO: fix tarpc
```
gknopf-aranya: What needs to be fixed specifically?

crates/aranya-daemon/src/api.rs
```diff
@@ -544,6 +573,7 @@ impl DaemonApi for Api {
         Ok(())
     }
 
+    #[cfg(feature = "aqc")]
```
gknopf-aranya, line 605: We used to have to implement an empty method that does nothing because the IPC API trait needed an implementation for each method in the trait. How did you work around this?

examples/c/example.bash
```diff
@@ -63,7 +63,7 @@ EOF
 done
 
 # build the daemon.
-cargo build --bin aranya-daemon --release
+cargo build -p aranya-daemon --bin aranya-daemon --features unstable,aqc --release
```
gknopf-aranya: As we start to add more features, we'll probably want a way to return a list of the default features they can be passed into the examples automatically. Not sure if we need a CLI for this or if `cargo` can do that.

jdygert-spok: We can make it a default feature in `Cargo.toml` and not have to pass the flags. But for AQC I don't think we want it as a default now since we are de-emphasizing.

crates/aranya-daemon-api/src/service.rs
jdygert-spok: The return type of cfg'd methods gets used unconditionally in some generated code, which is why I need to define those types here.

crates/aranya-daemon/src/api.rs
jdygert-spok: The trait and impl are both cfg'd. What was the issue before? IIRC we did it that way to get slightly nicer error messages.

crates/aranya-client-capi/cbindgen.toml
jdygert-spok: Added check for `ENABLE_ARANYA_UNSTABLE`

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Agreed, I had tried having it in a separate module but our capi tooling doesn't recurse down modules.

examples/c/example.bash
djohnson-spideroak: I agree we will not want it enabled by default. I am on the fence for the current moment though. I would like to avoid having no data plane enabled by default, so we may consider having it default until AFC is added back in.

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: Do we have an issue for this?

crates/aranya-daemon/src/api.rs
```diff
@@ -201,16 +224,21 @@ impl EffectHandler {
                 LabelDeleted(_) => {}
                 LabelAssigned(_) => {}
                 LabelRevoked(_) => {}
-                AqcNetworkNameSet(e) => {
+                AqcNetworkNameSet(_e) => {
```
djohnson-spideroak: Thoughts on some logging if a client receives AQC effects but the feature is not enabled?

djohnson-spideroak (approved):

crates/aranya-daemon/src/api.rs
gknopf-aranya: If I add the same feature flag config to a method in `service.rs` and `api.rs` it does not compile. Not sure why.

Makefile.toml
```diff
@@ -40,14 +47,17 @@ run_task = { name = [
...
 # Build the `aranya-client-capi` C API `aranya-client.h` header file.
+# NOTE: From 1.87 onward, cargo-expand doesn't include cfg's, so cbindgen won't
+# see them to include in the header. Until this is fixed, we cannot increase
+# our MSRV beyond 1.86. See https://github.com/rust-lang/rust/issues/139715.
 [tasks.build-capi-header]
 category = "build"
 description = "Build C API header"
-toolchain = "${NIGHTLY}"
 cwd = "crates/aranya-client-capi"
+env.RUSTC_BOOTSTRAP = "1"
```
nikki-aranya, line 58: can you add a note about what this is needed for

crates/aranya-client-capi/Cargo.toml
```diff
@@ -21,6 +21,11 @@ doctest = false
 [features]
 default = []
 
+experimental = [
+    "aranya-client/experimental",
```
gknopf-aranya: What is the purpose of defining features with paths instead of calling them "experimental" and "aqc"?

Makefile.toml
nikki-aranya: lgtm otherwise

examples/rust/run.bash
gknopf-aranya, line 1: @djohnson-spideroak, @jdygert-spok should we put the AQC example behind a feature flag in the Rust example?
That way the example can be compiled with different feature flags enabled depending on what users want to see an example of.

examples/c/CMakeLists.txt
gknopf-aranya, line 1: We may want to put the AQC C example behind a compiler flag so users can build the example with and without AQC enabled. Currently, the example assumes that AQC is always enabled.

examples/rust/run.bash
djohnson-spideroak: I think we can do this in a followup. We still want to eventually split the example into smaller more focused examples either way. That would be easier to deal with in terms of features.

djohnson-spideroak: https://github.com/aranya-project/aranya/issues/442 is the issue for splitting the examples

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: https://github.com/aranya-project/aranya-core/issues/401

crates/aranya-client-capi/Cargo.toml
jdygert-spok: This is just enabling the feature in aranya-client when the same feature in aranya-client-capi is enabled.

examples/c/CMakeLists.txt
djohnson-spideroak: https://github.com/aranya-project/aranya/issues/482

Cargo.toml
```diff
@@ -105,3 +106,6 @@ allow-branch = ["main"]
 shared-version = true
 tag-message = ""
 tag-prefix = ""
+
+[patch.crates-io]
+aranya-capi-codegen = { git = "https://github.com/aranya-project/aranya-core.git", branch = "main" }
```
djohnson-spideroak: I am guessing we need to do a release of this crate?

djohnson-spideroak (approved):

gknopf-aranya (approved):

crates/aranya-daemon/src/api.rs
gknopf-aranya: This was because I needed the same feature set enabled for the api and daemon crates.

crates/aranya-daemon/src/daemon.rs
ssweetney-spok, line 1: Should these lines:

```rust
// TODO: Make aqc optional.
let Toggle::Enabled(_) = &cfg.aqc else {
    anyhow::bail!(
        "AQC is currently required, set `aqc.enable = true` in daemon config."
    )
};
```

get removed or should it happen in one of the AFC PRs?

jdygert-spok: Good catch, just made it optional and added cfg feature to the config

crates/aranya-client/src/client.rs
```diff
@@ -76,26 +79,113 @@ impl ClientBuilder<'_> {
...
+                return Err(IpcError::new(io::Error::new(
+                    io::ErrorKind::Unsupported,
+                    format!("version mismatch: `{got}` != `{want}`"),
+                ))
+                .into());
+            }
+            debug!(client = ?want, daemon = ?got, "versions");
+
+            #[cfg(feature = "aqc")]
+            let aqc = {
```
jdygert-spok: Probably need to make the client parts optional too.

PR #489, opened 2025-09-17
afc: C API Bindings

nikki-aranya:
- [x] create_channel()
- [x] recv_ctrl()
- [x] get_channel_type()
- [x] open()
- [x] seal()
- [ ] `seq_cmp(seq1, seq2) -> int` to compare `Seq`
- [x] delete_channel()
- [x] Feature flags for afc, preview
- [x] label_id()
- [ ] Update C example with AFC example
- [ ] unlink afc shm files in example.c https://github.com/aranya-project/aranya/issues/494

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -11,6 +11,8 @@ use anyhow::Context as _;
 #[cfg(feature = "aqc")]
 use aranya_capi_core::opaque::Opaque;
 use aranya_capi_core::{prelude::*, ErrorCode, InvalidArg};
+#[cfg(feature = "afc")]
+use aranya_client::afc;
 #[cfg(feature = "aqc")]
 use aranya_client::aqc::{self, AqcPeerStream};
 use aranya_daemon_api::Text;
@@ -68,6 +70,10 @@ pub enum Error {
     #[capi(msg = "Aranya error")]
     Aranya,
 
+    #[cfg(feature = "afc")]
+    #[capi(msg = "wrong channel type provided")]
+    WrongChannelType,
+
     /// AQC library error.
     #[cfg(feature = "aqc")]
     #[capi(msg = "AQC library error")]
@@ -118,6 +124,8 @@ impl From<&imp::Error> for Error {
                     Self::Bug
                 }
             },
+            #[cfg(feature = "afc")]
+            imp::Error::WrongChannelType => Self::WrongChannelType,
             imp::Error::WouldBlock => Self::WouldBlock,
             imp::Error::Closed => Self::Closed,
             imp::Error::Config(_) => Self::Config,
@@ -2258,4 +2266,289 @@ pub unsafe fn aqc_recv_stream_try_recv(
     }
     *buffer_len = written;
     Ok(())
+}
+
+/// An AFC Channel Object.
+///
+/// This can contain a bidirectional, receiver, or sender channel.
+///
+/// Note that only bidi and sender channels can call `aranya_afc_seal()`, and
+/// only bidi and receiver channels can call `aranya_afc_open()`.
+#[cfg(feature = "afc")]
+#[aranya_capi_core::derive(Cleanup)]
+#[aranya_capi_core::opaque(size = 96, align = 8)]
+pub type AfcChannel = Safe<imp::AfcChannel>;
+
+/// An enum containing all [`AfcChannel`] variants.
+#[cfg(feature = "afc")]
+#[derive(Copy, Clone, Debug)]
+#[repr(u8)]
+pub enum AfcChannelType {
+    Bidirectional,
+    Sender,
+    Receiver,
+}
+
+/// An AFC Control Message, used to create the other end of a channel.
+#[cfg(feature = "afc")]
+#[aranya_capi_core::derive(Cleanup)]
+#[aranya_capi_core::opaque(size = 32, align = 8)]
+pub type AfcCtrlMsg = Safe<imp::AfcCtrlMsg>;
+
+/// An AFC Sequence Number, for reordering messages.
+#[cfg(feature = "afc")]
+#[aranya_capi_core::derive(Cleanup)]
+#[aranya_capi_core::opaque(size = 24, align = 8)]
+pub type AfcSeq = Safe<imp::AfcSeq>;
+
+/// Create a bidirectional AFC channel between this device and a peer.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  team_id the team's identifier [`TeamId`].
+/// @param[in]  peer_id the peer's identifier [`DeviceId`].
+/// @param[in]  label_id the label identifier [`LabelId`] to create the channel with.
+/// @param[out] channel the AFC channel object [`AfcChannel`].
+/// @param[out] control the AFC control message [`AfcCtrlMsg`]
+///
+/// @relates AranyaClient.
+#[cfg(feature = "afc")]
+pub fn afc_create_bidi_channel(
+    client: &Client,
+    team_id: &TeamId,
+    peer_id: &DeviceId,
+    label_id: &LabelId,
+    channel: &mut MaybeUninit<AfcChannel>,
+    control: &mut MaybeUninit<AfcCtrlMsg>,
+) -> Result<(), imp::Error> {
+    let (chan, ctrl) = client.rt.block_on(client.inner.afc().create_bidi_channel(
+        team_id.into(),
+        peer_id.into(),
+        label_id.into(),
+    ))?;
+
+    AfcChannel::init(channel, chan.into());
+    AfcCtrlMsg::init(control, ctrl.into());
+    Ok(())
+}
+
+/// Create a send-only AFC channel between this device and a peer.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  team_id the team's identifier [`TeamId`].
+/// @param[in]  peer_id the peer's identifier [`DeviceId`].
+/// @param[in]  label_id the label identifier [`LabelId`] to create the channel with.
+/// @param[out] channel the AFC channel object [`AfcChannel`].
+/// @param[out] control the AFC control message [`AfcCtrlMsg`]
+///
+/// @relates AranyaClient.
+#[cfg(feature = "afc")]
+pub fn afc_create_uni_send_channel(
+    client: &Client,
+    team_id: &TeamId,
+    peer_id: &DeviceId,
+    label_id: &LabelId,
+    channel: &mut MaybeUninit<AfcChannel>,
+    control: &mut MaybeUninit<AfcCtrlMsg>,
+) -> Result<(), imp::Error> {
+    let (chan, ctrl) = client
+        .rt
+        .block_on(client.inner.afc().create_uni_send_channel(
+            team_id.into(),
+            peer_id.into(),
+            label_id.into(),
+        ))?;
+
+    AfcChannel::init(channel, chan.into());
+    AfcCtrlMsg::init(control, ctrl.into());
+    Ok(())
+}
+
+/// Create a receive-only AFC channel between this device and a peer.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  team_id the team's identifier [`TeamId`].
+/// @param[in]  peer_id the peer's identifier [`DeviceId`].
+/// @param[in]  label_id the label identifier [`LabelId`] to create the channel with.
+/// @param[out] channel the AFC channel object [`AfcChannel`].
+/// @param[out] control the AFC control message [`AfcCtrlMsg`]
+///
+/// @relates AranyaClient.
+#[cfg(feature = "afc")]
+pub fn afc_create_uni_recv_channel(
+    client: &Client,
+    team_id: &TeamId,
+    peer_id: &DeviceId,
+    label_id: &LabelId,
+    channel: &mut MaybeUninit<AfcChannel>,
+    control: &mut MaybeUninit<AfcCtrlMsg>,
+) -> Result<(), imp::Error> {
+    let (chan, ctrl) = client
+        .rt
+        .block_on(client.inner.afc().create_uni_recv_channel(
+            team_id.into(),
+            peer_id.into(),
+            label_id.into(),
+        ))?;
+
+    AfcChannel::init(channel, chan.into());
+    AfcCtrlMsg::init(control, ctrl.into());
+    Ok(())
+}
+
+/// Use an ephemeral command to create an AFC channel between this device and a peer.
+///
+/// Note that this function takes ownership of the [`AfcCtrlMsg`] and invalidates any further use.
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  team_id the team's identifier [`TeamId`].
+/// @param[in]  control the AFC control message [`AfcCtrlMsg`].
+/// @param[out] channel the AFC channel object [`AfcChannel`].
+/// @param[out] __output the corresponding AFC channel type [`AfcChannelType`].
+///
+/// @relates AranyaClient.
+#[cfg(feature = "afc")]
+pub fn afc_recv_ctrl(
+    client: &Client,
+    team_id: &TeamId,
+    control: OwnedPtr<AfcCtrlMsg>,
+    channel: &mut MaybeUninit<AfcChannel>,
+) -> Result<AfcChannelType, imp::Error> {
+    // SAFETY: the user is responsible for passing in a valid `AfcCtrlMsg` pointer.
+    let ctrl = unsafe { Opaque::into_inner(control.read()).into_inner().0 };
+    let chan = client
+        .rt
+        .block_on(client.inner.afc().recv_ctrl(team_id.into(), ctrl))?;
+
+    let channel_type = match chan {
+        afc::Channel::Bidi(_) => AfcChannelType::Bidirectional,
+        afc::Channel::Send(_) => AfcChannelType::Sender,
+        afc::Channel::Recv(_) => AfcChannelType::Receiver,
+    };
+    AfcChannel::init(channel, chan.into());
+    Ok(channel_type)
+}
+
+/// Returns the [`AfcChannelType`] for a given [`AfcChannel`] container.
+///
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[out] __output the corresponding AFC channel type [`AfcChannelType`].
+#[cfg(feature = "afc")]
+pub fn afc_get_channel_type(channel: &AfcChannel) -> AfcChannelType {
+    match channel.deref().deref() {
+        imp::AfcChannel::Bidi(_) => AfcChannelType::Bidirectional,
+        imp::AfcChannel::Send(_) => AfcChannelType::Sender,
+        imp::AfcChannel::Receive(_) => AfcChannelType::Receiver,
+    }
+}
+
+/// Returns the [`LabelId`] for the associated [`AfcChannel`].
+///
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[out] __output the corresponding label ID [`LabelId`].
+#[cfg(feature = "afc")]
+pub fn afc_get_label_id(channel: &AfcChannel) -> LabelId {
+    match &channel.deref().deref() {
+        imp::AfcChannel::Bidi(c) => c.label_id().into(),
+        imp::AfcChannel::Send(c) => c.label_id().into(),
+        imp::AfcChannel::Receive(c) => c.label_id().into(),
+    }
+}
+
+/// Returns the raw data for a given [`AfcCtrlMsg`].
+///
+/// Note that the lifetime of the pointer is tied to the [`AfcCtrlMsg`].
+///
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[out] ptr the raw pointer of the stored buffer.
+/// @param[out] len the raw length of the stored buffer.
+#[cfg(feature = "afc")]
+pub fn afc_get_msg_data(
+    control: &AfcCtrlMsg,
+    ptr: &mut MaybeUninit<*const u8>,
+    len: &mut MaybeUninit<usize>,
+) {
+    let slice = control.0.as_bytes();
+    ptr.write(slice.as_ptr());
+    len.write(slice.len());
+}
+
+/// Returns the size of the overhead needed for a channel message.
+#[cfg(feature = "afc")]
+pub fn afc_channel_overhead() -> usize {
+    afc::Channels::OVERHEAD
+}
+
+/// Encrypts and authenticates `plaintext`, and writes it to `dst`.
+///
+/// Note that `dst` must be at least `plaintext.len()` + `aranya_afc_channel_overhead()`.
+///
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[in]  plaintext the message being encrypted.
+/// @param[out] dst the output buffer the ciphertext is written to.
+#[cfg(feature = "afc")]
+pub fn afc_channel_seal(
+    channel: &AfcChannel,
+    plaintext: &[u8],
+    dst: &mut [u8],
+) -> Result<(), imp::Error> {
+    match &channel.deref().deref() {
+        imp::AfcChannel::Bidi(c) => c.seal(dst, plaintext)?,
+        imp::AfcChannel::Send(c) => c.seal(dst, plaintext)?,
+        imp::AfcChannel::Receive(_) => return Err(imp::Error::WrongChannelType),
+    }
+
+    Ok(())
+}
+
+/// Decrypts and authenticates `ciphertext`, and writes it to `dst`.
+///
+/// Note that `dst` must be at least `ciphertext.len()` - `aranya_afc_channel_overhead()`.
+///
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[in]  ciphertext the message being decrypted.
+/// @param[out] dst the output buffer the message is written to.
+/// @param[out] seq the sequence number for the opened message, for reordering.
+#[cfg(feature = "afc")]
+pub fn afc_channel_open(
+    channel: &AfcChannel,
+    ciphertext: &[u8],
+    dst: &mut [u8],
+    seq: &mut MaybeUninit<AfcSeq>,
+) -> Result<(), imp::Error> {
+    let seq_raw = match &channel.deref().deref() {
+        imp::AfcChannel::Bidi(c) => c.open(dst, ciphertext)?,
+        imp::AfcChannel::Receive(c) => c.open(dst, ciphertext)?,
+        imp::AfcChannel::Send(_) => return Err(imp::Error::WrongChannelType),
+    };
+    AfcSeq::init(seq, seq_raw.into());
+
+    Ok(())
+}
+
+/// Removes an [`AfcChannel`] from use.
+///
+/// Note that this function takes ownership of the [`AfcChannel`] and invalidates
+/// any further use (i.e. calling open/seal).
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+///
+/// @relates AranyaClient.
+#[cfg(feature = "afc")]
+pub fn afc_channel_delete(
+    client: &Client,
+    channel: OwnedPtr<AfcChannel>,
+) -> Result<(), imp::Error> {
+    // SAFETY: the user is responsible for passing in a valid `AfcChannel` pointer.
+    match unsafe { channel.read().into_inner().into_inner() } {
+        imp::AfcChannel::Bidi(c) => Ok(client.rt.block_on(c.delete())?),
+        imp::AfcChannel::Receive(c) => Ok(client.rt.block_on(c.delete())?),
+        imp::AfcChannel::Send(c) => Ok(client.rt.block_on(c.delete())?),
+    }
 }
--- a/crates/aranya-client-capi/src/imp/afc.rs
+++ b/crates/aranya-client-capi/src/imp/afc.rs
@@ -1,0 +1,75 @@
+//! AFC-specific utilities for the Aranya Client C API.
+#![cfg(feature = "afc")]
+
+use aranya_capi_core::safe::{TypeId, Typed};
+use aranya_client::afc;
+
+/// All channel variants.
+#[derive(Debug)]
+pub enum AfcChannel {
+    Bidi(afc::BidiChannel),
+    Send(afc::SendChannel),
+    Receive(afc::ReceiveChannel),
+}
+
+impl Typed for AfcChannel {
+    const TYPE_ID: TypeId = TypeId::new(0xDC3130B2);
+}
+
+impl From<afc::Channel> for AfcChannel {
+    fn from(channel: afc::Channel) -> Self {
+        match channel {
+            afc::Channel::Bidi(c) => Self::Bidi(c),
+            afc::Channel::Send(c) => Self::Send(c),
+            afc::Channel::Recv(c) => Self::Receive(c),
+        }
+    }
+}
+
+impl From<afc::BidiChannel> for AfcChannel {
+    fn from(channel: afc::BidiChannel) -> Self {
+        Self::Bidi(channel)
+    }
+}
+
+impl From<afc::SendChannel> for AfcChannel {
+    fn from(channel: afc::SendChannel) -> Self {
+        Self::Send(channel)
+    }
+}
+
+impl From<afc::ReceiveChannel> for AfcChannel {
+    fn from(channel: afc::ReceiveChannel) -> Self {
+        Self::Receive(channel)
+    }
+}
+
+/// A control message, for creating the other end of a channel.
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct AfcCtrlMsg(pub(crate) afc::CtrlMsg);
+
+impl Typed for AfcCtrlMsg {
+    const TYPE_ID: TypeId = TypeId::new(0xB421D1CE);
+}
+
+impl From<afc::CtrlMsg> for AfcCtrlMsg {
+    fn from(msg: afc::CtrlMsg) -> Self {
+        Self(msg)
+    }
+}
+
+/// A sequence number, for reordering messages.
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct AfcSeq(afc::Seq);
+
+impl Typed for AfcSeq {
+    const TYPE_ID: TypeId = TypeId::new(0xC4DCE0C0);
+}
+
+impl From<afc::Seq> for AfcSeq {
+    fn from(seq: afc::Seq) -> Self {
+        Self(seq)
+    }
+}
--- a/crates/aranya-client-capi/src/imp/mod.rs
+++ b/crates/aranya-client-capi/src/imp/mod.rs
@@ -1,8 +1,11 @@
+pub mod afc;
 pub mod aqc;
 pub mod client;
 pub mod config;
 pub mod error;
 
+#[cfg(feature = "afc")]
+pub use afc::*;
 #[cfg(feature = "aqc")]
 pub use aqc::*;
 pub use client::*;
--- a/crates/aranya-client-capi/src/imp/error.rs
+++ b/crates/aranya-client-capi/src/imp/error.rs
@@ -4,6 +4,8 @@ use aranya_capi_core::{
     safe::{TypeId, Typed},
     write_c_str, ExtendedError, InvalidArg, WriteCStrError,
 };
+#[cfg(feature = "afc")]
+use aranya_client::afc;
 #[cfg(feature = "aqc")]
 use aranya_client::{aqc::TryReceiveError, error::AqcError};
 use buggy::Bug;
@@ -33,6 +35,10 @@ pub enum Error {
     #[error("connection was unexpectedly closed")]
     Closed,
 
+    #[cfg(feature = "afc")]
+    #[error("wrong channel type provided")]
+    WrongChannelType,
+
     #[error(transparent)]
     Utf8(#[from] core::str::Utf8Error),
 
@@ -50,6 +56,13 @@ pub enum Error {
 
     #[error(transparent)]
     Other(#[from] anyhow::Error),
+}
+
+#[cfg(feature = "afc")]
+impl From<afc::Error> for Error {
+    fn from(value: afc::Error) -> Self {
+        Self::Client(aranya_client::Error::Afc(value))
+    }
 }
 
 #[cfg(feature = "aqc")]


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2150,3 +2122,277 @@ pub unsafe fn aqc_recv_stream_try_recv(
     *buffer_len = written;
     Ok(())
 }
+
+/// An AFC Bidirectional Channel Object.
```
gknopf-aranya: For each channel type, we should mention which `open()` `seal()` operations it can perform.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2150,3 +2122,277 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+pub type AfcChannel = Safe<imp::AfcChannel>;
+
+/// An AFC Control Message.
+#[aranya_capi_core::derive(Cleanup)]
+#[aranya_capi_core::opaque(size = 40, align = 8)]
+pub type AfcCtrl = Safe<imp::AfcCtrl>;
+
+/// Create a bidirectional AFC channel between this device and a peer.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
```
gknopf-aranya, line 2296: I've added more docs to the Rust API regarding what to do with the ctrl message and the type of channel the peer will receive once it processes it. We should include those here for user's of the C API.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2150,3 +2122,277 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+/// @param[out] control the AFC control message [`AfcCtrl`]
+///
+/// @relates AranyaClient.
+pub fn afc_create_bidi_channel(
+    client: &Client,
+    team_id: &TeamId,
+    peer_id: &DeviceId,
+    label_id: &LabelId,
+    channel: &mut MaybeUninit<AfcBidiChannel>,
+    control: &mut MaybeUninit<AfcCtrl>,
```
gknopf-aranya: I wonder if we should group these fields into a single field with `MaybeUninit`. Because either neither one should be initialized or both should be initialized.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2150,3 +2122,277 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+        ))?;
+
+    AfcReceiveChannel::init(channel, imp::AfcReceiveChannel::new(chan));
+    AfcCtrl::init(control, imp::AfcCtrl::new(ctrl));
+    Ok(())
+}
+
+/// Use an ephemeral command to create an AFC channel between this device and a peer.
+///
+/// Note that the [`AfcChannel`] must be converted before it can be used:
```
gknopf-aranya: An alternative to this could be to always return the generic `AfcChannel` type via the C API and use that as a parameter to all afc methods. Then internally, we use a Rust match to determine the specific channel type and whether we can perform open/seal operations on that channel type. We can still provide the channel type to the user, but they can use it to determine whether to call open/seal instead of needing to get a new more specific channel type later.

crates/aranya-client-capi/src/imp/afc.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/// An AQC stream that can both send and receive data.
+#[derive(Debug)]
+pub struct AqcBidiStream {
+    pub(crate) inner: aqc::AqcBidiStream,
+    pub(crate) data: Bytes,
+}
+
+impl Typed for AqcBidiStream {
```
gknopf-aranya: We won't have streams for AFC so that code can be removed

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I think this may be a simpler API for the application to use. We'd probably want to add a `get_channel_type(channel) -> channel_type` so the channel type can be obtained directly from the channel object.

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: I can't use a tuple since it's not repr(C), my choices are either a wrapper struct or leave it as-is

gknopf-aranya: @jdygert-spok do you have a preference for this?

jdygert-spok: Looking at it from the C API it makes more sense to me to have as two separate out params.

Also, should control just be an out buf?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -58,10 +58,19 @@ pub enum Error {
...
     Aranya,
 
+    /// Unable to create configuration info.
+    #[capi(msg = "invalid config")]
+    Config,
+
     /// AQC library error.
+    #[cfg(feature = "afc")]
     #[capi(msg = "AQC library error")]
     Aqc,
```
jdygert-spok, line 80: Wrong one

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -557,7 +569,6 @@ pub unsafe fn get_key_bundle(
     keybundle: *mut MaybeUninit<u8>,
     keybundle_len: &mut usize,
 ) -> Result<(), imp::Error> {
-    let client = client.imp();
```
jdygert-spok: neat, don't need this now that we are using shared refs.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2178,3 +2157,267 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+    Bidirectional,
+    Sender,
+    Receiver,
+}
+
+/// An AFC Control Message.
+#[cfg(feature = "afc")]
+#[aranya_capi_core::derive(Cleanup)]
+#[aranya_capi_core::opaque(size = 40, align = 8)]
+pub type AfcCtrl = Safe<imp::AfcCtrl>;
```
jdygert-spok: Need a way to access the bytes.

crates/aranya-client-capi/src/imp/afc.rs
```diff
@@ -0,0 +1,67 @@
...
+    Bidi(afc::BidiChannel),
+    Send(afc::SendChannel),
+    Receive(afc::ReceiveChannel),
+}
+
+/// An AFC channel.
+#[derive(Debug)]
+pub struct AfcChannel {
+    pub(crate) inner: ChannelType,
+}
```
jdygert-spok: Could simplify to just the enum, the rust api of aranya-client-capi doesn't need to be encapsulated since it isn't published.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Hard to know the size of buffer needed though :thinking:

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2178,3 +2157,267 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+///
+/// Note that `dst` must be at least `plaintext.len()` + `aranya_afc_channel_overhead()`.
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[in]  plaintext the message being encrypted.
+/// @param[out] dst the output buffer the ciphertext is written to.
+#[cfg(feature = "afc")]
+pub fn afc_seal(
+    client: &Client,
```
jdygert-spok: We could remove the client from seal/open if we make the methods non-async. Looks like they are only async for the mutex.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Should we provide a method in the C API that serializes the `AfcCtrl` object?

jdygert-spok: No, it is already a serialized command. I'm thinking the recieve ctrl function should just take `&[u8]`. The functions which return the ctrl are trickier though, since it is hard to know the size needed.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -70,9 +79,9 @@ pub enum Error {
     #[capi(msg = "connection got closed")]
     Closed,
 
-    /// Unable to create configuration info.
-    #[capi(msg = "invalid config")]
-    Config,
+    /// Tried to call a function with the wrong channel (i.e. seal on a ReceiveChannel).
+    #[capi(msg = "wrong channel type provided")]
+    WrongChannelType,
```
gknopf-aranya: Gate behind afc feature flag

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -58,10 +58,19 @@ pub enum Error {
...
+    Config,
+
     /// AQC library error.
+    #[cfg(feature = "afc")]
     #[capi(msg = "AQC library error")]
     Aqc,
 
+    /// AFC library error.
+    #[capi(msg = "AFC library error")]
+    Afc,
```
gknopf-aranya: Gate behind afc feature flag

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I made this change to naming on the client PR:
```suggestion
pub type AfcCtrlMsg = Safe<imp::AfcCtrlMsg>;
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2178,3 +2157,267 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+    Receiver,
+}
+
+/// An AFC Control Message.
+#[cfg(feature = "afc")]
+#[aranya_capi_core::derive(Cleanup)]
+#[aranya_capi_core::opaque(size = 40, align = 8)]
+pub type AfcCtrl = Safe<imp::AfcCtrl>;
+
+/// An AFC Sequence Number, for reordering incoming messages.
```
gknopf-aranya: May want to mention how to compare `AfcSeq` with `seq_cmp()` here

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2178,3 +2157,267 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+    channel: &mut MaybeUninit<AfcChannel>,
+    control: &mut MaybeUninit<AfcCtrl>,
+) -> Result<(), imp::Error> {
+    let (chan, ctrl) = client.rt.block_on(client.inner.afc().create_bidi_channel(
+        team_id.into(),
+        peer_id.into(),
+        label_id.into(),
+    ))?;
+
+    AfcChannel::init(channel, imp::AfcChannel::new_bidi(chan));
```
gknopf-aranya: How about implementing a `From` conversion so this becomes:
```suggestion
    AfcChannel::init(channel, chan.into()));
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2178,3 +2157,267 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+    let seq = match &channel.deref().deref().inner {
+        imp::ChannelType::Bidi(c) => client.rt.block_on(c.open(dst, ciphertext))?,
+        imp::ChannelType::Receive(c) => client.rt.block_on(c.open(dst, ciphertext))?,
+        imp::ChannelType::Send(_) => return Err(imp::Error::WrongChannelType),
+    };
+
+    Ok(AfcSeq { seq })
+}
+
+/// Removes the current [`AfcChannel`] from use.
```
gknopf-aranya: Should mention that open/seal methods should not be invoked after deleting an afc channel.
There are docs in the Rust API for channel deletion that can be referenced for this.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Should we block here in the C API or make the seal/open methods in the Rust API non-async?

crates/aranya-client-capi/src/imp/afc.rs
gknopf-aranya, line 1: nit: add a doc like this to the top of the file:
`//! AFC-specific utilities for the Aranya client C API.`
Then docs like `/// An AFC channel.` become `/// A channel.`

crates/aranya-client-capi/src/imp/afc.rs
```diff
@@ -0,0 +1,72 @@
...
+/// An AFC channel.
+#[derive(Debug)]
+pub enum AfcChannel {
+    Bidi(afc::BidiChannel),
+    Send(afc::SendChannel),
+    Receive(afc::ReceiveChannel),
+}
+
+impl Typed for AfcChannel {
+    const TYPE_ID: TypeId = TypeId::new(0xDC3130B2);
```
gknopf-aranya, line 16: I wonder if we should move all the type IDs to a single module.
It would be easier to verify they are all unique that way.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -7,7 +7,9 @@ use std::{ffi::OsStr, ops::Deref, os::unix::ffi::OsStrExt, str::FromStr};
 
 use anyhow::Context as _;
 use aranya_capi_core::{opaque::Opaque, prelude::*, ErrorCode, InvalidArg};
-use aranya_client::aqc::{self, AqcPeerStream};
+#[cfg(feature = "afc")]
+use aranya_client::afc;
+use aranya_client::aqc;
```
gknopf-aranya, line 17: Need to bring in the AQC feature flag from upstream:
`#[cfg(feature = "aqc")]`

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2178,3 +2161,273 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+/// Encrypts and authenticates `plaintext`, and writes it to `dst`.
+///
+/// Note that `dst` must be at least `plaintext.len()` + `aranya_afc_channel_overhead()`.
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[in]  plaintext the message being encrypted.
+/// @param[out] dst the output buffer the ciphertext is written to.
+#[cfg(feature = "afc")]
+pub fn afc_seal(
```
gknopf-aranya: nit: perhaps call these `afc_channel_seal()` and `afc_channel_open()` since they are operations on the channel object.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: @nikki-aranya the base PR supports non-async for open/seal now with one of Dygert's latest commits

crates/aranya-client-capi/src/imp/afc.rs
jdygert-spok: Been wanting to remove these manual impls, was worried about some restrictions but got it working:
https://github.com/aranya-project/aranya-core/pull/417

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Agree that ctrl should be an out buf, then we don't need to add a separate conversion/serialization API for it before it can be sent to a peer.

crates/aranya-client-capi/src/imp/afc.rs
```diff
@@ -0,0 +1,75 @@
...
+impl From<afc::ReceiveChannel> for AfcChannel {
+    fn from(channel: afc::ReceiveChannel) -> Self {
+        Self::Receive(channel)
+    }
+}
+
+/// A control message, for creating the other end of a channel.
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct AfcCtrlMsg(pub(crate) afc::CtrlMsg);
```
gknopf-aranya, line 50: `AfcCtrlMsg` won't be needed anymore if we return the ctrl message via an out buf from the channel creation methods.

crates/aranya-client-capi/src/imp/mod.rs
```diff
@@ -1,8 +1,11 @@
+pub mod afc;
 pub mod aqc;
```
gknopf-aranya, line 2: Should these go behind feature flags?
```suggestion
#[cfg(feature = "afc")]
pub mod afc;
#[cfg(feature = "aqc")]
pub mod aqc;
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -591,7 +599,6 @@ pub unsafe fn get_key_bundle(
     keybundle: *mut MaybeUninit<u8>,
     keybundle_len: &mut usize,
 ) -> Result<(), imp::Error> {
-    let client = client.imp();
```
gknopf-aranya: These are contributing to diff noise. Could be merged in earlier on a different PR.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1224,9 +1223,9 @@ pub fn create_team(client: &Client, cfg: &CreateTeamConfig) -> Result<TeamId, im
 /// @param[in] client the Aranya Client [`Client`].
 /// @param[out] buf buffer where random bytes are written to.
 /// @param[in] buf_len the size of the buffer.
+///
+/// @relates AranyaClient.
```
gknopf-aranya, line 1227: These are also contributing to diff noise and could be brought in earlier on another PR

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2259,3 +2224,288 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+
+/// Returns the raw data for a given [`AfcCtrlMsg`].
+///
+/// Note that the lifetime of the pointer is tied to the [`AfcCtrlMsg`].
+///
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[out] ptr the raw pointer of the stored buffer.
+/// @param[out] len the raw length of the stored buffer.
+#[cfg(feature = "afc")]
+pub fn afc_get_msg_data(
```
gknopf-aranya: This is not needed if we return an out buf directly from the channel creation methods

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2259,3 +2224,288 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+    control: &AfcCtrlMsg,
+    ptr: &mut MaybeUninit<*const u8>,
+    len: &mut MaybeUninit<usize>,
+) {
+    let slice = control.0.as_bytes();
+    ptr.write(slice.as_ptr());
+    len.write(slice.len());
+}
+
+/// Returns the size of the overhead needed for a channel message.
```
```suggestion
/// Returns the overhead bytes to allocate to the ciphertext buffer.
/// Note that the ciphertext buffer must be at least `plaintext_len` + `aranya_afc_channel_overhead()` long.
```

crates/aranya-client-capi/src/imp/afc.rs
ssweetney-spok: We'd lose some of the type safety that the Rust API has but dropping this would make things a bit simpler

gknopf-aranya: The type is converted to a buffer here already:
https://github.com/aranya-project/aranya/pull/489#discussion_r2379756138
The difference is returning it directly as a buffer rather than returning a type, then converting to a buffer later.

ssweetney-spok: Oh, I meant this function `afc_recv_ctrl` when I meant "type safety"

```rust
pub fn afc_recv_ctrl(
    client: &Client,
    team_id: &TeamId,
    control: OwnedPtr<AfcCtrlMsg>,
    channel: &mut MaybeUninit<AfcChannel>,
) -> ...
```

It requires that specific type but now we'd update it to take a raw buffer of bytes

crates/aranya-client-capi/src/imp/mod.rs
ssweetney-spok: They are. It's set using inner attributes

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Solved by: https://github.com/aranya-project/aranya/pull/500

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Solved by: https://github.com/aranya-project/aranya/pull/500

crates/aranya-client-capi/src/imp/afc.rs
gknopf-aranya: Ah, we would have needed to serialize/deserialize the `AfcCtrlMsg` to send it to a peer. So the API as it is now wouldn't have been able to call `afc_recv_ctrl()` if the peer was on a different machine. Taking the buffer in `afc_recv_ctrl()` seems fine.

nikki-aranya: afaik it's not "converted", the AfcCtrlMsg is the Box which just holds the data pointing to the buffer, and then we return the ptr+len to the underlying buffer it points to which is a separate allocation that the C app can then send over transport, probably should update recv_ctrl though

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,305 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+
+/// Returns the raw data for a given [`AfcCtrlMsg`].
+///
+/// Note that the lifetime of the pointer is tied to the [`AfcCtrlMsg`].
+///
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[out] ptr the raw pointer of the stored buffer.
+/// @param[out] len the raw length of the stored buffer.
+#[cfg(feature = "afc")]
+pub fn afc_get_msg_data(
```
jdygert-spok: nit: afc_ctrl_msg_get_bytes

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,305 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+    afc::Channels::OVERHEAD
+}
+
+/// Returns the three-way comparison between `seq1` and `seq2`.
+///
+/// @param[in]  seq1 the first sequence number to compare.
+/// @param[in]  seq1 the second sequence number to compare.
+/// @param[out] __output the comparison result (-1 is <, 0 is =, 1 is >).
+#[cfg(feature = "afc")]
+pub fn afc_seq_cmp(seq1: &AfcSeq, seq2: &AfcSeq) -> isize {
```
jdygert-spok: nit: this should return c_int

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,305 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+    let seq1 = &seq1.deref().deref().0;
+    let seq2 = &seq2.deref().deref().0;
+    Ord::cmp(seq1, seq2) as isize
```
jdygert-spok: Note you usually don't have to deref manually. If you want to be somewhat explicit here I'd probably do it like this:
```suggestion
    afc::Seq::cmp(&seq1.0, &seq2.0) as isize
```

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+    size_t len;
+
+    Client* operator= & t->clients.operator;
+    Client* membera = &t->clients.membera;
+    Client* memberb = &t->clients.memberb;
+
+    AranyaLabelId label1_id;
+    AranyaChanOp op = ARANYA_CHAN_OP_SEND_RECV;
+    err = aranya_allocate_label(&operator->client, &t->id, "label1", &label1_id,
+                                2, &membera->id, op, &memberb->id, op);
```
jdygert-spok: You're passing ID pointers here but taking them by value with `va_arg(args, AranyaDeviceId)`

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,303 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+}
+
+/// Returns the size of the overhead needed for a channel message.
+///
+/// Note that the ciphertext buffer must be at least `plaintext_len` +
+/// `aranya_afc_channel_overhead()` long.
+#[cfg(feature = "afc")]
+pub fn afc_channel_overhead() -> usize {
+    afc::Channels::OVERHEAD
+}
```
jdygert-spok: I think we decided on a call to just make this a constant.

crates/aranya-client-capi/cbindgen.toml
```diff
@@ -100,6 +104,7 @@ style = "both"
 "target_os = linux" = "__linux__"
 "target_os = vxworks" = "__VXWORKS__"
 "target_arch = x86_64" = "__x86_64__"
+"feature = afc" = "ENABLE_AFC"
 "feature = aqc" = "ENABLE_AQC"
```
jdygert-spok: Wondering now if these should both be e.g. `ENABLE_ARANYA_AQC` or even `ARANYA_ENABLE_AQC`?

gknopf-aranya: Or `ARANYA_ENABLE_AFC`, `ARANYA_ENABLE_AQC`. We may want to keep the namespace we've been using for other Aranya related things.

examples/c/example.bash
```diff
@@ -68,7 +68,7 @@ EOF
 done
 
 # build the daemon.
-cargo build -p aranya-daemon --bin aranya-daemon --features experimental,aqc --release
+cargo build -p aranya-daemon --bin aranya-daemon --package aranya-daemon --features experimental,aqc,preview,afc --release
```
gknopf-aranya, line 71: Ideally, we should use a `cargo make` command to build the daemon and client lib in this script

examples/c/example.c
```diff
@@ -75,13 +76,13 @@ typedef enum {
 } Members;
 
 // List of Unix domain socket paths for the Aranya clients.
-const char *daemon_socks[] = {
+const char* daemon_socks[] = {
```
gknopf-aranya, line 79: We need a better way for each developer's IDE to format this file more consistently so there is less formatting diff noise on PRs.

examples/c/example.bash
nikki-aranya: That's fixed in #504

examples/c/example.c
nikki-aranya: We'll need a dedicated clang-format for that, that'll have to be a separate issue

jdygert-spok: There's already a `.clang-format` config, it just doesn't seem to be getting applied consistently for some reason.

nikki-aranya: #508

examples/c/example.c
```diff
@@ -139,17 +140,22 @@ typedef struct {
...
+AranyaError init_team(Team* t);
+AranyaError add_sync_peers(Team* t, AranyaSyncPeerConfig* cfg);
+AranyaError run(Team* t);
+AranyaError run_afc_example(Team* t);
+AranyaError run_aqc_example(Team* t);
+AranyaError cleanup_team(Team* t);
+
+AranyaError aranya_allocate_label(AranyaClient* client, AranyaTeamId* id,
+                                  const char* label_name,
+                                  AranyaLabelId* label_id, int num_peers, ...);
```
gknopf-aranya: Or we could take an array of device ids and an array of channel operations to iterate over, so this function doesn't take a variable list of arguments.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -3455,6 +3552,468 @@ AranyaError aranya_aqc_recv_stream_try_recv_ext(struct AranyaAqcReceiveStream *s
...
+ * @param[in]  ciphertext the message being decrypted.
+ * @param[out] dst the output buffer the message is written to.
+ * @param[out] seq the sequence number for the opened message, for reordering.
+ */
+AranyaError aranya_afc_channel_open(const struct AranyaAfcChannel *channel,
+                                    const uint8_t *ciphertext,
+                                    size_t ciphertext_len,
+                                    uint8_t *dst,
+                                    size_t dst_len,
+                                    struct AranyaAfcSeq *seq);
```
jdygert-spok, line 3994: The dst len for seal/open seems kinda funky currently. If someone makes dst too large, then after the seal/open they would assume `dst[..dst_len]` is the output but the end of it would be junk. I guess we should either use our typical outbuf pattern where we adjust the length, or return an error if `dst_len != ciphertext_len + overhead`.

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
     return err;
 }
 
+AranyaError aranya_allocate_label(AranyaClient* client, AranyaTeamId* id,
```
gknopf-aranya: The name of this function is a bit misleading because it sounds like the function will only allocate/create a new label, but it is also assigning the label to devices. I'd rather keep these operations separate because we want the AFC example to show that a label must be allocated to each member device before they can create an AFC channel with each other. Combining this functionality into a helper function hides some of the steps we want to show in the higher level AFC example. Here are some of the higher level steps we want to show:
- Create label
- Assign label to membera
- Assign label to memberb
- Create AFC channel
- Send/recv ctrl message
- Send/recv data
- Delete AFC channel

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+}
+
+// Run the AFC example.
+AranyaError run_afc_example(Team* t) {
+    AranyaError err = ARANYA_ERROR_OTHER;
+
+    unsigned char* ciphertext = NULL;
+    unsigned char* plaintext  = NULL;
+    size_t ciphertext_len;
+    size_t plaintext_len;
```
gknopf-aranya: These could be defined/allocated right before use which would make the code a bit less verbose.

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+    }
+
+    err = aranya_sync_now(&membera->client, &t->id, sync_addrs[OPERATOR], NULL);
+    EXPECT("error calling `sync_now` to sync with peer", err);
+
+    err = aranya_sync_now(&memberb->client, &t->id, sync_addrs[OPERATOR], NULL);
+    EXPECT("error calling `sync_now` to sync with peer", err);
+
+    AranyaAfcChannel channel_a;
+    AranyaAfcCtrlMsg message;
```
gknopf-aranya: Calling this `message` could be confusing since the example also sends the `afc_msg` data message.
```suggestion
    AranyaAfcCtrlMsg ctrl;
```
Or:
```suggestion
    AranyaAfcCtrlMsg ctrl_msg;
```

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+        goto exit;
+    }
+
+    err = aranya_sync_now(&membera->client, &t->id, sync_addrs[OPERATOR], NULL);
+    EXPECT("error calling `sync_now` to sync with peer", err);
+
+    err = aranya_sync_now(&memberb->client, &t->id, sync_addrs[OPERATOR], NULL);
+    EXPECT("error calling `sync_now` to sync with peer", err);
+
+    AranyaAfcChannel channel_a;
```
gknopf-aranya: nit: could call this:
```suggestion
    AranyaAfcChannel afc_bidi_chan_a;
```

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+    AranyaError err = ARANYA_ERROR_OTHER;
+
+    unsigned char* ciphertext = NULL;
+    unsigned char* plaintext  = NULL;
+    size_t ciphertext_len;
+    size_t plaintext_len;
+
+    AranyaAfcChannelType chan_type;
+    AranyaAfcSeq seq;
+    const uint8_t* ptr;
```
```suggestion
    const uint8_t* ctrl_msg_bytes;
```

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+    err = aranya_sync_now(&memberb->client, &t->id, sync_addrs[OPERATOR], NULL);
+    EXPECT("error calling `sync_now` to sync with peer", err);
+
+    AranyaAfcChannel channel_a;
+    AranyaAfcCtrlMsg message;
+    err = aranya_afc_create_bidi_channel(&membera->client, &t->id, &memberb->id,
+                                         &label1_id, &channel_a, &message);
+    EXPECT("error creating a bidi channel from membera", err);
+
+    err = aranya_afc_ctrl_msg_get_bytes(&message, &ptr, &len);
```
gknopf-aranya: We should add a comment here about which peer should call `aranya_afc_ctrl_msg_get_bytes()` and `aranya_afc_recv_ctrl()`. It would also be helpful to mention that the application is responsible for transmitting the ctrl message bytes to the peer via their own transport since that is not handled by Aranya.

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+
+    unsigned char* ciphertext = NULL;
+    unsigned char* plaintext  = NULL;
+    size_t ciphertext_len;
+    size_t plaintext_len;
+
+    AranyaAfcChannelType chan_type;
+    AranyaAfcSeq seq;
+    const uint8_t* ptr;
+    size_t len;
```
```suggestion
    size_t ctrl_msg_bytes_len;
```

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+    AranyaAfcCtrlMsg message;
+    err = aranya_afc_create_bidi_channel(&membera->client, &t->id, &memberb->id,
+                                         &label1_id, &channel_a, &message);
+    EXPECT("error creating a bidi channel from membera", err);
+
+    err = aranya_afc_ctrl_msg_get_bytes(&message, &ptr, &len);
+    EXPECT("error getting ptr+len from `AranyaAfcCtrlMsg`", err);
+
+    AranyaAfcChannel channel_b;
+    err = aranya_afc_recv_ctrl(&memberb->client, &t->id, ptr, len, &channel_b,
```
gknopf-aranya: We should check that `chan_type` is set to the expected bidi channel type after calling this method.

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+                               &chan_type);
+    EXPECT("error creating a bidi channel from memberb", err);
```
```suggestion
    EXPECT("error receiving a bidi channel ctrl message from membera", err);
```

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: The first option, adjusting the length, seems fine since that's what we've been doing

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+
+    const char* afc_msg = "afc msg";
+    size_t afc_msg_len  = strlen(afc_msg);
+    ciphertext_len      = afc_msg_len + ARANYA_AFC_CHANNEL_OVERHEAD;
+    ciphertext          = calloc(ciphertext_len, 1);
+    if (ciphertext == NULL) {
+        err = ARANYA_ERROR_BUG;
```
gknopf-aranya: This would be a memory allocation error possibly due to running out of memory, not an Aranya bug.

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+        goto exit;
+    }
+
+    err = aranya_afc_channel_seal(&channel_a, (const uint8_t*)afc_msg,
```
gknopf-aranya: Note that this is sending the null terminator byte which is probably fine

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+    }
+
+    err = aranya_afc_channel_delete(&membera->client, &channel_a);
+    EXPECT("error deleting membera's channel", err);
+
+    err = aranya_afc_channel_delete(&memberb->client, &channel_b);
+    EXPECT("error deleting memberb's channel", err);
+
+    free(ciphertext);
+    free(plaintext);
```
gknopf-aranya, line 909: Is `ptr` freed anywhere?

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+
+    err = aranya_sync_now(&membera->client, &t->id, sync_addrs[OPERATOR], NULL);
+    EXPECT("error calling `sync_now` to sync with peer", err);
+
+    err = aranya_sync_now(&memberb->client, &t->id, sync_addrs[OPERATOR], NULL);
+    EXPECT("error calling `sync_now` to sync with peer", err);
+
+    AranyaAfcChannel send_channel;
+    AranyaAfcCtrlMsg recv_message;
+    err = aranya_afc_create_uni_send_channel(&membera->client, &t->id,
```
gknopf-aranya, line 947: Should we add an example of `create_uni_recv_channel()`?

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+        EXPECT("plaintext does not match input text", ARANYA_ERROR_BUG);
+    }
+
+    err = aranya_afc_channel_delete(&membera->client, &send_channel);
+    EXPECT("error deleting membera's channel", err);
+
+    err = aranya_afc_channel_delete(&memberb->client, &recv_channel);
+    EXPECT("error deleting memberb's channel", err);
+
+exit:
```
gknopf-aranya, line 1023: I don't see where the labels are deallocated/freed/cleaned up. Since they are allocated by Aranya, there should be a `_cleanup()` method for them.

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+    EXPECT("error calling `sync_now` to sync with peer", err);
+
+    AranyaAfcChannel send_channel;
+    AranyaAfcCtrlMsg recv_message;
+    err = aranya_afc_create_uni_send_channel(&membera->client, &t->id,
+                                             &memberb->id, &label2_id,
+                                             &send_channel, &recv_message);
+    EXPECT("error creating a uni send channel from membera", err);
+
+    err = aranya_afc_ctrl_msg_get_bytes(&recv_message, &ptr, &len);
```
gknopf-aranya: Reusing `ptr` here may not be the best idea. We don't want to accidentally reuse the ctrl message from a different channel creation command or forget to properly deallocate/cleanup each one.

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+
+    AranyaAfcChannel recv_channel;
+    err = aranya_afc_recv_ctrl(&memberb->client, &t->id, ptr, len,
+                               &recv_channel, &chan_type);
+    EXPECT("error creating a uni recv channel from memberb", err);
+
+    const char* one_way_msg = "one way msg";
+    size_t one_way_msg_len  = strlen(one_way_msg);
+    ciphertext_len          = one_way_msg_len + ARANYA_AFC_CHANNEL_OVERHEAD;
+    ciphertext              = calloc(ciphertext_len, 1);
```
gknopf-aranya: Ditto about not reusing `ciphertext` and `plaintext` for open/seal on a different channel.

examples/c/example.c
```diff
@@ -763,17 +772,196 @@ AranyaError run(Team *t) {
...
+
+    plaintext_len = ciphertext_len - ARANYA_AFC_CHANNEL_OVERHEAD;
+    plaintext     = calloc(plaintext_len, 1);
+    if (plaintext == NULL) {
+        err = ARANYA_ERROR_BUG;
+        goto exit;
+    }
+
+    err = aranya_afc_channel_open(&recv_channel, ciphertext, ciphertext_len,
+                                  plaintext, plaintext_len, &seq);
```
gknopf-aranya: Could perform another `seal()/open()` after this to show an example of using `afc_seq_cmp()` with ascending sequence numbers.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,307 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+#[aranya_capi_core::derive(Cleanup)]
+#[aranya_capi_core::opaque(size = 24, align = 8)]
+pub type AfcSeq = Safe<imp::AfcSeq>;
+
+/// The overhead needed for a channel message.
+///
+/// Note that the ciphertext buffer must be at least `plaintext_len` +
+/// `aranya_afc_channel_overhead()` long.
+#[cfg(feature = "afc")]
+pub const ARANYA_AFC_CHANNEL_OVERHEAD: usize = 24;
```
gknopf-aranya, line 2283: Could we generate 24 with a macro so the assert below would no longer be required?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,307 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+/// @relates AranyaClient.
+#[cfg(feature = "afc")]
+pub fn afc_create_bidi_channel(
+    client: &Client,
+    team_id: &TeamId,
+    peer_id: &DeviceId,
+    label_id: &LabelId,
+    channel: &mut MaybeUninit<AfcChannel>,
+    control: &mut MaybeUninit<AfcCtrlMsg>,
+) -> Result<(), imp::Error> {
```
gknopf-aranya, line 2316: Could return `AfcChannelType` from the channel creation methods like we do for `afc_recv_ctrl()`.
This isn't strictly needed because the caller should already know the type of channel they created or they could call `afc_get_channel_type()` to figure it out. But it wouldn't hurt to return it anyways for convenience.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,307 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+
+#[allow(unused_qualifications)]
+#[cfg(feature = "afc")]
+const _: () = {
+    assert!(ARANYA_AFC_CHANNEL_OVERHEAD == aranya_client::afc::Channels::OVERHEAD);
+};
+
+/// Create a bidirectional AFC channel between this device and a peer.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
```
gknopf-aranya, line 2296: Docs for channel creation methods should mention what must be done with the `ctrl` message including calling `aranya_afc_ctrl_msg_get_bytes()` and sending it to the peer via the network.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,307 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+///
+/// @param[in]  client the Aranya Client [`Client`].
+/// @param[in]  team_id the team's identifier [`TeamId`].
+/// @param[in]  peer_id the peer's identifier [`DeviceId`].
+/// @param[in]  label_id the label identifier [`LabelId`] to create the channel with.
```
gknopf-aranya, line 2303: We should mention that both peers must have the label assigned to them before attempting to create an AFC channel with each other.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,307 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+/// Note that the lifetime of the pointer is tied to the [`AfcCtrlMsg`].
+///
+/// @param[in]  channel the AFC channel object [`AfcChannel`].
+/// @param[out] ptr the raw pointer of the stored buffer.
+/// @param[out] len the raw length of the stored buffer.
+#[cfg(feature = "afc")]
+pub fn afc_ctrl_msg_get_bytes(
+    control: &AfcCtrlMsg,
+    ptr: &mut MaybeUninit<*const u8>,
+    len: &mut MaybeUninit<usize>,
```
gknopf-aranya, line 2473:
```suggestion
    bytes_ptr: &mut MaybeUninit<*const u8>,
    bytes_len: &mut MaybeUninit<usize>,
```

examples/c/example.c
gknopf-aranya: Read the docs and realized that the lifetime of `ptr` is tied to `AfcCtrlMsg` so while `ptr` doesn't need to be freed, the `AfcCtrlMsg` object should be cleaned up with the corresponding `_cleanup()` method.

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: That sounds good to me

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,307 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+pub enum AfcChannelType {
+    Bidirectional,
+    Sender,
+    Receiver,
+}
+
+/// An AFC Control Message, used to create the other end of a channel.
+///
+/// In order to access the underlying buffer to send to a peer, you'll need to
+/// call `aranya_afc_get_msg_data()`.
```
ssweetney-spok:
```suggestion
/// call `aranya_afc_ctrl_msg_get_bytes()`.
```

examples/c/example.c
nikki-aranya: The problem is once we start divorcing from our owner/admin/operator/membera/memberb structure we've been using for so long (see: custom roles)
Also, duplicating this everywhere (over two examples now and more in the future once we add a test suite) goes against DRY, I'd be open to different names though

examples/c/example.c
nikki-aranya: I'd have to refactor this into two functions which I didn't really feel like doing, they're used twice so I have to define at the top

examples/c/example.c
nikki-aranya: I'm still not convinced "control message" is the best descriptor of "hey, give this to the peer you want to talk with so they can create a channel", it's way too vague

examples/c/example.c
nikki-aranya: see above, I don't think calling it a control message is helpful

examples/c/example.c
gknopf-aranya: That's why I left this comment about docs here:
https://github.com/aranya-project/aranya/pull/489/files#r2392650863

examples/c/example.c
nikki-aranya: I'd have to make a brand new error type, not sure if it's worth it though

examples/c/example.c
nikki-aranya: This comes back to the debate of whether this should be an example or tests, if it's an example then it's obvious that you swap the operations to send data in the other direction

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: Nope, cbindgen complains if it's anything other than a plain integer afaict

examples/c/example.c
gknopf-aranya: I left another comment about using different variable names so these aren't accidentally reused without reallocating or clearing the old values. E.g. `ciphertext1`, `plaintext1`, `ciphertext2`, `plaintext2`.

examples/c/example.c
gknopf-aranya: Right, and ideally we wouldn't need complete test coverage in an example. However, since we don't have a separate test suite for the C API we are forced to use the C example for a dual purpose of testing the API and providing an example of how to use the C API which is unfortunate. I'd prefer to tackle adding a test suite for the C API on a different PR, which means we should try our best to test each AFC method added on this PR.

examples/c/example.c
gknopf-aranya: `aranya_allocate_and_assign_label()` would also work

examples/c/example.c
gknopf-aranya: We use the term control message throughout our docs/implementation and in both the Rust and C API docs. I think we should at least use a consistent name for now. Calling the ctrl message something else would be a larger refactor. Perhaps we can have more discussion on a better name for this.

examples/c/example.c
ssweetney-spok: The label IDs are arrays that are allocated on the stack so I don't think we need to call a cleanup method for them

examples/c/example.c
```diff
@@ -763,17 +786,272 @@ AranyaError run(Team *t) {
...
+        EXPECT("error assigning label", err);
+    }
+
+exit:
+    return err;
+}
+
+// Run the AFC unidirectional example.
+AranyaError run_afc_bidi_example(Team* t) {
+    Client* operator= & t->clients.operator;
```
gknopf-aranya, line 800: Formatting:
```suggestion
    Client* operator = &t->clients.operator;
```

examples/c/example.c
```diff
@@ -772,179 +786,247 @@ AranyaError run(Team* t) {
...
+    // of your choice. Aranya Fast Channels (AFC) does not provide a transport,
+    // only the encryption capabilities to make such an operation safe.
+
+    // The peer needs to allocate a buffer to decrypt the data back into, minus
+    // channel overhead. This allows it to calculate the original data's length.
+    size_t plaintext_len     = ciphertext_len - ARANYA_AFC_CHANNEL_OVERHEAD;
+    unsigned char* plaintext = calloc(plaintext_len, 1);
     if (plaintext == NULL) {
-        err = ARANYA_ERROR_BUG;
+        err = ARANYA_ERROR_MEMORY;
```
jdygert-spok: We don't need to add a new variant for this, just abort like the aqc example does.

examples/c/example.c
jdygert-spok: The formatter just does that for some reason, we can try to resolve later.

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: The destination buffer can be larger than what's needed, so we can return an error if:
`dst_len < ciphertext_len + overhead`
When we write to the out buffer, `dst_len` will be updated with the value of the number of bytes written.
If the output buffer is too small, we'll return an error and allow the application to reallocate a larger buffer before calling `open()` again.

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: "Permission to perform this operation is checked against the Aranya policy."

gknopf-aranya: True, but requiring a user to review the entire `policy.md` before using the API is a big ask. Providing useful tips like assigning labels to devices first will improve ease of adoption.

examples/c/example.c
```diff
@@ -11,13 +11,18 @@
...
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 
-#include "aranya-client.h"
+#define ENABLE_ARANYA_AFC 1
+#define ENABLE_ARANYA_PREVIEW 1
+#define ENABLE_ARANYA_AQC 1
+#define ENABLE_ARANYA_EXPERIMENTAL 1
+#include "../../crates/aranya-client-capi/output/aranya-client.h"
```
jdygert-spok: revert

nikki-aranya: dangit, I try to be so careful and it always slips in 😆

jdygert-spok (dismissed):

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: They can see in the example code that we assign labels to peers and if they don't then it will give them an error about it, I don't like adding more than one note per function if I can help it, gets too noisy.

nikki-aranya: Also, permissions are literally the selling point of aranya, it should be obvious to anyone using the system that you need the label to be able to do things, otherwise why would we take it in as a param

gknopf-aranya: My opinion is that API docs should be detailed enough that the API can be used without an example.
@jdygert-spok, @ssweetney-spok do you have more thoughts on this?

ssweetney-spok: The docs could get very verbose if we want them to be so detailed that users can forgo reading the policy.

I think it's worth mentioning the label assignment here though. Maybe add an "Errors" section and include something like "Returns an error if the device has not been assigned the label associated with `label_id`"

gknopf-aranya (approved):
Overall looks good to me. I think the API docs could be improved so we rely less on the C example and `policy.md` to explain how to use the API but I don't think we should block the release on that.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2228,3 +2236,354 @@ pub unsafe fn aqc_recv_stream_try_recv(
...
+    }
+    *dst_len = plaintext.len() + ARANYA_AFC_CHANNEL_OVERHEAD;
+
+    Ok(())
+}
+
+/// Decrypts and authenticates `ciphertext`, and writes it to `dst`.
+///
+/// Note that `dst` must be at least `ciphertext.len()` - `aranya_afc_channel_overhead()`,
+/// or the function will return an error (`InvalidArgument` or `BufferTooSmall`).
```
gknopf-aranya, line 2530: Should we mention that if this returns `BufferTooSmall` the caller is expected to allocate a larger buffer and call it again?

examples/c/example.c
```diff
@@ -756,17 +775,268 @@ AranyaError run(Team *t) {
...
+    plaintext            = calloc(plaintext_len, 1);
+    if (plaintext == NULL) {
+        abort();
+    }
+
+    // Here, we open the message and get back the original data, as well as a
+    // sequence number, which allows you to reorder messages that may have been
+    // received out-of-order using `aranya_afc_seq_cmp()` to compare two seq.
+    AranyaAfcSeq seq;
+    err = aranya_afc_channel_open(&afc_bidi_chan_b, ciphertext, ciphertext_len,
```
gknopf-aranya, line 889: Should we show an example of reallocating the plaintext buffer if it is too small?

gknopf-aranya: I suppose its not required to have an example of this because there are examples of how to reallocate the buffer elsewhere. The benefit would be to validate the error handling path works.

nikki-aranya: You can in another PR, I can't be arsed when we've talked so much about testing suites and refactoring this entire file so it's not 1500LOC

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: if buffer too small, make buffer larger, I'd sure hope any developers calling this API aren't too smoothbrained/AI-brained to not realize that, I would take a new PR that re-words it to something like "if buffer's too small, no operations will be performed" or something

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: I added a section, we can improve it in the future

PR #495, opened 2025-09-23
Canonicalize uds path

jdygert-spok:


crates/aranya-daemon/src/api.rs
```diff
@@ -95,6 +95,9 @@ impl DaemonApiServer {
         quic: Option<quic_sync::Data>,
     ) -> anyhow::Result<Self> {
         let listener = UnixListener::bind(&uds_path)?;
+        let uds_path = uds_path
+            .canonicalize()
```
jdygert-spok, line 120: On the daemon side we can't resolve until after we create the socket. This means that the resolved path could technically be different from the bound path if someone swapped out some symlinks inbetween or something. But I don't think this actually worsens our reliance on a secure filesystem.

gknopf-aranya (approved):

PR #503, opened 2025-09-26
DO NOT MERGE: repro crash

gknopf-aranya:


examples/c/example.bash
```diff
@@ -57,7 +57,7 @@ aqc.enable = true
 
 [afc]
 enable = true
-shm_path = "/shm_${device}"
+shm_path = "/shm_test_${device}"
```
gknopf-aranya: Had to do this because there are leftover shm files causing the example to crash on my machine. Rename works around this.

PR #505, opened 2025-09-26
c-example: fix crash on errors in example

jdygert-spok:
This should fix the crashes that occurred when we had other errors.

We were "cleaning up" uninitialized objects when an error occurred. We now basically just exit on an error. There isn't really a need to do the cleanup here on errors. It's probably worth having in a more realistic example but this example is basically just a test script.

### DIFF
--- a/examples/c/example.c
+++ b/examples/c/example.c
@@ -881,48 +881,36 @@ static void *membera_aqc_thread(void *arg) {
 
     sleep(2);
 
-exit:
     printf("membera: cleanup bidi stream\n");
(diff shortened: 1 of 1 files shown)


gknopf-aranya:
Could this cause a memory leak?

nikki-aranya (approved):

jdygert-spok:
Yes, but it doesn't really matter. All memory is reclaimed on process exit. We don't need to see the memory freed in the unexpected test failure.

PR #510, opened 2025-10-01
Add client and daemon CMake targets

jdygert-spok:


crates/aranya-client-capi/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+option(ARANYA_ENABLE_EXPERIMENTAL "Enable experimental Aranya features." OFF)
+
+option(ARANYA_ENABLE_AFC "Enable Aranya Fast Channels." ON)
+
+# ==== Daemon executable ====
+
+corrosion_import_crate(
+    MANIFEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../aranya-daemon/Cargo.toml
+	FEATURES
+        $<$<BOOL:${ARANYA_ENABLE_PREVIEW}>:preview>
```
gknopf-aranya: Is there a way to not duplicate the feature code 3 times in this file?

jdygert-spok: We might be able to use a list variable to dedup the two crate uses but the header usage expands to e.g. `ENABLE_ARANYA_PREVIEW` instead of `preview`

gknopf-aranya: Maybe we could map each feature to a tuple/list:
`ARANYA_ENABLE_PREVIEW` -> (`preview`, `ENABLE_ARANYA_PREVIEW`)

jdygert-spok: I used some lists.

CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+project(aranya-client LANGUAGES C)
+
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+include(FetchContent)
+
+FetchContent_Declare(
+    Corrosion
+    GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
+    GIT_TAG v0.5.2
```
nikki-aranya: needs to be bumped to 0.6

nikki-aranya (approved):

CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 3.24...3.31)
```
gknopf-aranya, line 1: Why is there a range if 3.24 is the minimum required version?

jdygert-spok: > Always set the maximum version to the highest version you test (like 4.1). This will continue to work until CMake drops the maximum version from it’s support window, rather than the minimum!

https://cliutils.gitlab.io/modern-cmake/README.html

3.31 is the version we currently use in CI.

Note this doesn't prevent you from using it with a newer cmake.

PR #513, opened 2025-10-01
Use CTest for C API tests

jdygert-spok:


crates/aranya-client-capi/ctest/CMakeLists.txt
```diff
@@ -0,0 +1,115 @@
...
+endif()
+
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+    set(WINDOWS TRUE)
+endif()
+
+# Test builds run with extra runtime checks.
+if(NOT DEFINED ENV{NO_TEST_BUILD})
+    # TODO(declan): reenable fsanitize. not sure if the leak is occuring in our code.
+    # could be in the memtxp but the stack trace hints to it being a cgo malloc.
```
djohnson-spideroak: we can probably remove this and enable fsanitize

crates/aranya-client-capi/ctest/CMakeLists.txt
```diff
@@ -0,0 +1,115 @@
...
+
+enable_testing()
+
+add_executable(TestSimple test_simple.c)
+target_link_libraries(TestSimple PRIVATE Aranya::Client)
+add_test(NAME TestSimple COMMAND TestSimple $<TARGET_FILE:Aranya::Daemon>)
+
+# add_test(NAME TestSetup
+#     COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/accounts")
+# add_test(NAME TestCleanup
```
gknopf-aranya, line 107: Would prefer to put test artifacts in a tmp dir so a cleanup method is not required after tests run

crates/aranya-client-capi/ctest/test_simple.c
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    char *env[] = {
+        "ARANYA_DAEMON=debug",
+        NULL,
+    };
+    int ret = execle(path, path, "--help", NULL, env);
+    fprintf(stderr, "unexpected return %d", ret);
+    abort();
+}
```
jdygert-spok, line 29: Unsure if we want to try and manage the daemons within the tests or from shell scripts.

gknopf-aranya: Which approach do you think is more portable considering we may deploy to a variety of different OS?

gknopf-aranya: I would imagine users would want to run the test suite before attempting to develop their own applications on a new OS we don't support in CI.

crates/aranya-client-capi/ctest/CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
+cmake_minimum_required(VERSION 3.11)
```
gknopf-aranya, line 1: No version range like before?

gknopf-aranya (approved):

crates/aranya-client-capi/ctest/CMakeLists.txt
jdygert-spok: I don't think this call or the call to project below are needed but they don't seem to hurt anything :shrug:

crates/aranya-client-capi/ctest/test_simple.c
```diff
@@ -0,0 +1,48 @@
...
+        fprintf(stderr, "usage: `%s <daemon>`\n", argv[0]);
+        return EXIT_FAILURE;
+    }
+
+    // Show that we can use the client library.
+    AranyaExtError ext;
+    TRY(aranya_ext_error_init(&ext));
+    TRY(aranya_ext_error_cleanup(&ext));
+
+    spawn_daemon(argv[1]);
```
gknopf-aranya, line 42: It does not appear that anything kills the daemon process after the test exits.

jdygert-spok: This doesn't include managing the daemon, it just runs it with the `--help` flag to show that it was built and passed to the test correctly.

crates/aranya-client-capi/ctest/CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+
+# Apple and Windows system linkers do not recognize these options.
+if(NOT APPLE AND NOT WINDOWS)
+    add_link_options(
+        -Wl,-z,relro
+        -Wl,-z,now
+        -Wl,-z,defs
+        -Wl,-z,noexecstack
+    )
+endif()
```
jdygert-spok, line 95: FYI all these extra flags aren't "needed" but have been carried over from older code. Some enable more warnings, some enable security features, some enable sanitizers which detect bad behavior at runtime. They could possibly be cleaned up but I just copied them over mostly unchanged for now.

gknopf-aranya: Seems worth adding a TODO and possibly opening an issue to simplify this later

jdygert-spok: #676

PR #527, opened 2025-10-08
afc: remove channel from shm when label is revoked

gknopf-aranya:
Remove AFC channels from shared-memory when the corresponding label ID is revoked.

Blocked by:
- https://github.com/aranya-project/aranya-core/issues/429
- https://github.com/aranya-project/aranya-core/issues/434

### DIFF
--- a/crates/aranya-daemon/src/api.rs
+++ b/crates/aranya-daemon/src/api.rs
@@ -222,22 +222,16 @@ impl EffectHandler {
                 TeamCreated(_team_created) => {}
                 TeamTerminated(_team_terminated) => {
                     #[cfg(feature = "afc")]
-                    {
-                        self.afc.delete_channels().await?;
-                        continue;
-                    }
-
-                    #[cfg(not(feature = "afc"))]
-                    tracing::warn!(effect = ?_team_terminated, "received TeamTerminated effect");
+                    self.afc.delete_channels().await?;
                 }
                 MemberAdded(_member_added) => {}
-                MemberRemoved(member_removed) => {
+                MemberRemoved(_member_removed) => {
                     #[cfg(feature = "afc")]
                     {
-                        if self.device_id == member_removed.device_id.into() {
+                        if self.device_id == _member_removed.device_id.into() {
                             self.afc.delete_channels().await?;
                         } else {
-                            let peer_id = Some(member_removed.device_id.into());
+                            let peer_id = Some(_member_removed.device_id.into());
                             self.afc
                                 .remove_if(RemoveIfParams {
                                     peer_id,
@@ -245,11 +239,7 @@ impl EffectHandler {
                                 })
                                 .await?;
                         }
-                        continue;
                     }
-
-                    #[cfg(not(feature = "afc"))]
-                    tracing::warn!(effect = ?member_removed, "received MemberRemoved effect");
                 }
                 OwnerAssigned(_owner_assigned) => {}
                 AdminAssigned(_admin_assigned) => {}
@@ -258,29 +248,23 @@ impl EffectHandler {
                 AdminRevoked(_admin_revoked) => {}
                 OperatorRevoked(_operator_revoked) => {}
                 LabelCreated(_) => {}
-                LabelDeleted(label_deleted) => {
+                LabelDeleted(_label_deleted) => {
                     #[cfg(feature = "afc")]
-                    {
-                        self.afc
-                            .remove_if(RemoveIfParams {
-                                label_id: Some(label_deleted.label_id.into()),
-                                ..Default::default()
-                            })
-                            .await?;
-                        continue;
-                    }
-
-                    #[cfg(not(feature = "afc"))]
-                    tracing::warn!(effect = ?label_deleted, "received LabelDeleted effect");
+                    self.afc
+                        .remove_if(RemoveIfParams {
+                            label_id: Some(_label_deleted.label_id.into()),
+                            ..Default::default()
+                        })
+                        .await?;
                 }
                 LabelAssigned(_) => {}
-                LabelRevoked(label_revoked) => {
+                LabelRevoked(_label_revoked) => {
                     #[cfg(feature = "afc")]
                     {
-                        let label_id = Some(label_revoked.label_id.into());
+                        let label_id = Some(_label_revoked.label_id.into());
                         let mut peer_id = None;
-                        if self.device_id != label_revoked.device_id.into() {
-                            peer_id = Some(label_revoked.device_id.into());
+                        if self.device_id != _label_revoked.device_id.into() {
+                            peer_id = Some(_label_revoked.device_id.into());
                         };
                         self.afc
                             .remove_if(RemoveIfParams {
@@ -289,11 +273,7 @@ impl EffectHandler {
                                 ..Default::default()
                             })
                             .await?;
-                        continue;
                     }
-
-                    #[cfg(not(feature = "afc"))]
-                    tracing::warn!(effect = ?label_revoked, "received LabelRevoked effect");
                 }
                 QueriedLabel(_) => {}
                 AfcUniChannelCreated(_) => {}


crates/aranya-daemon/src/afc.rs
```diff
@@ -293,6 +293,15 @@ where
             .map_err(|err| anyhow!("unable to remove AFC channel: {err}"))
     }
 
+    pub(crate) async fn label_revoked(&self, label_id: LabelId) -> Result<()> {
+        let shm = self.shm.lock().await;
+
+        // TODO: aranya-core/issues/429
```
gknopf-aranya: Blocked by adding the `label_id` parameter in:
https://github.com/aranya-project/aranya-core/issues/429

crates/aranya-daemon/src/api.rs
```diff
@@ -236,9 +244,24 @@ impl EffectHandler {
...
+                    #[cfg(feature = "afc")]
+                    self.afc
+                        .label_deleted(label_deleted.label_id.into())
+                        .await?;
+                    tracing::warn!(effect = ?label_deleted, "received LabelDeleted effect");
+                }
                 LabelAssigned(_) => {}
-                LabelRevoked(_) => {}
+                LabelRevoked(label_revoked) => {
+                    // TODO: delete AFC channel if label is revoked from peer.
```
gknopf-aranya: @djohnson-spideroak, @jdygert-spok do we plan to handle the case of deleting a channel when a label is revoked from a peer? I'm not sure how we would accomplish this currently because the daemon doesn't keep track of peer device IDs and we can't look that up from the share memory.

jdygert-spok: Seems like we need to track it in either the daemon or shm

gknopf-aranya: The shm seems like the right place for this because that will ensure the peer device IDs have the same lifecycle as the channels in the shm. That being said, I think we could more easily prototype a solution that keeps track of peer device IDs in memory in the daemon for this.

What if a device has an AFC channel with 2 different peers using the same label?
If we detect that one of the peer's label was revoked, we only want to delete the channel with that peer and not all channels with that label.
`remove_if(channel_id, label_id)` wouldn't support this operation unless it became `remove_if(channel_id, label_id, peer_id)`.
So that may force us to put the peer's device ID in the shm for the delete operation.

jdygert-spok: You could keep a `Map<(Peer, Label), ChannelId>`, then when you see `effect Unassigned { device, label }` you can get the channel ID and just call `AranyaState::remove`.

It is probably simpler to keep it in the shm, I like keeping the amount of state that needs to be kept in sync between components low.

gknopf-aranya: It seems like we continue adding more info to the shm to delete on. May be worth considering generalizing the `remove_if()` param so new fields can be added in the future.
https://github.com/aranya-project/aranya-core/pull/428#discussion_r2414965810

gknopf-aranya: @jdygert-spok, @ssweetney-spok here's an issue to add the peer's id to the shm:
https://github.com/aranya-project/aranya-core/issues/434

gknopf-aranya: Starting to integrate with the WIP PR with these aranya-core changes.

gknopf-aranya: Verified that open/seal no longer work for either peer after revoking a label from one peer

crates/aranya-daemon/src/afc.rs
gknopf-aranya: Resolved

Cargo.lock
ssweetney-spok, line 1: The latest commit hash is [`f304ad232d263978aa46da2098e3fc3eb1290037`](https://github.com/aranya-project/aranya-core/pull/432/commits/f304ad232d263978aa46da2098e3fc3eb1290037), which includes the `RemoveIfParams` change

gknopf-aranya: Pulled in that commit. Looks like you'll need to resolve merge conflicts with `main` for me to be able to test with it:
```
error[E0432]: unresolved import `aranya_id::BaseId`
  --> crates/aranya-daemon-api/src/service/aqc.rs:13:28
   |
13 | use aranya_id::{custom_id, BaseId};
   |                            ^^^^^^ no `BaseId` in the root
```

ssweetney-spok: Oh, that PR is already merged now. You should be able to use the main branch instead of "429-remove_if"

crates/aranya-daemon/src/api.rs
```diff
@@ -236,9 +244,30 @@ impl EffectHandler {
                 AdminRevoked(_admin_revoked) => {}
                 OperatorRevoked(_operator_revoked) => {}
                 LabelCreated(_) => {}
-                LabelDeleted(_) => {}
+                LabelDeleted(label_deleted) => {
+                    #[cfg(feature = "afc")]
+                    self.afc
+                        .label_deleted(label_deleted.label_id.into())
+                        .await?;
+                    tracing::warn!(effect = ?label_deleted, "received LabelDeleted effect");
```
ssweetney-spok: Do we want this to be `warn` instead of `trace`?

crates/aranya-daemon/src/policy.md
ssweetney-spok, line 1: The changes are small but will this cause a merge conflict for the custom roles PR?

crates/aranya-daemon/src/api.rs
```diff
@@ -236,9 +244,30 @@ impl EffectHandler {
...
+                            .label_deleted(label_revoked.label_id.into())
+                            .await?;
+                    } else {
+                        self.afc
+                            .label_revoked(
+                                label_revoked.label_id.into(),
+                                label_revoked.device_id.into(),
+                            )
+                            .await?;
+                    }
```
ssweetney-spok: Calling `Afc::label_deleted` in the branch for handling the `LabelRevoked` effect looked confusing at first glance.

There isn't much difference between `Afc::label_deleted` and `Afc::label_revoked` so they could be combined into a single function

crates/aranya-daemon/src/policy.md
gknopf-aranya: Yes, almost any change to the policy will cause a merge conflict with the custom roles PR. It should be easy to resolve though. cc: @chip-so

chip-so: Yeah this should be easy enough to handle. There are two kinds of "label revoked" in the custom roles policy, one for revoking from roles, and one for revoking from devices (which I think is what's happening here). I assume we want this change in both?

gknopf-aranya: We'll likely need to update the channel deletion implementation after custom roles is added. I assume we'll want to delete the corresponding channels when a label is revoked from a role.

crates/aranya-daemon/src/api.rs
gknopf-aranya: I'm implementing a more generalized way to delete channels based on optional parameters to a single `remove_if()` method.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Unless we put the role itself in the shm (which I don't think we should do because the role may change later), we'll likely end up querying for a list of devices that have the role from which the label was revoked, then delete all channels corresponding to those devices from the shm.

gknopf-aranya: @jdygert-spok, @djohnson-spideroak looping you in on this discussion

jdygert-spok: Querying for affected `(DeviceId, LabelId)` pairs and then deleting those seems like a natural approach.

These are the possible triggers for channel removal that I can think of (after custom roles).

1. Role loses label
2. Role loses afc perm
3. User loses role
4. Role is deleted
5. User loses label
6. Label is deleted

It's also worth noting that these could all also happen in the recall of the inverse operation (e.g. for 6 if `CreateLabel` is recalled because the author lost permission to create labels "at the same time" so it gets braided before).

I'm rather wary of how complicated it seems to be getting...

gknopf-aranya: Agreed, it seems far too easy for the policy and the channel deletion implementation to get out of sync with this much complexity.

gknopf-aranya: Perhaps we should periodically query the shm for a list of channels, then pass the info for those channels into a policy query that returns whether the channels are valid. This would allow channel deletion to be more closely tied to the policy rather than relying on a complex channel deletion implementation in the daemon.

gknopf-aranya: We have the ability to return several effects from a single query (via `map`), but we don't currently have an elegant way to pass a vector of elements as input to a query action.

gknopf-aranya: An alternative would be to emit `ChannelDeleted` effects alongside any effects that could result in the deletion of a channel. This would avoid the need to query/poll the factdb to see which channels are valid.

gknopf-aranya: I'll open a new issue for this since this is out-of-scope of the current channel deletion implementation with the existing AFC policy.

gknopf-aranya: Resolving this discussion in favor of a new issue:
https://github.com/aranya-project/aranya/issues/533

crates/aranya-daemon/src/api.rs
```diff
@@ -226,19 +234,65 @@ impl EffectHandler {
             trace!(?effect, "handling effect");
             match effect {
                 TeamCreated(_team_created) => {}
-                TeamTerminated(_team_terminated) => {}
+                TeamTerminated(team_terminated) => {
+                    #[cfg(feature = "afc")]
+                    self.afc.delete_channels().await?;
+                    tracing::trace!(effect = ?team_terminated, "received TeamTerminated effect");
```
jdygert-spok: These traces don't add anything, we have `trace!(?effect, "handling effect");` above.

gknopf-aranya: They prevent an `unused` warning if the `afc` feature isn't enabled

crates/aranya-daemon/src/afc.rs
```diff
@@ -185,6 +194,45 @@ where
...
+                        return false;
+                    }
+                }
+                if let Some(peer_id) = params.peer_id {
+                    if chan_params.peer_id != peer_id {
+                        return false;
+                    }
+                }
+                true
+            })
```
jdygert-spok:
```suggestion
            .remove_if(|chan| {
                params.channel_id.is_none_or(|id| chan.channel_id == id)
                    && params.label_id.is_none_or(|id| chan.label_id == id)
                    && params.peer_id.is_none_or(|id| chan.peer_id == id)
            })
```

crates/aranya-daemon/src/afc.rs
```diff
@@ -185,6 +194,45 @@ where
...
+    pub(crate) async fn delete_channels(&self) -> Result<()>
+    where
+        E: Engine<CS = C>,
+    {
+        self.shm
+            .lock()
+            .await
+            .write
+            .remove_all()
+            .map_err(|err| anyhow!("unable to remove AFC channels: {err}"))
```
jdygert-spok: These should each just use context, noticed there's some existing ones.

```suggestion
            .context("unable to remove AFC channels")
```

crates/aranya-client/Cargo.toml
```diff
@@ -53,6 +53,7 @@ tracing = { workspace = true }
 
 [dev-dependencies]
 aranya-daemon = { path = "../aranya-daemon", features = ["aqc", "afc", "experimental", "preview"] }
+spideroak-base58 = { workspace = true }
```
jdygert-spok, line 50: Oh, this is only used in tests so we can remove the one under regular dependencies.

crates/aranya-daemon/src/afc.rs
gknopf-aranya: I had a feeling there would be a simpler way to write this part

djohnson-spideroak (dismissed):

crates/aranya-daemon/src/api.rs
jdygert-spok: I'd prefer to have them `warn` when unhandled like the aqc effects.

crates/aranya-daemon/src/afc.rs
```diff
@@ -191,7 +202,33 @@ where
...
+        self.shm
+            .lock()
+            .await
+            .write
+            .remove_all()
+            .context("unable to remove AFC channels")
+    }
+
+    /// Remove channels matching criteria.
+    pub(crate) async fn remove_if(&self, params: RemoveIfParams) -> Result<()> {
```
djohnson-spideroak, line 222: Does it need to match all criteria or just one?

gknopf-aranya: In order for a channel to be deleted, it must match all criteria set to `Some` in `RemoveIfParams`

djohnson-spideroak (dismissed):

crates/aranya-daemon/src/api.rs
jdygert-spok: My bad, I just realized none of these effects are actually gated behind the afc feature, so it shouldn't be a warning when they are unhandled. We can instead either keep them named starting with a `_`, or have e.g. `_ = team_terminated` at the end to explicitly ignore them.

gknopf-aranya: I underscored the effects

PR #531, opened 2025-10-15
chore: update to latest aranya-core

jdygert-spok:


Cargo.toml
```diff
@@ -109,3 +110,21 @@ allow-branch = ["main"]
 shared-version = true
 tag-message = ""
 tag-prefix = ""
+
+[patch.crates-io]
```
gknopf-aranya: Should we add a TODO here to remove this before the next release?

crates/aranya-client-capi/src/imp/afc.rs
gknopf-aranya, line 1: Why is `afc.rs` being deleted?

Cargo.toml
jdygert-spok: The security checks won't let us merge this.

crates/aranya-client-capi/src/imp/afc.rs
jdygert-spok: There's no need to wrap the types now that we don't need to implement `Typed`.

Cargo.toml
gknopf-aranya: In that case, we may want to do an incremental release of aranya-core so we can continue testing feature branches in aranya-core

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -292,7 +292,7 @@ pub type Client = Safe<imp::Client>;
 pub const ARANYA_ID_LEN: usize = 32;
 
 const _: () = {
-    assert!(ARANYA_ID_LEN == size_of::<aranya_crypto::Id>());
+    assert!(ARANYA_ID_LEN == size_of::<aranya_id::BaseId>());
```
gknopf-aranya, line 295: What's the reason for calling this `BaseId` vs. `Id`?

jdygert-spok: See https://github.com/aranya-project/aranya-core/pull/329

gknopf-aranya (approved):

ssweetney-spok (approved):

PR #532, opened 2025-10-15
afc: apis for getting globally unique channel IDs

gknopf-aranya:
Add APIs for getting globally unique AFC channel IDs.

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -529,22 +529,20 @@ pub struct AqcBidiChannelId {
 }
 
 #[cfg(feature = "aqc")]
-impl From<aqc::BidiChannelId> for AqcBidiChannelId {
-    fn from(value: aqc::BidiChannelId) -> Self {
+impl From<aranya_daemon_api::AqcBidiChannelId> for AqcBidiChannelId {
+    fn from(value: aranya_daemon_api::AqcBidiChannelId) -> Self {
         Self {
             id: Id {
-                bytes: value.__id.into(),
+                bytes: value.into(),
             },
         }
     }
 }
 
 #[cfg(feature = "aqc")]
-impl From<&AqcBidiChannelId> for aqc::BidiChannelId {
+impl From<&AqcBidiChannelId> for aranya_daemon_api::AqcBidiChannelId {
     fn from(value: &AqcBidiChannelId) -> Self {
-        Self {
-            __id: value.id.bytes.into(),
-        }
+        value.id.bytes.into()
     }
 }
 
@@ -557,22 +555,20 @@ pub struct AqcUniChannelId {
 }
 
 #[cfg(feature = "aqc")]
-impl From<aqc::UniChannelId> for AqcUniChannelId {
-    fn from(value: aqc::UniChannelId) -> Self {
+impl From<aranya_daemon_api::AqcUniChannelId> for AqcUniChannelId {
+    fn from(value: aranya_daemon_api::AqcUniChannelId) -> Self {
         Self {
             id: Id {
-                bytes: value.__id.into(),
+                bytes: value.into(),
             },
         }
     }
 }
 
 #[cfg(feature = "aqc")]
-impl From<&AqcUniChannelId> for aqc::UniChannelId {
+impl From<&AqcUniChannelId> for aranya_daemon_api::AqcUniChannelId {
     fn from(value: &AqcUniChannelId) -> Self {
-        Self {
-            __id: value.id.bytes.into(),
-        }
+        value.id.bytes.into()
     }
 }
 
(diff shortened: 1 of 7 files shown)


gknopf-aranya:
Actually, I think we can work around this by adding the `encap` field to the `AfcUniChannelCreated` effect.

gknopf-aranya:
IDs match now:
```
2025-10-15T23:43:06.958656Z  INFO creating afc send channel
2025-10-15T23:43:06.970339Z  INFO created afc channel: DgiRmgh7KTFZjaH8oUpuisNYmBooSSbPdGww8WiDm4hY
2025-10-15T23:43:06.970360Z  INFO receiving afc recv channel
2025-10-15T23:43:06.972951Z  INFO received afc channel: DgiRmgh7KTFZjaH8oUpuisNYmBooSSbPdGww8WiDm4hY
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -572,6 +572,32 @@ impl From<&AqcUniChannelId> for aranya_daemon_api::AqcUniChannelId {
...
+pub struct AfcChannelId {
+    id: Id,
+}
+
+#[cfg(feature = "afc")]
+impl From<afc::ChannelId> for AfcChannelId {
+    fn from(value: afc::ChannelId) -> Self {
+        Self {
+            id: Id {
+                bytes: value.into(),
```
gknopf-aranya: Doesn't look like we currently have a way to convert the id into bytes

gknopf-aranya: Added conversions for this

djohnson-spideroak (approved):

djohnson-spideroak (approved):

crates/aranya-client/src/afc.rs
```diff
@@ -54,6 +54,41 @@ impl From<Box<[u8]>> for CtrlMsg {
...
+    fn from(id: ChannelId) -> Self {
+        id.0.into()
+    }
+}
+
+impl fmt::Display for ChannelId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        std::fmt::Display::fmt(&self.0, f)
+    }
+}
```
jdygert-spok, line 69: Let's follow the same pattern as `aranya_client::DeviceId` etc

crates/aranya-client/src/afc.rs
```diff
@@ -54,6 +54,41 @@ impl From<Box<[u8]>> for CtrlMsg {
     }
 }
 
+/// A globally unique channel ID.
+#[derive(Copy, Clone, Debug)]
+pub struct ChannelId(aranya_daemon_api::AfcGlobalChannelId);
```
jdygert-spok: Should we call it `GlobalChannelId`?

gknopf-aranya: I considered this but since we only expose the global ID via the API and the local ID is an implementation detail, it seems clear as-is.

jdygert-spok: My concern is for us devs and anyone else who looks at the internals. Maybe we should rename the local one to `LocalChannelId`. I guess that's more consistent since most of our IDs are global.

gknopf-aranya: That works for me

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -529,8 +529,8 @@ pub struct AqcBidiChannelId {
 }
 
 #[cfg(feature = "aqc")]
-impl From<aranya_daemon_api::AqcBidiChannelId> for AqcBidiChannelId {
-    fn from(value: aranya_daemon_api::AqcBidiChannelId) -> Self {
+impl From<aranya_crypto::aqc::BidiChannelId> for AqcBidiChannelId {
+    fn from(value: aranya_crypto::aqc::BidiChannelId) -> Self {
```
jdygert-spok: Oh, the AQC IDs need to be wrapped too.

gknopf-aranya: Ugh, that does appear to be needed. Wrapping them.

crates/aranya-client/src/aqc/api.rs
```diff
@@ -14,6 +16,34 @@ use crate::{
...
+pub struct BidiChannelId {
+    #[doc(hidden)]
+    pub __id: AqcBidiChannelId,
+}
+
+impl Display for BidiChannelId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        Display::fmt(&self.__id, f)
+    }
+}
```
djohnson-spideroak: I think we can remove the bidi channel IDs correct?

gknopf-aranya: This is AQC which still has uni and bidi channels. Bidi channel IDs are being removed for AFC.

gknopf-aranya: Or more accurately, since the ID getters were not implemented for AFC previously, I'm only adding getters for AFC uni channels and not for bidi channels because we are removing support for AFC bidi channels.

gknopf-aranya: Now that there are plans to remove AQC, there is no need to add the AQC channel ID getters on this PR:
https://github.com/aranya-project/aranya/issues/535

gknopf-aranya: AQC ID getters have been removed from this PR

crates/aranya-client/src/afc.rs
```diff
@@ -54,16 +54,30 @@ impl From<Box<[u8]>> for CtrlMsg {
     }
 }
 
+/// A globally unique channel ID.
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct ChannelId {
+    #[doc(hidden)]
+    pub __id: AfcChannelId,
+}
```
djohnson-spideroak, line 63: Can you add some docs here about ChannelId vs local channel ID?

jdygert-spok: I don't think we really need them here if only the global id is exposed to the user, beyond specifying this is globally unique like it does.

But I do also want to rename the local channel ID all the way down through aranya-core to `LocalChannelId` to make it more clear for us.

jdygert-spok: https://github.com/aranya-project/aranya-core/pull/453

Issue #522, opened 2025-10-07
example: show example of using afc with custom transport


gknopf-aranya:
Demonstrate using AFC with a custom transport such as: UDP, TCP, QUIC.
This will unblock removal of AQC.
Before starting on this issue, we will need to confirm intent to remove AQC.

cc: @djohnson-spideroak, @jdygert-spok

jdygert-spok:
> Before starting on this issue, we will need to confirm intent to remove AQC.

I don't see why. Maybe it is slightly lower priority if not, but this seems like a rather realistic/useful example regardless.

gknopf-aranya:
Decision has been made to start with UDP instead of TCP

cc: @djohnson-spideroak, @jdygert-spok

gknopf-aranya:
The multi-node example is showing an example of using AFC with TCP. So this seems less important now.

PR #536, opened 2025-10-20
aqc: remove

### LINKED ISSUE #535: aqc: remove aqc
We dont plan to move forward with promoting AQC to preview/production. Remove AQC from Aranya including the public API surface and internal management code.

jdygert-spok:
Closes #535.

### DIFF
--- a/crates/aranya-daemon/src/policy.md
+++ b/crates/aranya-daemon/src/policy.md
@@ -957,6 +957,8 @@ effect AfcUniChannelCreated {
     label_id id,
     // The channel key ID.
     channel_key_id id,
+    // The channel peer's encapsulated KEM shared secret.
+    encap bytes,
 }
 
 // The effect that is emitted when the peer of a unidirectional
@@ -1027,6 +1029,7 @@ ephemeral command AfcCreateUniChannel {
                     peer_enc_pk: peer_enc_pk,
                     label_id: this.label_id,
                     channel_key_id: this.channel_key_id,
+                    encap: this.peer_encap,
                 }
             }
         }
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -535,7 +535,7 @@ typedef struct ARANYA_ALIGNED(8) AranyaAfcSendChannel {
      * UNDEFINED BEHAVIOR to read from or write to it.
      * @private
      */
-    uint8_t __for_size_only[96];
+    uint8_t __for_size_only[120];
 } AranyaAfcSendChannel;
 #endif
 
@@ -549,7 +549,7 @@ typedef struct ARANYA_ALIGNED(8) AranyaAfcReceiveChannel {
      * UNDEFINED BEHAVIOR to read from or write to it.
      * @private
      */
-    uint8_t __for_size_only[96];
+    uint8_t __for_size_only[120];
 } AranyaAfcReceiveChannel;
 #endif
 
@@ -584,6 +584,15 @@ typedef struct ARANYA_ALIGNED(8) AranyaAfcSeq {
      */
     uint8_t __for_size_only[24];
 } AranyaAfcSeq;
+#endif
+
+#if defined(ENABLE_ARANYA_AFC)
+/**
+ * Channel ID for AFC channel.
+ */
+typedef struct AranyaAfcChannelId {
+    struct AranyaId id;
+} AranyaAfcChannelId;
 #endif
 
 #ifdef __cplusplus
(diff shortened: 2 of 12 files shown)


examples/rust/aranya-example-multi-node/example.env
```diff
@@ -1,17 +1,17 @@
 export ARANYA_EXAMPLE=info
 export ARANYA_ONBOARDING_PASSPHRASE=passphrase
 export ARANYA_SYNC_ADDR_OWNER=127.0.0.1:10002
-export ARANYA_AQC_ADDR_OWNER=127.0.0.1:10000
+export ARANYA_AFC_ADDR_OWNER=127.0.0.1:10000
 export ARANYA_TCP_ADDR_OWNER=127.0.0.1:10001
```
gknopf-aranya, line 5: If we're sending AFC messages with UDP, is there a reason to onboard with TCP? Seems simpler to limit the example to 1 protocol.

examples/rust/aranya-example-multi-node/src/membera/main.rs
```diff
@@ -156,41 +152,38 @@ async fn main() -> Result<()> {
...
-        .expect("expected to create bidi channel");
+    let memberb = get_member_peer(&client, team_info.team_id).await?;
 
-    // Create stream on bidi channel.
-    info!("membera: creating bidi stream");
-    let mut stream = chan
-        .create_bidi_stream()
-        .await
-        .expect("expected to create bidi stream");
+    let receiver = TcpServer::bind(env.membera.afc_addr).await?;
```
gknopf-aranya, line 157: @djohnson-spideroak, @jdygert-spok I thought we agreed to use UDP rather than TCP for the examples:
https://github.com/aranya-project/aranya/issues/522

examples/rust/aranya-example-multi-node/src/tcp.rs
```diff
@@ -35,28 +35,27 @@ impl TcpServer {
 }
 
 /// Simple TCP client for demo purposes.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct TcpClient {
-    stream: TcpStream,
+    _priv: (),
```
gknopf-aranya, line 40: What is this field for?

examples/rust/aranya-example-multi-node/src/membera/main.rs
jdygert-spok: I wanted to do the minimal changes here to get the example working. If the reason for choosing udp was simplicity then I think this is simpler.

examples/rust/aranya-example-multi-node/src/tcp.rs
jdygert-spok: Making the type internals private so you don't construct it manually.

I changed it to connect when sending since that matches up with what the server expects with read_to_end.

examples/rust/aranya-example-multi-node/src/lib.rs
```diff
@@ -6,3 +6,24 @@ pub mod env;
...
+    let this_device = client.get_device_id().await?;
+    let team = client.team(team);
+    let queries = team.queries();
+    let devices = queries.devices_on_team().await?;
+    for &device in devices.iter() {
+        if device.__id == this_device.__id {
+            continue;
+        }
+        let role = queries.device_role(device).await?;
+        if role == aranya_client::client::Role::Member {
```
gknopf-aranya, line 24: This assumes there are only 2 member devices on the team. I'd prefer to query based on which devices are assigned to the label.

jdygert-spok: That assumes only one other device is assigned the label, which isn't much better.

jdygert-spok: Thinking I will have member B just send its ID over tcp first.

crates/aranya-client/src/client.rs
gknopf-aranya, line 1: We can remove all `NetIdentifier` code from types, APIs, policy etc. since AFC doesn't use it:
```
/// A device's network identifier.
#[derive(Clone, Debug, Serialize, Deserialize, Eq, Ord, PartialEq, PartialOrd)]
pub struct NetIdentifier(pub(crate) api::NetIdentifier);

impl FromStr for NetIdentifier {
    fn from_str(s: &str) -> anyhow::Result<Self> {
        Ok(Self(api::NetIdentifier(s.to_string().try_into()?)))
    }

    type Err = anyhow::Error;
}

impl Display for NetIdentifier {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        Display::fmt(&self.0, f)
    }
}
```

crates/aranya-daemon/src/policy.md
gknopf-aranya, line 1: We should clarify that `SendRecv` means a device can either send or receive data in a channel with that operation permitted by the label assignment. The device cannot send and receive data in the same channel.

examples/rust/aranya-example-multi-node/src/lib.rs
gknopf-aranya: Will it be encrypted with `age`?

gknopf-aranya:
I took a closer look and other than some net identifier remnants, I don't see any other unused code leftover from AQC.
Our docs on `SendRecv` permissions are out of date because we can only send or receive on a channel not send and receive.

examples/rust/aranya-example-multi-node/src/lib.rs
jdygert-spok: I suppose. Need to add peer device id getter to channel too for the other side so lets do this in a follow up PR since this works for now.

crates/aranya-daemon/src/policy.md
jdygert-spok: Changed to "send or receive". I think that's fine since it's about the label anyway and not a given channel.

examples/rust/aranya-example-multi-node/src/lib.rs
jdygert-spok: #538

djohnson-spideroak (approved):

PR #539, opened 2025-10-21
chore: update to latest aranya-core

jdygert-spok:


supply-chain/config.toml
```diff
@@ -158,14 +158,6 @@ criteria = "safe-to-run"
 version = "0.2.2"
 criteria = "safe-to-deploy"
 
-[[exemptions.ciborium-io]]
```
gknopf-aranya, line 161: What caused these exemptions to no longer be needed?

crates/aranya-daemon/src/api.rs
```diff
@@ -324,7 +325,7 @@ impl Api {
     /// Checks wither a team's graph is valid.
     /// If the graph is not valid, return an error to prevent operations on the invalid graph.
     async fn check_team_valid(&self, team: api::TeamId) -> anyhow::Result<()> {
-        if self.invalid.contains(team.into_id().into()) {
+        if self.invalid.contains(GraphId::transmute(team)) {
```
gknopf-aranya, line 328: Seems like we could put `transmute` into a `From` impl so `.into()` could be used for the type conversion.

gknopf-aranya: Or is the reason for doing this to make the conversion more explicit?

crates/aranya-daemon/src/afc.rs
gknopf-aranya, line 1: We can integrate with this change to remove if params:
https://github.com/aranya-project/aranya-core/pull/466

supply-chain/config.toml
jdygert-spok: Reviews from our imported reviewers. I just did `cargo vet prune`

crates/aranya-daemon/src/api.rs
jdygert-spok: Yes, more explicit.

crates/aranya-keygen/src/keygen.rs
```diff
@@ -139,16 +139,11 @@ impl KeyBundle {
         macro_rules! gen {
             ($key:ident) => {{
                 let sk = $key::<E::CS>::new(eng);
-                let id = sk.id()?;
-                let wrapped =
-                    eng.wrap(sk)
-                        .context(concat!("unable to wrap `", stringify!($key), "`"))?;
-                store.try_insert(id.into(), wrapped).context(concat!(
-                    "unable to insert wrapped `",
+                store.insert_key(eng, sk).context(concat!(
```
djohnson-spideroak, line 142: doesnt need to be wrapped anymore?

jdygert-spok: `insert_key` does the wrapping

jdygert-spok: https://github.com/aranya-project/aranya-core/pull/361

crates/aranya-daemon/src/actions.rs
```diff
@@ -54,19 +50,16 @@ where
...
-            .new_graph(
-                &[0u8],
-                vm_action!(create_team(
-                    owner_keys,
-                    nonce.unwrap_or(&Rng.bytes::<[u8; 16]>()),
-                )),
-                &mut sink,
-            )
-            .context("unable to create new team")?;
+        let policy_data = &[0u8];
```
djohnson-spideroak, line 53: I see this is used in the id assignment below, what is it for?

jdygert-spok: It's eventually meant to be the serialized policy module. Currently it's just used to determine the policy ID.

(This is unchanged, I just moved it and gave it a name.)

chip-so: It is currently meaningless. 😅

Issue #533, opened 2025-10-17
custom roles: afc channel deletion with custom roles policy


gknopf-aranya:
We need a good way to delete AFC channels from the daemon shared-memory when an effect is processed that results in the channel no longer being valid.

With the AFC policy, there are 2 effects that could result in a channel no longer being valid:
- `LabelRevoked`
- `LabelDeleted`

Now that I'm thinking about it more, `TeamTerminated`, `MemberRemoved` are some more effects that could trigger channel deletion.

When the custom roles policy is added, there will be more possible triggers for channel deletion (because of roles and perms):
1. Role loses label
2. Role loses afc perm
3. User loses role
4. Role is deleted

It would be ideal if we could tie channel deletion directly to the policy rather than implementing a complex effects handler in the daemon for this. Whatever effects handler we implement could get out-of-date as the policy is updated.

One option is to periodically query the policy for which channels are valid, then delete any invalid channels.
The drawback of this is that this introduces querying/polling (less efficient) where we used to be able to handle effects right away.

Another option is for the policy to emit extra `ChannelDeleted` effects for any commands which could result in a channel being deleted. This would allow the daemon to continue handling channel deletion in the effects handler, while simplifying the implementation. The daemon would only need to know which effect(s) correspond to channel deletion and not all the effects that could have a side-affect of a channel no longer being valid.

Original discussion:
https://github.com/aranya-project/aranya/pull/527#discussion_r2436690849

cc: @djohnson-spideroak, @jdygert-spok, @chip-so

jdygert-spok:
> > Note that the remove_if callback is called twice for each entry and must return the same result both times.
> 
> How would it return different answers if it was done in the same session where we did not update the state?

Realistically, probably just a spurious IO error on the graph storage or keystore. If something like that occurred, we'd either need to retry until success or nuke the shm.

Except now that I am thinking about it, can't we just memcpy the updated half after swapping and locking? (Once we have the seq num fixes)

PR #267, opened 2025-05-27
roles: custom ID-based roles with RBAC permission controls

elagergren-spideroak:
Introduces custom ID-based roles with RBAC permission controls to the policy.
Exposes corresponding APIs to the application.

TODOs:
- [x] Resolve upstream merge conflicts
- [x] Rust example (https://github.com/aranya-project/aranya/pull/567, https://github.com/aranya-project/aranya/pull/556)
- [x] C example and C API (https://github.com/aranya-project/aranya/pull/568)
- [x] Resolve remaining PR feedback
- [x] Resolve open TODOs added to code on PR
- [x] Evaluate test coverage (https://github.com/aranya-project/aranya/issues/569)
- [x] Remove label assignments from roles (https://github.com/aranya-project/aranya/pull/561)
- [x] AFC channel deletion based on custom roles (https://github.com/aranya-project/aranya/pull/542)
- [x] Command priorities (https://github.com/aranya-project/aranya/pull/425)

PR #542, opened 2025-10-22
custom roles: afc channel deletion

gknopf-aranya:
Emits a `CheckValidAfcChannels` effect for any commands which could result in deletion of AFC channels:
- `TeamTerminated`
- `RoleAssigned`
- `RoleRevoked`
- `RoleChanged`
- `DeviceRemoved`
- `LabelDeleted`
- `LabelRevokedFromDevice`

When handling the `CheckValidAfcChannels` effect, the daemon iterates over AFC channel entries in the shared-memory (via the `remove_if()` operation), deleting any channels that are no longer valid according to the `query_afc_channel_is_valid(sender_id, receiver_id, label_id)` query.

### DIFF
--- a/crates/aranya-daemon/src/policy.md
+++ b/crates/aranya-daemon/src/policy.md
@@ -1306,6 +1306,7 @@ fact CanRevokeRole[target_role_id id, managing_role_id id]=>{}
 //
 // If true, `CanRevokeRole(target_role_id, device_role_id)` holds.
 function can_revoke_role(device_id id, target_role_id id) bool {
+    check device_has_simple_perm(device_id, SimplePerm::RevokeRole)
     let device_role_id = get_assigned_role_id(device_id)
 
     // At this point we believe the following to be true:
@@ -1466,7 +1467,7 @@ command AssignRoleManagementPerm {
         check team_exists()
 
         let author = get_author(envelope)
-        check device_has_simple_perm(author.device_id, SimplePerm::CanChangeRolePerms)
+        // TODO check device_has_simple_perm(author.device_id, SimplePerm::CanChangeRolePerms)
         check device_owns_role(author.device_id, this.target_role_id)
 
         // Make sure we uphold the invariants for
@@ -1612,7 +1613,7 @@ command RevokeRoleManagementPerm {
         check team_exists()
 
         let author = get_author(envelope)
-        check device_has_simple_perm(author.device_id, SimplePerm::CanChangeRolePerms)
+        // TODO check device_has_simple_perm(author.device_id, SimplePerm::CanChangeRolePerms)
         check device_owns_role(author.device_id, this.target_role_id)
 
         let perm = role_management_perm_to_str(this.perm)
@@ -1946,7 +1947,8 @@ command SetupDefaultRole {
 
 ### Role Deletion
 
-TODO
+Role deletion is waiting on prefix query deletion (see
+https://github.com/aranya-project/aranya-core/issues/229)
 
 ### Role Assignment
 
@@ -2251,7 +2253,6 @@ command ChangeRole {
         // The author must have permission to revoke the old role.
         let old_role = check_unwrap query Role[role_id: this.old_role_id]
         check can_revoke_role(author.device_id, this.old_role_id)
-        check device_has_simple_perm(author.device_id, SimplePerm::RevokeRole)
 
         // The author must have permission to assign the new role.
         check exists Role[role_id: this.new_role_id]
@@ -2354,7 +2355,6 @@ command RevokeRole {
         let author = get_author(envelope)
 
         // The author must have permission to revoke the role.
-        check device_has_simple_perm(author.device_id, SimplePerm::RevokeRole)
         check can_revoke_role(author.device_id, this.role_id)
 
         let role = check_unwrap query Role[role_id: this.role_id]


crates/aranya-daemon/src/api.rs
```diff
@@ -228,27 +236,44 @@ impl EffectHandler {
...
+                RoleAssigned(_role_assigned) => {
+                    #[cfg(feature = "afc")]
+                    {
+                        // Delete channels for devices where new role does not have the `CanUseAfc` permission)
+                        if !self
+                            .client
+                            .query_role_has_perm(
+                                graph.into_id().into(),
+                                _role_assigned.role_id.into(),
+                                "CanUseAfc".into(),
```
gknopf-aranya: Would prefer not to use a string here for the perm

crates/aranya-daemon/src/api.rs
```diff
@@ -287,21 +316,84 @@ impl EffectHandler {
...
                 PermAddedToRole(_) => {}
-                PermRemovedFromRole(_) => {}
+                PermRemovedFromRole(_perm_removed_from_role) => {
+                    #[cfg(feature = "afc")]
+                    {
+                        // Delete channels where `CanUseAfc` permission has been revoked.
+                        if _perm_removed_from_role.perm == "CanUseAfc" {
+                            let devices = self
+                                .client
+                                .query_devices_on_team(graph.into_id().into())
```
gknopf-aranya: This queries for a list of devices which means a device could be missed by the query if `PermRemovedFromRole` and `DeviceRemoved` happen back-to-back. This is acceptable because AFC channels will be deleted from the device when `DeviceRemoved` is handled later.

crates/aranya-daemon/src/api.rs
```diff
@@ -228,27 +236,44 @@ impl EffectHandler {
...
+                            .await?;
+                    }
+                }
+                RoleAssigned(_role_assigned) => {
+                    #[cfg(feature = "afc")]
+                    {
+                        // Delete channels for devices where new role does not have the `CanUseAfc` permission)
+                        if !self
+                            .client
+                            .query_role_has_perm(
```
gknopf-aranya: If a role is assigned then the `CanUseAfc` permission changed right after, this query could be out of date.
That is not a problem though, because if `CanUseAfc` is revoked from the role before the query is run, the corresponding AFC channels will be deleted here as expected.
If `CanUseAfc` is added to the role later, AFC channels will be deleted as expected (if you do not want any AFC channels to be deleted you must assign `CanUseAfc` to the role before assigning the role to a device).
We will also catch any AFC channel deletions that may have been missed when handling the `PermRemovedFromRole` effect.

djohnson-spideroak: Lets make sure we have clear documentation on when channels are/should be revoked, especially in those less intuitive cases.

gknopf-aranya: Will that documentation be inlined here? Or in a spec somewhere?

crates/aranya-daemon/src/api.rs
```diff
@@ -287,21 +316,84 @@ impl EffectHandler {
...
+                                        device.into_id().into(),
+                                    )
+                                    .await?
+                                {
+                                    if role.id == _perm_removed_from_role.role_id.into() {
+                                        self.delete_device_afc_channels(device.into_id().into())
+                                            .await?;
+                                    }
+                                }
+                            }
```
djohnson-spideroak: Is there a more efficient way to do this? If we query all devices, then query the role for each device, could we combine this in a better query?

gknopf-aranya: I think we could write a custom query in the policy that queries all the devices on a team, then looks up the role and whether it has the `CanUseAfc` perm before returning the effect.

crates/aranya-daemon/src/api.rs
gknopf-aranya: Implementing this more efficient query would simplify the rationale here by running both queries on the same graph state.
https://github.com/aranya-project/aranya/pull/542#discussion_r2461486724

crates/aranya-daemon/src/api.rs
gknopf-aranya: Added a `QueryDevicesWithRole` query to return devices on the team matching a particular role

crates/aranya-daemon/src/api.rs
gknopf-aranya: Added additional docs inline

crates/aranya-daemon/src/api.rs
gknopf-aranya: Implemented the `QueryDevicesWithRole` query so this comment no longer applies.

crates/aranya-daemon/src/api.rs
gknopf-aranya: Probably not worth changing this given the tight timeline we want to land custom roles on

gknopf-aranya: Adding some todos to resolve this later.

crates/aranya-client/tests/afc.rs
```diff
@@ -456,6 +456,21 @@ async fn test_afc_uni_chan_delete_label() -> Result<()> {
     Ok(())
 }
 
+// TODO: demonstrate terminating the team deletes the AFC channel.
```
gknopf-aranya: @djohnson-spideroak, @chip-so I was about to start implementing these tests but realized the afc unit tests are broken on the base custom roles PR. `cargo test afc --features afc,preview`:
```
2025-10-24T22:03:33.163569Z  INFO tarpc::client: Shutdown: write half closed, and no requests in flight.
2025-10-24T22:03:33.164351Z  INFO daemon{name="memberb"}:api-server: aranya_daemon::api: effect handler exiting
Error: daemon error

Caused by:
    unable to create label

    Caused by:
        not authorized


failures:
    test_afc_uni_chan_create
    test_afc_uni_chan_delete
    test_afc_uni_chan_delete_label
    test_afc_uni_chan_remove_device
    test_afc_uni_chan_revoke_label
    test_afc_uni_multi_send_chans
    test_afc_uni_send_chan_seal_open

test result: FAILED. 0 passed; 7 failed; 0 ignored; 0 measured; 0 filtered out; finished
```
This means at least the AFC unit tests will need to be fixed before I can start adding these new tests.

gknopf-aranya: Solved by https://github.com/aranya-project/aranya/issues/552

crates/aranya-client/src/client.rs
```diff
@@ -63,7 +63,7 @@ impl ClientBuilder<'_> {
     /// # #[tokio::main]
     /// # async fn main() -> anyhow::Result<()> {
     /// let client = Client::builder()
-    ///     .daemon_uds_path("/var/run/aranya/uds.sock".as_ref())
+    ///     .with_daemon_uds_path("/var/run/aranya/uds.sock".as_ref())
```
gknopf-aranya, line 66: @chip-so I put this change here to get the unit tests to compile. Would prefer this be done on the base PR.

chip-so: Alright, will do.

chip-so: Done. https://github.com/aranya-project/aranya/pull/267/commits/07e1d74a393a0f2dca1163d232175c94bef09f8b

gknopf-aranya: Thank you!

crates/aranya-client/tests/afc.rs
```diff
@@ -450,6 +450,347 @@ async fn test_afc_uni_chan_delete_label() -> Result<()> {
...
+        .context("unable to seal afc message")
+        .expect_err("expected seal to fail");
+    recv.open(&mut plaintext, &ciphertext)
+        .context("unable to open afc message")
+        .expect_err("expected open to fail");
+
+    Ok(())
+}
+
+// TODO: Demonstrate changing the device's role to a new role without `CanUseAfc` perm deletes the AFC channel.
```
gknopf-aranya: Blocked on adding `change_role()` to IPC API

crates/aranya-client/tests/afc.rs
```diff
@@ -450,6 +450,347 @@ async fn test_afc_uni_chan_delete_label() -> Result<()> {
...
+
+// TODO: Demonstrate removing the `CanUseAfc` perm from a device's role deletes the AFC channel.
```
gknopf-aranya: Blocked on adding `remove_perm_from_role()` to IPC API

crates/aranya-client/tests/afc.rs
```diff
@@ -450,6 +450,347 @@ async fn test_afc_uni_chan_delete_label() -> Result<()> {
...
+        .admin
+        .client
+        .team(team_id)
+        .sync_now(operator_addr, None)
+        .await?;
+
+    // Assign roles without `CanUseAfc` permission.
+    // TODO: this shouldn't work because assigning a role should be rejected if the device already has a role.
+    owner_team
+        .assign_role(devices.membera.id, default_roles.operator().id)
```
gknopf-aranya: @djohnson-spideroak, @chip-so this role assignment shouldn't have worked according to the `assign_role()` docs. Because the channel devices already have the member role assigned, assigning the operator role to them should fail. For devices that already have a role assigned, `change_role()` is the operation we should be using.
```
// Assigns the specified role to the device.
//
// It is an error if the device has already been assigned a role.
//
// # Required Permissions
//
// - `AssignRole`
// - `CanAssignRole(role_id)`
action assign_role(device_id id, role_id id) {
    publish AssignRole {
        device_id: device_id,
        role_id: role_id,
    }
}
```

crates/aranya-client/tests/afc.rs
gknopf-aranya: Closing since changing role perms isn't supported yet with default roles

crates/aranya-client/tests/afc.rs
gknopf-aranya: Filed a bug report here:
https://github.com/aranya-project/aranya/issues/557
Doesn't block this PR since I'm switching over to `change_role()`

crates/aranya-client/src/client/team.rs
```diff
@@ -302,6 +302,28 @@ impl Team<'_> {
             .map_err(aranya_error)
     }
 
+    /// Changes `role` assigned to `device`.
+    #[instrument(skip(self))]
+    pub async fn change_role(
```
gknopf-aranya: @chip-so this change could go on the base PR

crates/aranya-client/tests/afc.rs
```diff
@@ -450,6 +450,351 @@ async fn test_afc_uni_chan_delete_label() -> Result<()> {
...
+    let operator_addr = devices.operator.aranya_local_addr().await?.into();
+    devices
+        .admin
+        .client
+        .team(team_id)
+        .sync_now(operator_addr, None)
+        .await?;
+
+    // Assign roles without `CanUseAfc` permission.
+    // TODO: this shouldn't work because assigning a role should be rejected if the device already has a role.
```
gknopf-aranya:
```suggestion
```

crates/aranya-client/src/client/team.rs
chip-so: Just pushed this with the duplicate `assign_role()` test.

crates/aranya-daemon/src/api.rs
```diff
@@ -372,6 +465,239 @@ impl Api {
...
+        Ok(roles)
+    }
+
+    // Query role has permission.
+    #[cfg(feature = "afc")]
+    async fn query_role_has_perm(
+        &self,
+        team: api::TeamId,
+        role_id: api::RoleId,
+        perm: String,
```
chip-so: Maybe there's a convention I don't know about here. Does this need to be owned? Could it just be `&str`?

crates/aranya-daemon/src/api.rs
```diff
@@ -615,23 +941,11 @@ impl DaemonApi for Api {
...
-                if let Effect::QueryDevicesOnTeamResult(e) = e {
-                    Some(e.device_id.into())
-                } else {
-                    warn!(name = e.name(), "unexpected effect");
-                    None
-                }
-            })
-            .collect();
-        Ok(devices)
+            .query_devices_on_team(team)
```
chip-so: Can you explain a little about why this got pushed down into `Client`? I don't understand the architectural consequences here.

crates/aranya-daemon/src/policy.md
```diff
@@ -459,6 +459,67 @@ ephemeral command QueryDevicesOnTeam {
...
+```policy
+// Emits `QueryDevicesWithRole` for each device on the team with specified role.
+ephemeral action query_devices_with_role(role_id id) {
+    // Publishing `QueryDevicesWithRole` emits
+    // `QueryDevicesWithRoleResult`.
+    map Device[device_id: ?] as f {
+        publish QueryDevicesWithRole {
+            device_id: f.device_id,
+            role_id: role_id,
+        }
```
chip-so: Isn't this just `query_devices_on_team()` except it passes through `role_id` into every effect? Is that just because the effect is processed in a different place that doesn't have the `role_id` handy?

chip-so: Oh nevermind. I see the filtering below.

crates/aranya-daemon/src/policy.md
```diff
@@ -459,6 +459,67 @@ ephemeral command QueryDevicesOnTeam {
...
+
+// Emitted when a device is queried by `query_devices_with_role`.
+effect QueryDevicesWithRoleResult {
+    // The ID of a device on the team.
+    device_id id,
+    // The ID of the role.
+    role_id id,
+}
+
+// A trampoline that forwards `device_id` to the effect.
```
chip-so: Since this is doing some filtering logic, I wouldn't describe it as a trampoline.

crates/aranya-daemon/src/api.rs
gknopf-aranya: I moved these queries into the `Client` so I could run them from either the API object or the effect handler. The effect handler already has access to the `Client` object.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Good point. This was likely a copy-paste error from another query.

crates/aranya-daemon/src/api.rs
gknopf-aranya: Since this is part of the daemon's public IPC API, the string will need to be copied to into the client application. I don't think there's a way to pass `&str` between different applications via the API.

gknopf-aranya: I don't think it would hurt to make a type alias for this

jdygert-spok: This is on the `Client`, not in the api. As-is, this should be `Text`.

gknopf-aranya: True, that's what is passed into the API originally. Updating this.

chip-so (approved):

crates/aranya-daemon/src/policy.md
```diff
@@ -490,6 +490,67 @@ ephemeral command QueryDevicesOnTeam {
...
+ephemeral action query_devices_with_role(role_id id) {
+    // Publishing `QueryDevicesWithRole` emits
+    // `QueryDevicesWithRoleResult`.
+    map Device[device_id: ?] as f {
+        publish QueryDevicesWithRole {
+            device_id: f.device_id,
+            role_id: role_id,
+        }
+    }
+}
```
jdygert-spok: Use `map RoleAssignmentIndex[role_id: role_id, device_id: ?]`, then no filtering is needed.

gknopf-aranya: Good point. Updated this.

crates/aranya-daemon/src/policy.md
```diff
@@ -530,21 +530,10 @@ ephemeral command QueryDevicesWithRole {
...
-                        device_id: this.device_id,
-                        role_id: this.role_id,
-                    }
-                }
-            } else {
-                finish {}
+        finish {
+            emit QueryDevicesWithRoleResult {
+                device_id: this.device_id,
+                role_id: this.role_id,
```
chip-so: Okay now it's actually a trampoline. 😆

chip-so: Also since `this` and `QueryDevicesWithRoleResult` have the same shape you can use struct conversion to save a bit of work here:
```
let eff = this as QueryDevicesWithRoleResult
finish {
    emit eff
}
```

chip-so: (This can be applied in a lot of places in the policy so no worries if you just want to leave it. I'll be going over this in another pass later.)

gknopf-aranya: That's cool. Didn't realize this was a possibility.

gknopf-aranya: Updated

crates/aranya-daemon/src/api.rs
```diff
@@ -369,6 +458,239 @@ impl Api {
     }
 }
 
+impl Client {
+    // Query devices on team.
+    async fn query_devices_on_team(&self, team: api::TeamId) -> api::Result<Vec<api::DeviceId>> {
```
jdygert-spok: I don't like that we are adding methods on `Client` that use `aranya_daemon_api` types. IMO we should keep those to the API handler. The client methods can e.g. take `GraphId` and return the effects. (See #560.)

crates/aranya-daemon/src/api.rs
```diff
@@ -285,20 +311,83 @@ impl EffectHandler {
                 LabelManagingRoleAdded(_) => {}
                 LabelManagingRoleRevoked(_) => {}
                 PermAddedToRole(_) => {}
-                PermRemovedFromRole(_) => {}
+                PermRemovedFromRole(_perm_removed_from_role) => {
+                    #[cfg(feature = "afc")]
+                    {
+                        // Delete channels where `CanUseAfc` permission has been revoked.
+                        // This queries for a list of devices which means a device could be missed by the query if `PermRemovedFromRole`` and `DeviceRemoved`` happen back-to-back.
+                        // This is acceptable because AFC channels will be deleted from the device when `DeviceRemoved` is handled later.
```
jdygert-spok: These seem fine at a glance but I also remember "we shouldn't query as the result of an effect" being said in our last rust call. We should double check there isn't a weird case that will cause issues.

crates/aranya-daemon/src/api.rs
jdygert-spok: IOW conversions to/from API types should be done in the api handler methods.

crates/aranya-daemon/src/api.rs
gknopf-aranya: That was especially true if multiple queries needed to be run back-to-back. We've got each channel deletion down to a single query now.

gknopf-aranya: I've left docs here explaining why these queries should result in the expected channel deletions

crates/aranya-daemon/src/api.rs
gknopf-aranya: I need to run these queries from the effects handler. If they are implemented on the API object, it doesn't make sense to pass the entire API object into the effects handler in order to run the queries. That's why I implemented the queries on the `Client` object which is already available to the effect handler. If I get rid of the `api::*` types from these methods, then I don't have objects for `KeyBundle`, `Role` to put the parameters/results into without defining new types.

gknopf-aranya: I suppose I could return the effects for `KeyBundle` and `Role` to the API methods so the type conversion can be performed from the effect to the API type there.

gknopf-aranya: Removed api types from `Client` queries

crates/aranya-daemon/src/api.rs
gknopf-aranya: I've implemented a different approach discussed here:
https://github.com/aranya-project/aranya/issues/533

crates/aranya-client/tests/afc.rs
```diff
@@ -450,6 +450,350 @@ async fn test_afc_uni_chan_delete_label() -> Result<()> {
     Ok(())
 }
 
+// TODO: demonstrate terminating the team deletes the AFC channel once `close_team()` is implemented.
```
gknopf-aranya: Debatable if this is needed since a `CheckValidAfcChannels` effect is emitted right after `TeamTerminated`. The symmetry in how `CheckValidAfcChannels` is handled means the policy can be audited to determine this property rather than requiring a unit test for each case.

crates/aranya-daemon/src/actions.rs
```diff
@@ -353,6 +355,27 @@ where
         .in_current_span()
     }
 
+    /// Invokes `query_afc_channel_is_valid`.
+    #[allow(clippy::type_complexity)]
+    #[instrument(skip(self))]
+    fn query_afc_channel_is_valid(
```
gknopf-aranya: unused

crates/aranya-daemon/src/actions.rs
```diff
@@ -621,3 +644,35 @@ impl<CS: aranya_crypto::CipherSuite> TryFrom<&PublicKeys<CS>> for KeyBundle {
         })
     }
 }
+
+#[cfg(feature = "afc")]
+pub(crate) fn query_valid_afc<EN, SP, CE>(
```
gknopf-aranya:
```suggestion
pub(crate) fn query_afc_channel_is_valid<EN, SP, CE>(
```

crates/aranya-daemon/src/actions.rs
```diff
@@ -621,3 +644,35 @@ impl<CS: aranya_crypto::CipherSuite> TryFrom<&PublicKeys<CS>> for KeyBundle {
...
+    )?;
+    let effects = sink.collect()?;
+    for e in effects {
+        if let Effect::QueryAfcChannelIsValidResult(e) = e {
+            if e.is_valid {
+                return Ok(true);
+            }
+        }
+    }
+    Ok(false)
```
gknopf-aranya:
```suggestion
    Ok(effects.iter().any(|e| {
        if let Effect::QueryAfcChannelIsValidResult(e) = e {
            return e.is_valid;
        }
        false
    }))
```

crates/aranya-daemon/src/afc.rs
```diff
@@ -205,28 +200,41 @@ where
...
-            .context("unable to remove AFC channels")
-    }
-
-    /// Remove channels matching criteria.
-    pub(crate) async fn remove_if(&self, params: RemoveIfParams) -> Result<()> {
+    /// Remove channels that are no longer valid from shared-memory.
+    pub(crate) async fn remove_invalid_channels(
+        &self,
+        graph: GraphId,
+        device_id: DeviceId,
```
gknopf-aranya, line 211: Could make `device_id` a field of `self.device_id` since it will never change. I don't think this is worth spending time to do on the current PR.

crates/aranya-daemon/src/afc.rs
```diff
@@ -205,28 +200,41 @@ where
...
         let shm = self.shm.lock().await;
 
+        let mut client = self.client.aranya.lock().await;
         shm.write
             .remove_if(|chan| {
-                params.channel_id.is_none_or(|id| chan.channel_id == id)
-                    && params.label_id.is_none_or(|id| chan.label_id == id)
-                    && params.peer_id.is_none_or(|id| chan.peer_id == id)
+                // TODO: run a single query when channel direction is added to remove_if() params
+                if let Ok(is_valid) = crate::actions::query_valid_afc(
```
gknopf-aranya: Going to simplify this logic

crates/aranya-daemon/src/afc.rs
```diff
@@ -205,28 +200,41 @@ where
...
+                    graph,
+                    chan.peer_id,
+                    device_id,
+                    chan.label_id,
+                ) {
+                    if is_valid {
+                        return false;
+                    }
+                }
+                true
```
gknopf-aranya:
```suggestion
                let sender_valid = crate::actions::query_afc_channel_is_valid(
                    &mut client,
                    graph,
                    device_id,
                    chan.peer_id,
                    chan.label_id,
                )
                .is_ok_and(|v| v);

                let receiver_valid = crate::actions::query_afc_channel_is_valid(
                    &mut client,
                    graph,
                    chan.peer_id,
                    device_id,
                    chan.label_id,
                )
                .is_ok_and(|v| v);

                !sender_valid && !receiver_valid
```

crates/aranya-daemon/src/policy.md
```diff
@@ -490,6 +490,68 @@ ephemeral command QueryDevicesOnTeam {
...
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        check team_exists()
+
+        // Check that both devices have permission to create the AFC channel.
+        let is_valid = afc_uni_channel_is_valid(this.sender_id, this.receiver_id, this.label_id)
+        finish {
+            // TODO: substruct selection from `this`
```
gknopf-aranya, line 543: @jdygert-spok, @chip-so, @ssweetney-spok could the conversion from command inputs to effect be simplified with substruct selection?

crates/aranya-daemon/src/afc.rs
```diff
@@ -205,28 +200,38 @@ where
...
+                let receiver_valid = crate::actions::query_afc_channel_is_valid(
+                    &mut client,
+                    graph,
+                    chan.peer_id,
+                    device_id,
+                    chan.label_id,
+                )
+                .is_ok_and(|v| v);
+
+                !sender_valid && !receiver_valid
```
jdygert-spok: As mentioned in #533, if a query gets a spurious error from IO or something, the shm will end up in an inconsistent state. I think we could change AFC itself to avoid this. We can either accept this as is for now or we can add a cache here.

crates/aranya-daemon/src/policy.md
jdygert-spok:
```
            emit QueryAfcChannelIsValidResult {
                is_valid: is_valid,
                ...this,
            }
```

But I also don't think you need any of the other fields...

gknopf-aranya: The other fields aren't used at the moment. I'm following a convention in the policy which seems to be to include extra fields just in case.

crates/aranya-daemon/src/afc.rs
gknopf-aranya: I don't think we should land something on main that could result in an inconsistent shm state. If we add the cache in aranya-core, then the aranya-core release becomes a blocker for custom roles. This is something we've been trying to avoid. So it may be more expedient to implement the cache here.

djohnson-spideroak: How complex would the caching be vs the other AFC changes?

jdygert-spok: Added cache.

crates/aranya-daemon/src/afc.rs
```diff
@@ -205,28 +203,46 @@ where
...
+    /// Hold `Client` lock so all queries are run on the same graph state.
+    pub(crate) async fn remove_invalid_channels(
+        &self,
+        graph: GraphId,
+        device_id: DeviceId,
+    ) -> Result<()> {
         let shm = self.shm.lock().await;
+        let mut client = self.client.aranya.lock().await;
+
+        // Use cache since remove_if callback currently must always produce same result.
```
gknopf-aranya: Let's leave a todo referencing the aranya-core issue number here

crates/aranya-daemon/src/afc.rs
```diff
@@ -205,28 +203,46 @@ where
...
             .remove_if(|chan| {
-                params.channel_id.is_none_or(|id| chan.channel_id == id)
-                    && params.label_id.is_none_or(|id| chan.label_id == id)
-                    && params.peer_id.is_none_or(|id| chan.peer_id == id)
+                let e = match cache.entry((chan.peer_id, chan.label_id)) {
+                    Entry::Occupied(e) => return *e.get(),
+                    Entry::Vacant(e) => e,
+                };
+
+                // TODO: run a single query when channel direction is added to remove_if() params
```
gknopf-aranya: Let's reference the aranya-core issue number here

crates/aranya-daemon/src/afc.rs
```diff
@@ -210,8 +213,16 @@ where
         let shm = self.shm.lock().await;
         let mut client = self.client.aranya.lock().await;
 
+        // Use cache since remove_if callback currently must always produce same result.
+        let mut cache = HashMap::new();
+
         shm.write
             .remove_if(|chan| {
+                let e = match cache.entry((chan.peer_id, chan.label_id)) {
```
gknopf-aranya, line 221: What if we have more than one channel with the same peer? Do we need `channel_id` in the cache key?

jdygert-spok: We only need to cache on the inputs we actually use.

crates/aranya-daemon/src/afc.rs
```diff
@@ -231,7 +242,7 @@ where
                 )
                 .is_ok_and(|v| v);
 
-                !sender_valid && !receiver_valid
+                *e.insert(!sender_valid && !receiver_valid)
```
gknopf-aranya: It's a little less clear here that `!sender_valid && !receiver_valid` is being returned without looking up the `insert` docs.

crates/aranya-daemon/src/afc.rs
gknopf-aranya: We do use the current device's `device_id`

gknopf-aranya: That's a constant though across all the operations so shouldn't be needed in the key.

PR #543, opened 2025-10-23
Combine `ClientWithState` with `Client` and encapsulate fields

jdygert-spok:


jdygert-spok:
By not taking in or returning any `Arc`s, we can know that all access to the underlying client, caches, and subscriptions are done through `Client` rather than some arc cloned and stashed somewhere.

This seems iffy at first given that `DaemonApi` doesn't need those other parts, but it does seem convenient to be able to put other shared handlers in there. e.g. I think `InvalidGraphs` can be put there in the same way, which is shared by sync client and daemon.

jdygert-spok:
Probably want to see where @nikki-aranya 's branch ends up.

crates/aranya-daemon/src/aranya.rs
gknopf-aranya, line 1: Aranya refers to the entire product we have built including the Aranya client and Aranya daemon. The naming of this aranya.rs module seems to be from when the Aranya graph was the entire product. Would a more suitable name for this module be graph.rs? I'd prefer not to call it client.rs since we already have networking client modules in the codebase.

gknopf-aranya (approved):
The changes themselves look good. Naming of the aranya.rs module seems less meaningful given how our product has evolved. `Client` and `State` are somewhat vague given that they refer to the graph, associated memory/storage for the graph, and a client handle to the graph with synchronization utilities. `Client` could be confused with a network client such as the QUIC syncer client.

crates/aranya-daemon/src/aranya.rs
jdygert-spok: I agree it isn't great. Let's bikeshed here: #692.

PR #554, opened 2025-10-25
custom roles: fix afc tests

gknopf-aranya:
Fixes broken AFC tests on the custom roles PR.

Root cause:
The custom roles policy changed which roles have permission to create/assign/delete labels by default.
E.g. operator is no longer allowed to create labels, and admin is no longer allowed to revoke labels.
It is now required to specify which role is allowed to manage a label.
The owner can perform all label operations by default.

Solution:
Rather than dealing with the complexity of having different label operations performed by owner, admin, operator, have the owner manage labels for the AFC tests. This simplifies syncing and keeps the tests focused on AFC.
If we want to test more sophisticated label operation scenarios that can be done in dedicated tests.

### DIFF
--- a/crates/aranya-client/tests/common/mod.rs
+++ b/crates/aranya-client/tests/common/mod.rs
@@ -343,13 +343,13 @@ impl DeviceCtx {
         if grant_delegations {
             let mappings = [
                 // admin -> operator
-                ("admin -> operator", roles.operator().id, roles.admin().id),
+                ("admin -> operator", roles.admin().id, roles.operator().id),
                 // admin -> member
-                ("admin -> member", roles.member().id, roles.admin().id),
+                ("admin -> member", roles.admin().id, roles.member().id),
                 // operator -> member
-                ("operator -> member", roles.member().id, roles.operator().id),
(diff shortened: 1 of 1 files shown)


crates/aranya-client/src/client.rs
```diff
@@ -39,6 +28,18 @@ use {
     std::sync::{Arc, Mutex},
 };
 
+#[doc(inline)]
+pub use crate::client::{
```
gknopf-aranya, line 32: This is showing up on the diff because the base PR didn't run `cargo make fmt`

crates/aranya-metrics/src/main.rs
```diff
@@ -179,14 +179,10 @@ impl ClientCtx {
...
 
-        let client = (|| {
-            Client::builder()
-                .with_daemon_uds_path(&uds_sock)
-                .connect()
-        })
-        .retry(ExponentialBuilder::default())
-        .await
-        .context("unable to initialize client")?;
+        let client = (|| Client::builder().with_daemon_uds_path(&uds_sock).connect())
```
gknopf-aranya, line 182: Base PR should run `cargo make fmt` for this not to show up in the diff

PR #558, opened 2025-10-27
Test that you can't assign a role twice

chip-so:
Appears to be working as expected.

### DIFF
--- a/crates/aranya-client/tests/tests.rs
+++ b/crates/aranya-client/tests/tests.rs
@@ -860,6 +860,31 @@ async fn test_assign_and_revoke_role_management_permission() -> Result<()> {
     Ok(())
 }
 
+/// Test that a role cannot be assigned if a role is already assigned.
+#[test(tokio::test(flavor = "multi_thread"))]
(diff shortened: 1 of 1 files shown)


crates/aranya-client/tests/tests.rs
```diff
@@ -860,6 +860,31 @@ async fn test_assign_and_revoke_role_management_permission() -> Result<()> {
     Ok(())
 }
 
+/// Test that a role cannot be assigned if a role is already assigned.
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_cannot_assign_twice() -> Result<()> {
+    let mut devices = DevicesCtx::new("test_cannot_assign_twice").await?;
```
gknopf-aranya, line 866:
```suggestion
async fn test_cannot_assign_role_twice() -> Result<()> {
    let mut devices = DevicesCtx::new("test_cannot_assign_role_twice").await?;
```

chip-so: I actually made this change, but I forgot to push it. 🤦

PR #561, opened 2025-10-28
custom roles: remove label assignment to roles from policy

gknopf-aranya:
Remove support for assigning labels to roles from the policy.
This simplifies AFC channel deletion handling because we don't need to compare the set of labels assigned to a device's role to the set of labels assigned directly to the device when determining whether a channel can be deleted.
If a label is not assigned to a device, then any AFC channels the device is participating in can be deleted.


crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya, line 1: @chip-so this is showing up in the diff because the base PR didn't generate the C API header file.
If you run `cargo make run-capi-example` it will generate the C header.

chip-so: Regenerated and pushed to the base PR.

gknopf-aranya: Thanks for updating that, I've pulled in the changes

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok, @chip-so ready for a look

chip-so (approved):

PR #563, opened 2025-10-28
custom roles: fix compiler warnings

gknopf-aranya:
Fix compiler warnings on the custom roles PR.
Unit tests should pass in CI with this change.

### DIFF
--- a/crates/aranya-client/src/client.rs
+++ b/crates/aranya-client/src/client.rs
@@ -165,6 +165,12 @@ impl<'a> ClientBuilder<'a> {
     }
 }
 
+impl Default for ClientBuilder<'_> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 /// A client for invoking actions on and processing effects from
 /// the Aranya graph.
 ///


crates/aranya-daemon/src/actions.rs
```diff
@@ -336,32 +338,6 @@ where
         .in_current_span()
     }
 
-    /// Invokes `query_aqc_net_id`.
-    #[allow(clippy::type_complexity)]
-    #[instrument(skip(self))]
-    fn query_aqc_net_id(
```
gknopf-aranya, line 342: @chip-so these aqc methods could be removed on the base PR. I removed them here because they have `SessionData` which is related to the compiler warnings being addressed.

crates/aranya-client/src/client.rs
```diff
@@ -165,6 +165,12 @@ impl<'a> ClientBuilder<'a> {
     }
 }
 
+impl Default for ClientBuilder<'_> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
chip-so, line 172: I really don't see the point of the "new_without_default" clippy lint. Like, sure, `Default` is a good pattern. But you shouldn't have to make something like this, that an optimizer will no doubt throw away, just to get the compiler to shut up.

crates/aranya-daemon/src/actions.rs
chip-so: If they're related, I think it's fine to nuke them here. Same effect, anyway.

crates/aranya-client/src/client.rs
gknopf-aranya: The `ClientBuilder` is part of our public API. Perhaps it's slightly more convenient for users to have a `default()` implementation available for it.

PR #565, opened 2025-10-29
custom-roles: small cleanup

jdygert-spok:
Various small cleanup from some existing feedback and from me rather than opening feedback.


gknopf-aranya (approved):
Thanks, lgtm

chip-so (approved):

PR #567, opened 2025-10-29
custom-roles: fix multi-node rust example

jdygert-spok:
Closes #564.

PR #568, opened 2025-10-30
custom roles: C API

ssweetney-spok:
Use the following the C example:

- [x] `change_role`
- [x] `assign_role_management_permission`
- [x] `revoke_role_management_permission`
- [x] `role_owners`
- [x] `add_role_owner`
- [x] `remove_role_owner`
- [x] `assign_role`
- [x] `revoke_role`

### DIFF
--- a/examples/c/example.c
+++ b/examples/c/example.c
@@ -137,6 +137,9 @@ typedef struct {
     };
 } Team;
 
+// A zero-initialized struct for comparison
+static const AranyaRole ZERO_ROLE = {0};
+
 // Forward Declarations
 AranyaError init_client(Client* c, const char* name, const char* daemon_addr);
 AranyaError init_team(Team* t);
@@ -961,17 +964,35 @@ AranyaError run_custom_roles_example(Team* t) {
     err = aranya_team_device_role(&owner->client, &t->id, &membera->id, &member_role);
     EXPECT("unable to get role assigned to the 'membera' device", err);
 
+    if (memcmp(&member_role, &ZERO_ROLE, sizeof(AranyaRole)) == 0) {
+        printf("no role assigned to 'membera'.\n");
+        err = ARANYA_ERROR_OTHER;
+        goto exit;
+    }
+
     // Get the role assigned to 'owner'
     AranyaRole owner_role;
     printf("getting the role assigned to 'owner'.\n");
     err = aranya_team_device_role(&owner->client, &t->id, &owner->id, &owner_role);
     EXPECT("unable to get role assigned to the 'owner' device", err);
 
+    if (memcmp(&owner_role, &ZERO_ROLE, sizeof(AranyaRole)) == 0) {
+        printf("no role assigned to 'owner'.\n");
+        err = ARANYA_ERROR_OTHER;
+        goto exit;
+    }
+
     // Get the role assigned to 'admin'
     AranyaRole admin_role;
     printf("getting the role assigned to 'admin'.\n");
     err = aranya_team_device_role(&admin->client, &t->id, &admin->id, &admin_role);
     EXPECT("unable to get role assigned to the 'admin' device", err);
+
+    if (memcmp(&admin_role, &ZERO_ROLE, sizeof(AranyaRole)) == 0) {
+        printf("no role assigned to 'admin'.\n");
+        err = ARANYA_ERROR_OTHER;
+        goto exit;
+    }
 
     // Get the ID of the 'member' role.
     AranyaRoleId member_role_id;
@@ -995,6 +1016,12 @@ AranyaError run_custom_roles_example(Team* t) {
 
     err = aranya_team_device_role(&owner->client, &t->id, &membera->id, &member_role);
     EXPECT("unable to get role assigned to the 'membera' device", err);
+
+    if (memcmp(&member_role, &ZERO_ROLE, sizeof(AranyaRole)) == 0) {
+        printf("no role assigned to 'membera'.\n");
+        err = ARANYA_ERROR_OTHER;
+        goto exit;
+    }
 
     const char* role_name = NULL;
     err = aranya_role_get_name(&member_role, &role_name);


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -468,6 +468,40 @@ impl Addr {
...
 }
 
+/// The name of a permission.
+///
+/// E.g. "CanAssignRole"
+///
+/// Refer to the "Role Management" section of the policy for an exhaustive list.
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug)]
+pub struct Permission(*const c_char);
```
gknopf-aranya, line 489: Does the capi tooling care if we make all the types in this file `pub(crate)`? 
```suggestion
pub(crate) struct Permission(*const c_char);
```
We don't plan to import this module directly into any other crates other than the capi tooling if it relies on importing the module. If the tooling is simply parsing the AST, then it may not care.

ssweetney-spok: > Does the capi tooling care if we make all the types in this file pub(crate)?

Our capi-codegen crate doesn't but cbindgen does

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+        )?
+        .__into_data();
+
+    if *roles_len < default_roles.len() {
+        *roles_len = default_roles.len();
+        return Err(imp::Error::BufferTooSmall);
+    }
+    let out = aranya_capi_core::try_as_mut_slice!(roles_out, *roles_len);
+
+    for (dst, src) in out.iter_mut().zip(default_roles) {
```
gknopf-aranya, line 1015: Our label query checks the `*_len` after iterating, but I think the implementation you have here is more efficient. Either way, we should be consistent with the way we are writing to output buffers.
```
pub unsafe fn query_labels(
    client: &Client,
    team: &TeamId,
    labels: *mut MaybeUninit<LabelId>,
    labels_len: &mut usize,
) -> Result<(), imp::Error> {
    let data = client
        .rt
        .block_on(client.inner.team(team.into()).labels())?;
    let data = data.__data();
    let out = aranya_capi_core::try_as_mut_slice!(labels, *labels_len);
    for (dst, src) in out.iter_mut().zip(data) {
        dst.write(src.id.into());
    }
    if *labels_len < data.len() {
        *labels_len = data.len();
        return Err(imp::Error::BufferTooSmall);
    }
    *labels_len = data.len();
    Ok(())
}
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+    Ok(())
+}
+
+/// Removes an owning_role as an owner of role.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[in] role the ID of the subject role [`RoleId`].
+/// @param[in] owning_role ID of the owning role [`RoleId`].
+pub fn remove_role_owner(
```
gknopf-aranya, line 1054: @ssweetney-spok, @chip-so, @jdygert-spok the name of this function is confusing because it could be interpreted as removing the "owner" role rather than removing the owning role from a role. But that name wasn't introduced on this PR.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
             .inner
             .team(team.into())
-            .setup_default_roles(owner_role.id),
+            .remove_role_owner(role.into(), owning_role.into()),
     )?;
+
     Ok(())
 }
 
+/// Returns the roles that own role.
```
```suggestion
/// Returns the roles that own a role.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+    let out = aranya_capi_core::try_as_mut_slice!(roles_out, *roles_len);
+
+    for (dst, src) in out.iter_mut().zip(owning_roles) {
+        Role::init(dst, imp::Role(src));
+    }
+
+    Ok(())
+}
+
+pub fn assign_role_management_permission(
```
gknopf-aranya, line 1118: Needs docs

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+        client
+            .inner
+            .team(team.into())
+            .assign_role_management_permission(role.into(), managing_role.into(), perm),
+    )?;
+
+    Ok(())
+}
+
+pub fn revoke_role_management_permission(
```
gknopf-aranya, line 1147: Needs docs

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+    perm: Permission,
+) -> Result<(), imp::Error> {
+    // SAFETY: Caller must ensure `perm` is a valid C String.
+    let perm = unsafe { perm.as_underlying() }?;
+
+    client.rt.block_on(
+        client
+            .inner
+            .team(team.into())
+            .revoke_role_management_permission(role.into(), managing_role.into(), perm),
```
gknopf-aranya, line 1161: It seems way too easy to switch the `role` and `owning_role` params since they are the same type

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
 
+/// Returns the roles that own role.
+///
+/// Returns an `AranyaBufferTooSmall` error if the output buffer is too small to hold the roles.
+/// Writes the number of roles that would have been returned to `roles_len`.
+/// The application can use `roles_len` to allocate a larger buffer.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[in] role the ID of the subject role [`Role`].
```
```suggestion
/// @param[in] role the ID of the subject role [`RoleId`].
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+/// Writes the number of roles that would have been returned to `roles_len`.
+/// The application can use `roles_len` to allocate a larger buffer.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[in] roles_out returns a list of roles on the team [`Role`].
+/// @param[in,out] roles_len the number of roles written to the buffer.
+///
+/// @relates AranyaClient.
+pub fn roles(
```
gknopf-aranya: We previously had a convention of naming functions `query_*` that queried the factdb. If following that convention, this would become `query_roles()`. However, the custom roles PR appears to have changed that convention. So some cleanup of the function names in this file may be required.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+    client: &Client,
+    team: &TeamId,
+    roles_out: *mut MaybeUninit<Role>,
+    roles_len: &mut usize,
+) -> Result<(), imp::Error> {
+    let all_roles = client
+        .rt
+        .block_on(client.inner.team(team.into()).roles())?
+        .__into_data();
```
gknopf-aranya: nit: I don't see a reason to specify that this is all the roles. The `team.roles()` method doesn't.
```suggestion
    let roles = client
        .rt
        .block_on(client.inner.team(team.into()).roles())?
        .__into_data();
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+/// `__output` will be a valid pointer if the value returned
+/// by this function is `ARANYA_ERROR_SUCCESS`.
+///
+/// @param[in] role_list the list of roles [`Role`].
+/// @param[in] role_list_len the length of the list of roles [`Role`].
+/// @param[in] name the name used to search for the role [`RoleName`].
+/// @param[out] __output an out pointer to the ID of the [`Role`] that was found.
+///
+/// @relates Role.
+pub fn get_role_id_by_name(role_list: &[Role], name: RoleName) -> Result<RoleId, imp::Error> {
```
gknopf-aranya: @ssweetney-spok, @jdygert-spok we've had a couple discussions about not wanting to support searching for roles by name in the API because this is not a reliable way to lookup roles (e.g. multiple roles could have the same name). This was the most recent discussion:
https://github.com/aranya-project/aranya/pull/567#discussion_r2478976926
I think this search should be done in the C example itself.

ssweetney-spok: I was wondering about that. The `Role` is an opaque type so we still need a function in the API to extract the name.

Or we could make the C API behave differently and the functions that return `Role`s return `RoleId`s instead

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya, line 1: Missing a method for `change_role()`

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: Hmm, even if we did return `RoleId`s, the user would still need a way to know which human readable name corresponds to an ID. Then we just need a single function to extract the name from a `Role`?

gknopf-aranya: A getter method for the role name seems good. We do want users to have access to the name of the role. We just don't want them to depend on searching for role IDs based on the name.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: More generally, let's ensure method names in the C API mirror method names in the Rust API as much as possible. I noticed quite a few places where they are out-of-sync. That was introduced on the base custom roles PR but we might as well fix it here since it is in scope for the C API.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Huh, the timestamp shows I commented first, but my comment is showing last. Yep, just need a method for getting the name of a role. Then in the C example, we can search for a role name matching the expected name.

chip-so: > this is not a reliable way to lookup roles (e.g. multiple roles could have the same name)

IMO this is a design problem. If you don't want roles with the same name, don't do that. Does anyone know why that's allowed anyway? It's possible for the policy to enforce it by tracking role names in a fact.

jdygert-spok: Consolidating discussion: https://github.com/orgs/aranya-project/discussions/570

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,15 +997,207 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+    client: &Client,
+    team: &TeamId,
+    role: &RoleId,
+    roles_out: *mut MaybeUninit<Role>,
+    roles_len: &mut usize,
+) -> Result<(), imp::Error> {
+    let owning_roles = client
+        .rt
+        .block_on(client.inner.team(team.into()).role_owners(role.into()))?
+        .__into_data();
```
nikki-aranya, line 1093: not sure how much I like hidden methods like this, just means we're hiding jank usually

crates/aranya-client-capi/src/api/defs.rs
nikki-aranya: I agree, not sure the best way to address it though, NewType/Struct?

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: A role can have multiple owners right? The name describes the correct purpose IMO

Edit: Yea, this is the case

> As previously mentioned, each role is "owned" by zero or more
other roles, called the _owning roles_. The owning roles are
responsible for delegating management permissions of the role to
other roles.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: Yea, I do see that there were some naming changes made to the Rust client API.

My thinking is that the methods on the respective `Client`, `Team`, and device `Device` types make it clear to see which type is being operated on, e.g. `Device::keybundle` (Rust) -> `query_device_keybundle` (C)

So maybe this should be renamed to `query_roles_on_team`

examples/c/example.c
```diff
@@ -713,17 +726,48 @@ AranyaError run_afc_example(Team* t) {
...
+    //     printf("handling buffer too small error. roles_len: %zu\n", roles_len);
+    //     roles = realloc(roles, roles_len);
+    //     if (roles == NULL) {
+    //         abort();
+    //     }
+
+    //     printf("handled buffer reallocation. roles_len: %zu\n", roles_len);
+    //     err = aranya_roles(&owner->client, &t->id, roles, &roles_len);
+    //     printf("Re tried fefching list of roles\n");
+    // }
```
ssweetney-spok: Using a static array here since I get a `heap-buffer-overflow` when I try to use the code I commented out. I believe the mismatch between the sizes of the C and Rust types were resolved with `Opaque` fix so I'm not sure why the second call to `aranya_roles` is failing

cc: @jdygert-spok.

Edit: Resolved

> oh, roles_len is the number of elements, not the size to pass to realloc
> \- jdygert

ssweetney-spok:
```
==14683==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x000107f09600 at pc 0x000101422c64 bp 0x00016f567540 sp 0x00016f566cf0
WRITE of size 96 at 0x000107f09600 thread T0
    #0 0x101422c60 in wrap_memcpy+0x428 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x1ac60)
    #1 0x100e2331c  (libaranya_client_capi.dylib:arm64+0xab31c)
    #2 0x100d875dc in aranya_roles+0x180 (libaranya_client_capi.dylib:arm64+0xf5dc)
    #3 0x100897644 in run_afc_example+0x314 (example:arm64+0x100007644)
    #4 0x100896f8c in run+0x750 (example:arm64+0x100006f8c)
    #5 0x1008987b0 in main+0x1bc (example:arm64+0x1000087b0)
    #6 0x183229d50 in start+0x1c0c (dyld:arm64e+0x3d50)
 ```

File: crates/aranya-client-capi/src/api/defs.rs (hunk 3 above)
Comment (gknopf-aranya) [reply crates/aranya-client-capi/src/api/defs.rs:1054]: What I mean is this function name could be misinterpreted as removing the owner role from the team. Suggest calling this `remove_owning_role_from_role()`.

Comment (ssweetney-spok) [reply crates/aranya-client-capi/src/api/defs.rs:1054]: Ah, I didn't make that connection. It reads clearly to me but I think `remove_owner_role` would've been more ambiguous

File: crates/aranya-client-capi/src/api/defs.rs (hunk 12 above)
Comment (ssweetney-spok) [reply crates/aranya-client-capi/src/api/defs.rs:1093]: AFAICT, they're only hidden so they can be used in `aranya-client-capi`. I suppose we could resolve this by unifying the two crates and making the methods that were hidden `pub(crate)`.

File: crates/aranya-client-capi/src/api/defs.rs (hunk 9 above)
Comment (chip-so) [reply crates/aranya-client-capi/src/api/defs.rs]: I feel like `query` is redundant but the C function name should reflect the objects used in getting to this function. So this would be `team_roles()`.

File: crates/aranya-client-capi/src/api/defs.rs (hunk 12 above)
Comment (gknopf-aranya) [reply crates/aranya-client-capi/src/api/defs.rs:1093]: Seems reasonable to implement `aranya-client-capi` as a submodule of `aranya-client`. It is the C API of the client. And there shouldn't be a need to publish the crate separately since C applications use the compiled library rather than importing a Rust crate. Is there anything in the `capi` tooling that assumes the C API is implemented in a stand-alone crate?

File: crates/aranya-client-capi/src/api/defs.rs (hunk 11 above)
Comment (gknopf-aranya) [reply crates/aranya-client-capi/src/api/defs.rs]: I don't think we're going to resolve this discussion by the time custom roles needs to land this week. So I suggest this be handled in a follow-up PR.

Comment (jdygert-spok) [reply crates/aranya-client-capi/src/api/defs.rs]: I think doing the search in the example and not having that in the API is a good step for now.

File: crates/aranya-client-capi/src/api/defs.rs (hunk 9 above)
Comment (ssweetney-spok) [reply crates/aranya-client-capi/src/api/defs.rs]: Renamed the functions prefixed with "query"

File: crates/aranya-client-capi/src/api/defs.rs (hunk 7 above)
Comment (gknopf-aranya) [reply crates/aranya-client-capi/src/api/defs.rs:1161]: Creating a new type may not be as ergonomic because then the user would need to convert from a role ID into that type before using the method. If we return a specific error when a role that is not the owning_role is used rather than a generic policy authentication failure that would help.

File: examples/c/example.c [hunk 14]
```diff
@@ -144,6 +144,10 @@ AranyaError add_sync_peers(Team* t, AranyaSyncPeerConfig* cfg);
 AranyaError run(Team* t);
 AranyaError run_afc_example(Team* t);
 AranyaError cleanup_team(Team* t);
+AranyaError get_role_id_by_name(const struct AranyaRole *role_list,
+                                       size_t role_list_len,
```
gknopf-aranya: nit: could use slightly shorter names:
```suggestion
AranyaError get_role_id_by_name(const struct AranyaRole *roles,
                                       size_t roles_len,
```

examples/c/example.c
```diff
@@ -152,9 +156,35 @@ typedef struct AranyaChannelIdent {
...
+    for (size_t i = 0; i < role_list_len; i++) {
+        AranyaRole role = role_list[i];
+        const char* role_name = NULL;
+        err = aranya_role_get_name(&role, &role_name);
+
+        if (err != ARANYA_ERROR_SUCCESS) {
+            return err;
+        }
+
+        if (strcmp(name, role_name) == 0) {
```
gknopf-aranya, line 183: `strncmp` is more secure and has bounded execution

examples/c/example.c
```diff
@@ -144,6 +144,10 @@ AranyaError add_sync_peers(Team* t, AranyaSyncPeerConfig* cfg);
 AranyaError run(Team* t);
 AranyaError run_afc_example(Team* t);
 AranyaError cleanup_team(Team* t);
+AranyaError get_role_id_by_name(const struct AranyaRole *role_list,
+                                       size_t role_list_len,
+                                       const char* name,
+                                       struct AranyaRoleId *__output);
```
```suggestion
                                       struct AranyaRoleId *role_id);
```

examples/c/example.c
```diff
@@ -152,9 +156,35 @@ typedef struct AranyaChannelIdent {
...
                                        AranyaChannelIdent* idents,
                                        int num_peers);
 
+AranyaError get_role_id_by_name(const struct AranyaRole *role_list,
+                                       size_t role_list_len,
+                                       const char*  name,
+                                       struct AranyaRoleId *__output) {
+    
+    AranyaError err;
+
```
gknopf-aranya, line 173: Should set output role ID to NULL here so if there are no roles matching the name, the caller knows the result is invalid.
```
role_id = NULL;
```

examples/c/example.c
```diff
@@ -352,42 +382,114 @@ AranyaError init_team(Team* t) {
...
+    err = aranya_setup_default_roles(&owner->client, &t->id, default_roles, &default_roles_len);
+    if (err != ARANYA_ERROR_SUCCESS) {
+        fprintf(stderr,
+                "unable to set up default roles\n");
+        return err;
+    }
+
+    // Get the ID of the admin role.
+    AranyaRoleId admin_role_id;
+    err = get_role_id_by_name(default_roles, default_roles_len, "admin", &admin_role_id);
```
gknopf-aranya, line 406: Verify that `admin_role_id != NULL` before proceeding or return an error. There is no guarantee that there will be a role matching a certain string name.

examples/c/example.c
```diff
@@ -352,42 +382,114 @@ AranyaError init_team(Team* t) {
 
     // Team members are added to the team by first calling
     // `aranya_add_device_to_team`, passing in the submitter's client, the
-    // team ID and the public key of the device to be added. In a real world
+    // team ID ,the public key of the device to be added, and the role ID. In a real world
```
```suggestion
    // team ID, the public key of the device to be added, and the role ID. In a real world
```

examples/c/example.c
```diff
@@ -352,42 +382,114 @@ AranyaError init_team(Team* t) {
...
     // scenario, the keys would be exchanged outside of Aranya using
     // something like `scp`.
 
+    size_t default_roles_len = 3;
```
gknopf-aranya: There are currently 4 default roles: owner, admin, operator, member. It wouldn't hurt to allocate some extra space here in case we add more default roles later. Or handle the `AranyaBufferTooSmall` to reallocate a larger buffer at runtime.

examples/c/example.c
```diff
@@ -352,42 +382,114 @@ AranyaError init_team(Team* t) {
...
+    AranyaRole default_roles[default_roles_len];
+
+    // setup default roles.
+    err = aranya_setup_default_roles(&owner->client, &t->id, default_roles, &default_roles_len);
```
gknopf-aranya, line 397: Did you confirm that this function does not return an error? It should be returning a `AranyaBufferTooSmall` given the buffer allocated is too small to contain the default roles.

ssweetney-spok: Yea, it should be showing up in the diff

```
   // setup default roles.
    err = aranya_setup_default_roles(&owner->client, &t->id, default_roles, &default_roles_len);
        fprintf(stderr,
                "unable to set up default roles\n");
        return err;
    }
```
    

It's not returning an error because there are only 3 default roles returned. (owner isn't in this list)

examples/c/example.c
gknopf-aranya: Oh, I see that error handling is done via the `AranyaError` return type rather than a null pointer so this works as is.

examples/c/example.c
ssweetney-spok: https://github.com/aranya-project/aranya/pull/568#discussion_r2491304440

examples/c/example.c
```diff
@@ -352,42 +382,114 @@ AranyaError init_team(Team* t) {
...
-                                    admin->pk_len);
+                                    admin->pk_len, &admin_role_id);
     if (err != ARANYA_ERROR_SUCCESS) {
         fprintf(stderr, "unable to add admin to team\n");
         return err;
     }
 
+    // Get the ID of the operator role.
+    AranyaRoleId operator_role_id;
+    err = get_role_id_by_name(default_roles, default_roles_len, "operator", &operator_role_id);
```
gknopf-aranya, line 422: Seems like we could add the role IDs onto the `Client` struct since that's where we've been storing info for each device:
```
// Aranya client.
typedef struct {
    // Name of Aranya client.
    const char* name;
    // Pointer to Aranya client.
    AranyaClient client;
    // Aranya client's serialized public key bundle.
    uint8_t* pk;
    // Aranya client's serialized public key bundle length.
    size_t pk_len;
    // Aranya client's public id.
    AranyaDeviceId id;
    // Aranya client's role id.
    AranyaRoleId id,
} Client;
```

ssweetney-spok: Good point

examples/c/example.c
ssweetney-spok: I'd have to call `strlen` to get the right length to pass to `strncmp`. The execution wouldn't bounded anymore then

examples/c/example.c
gknopf-aranya: I see. The action itself only returns 3 roles in the policy:
```
// Setup default roles on a team.
action setup_default_roles(owning_role_id id) {
    publish SetupDefaultRole {
        name: DefaultRoleName::Admin,
        owning_role_id: owning_role_id,
    }
    publish SetupDefaultRole {
        name: DefaultRoleName::Operator,
        owning_role_id: owning_role_id,
    }
    publish SetupDefaultRole {
        name: DefaultRoleName::Member,
        owning_role_id: owning_role_id,
    }
}
```

examples/c/example.c
ssweetney-spok: Hmm, I suppose I could pass in a hardcoded value for the length of the string literals

examples/c/example.c
gknopf-aranya: That makes it tricky to search for the owner role ID by name.

gknopf-aranya: If we want to get the owner role ID, then it would be better to search the list of roles returned by `Team::roles()`.

gknopf-aranya: Or the list returned by `aranya_team_roles()` in the C API

examples/c/example.c
ssweetney-spok: Actually, I'm not sure if it's a good idea to keep it there since the assigned role can change. I assign and revoke a few places in the example

examples/c/example.c
ssweetney-spok: Yea, that's where it's pulled from in `run_afc_example`

examples/c/example.c
gknopf-aranya: True, unless you have a wrapper method around assign/revoke that updates it

examples/c/example.c
chip-so: Yeah, that's what you'd usually do. The `n` in `strncmp()` is usually an upper bound to protect from overrunning an input buffer, not an exact length of the longer string.

~~And as far as runtime, `strcmp()` is a linear-time comparison that only terminates if it hits a null, so I'm not sure how that would be bounded in the first place. Or at least, it's no more or less bounded than calling `strlen()` first, which does the same thing.~~ Nevermind, I misunderstood what was being argued.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Refactoring the `aranya-client-capi` is not something we should attempt on the current PR. We could open an issue for this.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,12 +980,240 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+///
+/// This will change the device's current role to the new role assigned.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[in] device the device's ID [`DeviceId`].
+/// @param[in] old_role the ID of the role currently assigned to the device.
+/// @param[in] new_role the ID of the role to assign to the device.
```
```suggestion
/// @param[in] old_role the ID of the role currently assigned to the device [`RoleId`].
/// @param[in] new_role the ID of the role to assign to the device [`RoleId`].
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,12 +980,240 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+            .team(team.into())
+            .revoke_role_management_permission(role.into(), managing_role.into(), perm),
+    )?;
+
+    Ok(())
+}
+
+/// Changes the `role` on a `device`
+///
+/// This will change the device's current role to the new role assigned.
```
gknopf-aranya, line 1169:
```suggestion
/// This will change the device's currently assigned role to the new role.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -952,12 +980,240 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
...
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[out] roles_out returns a list of roles on the team [`Role`].
+/// @param[in,out] roles_len the number of roles written to the buffer.
+///
+/// @relates AranyaClient.
+pub unsafe fn team_roles(
+    client: &Client,
+    team: &TeamId,
+    roles_out: *mut MaybeUninit<Role>,
+    roles_len: &mut usize,
```
gknopf-aranya: Either `roles_out` and `roles_out_len` or `roles` and `roles_len`
```suggestion
    roles: *mut MaybeUninit<Role>,
    roles_len: &mut usize,
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -967,6 +1223,8 @@ pub fn setup_default_roles(client: &mut Client, team: &TeamId) -> Result<(), imp
 ///
 /// Permission to perform this operation is checked against the Aranya policy.
 ///
+/// It is an error if the device has already been assigned a role.
```
gknopf-aranya, line 1241: In the policy we reference `change_role()` from `assign_role()` so users know to use `change_role()` if the device already has a role assigned. Wouldn't hurt to mention it in the C and Rust API docs as well since that's more likely to be what users are interfacing with.

gknopf-aranya: This will be true once this PR is merged:
https://github.com/aranya-project/aranya/pull/576

examples/c/example.c
```diff
@@ -352,42 +382,114 @@ AranyaError init_team(Team* t) {
...
     }
 
+    size_t member_owning_roles_len = 1;
+    AranyaRole *member_owning_roles = calloc(member_owning_roles_len, sizeof(AranyaRole));
+
+    // check that there is a single owner for the 'member' role.
+    err = aranya_role_owners(&owner->client, &t->id, &member_role_id, member_owning_roles, &member_owning_roles_len);
+    EXPECT("unable to get the owning roles for the 'member' role.", err);
+    if (member_owning_roles_len != 1) {
+        fprintf(stderr, "there should only be 1 owner for the 'member' role. Actual: %zu\n", member_owning_roles_len);
```
gknopf-aranya: This should return an error if there is an unexpected result. Printing this out is not enough to cause the CI job to fail.

examples/c/example.c
```diff
@@ -352,42 +382,114 @@ AranyaError init_team(Team* t) {
...
+    // check that there are 2 owners for the 'member' role.
+    err = aranya_role_owners(&owner->client, &t->id, &member_role_id, member_owning_roles, &member_owning_roles_len);
+    if (err == ARANYA_ERROR_BUFFER_TOO_SMALL) {
+        printf("handling buffer too small error\n");
+        member_owning_roles = realloc(member_owning_roles, member_owning_roles_len * sizeof(AranyaRole));
+        err = aranya_role_owners(&owner->client, &t->id, &member_role_id, member_owning_roles, &member_owning_roles_len);
+    }
+    EXPECT("unable to get the owning roles for the 'member' role", err);
+    if (member_owning_roles_len != 2) {
+        fprintf(stderr, "there should only be 2 owners for the 'member' role. Actual:%zu\n", member_owning_roles_len);
```
gknopf-aranya: Return error if unexpected number of owning roles

examples/c/example.c
```diff
@@ -352,42 +382,114 @@ AranyaError init_team(Team* t) {
...
+
+     // Remove an owning role from the 'member' role.
+    err = aranya_remove_role_owner(&owner->client, &t->id, &member_role_id, &admin_role_id);
+    EXPECT("unable to remove an owning role.", err);
+
+    // check that there is now just 1 owner for the 'member' role after removing an owning role.
+    err = aranya_role_owners(&owner->client, &t->id, &member_role_id, member_owning_roles, &member_owning_roles_len);
+    EXPECT("unable to get the owning roles for the 'member' role.", err);
+    if (member_owning_roles_len != 1) {
+        fprintf(stderr, "there should only be 1 owner for the 'member' role. Actual: %zu\n", member_owning_roles_len);
```
gknopf-aranya: Ditto

examples/c/example.c
gknopf-aranya, line 1: `init_team()` is for setting up the team before running examples/tests. Would prefer to move the custom roles tests into a new method such as `run_custom_roles_example()`

gknopf-aranya: `setup_default_roles()` seems like it belongs in `init_team()` since that is part of the default team setup.

examples/c/example.c
```diff
@@ -713,17 +816,46 @@ AranyaError run_afc_example(Team* t) {
...
+    if (err == ARANYA_ERROR_BUFFER_TOO_SMALL) {
+        printf("handling buffer too small error. roles_len: %zu\n", roles_len);
+        roles = realloc(roles, roles_len * sizeof(AranyaRole));
+        if (roles == NULL) {
+            abort();
+        }
+
+        printf("handled buffer reallocation. roles_len: %zu\n", roles_len);
+        err = aranya_team_roles(&owner->client, &t->id, roles, &roles_len);
+        printf("Re tried fefching list of roles\n");
```
```suggestion
        printf("Retried fetching list of roles\n");
```

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: Yeah, different PR.

examples/c/example.c
jdygert-spok: I don't think there's really a need for it when we are passing in `name` ourselves. The amount of `role_name` that we read will be limited by the length of `name` anyway, so the only thing we'd be protecting against is calling `get_role_id_by_name` with a non-null-terminated `name`. It'd be different if we had a buffer that `aranya_role_get_name` copied out the name to, since that would have a known size.

examples/c/example.c
```diff
@@ -352,42 +382,117 @@ AranyaError init_team(Team* t) {
...
-    // team ID and the public key of the device to be added. In a real world
+    // team ID, the public key of the device to be added, and the role ID. In a real world
     // scenario, the keys would be exchanged outside of Aranya using
     // something like `scp`.
 
+    size_t default_roles_len = 3;
+    AranyaRole default_roles[default_roles_len];
+
+    // setup default roles.
+    err = aranya_setup_default_roles(&owner->client, &t->id, default_roles, &default_roles_len);
```
gknopf-aranya, line 397: We should handle the `AranyaBufferTooSmall` and reallocate the buffer if it is too small. Then `free()` the buffer when it's no longer used.

examples/c/example.c
```diff
@@ -352,42 +382,117 @@ AranyaError init_team(Team* t) {
...
     err = aranya_add_device_to_team(&owner->client, &t->id, memberb->pk,
-                                    memberb->pk_len);
+                                    memberb->pk_len, &member_role_id);
     if (err != ARANYA_ERROR_SUCCESS) {
         fprintf(stderr, "unable to add memberb to team\n");
         return err;
     }
 
+    size_t member_owning_roles_len = 1;
+    AranyaRole *member_owning_roles = calloc(member_owning_roles_len, sizeof(AranyaRole));
```
gknopf-aranya: Missing `free()`

examples/c/example.c
```diff
@@ -713,17 +819,46 @@ AranyaError run_afc_example(Team* t) {
     AranyaLabelId label_id;
     AranyaChannelIdent idents[] = {{&membera->id, ARANYA_CHAN_OP_SEND_ONLY},
                                    {&memberb->id, ARANYA_CHAN_OP_RECV_ONLY}};
-    err = aranya_create_assign_label(&operator->client, &t->id, "uni_label",
-                                     &label_id, idents, 2);
+
+    size_t roles_len = 0;
+    roles = calloc(roles_len, sizeof(AranyaRole));
```
gknopf-aranya, line 788: Missing `free()`

crates/aranya-client-capi/src/imp/client.rs
```diff
@@ -40,7 +40,7 @@ pub fn key_bundle_deserialize(buf: &[u8]) -> Result<KeyBundle, imp::Error> {
 
 /// An Aranya role.
 #[derive(Debug)]
-pub struct Role(aranya_client::Role);
+pub struct Role(pub(crate) aranya_client::Role);
```
jdygert-spok: This wrapper can be removed.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -934,11 +957,22 @@ pub fn sync_peer_config_builder_set_sync_later(cfg: &mut SyncPeerConfigBuilder)
 /// - operator
 /// - member
 ///
+/// Returns an `AranyaBufferTooSmall` error if the output buffer is too small to hold the roles.
+/// Writes the number of roles that would have been returned to `roles_len`.
+/// The application can use `roles_len` to allocate a larger buffer.
```
jdygert-spok: This will probably fail if called a second time.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: I think it's fine as `role_owner` (and not `owner_role`)

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya, line 1: Missing:
- `add_label_managing_role()`
- `// TODO: query_device_role` This todo has existed for a long time before custom roles and the method can now be implemented.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1182,11 +1448,11 @@ pub unsafe fn encrypt_psk_seed_for_peer(
         *seed_len = wrapped_seed.len();
         return Err(imp::Error::BufferTooSmall);
     }
+    *seed_len = wrapped_seed.len();
     let out = aranya_capi_core::try_as_mut_slice!(seed, *seed_len);
```
jdygert-spok, line 1486: This overwrites `seed_len` before making the out slice, meaning we could write past the end of `seed`.

jdygert-spok: Oh, but we check it first.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Sounds good. Marking this as resolved.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: Oh, right. This is only meant to be called once. So you _have_ to call this with a buffer that's big enough

gknopf-aranya: Then I think the API should provide the user with the correct length to allocate for the buffer.

ssweetney-spok: The tricky part is that you don't know how many roles there are until you trigger the action that creates the default roles. We could return a hard-coded value but then there's the risk of things getting out of sync if the policy is changed but the function isn't 🤔

gknopf-aranya: So we allocate a large buffer in the application with 256 elements. Or we return a wrapper to a vector/iterator where `next()` or `get()` are called to get elements of it.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Marking this as resolved since this doesn't seem important enough to worry about right now.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Should this just write out to a `struct DefaultRoles { RoleID admin, RoleID operator, RoleID member }` ?

gknopf-aranya: In the interest of time, let's allocate a large buffer with an issue created to add a more flexible way to return lists for functions that can only be called once and therefore will not allow for reallocation.

gknopf-aranya: > Should this just write out to a `struct DefaultRoles { RoleID admin, RoleID operator, RoleID member }` ?

Probably. That's what our unit tests did.

ssweetney-spok: That'd make the example simpler since we wouldn't need to do any filtering. `aranya_role_get_name` and `aranya_role_get_id` wouldn't be used anymore though but I think it'd be fine to leave them in the API

gknopf-aranya: I'd keep `aranya_role_get_id` and `aranya_role_get_name` in the API & example so we can make sure they continue to work. Maybe just print them out instead of using the result somewhere.

jdygert-spok: If we change it to the struct it should be done all the way up to the daemon. I'm wondering if there are backwards compatibility concerns. I'm not really sure what the plan is around the default roles. I imagine most changes we could make that would break compatibility for the struct, would also break the current version at runtime.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1655,7 +1657,37 @@ pub unsafe fn team_devices(
...
+/// @param[out] device the ID of the device [`DeviceId`].
+/// @param[out] role_out the role assigned to the device. `role_out` will be NULL
+/// if a role was not assigned to the device. [`Role`].
+///
+/// @relates AranyaClient.
+pub unsafe fn team_device_role(
+    client: &Client,
+    team: &TeamId,
+    device: &DeviceId,
+    role_out: *mut MaybeUninit<Role>,
```
```suggestion
    role_out: &mut MaybeUninit<Role>,
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1655,7 +1657,37 @@ pub unsafe fn team_devices(
...
 }
 
-// TODO: query_device_role
+/// Returns the role assigned to the device, if any.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[out] device the ID of the device [`DeviceId`].
+/// @param[out] role_out the role assigned to the device. `role_out` will be NULL
+/// if a role was not assigned to the device. [`Role`].
```
jdygert-spok: "`*role_out` will be zeroed" if not assigned, matches the impl.

Although I don't love that way of signifying no role, but I don't think an out-param bool is that much better...

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: For now let's expose a constant for the number of roles and make an issue for this.

gknopf-aranya: I don't think it's necessary to return a list of default roles here. We can setup the default roles then query the roles on the team later when reallocation is possible. That would help us land this PR sooner.

chip-so: > We can setup the default roles then query the roles on the team later when reallocation is possible.

I think that's a very reasonable solution.

ssweetney-spok: [Adding the constant was painless](https://github.com/aranya-project/aranya/pull/568/commits/bf466cf3358ae6a99aaed52c71a0db16de79d8ea).

gknopf-aranya: This also works around backward compatibility concerns

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -267,6 +267,9 @@ const _: () = {
     assert!(ARANYA_ID_LEN == size_of::<aranya_id::BaseId>());
 };
 
+/// The number of roles returned from `setup_default_roles`.
```
gknopf-aranya, line 272: I'd reference the policy.md file from here to make sure this stays up-to-date

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1687,28 +1687,25 @@ pub unsafe fn team_devices(
 /// @param[in] client the Aranya Client [`Client`].
 /// @param[in] team the team's ID [`TeamId`].
 /// @param[out] device the ID of the device [`DeviceId`].
-/// @param[out] role_out the role assigned to the device. `role_out` will be NULL
+/// @param[out] role_out the role assigned to the device. `role_out` will be zeroed
```
gknopf-aranya, line 1697: Is it the caller's responsibility to check if `role_out` has been zeroed? Or will a returned error indicate this was zeroed?

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: https://github.com/aranya-project/aranya/issues/582

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: The error doesn't indicate that the value was zeroed. This is kind of similar to returning a `Result<Option<T>>`

examples/c/example.c
```diff
@@ -834,10 +937,117 @@ AranyaError run_afc_example(Team* t) {
...
-
 exit:
     free(ciphertext);
     free(plaintext);
+    free(roles);
+    return err;
+}
+
+// Run the custom roles example
+AranyaError run_custom_roles_example(Team* t) {
```
gknopf-aranya, line 951: The custom roles example doesn't currently have `printf()` to show which step is being executed like the other examples do.

gknopf-aranya (approved):
This looks good to me

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Then shouldn't we check whether it is set before using it in the example?

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: #583

crates/aranya-client-capi/src/api/defs.rs
djohnson-spideroak: Maybe a function like `is_no_role` to compare if the returned value is zeroized?

ssweetney-spok: That'd work. I just pushed a [commit](https://github.com/aranya-project/aranya/pull/568/commits/362233fa91d57a1724dd580b613dc2482c4a4a0b) that uses `memcmp` instead

jdygert-spok (approved):

PR #571, opened 2025-10-30
custom roles: improve unit test coverage

gknopf-aranya:
The goal of this PR is to improve unit test coverage for custom roles. Gaps in test coverage could include:
- No unit tests executing a particular method in the custom roles API
- A missing positive/negative test case of a particular method

A test pattern that is useful for ensuring coverage of role/perm/labels is:
- Verify operation fails
- Create role/perm/label as applicable
- Query to verify role/perm/label was created
- Verify operation fails
- Assign role/perm/label
- Query to verify assignment
- Verify operation succeeds
- Revoke role/perm/label as applicable
- Query to verify revocation
- Verify operation fails
- Delete role/perm/label as applicable
- Query to verify deletion of role/perm/label as applicable
- Verify operation fails

This ensures that operations are only allowed when valid permissions have been assigned and rejected otherwise.

Unit test improvements:
- Remove `sleep()` and `add_all_sync_peers()` from unit tests because they aren't needed when `sync_now()` is used. This will speed up the unit tests.
- `test_add_remove_sync_peers` - Demonstrates that after adding a sync peer with a sync interval, syncing will automatically occur after waiting longer than the sync interval. Demonstrates that when a sync peer is removed, syncing of new commands no longer occurs without an explicit `sync_now()`.
- `test_afc_create_assign_revoke_delete_label` - Verifies `CanUseAfc` permission is required before assigning a label
- Adds tests for role queries
- `test_change_role` - Tests change_role()
- `test_afc_create_assign_revoke_delete_label` - Tests create/assign/revoke/delete label. Verifies label cannot be assigned after it has been deleted.
- `test_role_create_assign_revoke` - Tests role create/assign/revoke
- `test_assign_and_revoke_role_management_permission` - Verify owning role cannot perform role management operations after it is removed as an owner from another role

Other changes:
- Remove obsolete `test_assign_label_to_device_requires_network_id()`. AQC and network identifiers have been removed from the codebase.
- Was going to add test coverage for `close_team()` but discovered it is unimplemented. Removing it from the APIs until we are able to implement it. This is unfortunate from a backward compatibility standpoint. However, we should have waited to add this API until it was implemented.

### DIFF
--- a/crates/aranya-daemon/src/api.rs
+++ b/crates/aranya-daemon/src/api.rs
@@ -485,6 +485,13 @@ impl DaemonApi for Api {
         Ok(team_id)
     }
 
+    #[instrument(skip(self), err)]
+    async fn close_team(self, _: context::Context, team: api::TeamId) -> api::Result<()> {
+        self.check_team_valid(team).await?;
+
+        todo!();
+    }
+
     //
     // Device onboarding
     //
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -1833,6 +1833,29 @@ AranyaError aranya_remove_team_ext(const struct AranyaClient *client,
                                    struct AranyaExtError *__ext_err);
 
 /**
+ * Close the team and stop all operations on the graph.
+ *
+ * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
+ * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
+ *
+ * @relates AranyaClient.
+ */
+AranyaError aranya_close_team(const struct AranyaClient *client,
+                              const struct AranyaTeamId *team);
+
+/**
+ * Close the team and stop all operations on the graph.
+ *
+ * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
+ * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
+ *
+ * @relates AranyaClient.
+ */
+AranyaError aranya_close_team_ext(const struct AranyaClient *client,
+                                  const struct AranyaTeamId *team,
+                                  struct AranyaExtError *__ext_err);
+
+/**
  * Add a device to the team with the default role.
  *
  * Permission to perform this operation is checked against the Aranya policy.
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1214,6 +1214,19 @@ pub fn remove_team(client: &Client, team: &TeamId) -> Result<(), imp::Error> {
     Ok(())
 }
 
+/// Close the team and stop all operations on the graph.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+///
+/// @relates AranyaClient.
+pub fn close_team(client: &Client, team: &TeamId) -> Result<(), imp::Error> {
+    client
+        .rt
+        .block_on(client.inner.team(team.into()).close_team())?;
+    Ok(())
+}
+
 /// Add a device to the team with the default role.
 ///
 /// Permission to perform this operation is checked against the Aranya policy.
--- a/crates/aranya-client/src/client/team.rs
+++ b/crates/aranya-client/src/client/team.rs
@@ -33,6 +33,17 @@ impl Team<'_> {
     pub fn team_id(&self) -> TeamId {
         TeamId::from_api(self.id)
     }
+
+    /// Closes the team, preventing any further operations on it.
+    #[instrument(skip(self))]
+    pub async fn close_team(&self) -> Result<()> {
+        self.client
+            .daemon
+            .close_team(context::current(), self.id)
+            .await
+            .map_err(IpcError::new)?
+            .map_err(aranya_error)
+    }
 }
 
 impl Team<'_> {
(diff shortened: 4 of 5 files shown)


crates/aranya-client/tests/afc.rs
```diff
@@ -11,6 +11,50 @@ use {
...
+
+    // Assigning labels to devices with the "operator" role should fail since it does not have `CanUseAfc` permission.
+    owner_team
+        .device(devices.operator.id)
+        .assign_label(label_id, op)
+        .await
+        .context("unable to assign label")
+        .expect_err("expected label assignment to fail");
+
+    // Assigning labels to devices with the "member" role should succeed since it does not have `CanUseAfc` permission.
```
chip-so: Confused by this comment. Do you mean that it does have `CanUseAfc` permission?

crates/aranya-client/tests/tests.rs
```diff
@@ -87,6 +122,82 @@ async fn test_sync_now() -> Result<()> {
...
+    let membera_addr = devices.membera.aranya_local_addr().await?.into();
+    owner_team.remove_sync_peer(membera_addr).await?;
+
+    // Create another label.
+    owner_team
+        .create_label(text!("label2"), roles.owner().id)
+        .await?;
+
+    // Wait for syncing.
+    sleep(SLEEP_INTERVAL).await;
```
chip-so, line 194: We've removed this in a lot of other tests. Why do we need it here?

crates/aranya-client/tests/afc.rs
gknopf-aranya: Yep

crates/aranya-client/tests/tests.rs
gknopf-aranya: This test is specifically for verifying sync peers are synced with periodically after they are added with a sync interval. This is the only unit test that verifies that functionality so it is needed for coverage. The rest of the unit tests use `sync_now()` for efficiency.

gknopf-aranya: I'm adding docs for this

crates/aranya-daemon/src/policy.md
```diff
@@ -1192,8 +1192,8 @@ command RemoveRoleOwner {
         check team_exists()
 
         let author = get_author(envelope)
-        check device_has_simple_perm(author.device_id, SimplePerm::ChangeRoleManagingRole)
-        check device_owns_role(author.device_id, this.target_role_id)
+        //check device_has_simple_perm(author.device_id, SimplePerm::ChangeRoleManagingRole)
```
gknopf-aranya: Revert these

crates/aranya-client/tests/tests.rs
```diff
@@ -87,6 +122,247 @@ async fn test_sync_now() -> Result<()> {
...
+        bail!("did not expect role to be assigned to admin device");
+    }
+
+    // Show that admin cannot create a label.
+    admin_team
+        .create_label(text!("label1"), roles.admin().id)
+        .await
+        .expect_err("expected label creation to fail");
+
+    // TODO: show operation fails after delete_role() (after it is added to the API).
```
djohnson-spideroak: Is this planned for before or after the initial custom roles version? If after, we need tickets to track all of the TODOs related to APIs we plan to add in the future.

gknopf-aranya: This would be after the initial custom roles version lands. Default roles are not allowed to be customized/deleted in this initial version.

gknopf-aranya: Opened an issue for implementing `role_delete()` and updating this unit test:
https://github.com/aranya-project/aranya/issues/575

crates/aranya-daemon/src/api.rs
```diff
@@ -485,13 +485,6 @@ impl DaemonApi for Api {
         Ok(team_id)
     }
 
-    #[instrument(skip(self), err)]
-    async fn close_team(self, _: context::Context, team: api::TeamId) -> api::Result<()> {
-        self.check_team_valid(team).await?;
-
-        todo!();
-    }
```
jdygert-spok, line 493: We have a `terminate_team` action now, should we just thread that through rather than deleting these and having to re-add them?

jdygert-spok: Or was `close_team` meant to be the opposite of `add_team`...?

gknopf-aranya: This is the description we have on the `CloseTeam` operation in the spec:
`CloseTeam(team_id) -> bool - close the team and stop all operations on the graph.`
https://github.com/aranya-project/aranya-docs/blob/main/docs/aranya-mvp.md

gknopf-aranya: I'd prefer to implement `close_team()` and `terminate_team()` on a different PR. It's not related to custom roles and we need more discussion on what they are supposed to do.

gknopf-aranya: Opened an issue to implement these once we know what they are supposed to do:
https://github.com/aranya-project/aranya/issues/577

jdygert-spok: > It's not related to custom roles

I would generally agree but the custom roles policy implemented everything needed for it. If it's not related then we also shouldn't be removing it on this PR... :shrug:

jdygert-spok (approved):

crates/aranya-daemon/src/api.rs
gknopf-aranya: I can revert that change. But that means we would be merging this PR with the caveat that there is no test coverage for those methods yet.

gknopf-aranya: The reason I noticed these methods and decided to remove them here is because they showed up while I was auditing to see if there were any gaps in test coverage. In order to resolve that, I would either need to implement them and add tests, or remove them. Removing them seemed like the simplest way to unblock these changes. I'm fine leaving it as is as long as it's clear this was an existing gap in coverage we do not intend to address yet for custom roles.

PR #576, opened 2025-11-03
Custom roles explanations

chip-so:
Expands on the TODOs and other deficiencies noted in the main [custom roles PR](#267). There are still a number of remaining TODOs for things that are either currently not possible or just future (or potential future) enhancements.

### DIFF
--- a/crates/aranya-client/tests/afc.rs
+++ b/crates/aranya-client/tests/afc.rs
@@ -11,6 +11,136 @@ use {
     aranya_daemon_api::text,
 };
 
+/// Demonstrate assigning/revoking a label requires `CanUseAfc` permission.
+#[cfg(feature = "afc")]
+#[test_log::test(tokio::test(flavor = "multi_thread"))]
+async fn test_afc_create_assign_revoke_delete_label() -> Result<()> {
+    let mut devices = DevicesCtx::new("test_afc_create_assign_revoke_delete_label").await?;
+
+    // create team.
+    let team_id = devices.create_and_add_team().await?;
+
+    // create default roles
+    let default_roles = devices.setup_default_roles(team_id).await?;
+
+    // Tell all peers to sync with one another, and assign their roles.
+    devices
+        .add_all_device_roles(team_id, &default_roles)
+        .await?;
+
+    let owner_addr = devices.owner.aranya_local_addr().await?.into();
+    let owner_team = devices.owner.client.team(team_id);
+    let membera_team = devices.membera.client.team(team_id);
+    let memberb_team = devices.memberb.client.team(team_id);
+
+    // Query team labels to show label has not been created yet.
+    membera_team.sync_now(owner_addr, None).await?;
+    assert_eq!(membera_team.labels().await?.iter().count(), 0);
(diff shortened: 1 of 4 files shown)


crates/aranya-daemon/src/policy.md
gknopf-aranya, line 1: @chip-so should we reference the policy book from this policy in case someone hasn't read it yet or doesn't know where to find it?

crates/aranya-daemon/src/policy.md
```diff
@@ -103,7 +130,9 @@ authorized to publish.
 
 ## Command Priorities
 
-Priorities can defined for commands by specifying the `priority` attribute in the command's `attributes` block:
+Priorities can defined for commands by specifying the `priority`
```
```suggestion
Priorities can be defined for commands by specifying the `priority`
```

crates/aranya-daemon/src/policy.md
```diff
@@ -979,7 +1006,7 @@ command RemovePermFromRole {
         check can_change_role_perms(author.device_id, this.role_id)
 
 
-        // TODO(eric): Should this case be a no-op or an error?
+        // It is an error to remove a permission that does not exist.
```
```suggestion
        // It is an error to remove a permission that was not previously added to the role.
```

crates/aranya-daemon/src/policy.md
```diff
@@ -3481,9 +3505,9 @@ permissive `ChanOp` is used.
 //
 // # Caveats
 //
-// We do not yet support prefix deletion, so this fact is NOT
-// deleted when the label or the role are deleted. Use TODO to
-// verify whether a role is allowed to use the label instead of
+// TODO(eric): We do not yet support prefix deletion, so this fact
```
```suggestion
// TODO: issue#<issue number> We do not yet support prefix deletion, so this fact
```

crates/aranya-daemon/src/policy.md
chip-so: Not a bad idea, but it is as yet unpublished.

gknopf-aranya: Ah, didn't realize we don't have a place to point people to yet. Let's create a todo/issue to reference the policy book once that is published so this doesn't get dropped.

chip-so: Roger: #578

ycarmy:
```suggestion
- [`idam`][idam-ffi]: IDAM functionality, such as access to device
```

ycarmy: I believe the answer to this should be "yes" if we still plan to support key rotations at some point. Might be worth resolving the TODO by opening an issue to discuss further and leaving the key IDs in the facts as they are currently for now.

ycarmy: Can you remind me what `debug_assert` does?

ycarmy:
```suggestion
// Returns a device if it exists, or raises a check error
```

ycarmy: Were you planning on resolving these TODOs in this PR too?

ycarmy: broken link

PR #586, opened 2025-11-05
custom-roles: use bool for optional role

jdygert-spok:


gknopf-aranya (approved):
Change itself looks good. Don't like the formatting noise in the diff.

PR #587, opened 2025-11-05
capi: use cbindgen without expansion and in build script

jdygert-spok:
After aranya-project/aranya-core#503, there is no need for cbindgen to expand the capi crate. Running without expansion is much faster, and also removes our limitation of 1.86 related to cfg expansion. We also can now run cbindgen in the build script since it doesn't need to run in nightly.


gknopf-aranya:
Looks good from what I can tell other than CI not passing yet

PR #593, opened 2025-11-06
Custom roles updates

### LINKED ISSUE #588: custom roles: document simple perms in the policy
Document what each perm is for in the `SimplePerm` enum:
```
enum SimplePerm {
    // Team management
    AddDevice,
    RemoveDevice,
    TerminateTeam,

    // Roles
    AssignRole,
    RevokeRole,
    SetupDefaultRole,
    ChangeRoleManagingRole,

    // Labels
    CreateLabel,
    DeleteLabel,
    ChangeLabelManagingRole,
    AssignLabel,
    RevokeLabel,

    // AFC
    CanUseAfc,
    CreateAfcUniChannel,
    CreateAfcBidiChannel,
}
```

cc: @djohnson-spideroak, @jdygert-spok, @chip-so

### LINKED ISSUE #591: custom roles: Grant missing AFC permissions
Currently the `owner` role has all permissions except for the three relating to AFC usage - `CanUseAfc`, `CreateAfcUniChannel`, and `CreateAfcBidiChannel`. Since I assume we want the `owner` to be an all-permissions superuser, we should assign those on team creation.

cc @djohnson-spideroak @jdygert-spok @gknopf-aranya

### LINKED ISSUE #592: custom roles: Remove defunct `SimplePerm::CreateAfcBidiChannel`
We don't have bidirectional channels and this permission is unused, so it can be removed.

chip-so:
A few minor fixes.

Closes #588
Closes #591
Closes #592

### DIFF
--- a/crates/aranya-daemon/src/policy.md
+++ b/crates/aranya-daemon/src/policy.md
@@ -3534,9 +3534,10 @@ fact LabelAssignedToDevice[label_id id, device_id id]=>{op enum ChanOp, device_g
 //
 // # Required Permissions
 //
-// - `CanUseAfc`
 // - `AssignLabel`
 // - `CanManageLabel(label_id)`
+//
+// Additionally, the target device must have `CanUseAfc` permissions
 action assign_label_to_device(device_id id, label_id id, op enum ChanOp) {
     publish AssignLabelToDevice {
         device_id: device_id,


crates/aranya-daemon/src/policy.md
```diff
@@ -771,27 +771,52 @@ cannot be created or deleted at runtime.
...
     AssignLabel,
+    // The role can revoke a label from a device. The role must
+    // also have label management permissions granted by a role
+    // with the `ChangeLabelManagingRole` permission above.
     RevokeLabel,
 
     // AFC
+    // The role can use AFC. This controls both the ability to
+    // receive a unidirectional AFC channels and the ability to
+    // assign a label to a device (but not revoke).
```
chip-so: Not needing `CanUseAfc` to revoke a label feels like an oversight.

crates/aranya-daemon/src/policy.md
```diff
@@ -771,27 +771,52 @@ cannot be created or deleted at runtime.
 ```policy
 enum SimplePerm {
     // Team management
```
gknopf-aranya: We probably want a different way to group the perm sections so they aren't confused as documenting individual permissions.
```suggestion
    //
    // Team management
    //
    
```

crates/aranya-daemon/src/policy.md
chip-so: And separately it kind of feels like overkill that you need both `AssignLabel` and `CanUseAfc` to assign a label. I understand the labels are exclusively used by AFC but maybe you shouldn't need to be able to use AFC to manage labels.

jdygert-spok: Yeah I wouldn't expect to need `CanUseAfc` to assign/revoke labels.

crates/aranya-daemon/src/policy.md
chip-so: Does this look alright?
```
enum SimplePerm {
    // # Team management
    //
    // The role can add a device to the team.
    AddDevice,
```

gknopf-aranya: Looks better

crates/aranya-daemon/src/policy.md
```diff
@@ -771,27 +771,52 @@ cannot be created or deleted at runtime.
...
     TerminateTeam,
 
     // Roles
+    // The role can assign a role to other devices.
     AssignRole,
+    // The role can revoke a role from other devices.
     RevokeRole,
+    // The role can set up default roles. This can only be done
+    // once, so this permission can only effectively be used by
+    // the `owner` role.
```
gknopf-aranya, line 791: I believe any role can setup the default roles

chip-so: Technically, yes. But as currently designed the only role that exists after creating a team is `owner` and the only way you can get more roles is by setting up the default roles.

crates/aranya-daemon/src/policy.md
```diff
@@ -771,27 +771,52 @@ cannot be created or deleted at runtime.
...
     SetupDefaultRole,
+    // The role can add or remove an owning role to a target
+    // role.
```
gknopf-aranya: "adding to" or "removing from" is more clear. I think "add/remove to/from" would be confusing as well so maybe this:
```suggestion
    // The role can add an owning role to or remove an owning role from a target role
```

crates/aranya-daemon/src/policy.md
gknopf-aranya: Yep, we should be able to create/assign/delete labels without `CanUseAfc`. With default roles, the `Admin` or `Operator` role would assign a label to a `Member` where the `Member` role is allowed to create/use an AFC channel.

crates/aranya-daemon/src/policy.md
```diff
@@ -771,27 +771,52 @@ cannot be created or deleted at runtime.
...
+    // with the `ChangeLabelManagingRole` permission above.
     AssignLabel,
+    // The role can revoke a label from a device. The role must
+    // also have label management permissions granted by a role
+    // with the `ChangeLabelManagingRole` permission above.
     RevokeLabel,
 
     // AFC
+    // The role can use AFC. This controls both the ability to
+    // receive a unidirectional AFC channels and the ability to
```
```suggestion
    // create/receive unidirectional AFC channels and the ability to
```

crates/aranya-daemon/src/policy.md
chip-so: Yeah I was trying to make it brief but you're right, it's not very clear.

crates/aranya-daemon/src/policy.md
chip-so: Alright. Removed the `CanUseAfc` permission requirement from assigning a label.

crates/aranya-daemon/src/policy.md
```diff
@@ -3568,9 +3594,6 @@ command AssignLabelToDevice {
 
         check can_manage_label(author.device_id, this.label_id)
 
-        // The target device must be able to use AFC.
-        check device_has_simple_perm(this.device_id, SimplePerm::CanUseAfc)
```
chip-so, line 3572: Hmm. I'm starting to question whether this did make sense. Because I think we all assumed it was checking whether the author had `CanUseAfc` permission, but it's actually checking the target device. Does it make sense to assign a label to a device that doesn't have `CanUseAfc`?

djohnson-spideroak: Yeah, I think requiring the target of AssignLabel has the CanUseAfc permission makes sense.

chip-so: Okay. I'll just revert this part of it, then.

crates/aranya-daemon/src/policy.md
```diff
@@ -3508,6 +3534,7 @@ fact LabelAssignedToDevice[label_id id, device_id id]=>{op enum ChanOp, device_g
 //
 // # Required Permissions
 //
+// - `CanUseAfc`
```
jdygert-spok: Should probably distinguish author vs target required permissions.

chip-so: Good call.

gknopf-aranya (approved):

PR #595, opened 2025-11-06
custom roles: unify owning_role, managing_role

gknopf-aranya:
Starting with the user-facing APIs and docs.
Internal renaming could happen on another PR or on this PR if there is time.


gknopf-aranya:
Owning roles and management permissions are different. Role owners can delegate role management permissions to other roles.

Issue #575, opened 2025-11-03
custom roles: add support for role_delete


gknopf-aranya:
The initial version of custom roles with default roles does not allow for customizing/deleting the default roles. Add support for `role_delete()` to the API so roles can be deleted.

After adding `role_delete()` to the API, test it in `test_role_create_assign_revoke()`. There is a TODO in the unit test for this:
`// TODO: show operation fails after delete_role() (after it is added to the API).`

cc: @djohnson-spideroak, @jdygert-spok, @chip-so

Issue #523, opened 2025-10-07
docs: remove redundant type links for parameters in C api


jdygert-spok:
We don't need to repeat the type of the parameter in the docs, the signature already has it as a link.

PR #598, opened 2025-11-07
custom roles: create/delete role, add/remove role perms

### LINKED ISSUE #594: custom roles: add in role definition
Add in the actions, commands, and API surface to support defining new custom roles.

chip-so:
Implements:
- `create_role()`
- `delete_role()`
- `add_perm_to_role()`
- `remove_perm_from_role()`

Scope of this PR has been expanded to include role permissions because custom roles are not very useful without any permissions assigned to them.

Closes #594, https://github.com/aranya-project/aranya/issues/606, https://github.com/aranya-project/aranya/issues/575

### DIFF
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -1776,11 +1776,11 @@ AranyaError aranya_team_roles_ext(const struct AranyaClient *client,
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
- * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param[in] role_name the name of the new role [`AranyaRoleName`](@ref AranyaRoleName).
- * @param[in] owning_role the [`AranyaRoleId`](@ref AranyaRoleId) that will own the new role.
- * @param[out] role_out the newly created [`AranyaRole`](@ref AranyaRole).
+ * @param[in] client the Aranya Client
+ * @param[in] team the team's ID
+ * @param[in] role_name the name of the new role
+ * @param[in] owning_role the role ID of the role that will own the new role
+ * @param[out] role_out the newly created role
  *
  * @relates AranyaClient
  */
@@ -1797,11 +1797,11 @@ AranyaError aranya_create_role(const struct AranyaClient *client,
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
- * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param[in] role_name the name of the new role [`AranyaRoleName`](@ref AranyaRoleName).
- * @param[in] owning_role the [`AranyaRoleId`](@ref AranyaRoleId) that will own the new role.
- * @param[out] role_out the newly created [`AranyaRole`](@ref AranyaRole).
+ * @param[in] client the Aranya Client
+ * @param[in] team the team's ID
+ * @param[in] role_name the name of the new role
+ * @param[in] owning_role the role ID of the role that will own the new role
+ * @param[out] role_out the newly created role
  *
  * @relates AranyaClient
  */
@@ -1820,9 +1820,9 @@ AranyaError aranya_create_role_ext(const struct AranyaClient *client,
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
- * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param[in] role the [`AranyaRoleId`](@ref AranyaRoleId) of the role to delete.
+ * @param[in] client the Aranya Client
+ * @param[in] team the team's ID
+ * @param[in] role the ID of the role to delete
  *
  * @relates AranyaClient
  */
@@ -1838,9 +1838,9 @@ AranyaError aranya_delete_role(const struct AranyaClient *client,
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
- * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param[in] role the [`AranyaRoleId`](@ref AranyaRoleId) of the role to delete.
+ * @param[in] client the Aranya Client
+ * @param[in] team the team's ID
+ * @param[in] role the ID of the role to delete
  *
  * @relates AranyaClient
  */
@@ -1856,10 +1856,10 @@ AranyaError aranya_delete_role_ext(const struct AranyaClient *client,
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
- * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param[in] role the [`AranyaRoleId`](@ref AranyaRoleId) the permission is being added to.
- * @param[in] perm a [`AranyaPermission`](@ref AranyaPermission) to add to the role.
+ * @param[in] client the Aranya Client
+ * @param[in] team the team's ID
+ * @param[in] role the role ID of the role the permission is being added to
+ * @param[in] perm a permission to add to the role
  */
 AranyaError aranya_add_perm_to_role(const struct AranyaClient *client,
                                     const struct AranyaTeamId *team,
@@ -1873,10 +1873,10 @@ AranyaError aranya_add_perm_to_role(const struct AranyaClient *client,
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
- * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param[in] role the [`AranyaRoleId`](@ref AranyaRoleId) the permission is being added to.
- * @param[in] perm a [`AranyaPermission`](@ref AranyaPermission) to add to the role.
+ * @param[in] client the Aranya Client
+ * @param[in] team the team's ID
+ * @param[in] role the role ID of the role the permission is being added to
+ * @param[in] perm a permission to add to the role
  */
 AranyaError aranya_add_perm_to_role_ext(const struct AranyaClient *client,
                                         const struct AranyaTeamId *team,
@@ -1891,10 +1891,10 @@ AranyaError aranya_add_perm_to_role_ext(const struct AranyaClient *client,
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
- * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param[in] role the [`AranyaRoleId`](@ref AranyaRoleId) the permission is being removed from.
- * @param[in] perm a [`AranyaPermission`](@ref AranyaPermission) to remove from the role.
+ * @param[in] client the Aranya Client
+ * @param[in] team the team's ID
+ * @param[in] role the role ID of the role the permission is being removed from
+ * @param[in] perm a permission to remove from the role
  */
 AranyaError aranya_remove_perm_from_role(const struct AranyaClient *client,
                                          const struct AranyaTeamId *team,
@@ -1908,10 +1908,10 @@ AranyaError aranya_remove_perm_from_role(const struct AranyaClient *client,
  *
  * Permission to perform this operation is checked against the Aranya policy.
  *
- * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
- * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
- * @param[in] role the [`AranyaRoleId`](@ref AranyaRoleId) the permission is being removed from.
- * @param[in] perm a [`AranyaPermission`](@ref AranyaPermission) to remove from the role.
+ * @param[in] client the Aranya Client
+ * @param[in] team the team's ID
+ * @param[in] role the role ID of the role the permission is being removed from
+ * @param[in] perm a permission to remove from the role
  */
 AranyaError aranya_remove_perm_from_role_ext(const struct AranyaClient *client,
                                              const struct AranyaTeamId *team,
@@ -2009,8 +2009,7 @@ AranyaError aranya_revoke_role_ext(const struct AranyaClient *client,
  * @param[in] client the Aranya Client
  * @param[in] team the team's ID
  * @param[in] name label name string
- * @param[in] managing_role_id the ID of the role that manages this
- *        label [`AranyaRoleId`](@ref AranyaRoleId).
+ * @param[in] managing_role_id the ID of the role that manages this label
  *
  * @relates AranyaClient.
  */
@@ -2028,8 +2027,7 @@ AranyaError aranya_create_label(const struct AranyaClient *client,
  * @param[in] client the Aranya Client
  * @param[in] team the team's ID
  * @param[in] name label name string
- * @param[in] managing_role_id the ID of the role that manages this
- *        label [`AranyaRoleId`](@ref AranyaRoleId).
+ * @param[in] managing_role_id the ID of the role that manages this label
  *
  * @relates AranyaClient.
  */
@@ -2585,8 +2583,8 @@ AranyaError aranya_team_devices_ext(const struct AranyaClient *client,
  * @param[in] team the team's ID
  * @param[out] device the ID of the device
  * @param[out] role_out the role assigned to the device. `role_out` will be zeroed
- * if a role was not assigned to the device. [`AranyaRole`](@ref AranyaRole).
- * @param[out] has_role whether a role is assigned to the device.
+ * if a role was not assigned to the device
+ * @param[out] has_role whether a role is assigned to the device
  *
  * @relates AranyaClient.
  */
@@ -2603,8 +2601,8 @@ AranyaError aranya_team_device_role(const struct AranyaClient *client,
  * @param[in] team the team's ID
  * @param[out] device the ID of the device
  * @param[out] role_out the role assigned to the device. `role_out` will be zeroed
- * if a role was not assigned to the device. [`AranyaRole`](@ref AranyaRole).
- * @param[out] has_role whether a role is assigned to the device.
+ * if a role was not assigned to the device
+ * @param[out] has_role whether a role is assigned to the device
  *
  * @relates AranyaClient.
  */
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1242,11 +1242,11 @@ pub unsafe fn team_roles(
 ///
 /// Permission to perform this operation is checked against the Aranya policy.
 ///
-/// @param[in] client the Aranya Client [`Client`].
-/// @param[in] team the team's ID [`TeamId`].
-/// @param[in] role_name the name of the new role [`RoleName`].
-/// @param[in] owning_role the [`RoleId`] that will own the new role.
-/// @param[out] role_out the newly created [`Role`].
+/// @param[in] client the Aranya Client
+/// @param[in] team the team's ID
+/// @param[in] role_name the name of the new role
+/// @param[in] owning_role the role ID of the role that will own the new role
+/// @param[out] role_out the newly created role
 ///
 /// @relates AranyaClient
 pub fn create_role(
@@ -1275,9 +1275,9 @@ pub fn create_role(
 ///
 /// Permission to perform this operation is checked against the Aranya policy.
 ///
-/// @param[in] client the Aranya Client [`Client`].
-/// @param[in] team the team's ID [`TeamId`].
-/// @param[in] role the [`RoleId`] of the role to delete.
+/// @param[in] client the Aranya Client
+/// @param[in] team the team's ID
+/// @param[in] role the ID of the role to delete
 ///
 /// @relates AranyaClient
 pub fn delete_role(client: &Client, team: &TeamId, role: &RoleId) -> Result<(), imp::Error> {
@@ -1293,10 +1293,10 @@ pub fn delete_role(client: &Client, team: &TeamId, role: &RoleId) -> Result<(), 
 ///
 /// Permission to perform this operation is checked against the Aranya policy.
 ///
-/// @param[in] client the Aranya Client [`Client`].
-/// @param[in] team the team's ID [`TeamId`].
-/// @param[in] role the [`RoleId`] the permission is being added to.
-/// @param[in] perm a [`Permission`] to add to the role.
+/// @param[in] client the Aranya Client
+/// @param[in] team the team's ID
+/// @param[in] role the role ID of the role the permission is being added to
+/// @param[in] perm a permission to add to the role
 pub fn add_perm_to_role(
     client: &Client,
     team: &TeamId,
@@ -1321,10 +1321,10 @@ pub fn add_perm_to_role(
 ///
 /// Permission to perform this operation is checked against the Aranya policy.
 ///
-/// @param[in] client the Aranya Client [`Client`].
-/// @param[in] team the team's ID [`TeamId`].
-/// @param[in] role the [`RoleId`] the permission is being removed from.
-/// @param[in] perm a [`Permission`] to remove from the role.
+/// @param[in] client the Aranya Client
+/// @param[in] team the team's ID
+/// @param[in] role the role ID of the role the permission is being removed from
+/// @param[in] perm a permission to remove from the role
 pub fn remove_perm_from_role(
     client: &Client,
     team: &TeamId,
@@ -1406,8 +1406,7 @@ pub fn revoke_role(
 /// @param[in] client the Aranya Client
 /// @param[in] team the team's ID
 /// @param[in] name label name string
-/// @param[in] managing_role_id the ID of the role that manages this
-///        label [`RoleId`].
+/// @param[in] managing_role_id the ID of the role that manages this label
 ///
 /// @relates AranyaClient.
 pub fn create_label(
@@ -1806,8 +1805,8 @@ pub unsafe fn team_devices(
 /// @param[in] team the team's ID
 /// @param[out] device the ID of the device
 /// @param[out] role_out the role assigned to the device. `role_out` will be zeroed
-/// if a role was not assigned to the device. [`Role`].
-/// @param[out] has_role whether a role is assigned to the device.
+/// if a role was not assigned to the device
+/// @param[out] has_role whether a role is assigned to the device
 ///
 /// @relates AranyaClient.
 pub fn team_device_role(


crates/aranya-daemon/src/actions.rs
```diff
@@ -185,6 +185,17 @@ where
         .in_current_span()
     }
 
+    /// Invokes `create_role`
+    #[instrument(skip(self))]
+    fn create_role(
+        &self,
+        role_name: Text,
+        owning_role_id: RoleId,
```
chip-so, line 193: This does require an owning role. I thought that made sense since every default role has an owning role and I guess that means we want a hierarchy. Could make it optional if we don't.

jdygert-spok (approved):

crates/aranya-client/tests/tests.rs
```diff
@@ -981,6 +981,34 @@ async fn test_setup_default_roles_rejects_unknown_owner() -> Result<()> {
...
+        .await?
+        .into_iter()
+        .find(|r| r.name == "owner")
+        .ok_or_else(|| anyhow::anyhow!("no owner role!?"))?;
+
+    owner_team
+        .create_role(text!("test_role"), owner_role.id)
+        .await?;
+
+    owner_team
```
gknopf-aranya, line 1007: May want to check that other devices can see this role after syncing with the device that created the role

crates/aranya-client/tests/tests.rs
```diff
@@ -981,6 +981,34 @@ async fn test_setup_default_roles_rejects_unknown_owner() -> Result<()> {
...
+        .roles()
+        .await?
+        .into_iter()
+        .find(|r| r.name == "test_role")
+        .ok_or_else(|| anyhow::anyhow!("no test role found"))?;
+
+    Ok(())
```
gknopf-aranya, line 1073: I'd like to test that the new role can have permissions granted to it and be assigned to a device on the team.

crates/aranya-client/tests/tests.rs
gknopf-aranya, line 1: @chip-so, @djohnson-spideroak, @jdygert-spok thoughts on adding a test case that attempts to escalate privileges on a newly created role to demonstrate it is rejected?

crates/aranya-daemon/src/actions.rs
gknopf-aranya: Yes, I believe creating a new role does require an owning role. We could default to having the current device's role be the owning role to simplify this. Although allowing any owning role to be specified here seems more flexible.

gknopf-aranya: I don't think we need to worry about privilege escalation if the role that creates the new role assigns an owning role with additional permissions it does not have. The owning role should be controlled by a device that is independent from the device that created the role. If the owning role is controlled by the device that created the role, we've already shown that privilege escalation is not possible because a role cannot grant permissions to another role it doesn't have.

crates/aranya-daemon/src/api.rs
```diff
@@ -677,6 +677,37 @@ impl DaemonApi for Api {
...
+        self.check_team_valid(team).await?;
+
+        let effects = self
+            .client
+            .actions(GraphId::transmute(team))
+            .create_role(role_name, RoleId::transmute(owning_role))
+            .await
+            .context("unable to create role")?;
+
+        if effects.len() != 1 {
```
gknopf-aranya: It's possible we could add other effects to this action in the future. We should search for the specific effect we're interested in here rather than requiring only a single effect to be emitted.

crates/aranya-daemon/src/policy.md
```diff
@@ -1725,6 +1728,8 @@ struct DefaultRole {
     author_id id,
     // The ID of the initial role owner.
     owning_role_id id,
+    // Is this a default role?
+    default bool,
 }
 
 // Creates the following facts for a default role
```
gknopf-aranya: Doc is out-of-date

crates/aranya-daemon/src/policy.md
```diff
@@ -1778,8 +1783,55 @@ effect RoleCreated {
...
+action create_role(role_name string, owning_role_id id) {
+    publish CreateRole {
+        role_name: role_name,
+        owning_role_id: owning_role_id,
+    }
+}
+
+command CreateRole {
+    attributes {
+        priority: 110
```
gknopf-aranya, line 1859: May be worth adding a note here that `DeleteRole` should be higher priority than this when implemented

crates/aranya-daemon/src/api.rs
chip-so: Is this because we might some day be able to update the policy without updating the daemon? It doesn't seem problematic to check that we receive exactly the number of effects we expect.

gknopf-aranya: The policy may currently only return a single effect. But it could potentially add another effect in the future. Elsewhere in this API we check for particular effects with `find_effect!`. An example of where an additional effect was added into the policy later is `CheckValidAfcChannels`.

jdygert-spok: I agree with chip that we should be more strict about the effects we expect from actions. I started doing this systematically for all actions in #560 but that's now outdated and getting pushed til after release.

chip-so: > The policy may currently only return a single effect. But it could potentially add another effect in the future.

I understand this but my question is why would we update the policy to produce another effect but not update this code that receives it?

jdygert-spok: I suppose some awkwardness comes from there essentially being two different kinds of effects: ones that need to be sent to the handler to affect "global" state, and ones that "return" values from an action.

crates/aranya-daemon/src/policy.md
```diff
@@ -1734,14 +1739,14 @@ struct DefaultRole {
 // - CanAssignRole
 // - CanRevokeRole
 // - CanChangeRolePerms
-finish function create_default_role(role struct DefaultRole) {
+finish function create_role_facts(role struct RoleInfo) {
```
gknopf-aranya, line 1789: We should be able to create the `DefaultRoleSeeded` fact inside `create_role_facts()` now that `RoleInfo` contains a bool indicating whether the role is a default role. This assumes that all the `assign_perm_to_role()` operations succeed. They should since they are defined as:
```
finish function assign_perm_to_role(role_id id, perm enum SimplePerm) {
    create RoleHasPerm[
        role_id: role_id,
        perm: perm,
    ]=>{}
}
```

crates/aranya-daemon/src/api.rs
gknopf-aranya: > I understand this but my question is why would we update the policy to produce another effect but not update this code that receives it?

Different code may be interested in different effects. For example, certain effects may be processed immediately when executing actions. Other effects may be handled in the background when `handle_effects()` is called by a peer.

gknopf-aranya (dismissed):
Lgtm. I left a few comments. Nothing that seems worth blocking on.

crates/aranya-client/tests/tests.rs
djohnson-spideroak: Agreed, if needed we can assign to someone else as a parallel followup

gknopf-aranya: I'll open a new issue for this

crates/aranya-daemon/src/api.rs
chip-so: > Different code may be interested in different effects. For example, certain effects may be processed immediately when executing an actions. Other effects may be handled in the background when `handle_effects()` is called by a peer.

So you're saying someone might update the policy to make `create_role()` emit another effect for an unrelated part of the code, and we want to relieve them of the burden of updating this part of the code?

crates/aranya-client/tests/tests.rs
gknopf-aranya: https://github.com/aranya-project/aranya/issues/604

crates/aranya-daemon/src/api.rs
gknopf-aranya: Yes, separation of concerns for independent effects

chip-so: I suppose that's a good reason. I'll update it.

crates/aranya-client/tests/tests.rs
```diff
@@ -981,6 +981,50 @@ async fn test_setup_default_roles_rejects_unknown_owner() -> Result<()> {
...
+async fn test_delete_role() -> Result<()> {
+    let mut devices = DevicesCtx::new("test_delete_role").await?;
+
+    let team_id = devices.create_and_add_team().await?;
+    let roles = devices
+        .setup_default_roles_without_delegation(team_id)
+        .await?;
+
+    let owner_team = devices.owner.client.team(team_id);
+    owner_team.delete_role(roles.member().id).await?;
```
gknopf-aranya: The internal policy error may not be as clear as:
```suggestion
    owner_team.delete_role(roles.member().id).await.expect("expected to delete role");
```

crates/aranya-client/tests/tests.rs
```diff
@@ -981,6 +981,50 @@ async fn test_setup_default_roles_rejects_unknown_owner() -> Result<()> {
...
+    let owner_role = owner_team
+        .roles()
+        .await?
+        .into_iter()
+        .find(|r| r.name == "owner")
+        .ok_or_else(|| anyhow::anyhow!("no owner role!?"))?;
+
+    owner_team
+        .create_role(text!("test_role"), owner_role.id)
+        .await?;
```
```suggestion
    owner_team
        .create_role(text!("test_role"), owner_role.id)
        .await.expect("expected to create role");
```

crates/aranya-daemon/src/policy.md
chip-so: I don't think you can because there's no way for a finish function to branch based on `role.default`.

crates/aranya-daemon/src/actions.rs
```diff
@@ -185,6 +185,24 @@ where
         .in_current_span()
     }
 
+    /// Invokes `create_role`
```
gknopf-aranya: nit:
```suggestion
    /// Invokes `create_role`.
```

crates/aranya-daemon/src/actions.rs
```diff
@@ -185,6 +185,24 @@ where
...
+    fn create_role(
+        &self,
+        role_name: Text,
+        owning_role_id: RoleId,
+    ) -> impl Future<Output = Result<Vec<Effect>>> + Send {
+        self.call_persistent_action(policy::create_role(role_name, owning_role_id.as_base()))
+            .in_current_span()
+    }
+
+    /// Invokes `delete_role`
```
gknopf-aranya: nit:
```suggestion
    /// Invokes `delete_role`.
```

crates/aranya-daemon/src/policy.md
```diff
@@ -1083,6 +1091,8 @@ other roles.
 // - `target_role_id` refers to the `Role` fact
 // - `owning_role_id` refers to the `Role` fact
 fact OwnsRole[target_role_id id, owning_role_id id]=>{}
+// Secondary index that tracks whether a role owns other roles
+fact RoleOwned[owner_role_id id, target_role_id id]=>{}
```
chip-so, line 1095: I added this secondary index so we could know that a role owns other roles. Otherwise you have to iterate over `OwnsRole` and that's not possible.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Ah, didn't realize a finish function couldn't handle an `if` statement. In that case, this seems fine to leave as is. We could work around that by changing `DefaultRoleSeeded[name]` to `RoleSeeded[name]=>{default bool}` but that may not be worth the effort.

crates/aranya-client/tests/tests.rs
chip-so: This is going to have to be daemon-side for now as granting permissions is not yet exposed in the client.

crates/aranya-daemon/src/policy.md
gknopf-aranya, line 1: It looks like we need to expose an action for assigning perms to roles. This wasn't needed previously because the permissions were set up internally when initializing the default roles.

gknopf-aranya: I see it now. It's called `add_perm_to_role` not `assign_perm_to_role`.

gknopf-aranya: It is currently missing from the daemon IPC API. The role perm tests can be moved from the daemon to the client once that API method is added.

crates/aranya-client/tests/tests.rs
gknopf-aranya: I think we should add methods to the API for adding/removing permissions from roles on this PR. There's no point in creating roles without any permissions assigned to them.

gknopf-aranya:
Missing updated Rust and C examples of `create_role()` `delete_role()`

crates/aranya-daemon/src/policy.md
gknopf-aranya: Opened issue for this:
https://github.com/aranya-project/aranya/issues/606

crates/aranya-client/tests/tests.rs
gknopf-aranya: Opened issue for this:
https://github.com/aranya-project/aranya/issues/606

crates/aranya-client/tests/tests.rs
```diff
@@ -981,6 +981,117 @@ async fn test_setup_default_roles_rejects_unknown_owner() -> Result<()> {
...
+        .await?
+        .into_iter()
+        .find(|r| r.name == "owner")
+        .expect("no owner role!?");
+
+    // Create a custom admin type role
+    let admin_role = owner_team
+        .create_role(text!("admin"), owner_role.id)
+        .await?;
+    owner_team
```
gknopf-aranya, line 1053: I thought the `CanChangeRolePerms` role management permission was required before changing role permissions.

chip-so: It is, but it's assigned by default in `create_role_facts()` when the role is created.

crates/aranya-client/tests/tests.rs
chip-so: APIs added.

crates/aranya-client/tests/tests.rs
```diff
@@ -981,6 +981,154 @@ async fn test_setup_default_roles_rejects_unknown_owner() -> Result<()> {
...
+
+    admin_team
+        .add_device(devices.operator.pk, None)
+        .await
+        .expect("admin should be able to add operator");
+
+    Ok(())
+}
+
+/// Tests that role creation works.
```
gknopf-aranya, line 1149: Copy-paste typo from another test

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1232,6 +1232,34 @@ pub unsafe fn team_roles(
...
+/// Create a role.
+pub fn create_role(
+    client: &Client,
+    team: &TeamId,
+    role_name: &str,
+    owning_role: &RoleId,
+    role_out: &mut MaybeUninit<Role>,
+) -> Result<(), imp::Error> {
+    let role_name = Text::from_str(role_name)
+        .map_err(|_| imp::Error::Other(anyhow!("bad role name string")))?;
```
chip-so: Still working on this, but I wanted some feedback. I don't think there's anywhere else we're accepting arbitrary text so I wasn't sure if this was a reasonable approach.

jdygert-spok: See `LabelName`. It is basically the same but gets a type alias. :shrug:

chip-so: Aaah, right. I forgot about label names. Guess I'll make `RoleName`, then.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1232,6 +1245,72 @@ pub unsafe fn team_roles(
...
+    // SAFETY: Caller must ensure `perm` is a valid C String.
+    let perm = unsafe { perm.as_underlying() }?;
+
+    client.rt.block_on(
+        client
+            .inner
+            .team(team.into())
+            .add_perm_to_role(role.into(), perm),
+    )?;
+    Ok(())
```
gknopf-aranya, line 1315: We could probably map errors so the result could be returned directly:
```suggestion
    client.rt.block_on(
        client
            .inner
            .team(team.into())
            .add_perm_to_role(role.into(), perm),
    ).map_err(...)
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1232,6 +1245,72 @@ pub unsafe fn team_roles(
     Ok(())
 }
 
+/// Create a role.
```
gknopf-aranya, line 1239: We usually have more detailed docs for each parameter

crates/aranya-client/src/client/team.rs
```diff
@@ -187,6 +187,67 @@ impl Team<'_> {
...
+            )
+            .await
+            .map_err(IpcError::new)?
+            .map_err(aranya_error)?;
+        Ok(Role::from_api(role))
+    }
+
+    /// Deletes a role.
+    ///
+    /// The role must not be assigned to any devices, nor should it own
```
gknopf-aranya, line 214: Does this include not having any role management perms assigned to it?

examples/c/example.c
```diff
@@ -1114,6 +1114,33 @@ AranyaError run_custom_roles_example(Team* t) {
...
+    AranyaRoleId buddy_role_id;
+    err = aranya_role_get_id(&buddy_role, &buddy_role_id);
+    EXPECT("unable to get role id", err);
+    printf("Role 'buddy' created\n");
+
+    // Add `CanUseAfc` permission to the custom role.
+    err = aranya_add_perm_to_role(&owner->client, &t->id, &buddy_role_id,
+                                  "CanUseAfc");
+    EXPECT("unable to assign 'CanUseAfc' to 'buddy'", err);
+    printf("Assigned 'buddy' the 'CanUseAfc' permission\n");
```
gknopf-aranya, line 1149: May be worth assigning the custom role to a device here to demonstrate the full lifecycle

gknopf-aranya:
Looks fairly close to being complete from what I can tell

crates/aranya-client/src/client/team.rs
chip-so: No. Role management perms are not checked because they would require their own reverse indexes.

examples/c/example.c
chip-so: I tried to reassign `memberb` but I couldn't get it to work for reasons I didn't feel like tracking down. I'll see if I can figure it out tomorrow.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -464,6 +464,19 @@ impl LabelName {
     }
 }
 
+/// A role name.
```
gknopf-aranya, line 467:
```suggestion
/// A role name.
///
/// E.g. "owner"
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I don't think we need to make this change on this PR. Just an observation of a potential improvement to `defs.rs`

jdygert-spok: I like the old way. We also couldn't always do the new way like when we need to write an out param or something. So it's more consistent to keep it that way.

crates/aranya-daemon/src/policy.md
```diff
@@ -1804,8 +1828,67 @@ effect RoleCreated {
...
 
-> **Note**: Custom roles are under development and will be
-> generally available after MVP.
+```policy
+// Creates a role
+//
+// This action does not (and cannot usefully) check for name
+// overlap. Do not assume that role names are unique.
+//
+// The owning role is by default granted all role management
```
gknopf-aranya, line 1843: Let's mention that each role requires at least 1 owning role and an owning role can add/remove other owning roles.

crates/aranya-daemon/src/policy.md
```diff
@@ -1804,8 +1828,67 @@ effect RoleCreated {
...
+        check team_exists()
+
+        let author = get_author(envelope)
+        // The author must have the permission to create a role
+        check device_has_simple_perm(author.device_id, SimplePerm::CreateRole)
+
+        // The owning role must exist
+        check exists Role[role_id: this.owning_role_id]
+
+        let role_id = derive_role_id(envelope)
```
gknopf-aranya, line 1880: We should add a check that the role being created does not already exist

gknopf-aranya: Unless, `let role_id = derive_role_id(envelope)` is guaranteed to always return a unique role ID.

crates/aranya-client/src/client/team.rs
gknopf-aranya: So if we delete a role that has management permissions, the management permissions are left in the factdb forever?

crates/aranya-daemon/src/policy.md
```diff
@@ -1984,8 +2072,72 @@ command SetupDefaultRole {
 
 ### Role Deletion
 
-Role deletion is waiting on prefix query deletion (see
```
gknopf-aranya, line 1987: Would be good to keep this note in whatever locations in the policy are waiting on prefix query deletion

examples/c/example.c
gknopf-aranya: A device that already has a role assigned to it will reject a new `assign_role()`. You could either call `change_role()` or `revoke_role()` followed by `assign_role()`. Another option would be to onboard a new device that doesn't already have a role assigned to it then `assign_role()` to it.

examples/c/example.c
```diff
@@ -1114,6 +1114,33 @@ AranyaError run_custom_roles_example(Team* t) {
...
+    EXPECT("unable to create role", err);
+    AranyaRoleId buddy_role_id;
+    err = aranya_role_get_id(&buddy_role, &buddy_role_id);
+    EXPECT("unable to get role id", err);
+    printf("Role 'buddy' created\n");
+
+    // Add `CanUseAfc` permission to the custom role.
+    err = aranya_add_perm_to_role(&owner->client, &t->id, &buddy_role_id,
+                                  "CanUseAfc");
+    EXPECT("unable to assign 'CanUseAfc' to 'buddy'", err);
```
```suggestion
    EXPECT("unable to add 'CanUseAfc' permission to 'buddy' role", err);
```

examples/c/example.c
```diff
@@ -1114,6 +1114,33 @@ AranyaError run_custom_roles_example(Team* t) {
...
+    err = aranya_remove_perm_from_role(&owner->client, &t->id, &buddy_role_id,
+                                       "CanUseAfc");
+    EXPECT("unable to remove 'CanUseAfc' permission from 'buddy' role", err);
+    printf("Removed 'CanUseAfc' permission from 'buddy' role\n");
+
+    // Delete the custom role.
+    err = aranya_delete_role(&owner->client, &t->id, &buddy_role_id);
+    EXPECT("unable to remove 'buddy' role", err);
+    printf("Removed 'buddy' role\n");
+
```
gknopf-aranya, line 1161: In a later PR, we could improve this example by querying before/after creating/deleting the role to show the expected number of roles are on the team. We could also demonstrate that the custom role allows a device it is assigned to to create an AFC channel. This isn't needed for coverage of the API methods themselves since that already happens elsewhere with the default roles demo. It is more so to demonstrate that we can exercise the full lifecycle with custom roles instead of the default roles we've been using previously.

crates/aranya-client/src/client/team.rs
chip-so: Maybe not forever, but as currently designed, they do remain. We could add some management commands to clean this up later, but because the role is the second key field, we'd have to iterate over them in an action. So we can't clean them up automatically in the command policy. And we would need a reverse index fact to detect them.

examples/c/example.c
gknopf-aranya: Opened an issue to implement this later:
https://github.com/aranya-project/aranya/issues/613

crates/aranya-daemon/src/policy.md
chip-so: I've added adding/removing other roles to the earlier "Role Ownership" section. And that section explains you can actually have zero owning roles (for some reason).

crates/aranya-daemon/src/policy.md
chip-so: It should be unique, yes.

crates/aranya-daemon/src/policy.md
chip-so: Since we've agreed to just decline the operation on blocking facts rather than clean them up, I don't think this is actually waiting on it anymore. And the removal of role management permission facts are not solved by prefix deletion. But I will leave a note about their fate.

examples/c/example.c
chip-so: Sorry, didn't see yours. Deleting.

examples/c/example.c
chip-so: Yeah, `change_role()` failed and I didn't have time to track it down. ~~I've lumped this into #615.~~

Nevermind I see you made #613.

crates/aranya-client/src/client/team.rs
chip-so: I have added a note about deleting these as a future enhancement.

crates/aranya-daemon/src/policy.md
gknopf-aranya: I thought there was a check in the policy that prevents the number of role owners from being less than 1.

chip-so: There is a check specifically for the `owner` role, but otherwise no.

gknopf-aranya: In `remove_role_owner()` there is this check:
```
        check at_least 2 OwnsRole[
            target_role_id: this.target_role_id,
            owning_role_id: ?,
        ]=>{}
```

chip-so: Oh. Huh. Well I'll update the docs then.

gknopf-aranya: Seems like a good way to prevent losing control of a role. Although if the owning role isn't assigned to any devices, we'd lose control anyways.

gknopf-aranya: Oh, looks like there's a check to make sure an owning role is assigned to at least one device:
```
if is_owner(role) {
    check at_least 2 RoleAssignmentIndex[
        role_id: this.role_id,
        device_id: ?,
    ]=>{}
}
```

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1762,6 +1769,156 @@ AranyaError aranya_team_roles_ext(const struct AranyaClient *client,
...
+/**
+ * Create a role.
+ *
+ * The `owning_role` is the initial owner of the new role.
+ *
+ * Permission to perform this operation is checked against the Aranya policy.
+ *
+ * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
+ * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
+ * @param[in] role_name the name of the new role.
```
gknopf-aranya: Should this have a reference like the other params? [`AranyaRoleName`]

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1221,6 +1236,113 @@ pub unsafe fn team_roles(
...
 
+/// Create a role.
+///
+/// The `owning_role` is the initial owner of the new role.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[in] role_name the name of the new role.
```
```suggestion
/// @param[in] role_name the name of the new role [`RoleName`].
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1221,6 +1236,113 @@ pub unsafe fn team_roles(
...
+/// Assign a permission to a role.
+///
+/// It is an error to add a permission already assigned to the role.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[in] role the [`RoleId`] the permission is being added to.
+/// @param[in] perm A [`Permission`] to add to the role.
```
```suggestion
/// @param[in] perm a [`Permission`] to add to the role.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1221,6 +1236,113 @@ pub unsafe fn team_roles(
...
+///
+/// @relates AranyaClient
+pub fn delete_role(client: &Client, team: &TeamId, role: &RoleId) -> Result<(), imp::Error> {
+    client
+        .rt
+        .block_on(client.inner.team(team.into()).delete_role(role.into()))?;
+    Ok(())
+}
+
+/// Assign a permission to a role.
```
gknopf-aranya: Docs and function names should be consistent:
```suggestion
/// Add a permission to a role.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1221,6 +1236,113 @@ pub unsafe fn team_roles(
...
+/// Remove a permission from a role.
+///
+/// It is an error to remove a permission not assigned to the role.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[in] role the [`RoleId`] the permission is being removed from.
+/// @param[in] perm A [`Permission`] to remove from the role.
```
```suggestion
/// @param[in] perm a [`Permission`] to remove from the role.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1221,6 +1236,113 @@ pub unsafe fn team_roles(
...
+pub fn delete_role(client: &Client, team: &TeamId, role: &RoleId) -> Result<(), imp::Error> {
+    client
+        .rt
+        .block_on(client.inner.team(team.into()).delete_role(role.into()))?;
+    Ok(())
+}
+
+/// Add a permission to a role.
+///
+/// It is an error to add a permission already assigned to the role.
```
```suggestion
/// It is an error to add a permission already added to the role.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1221,6 +1236,113 @@ pub unsafe fn team_roles(
...
+            .inner
+            .team(team.into())
+            .add_perm_to_role(role.into(), perm),
+    )?;
+    Ok(())
+}
+
+/// Remove a permission from a role.
+///
+/// It is an error to remove a permission not assigned to the role.
```
```suggestion
/// It is an error to remove a permission not added to the role.
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1221,6 +1236,113 @@ pub unsafe fn team_roles(
...
+///
+/// The `owning_role` is the initial owner of the new role.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
+///
+/// @param[in] client the Aranya Client [`Client`].
+/// @param[in] team the team's ID [`TeamId`].
+/// @param[in] role_name the name of the new role [`RoleName`].
+/// @param[in] owning_role the [`RoleId`] that will own the new role.
+/// @param[out] role_out the newly created [`Role`].
```
jdygert-spok: Please remove the type links ala #523.

gknopf-aranya: Done

chip-so:
I can't approve my own PR but the changes look good to me.

ssweetney-spok (approved):

PR #599, opened 2025-11-12
afc: use new seal ctx

jdygert-spok:


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1850,7 +1850,7 @@ pub unsafe fn team_label_exists(
 /// An AFC Sending Channel Object.
 #[cfg(feature = "afc")]
 #[aranya_capi_core::derive(Cleanup)]
-#[aranya_capi_core::opaque(size = 152, align = 8)]
+#[aranya_capi_core::opaque(size = 160, align = 8)]
```
gknopf-aranya, line 1853: What caused the size of this to change?

crates/aranya-client/src/afc.rs
```diff
@@ -248,13 +264,13 @@ impl SendChannel {
     /// # Panics
     ///
     /// Will panic on poisoned internal mutexes.
-    pub fn seal(&self, dst: &mut [u8], plaintext: &[u8]) -> Result<(), Error> {
+    pub fn seal(&mut self, dst: &mut [u8], plaintext: &[u8]) -> Result<(), Error> {
         debug!(?self.local_channel_id, ?self.label_id, "seal");
```
gknopf-aranya, line 268: Should we still trace `local_channel_id` if it's no longer used here?

gknopf-aranya (approved):

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Storing the `SealCtx` in the send channel object.

crates/aranya-client/src/afc.rs
jdygert-spok: Yeah still seems like useful debug info.

PR #603, opened 2025-11-12
custom roles capi: update setup_default_roles

### LINKED ISSUE #584: capi: `setup_default_roles` does not match Rust API
The rust API takes the owning role as a parameter while the C API looks up the owner.

ssweetney-spok:
closes #584


examples/c/example.c
```diff
@@ -384,11 +384,33 @@ AranyaError init_team(Team* t) {
     // real world scenario, the keys would be exchanged outside of Aranya using
     // something like `scp`.
 
+    size_t owner_role_len = 1;
+    AranyaRole owner_role[owner_role_len];
+
+    err = aranya_team_roles(&owner->client, &t->id, owner_role, &owner_role_len);
+    if (owner_role_len != 1) {
+        printf("There should only be 1 role after creating a team but there are %zu roles.\n", owner_role_len);
```
gknopf-aranya: More than 1 role here should print to `stderr` and return an error rather than attempting to continue.
```
fprintf(stderr, ...
return err
```

examples/c/example.c
```diff
@@ -384,11 +384,33 @@ AranyaError init_team(Team* t) {
...
+    }
+    if (err != ARANYA_ERROR_SUCCESS) {
+        fprintf(stderr, "unable to get list of roles\n");
+        return err;
+    }
+
+    // Get the ID of the owner role.
+    AranyaRoleId owner_role_id;
+    err = get_role_id_by_name(owner_role, owner_role_len, "owner",
```
gknopf-aranya: There's no need to iterate over the role list searching for a matching name if there's only one role returned. `aranya_role_get_name()` and `aranya_role_get_id()` can be called directly on `Role owner_role = team_roles[0]`

examples/c/example.c
```diff
@@ -384,11 +384,33 @@ AranyaError init_team(Team* t) {
     // real world scenario, the keys would be exchanged outside of Aranya using
     // something like `scp`.
 
+    size_t owner_role_len = 1;
+    AranyaRole owner_role[owner_role_len];
```
gknopf-aranya: Would prefer to call this `team_roles` then set `Role owner_role = team_roles[0]` after checking that only 1 role was returned.
```suggestion
    AranyaRole team_roles[owner_role_len];
```

examples/c/example.c
```diff
@@ -384,27 +384,23 @@ AranyaError init_team(Team* t) {
...
+    size_t team_roles_len = 1;
+    AranyaRole team_roles[team_roles_len];
 
-    err = aranya_team_roles(&owner->client, &t->id, owner_role, &owner_role_len);
-    if (owner_role_len != 1) {
-        printf("There should only be 1 role after creating a team but there are %zu roles.\n", owner_role_len);
+    err = aranya_team_roles(&owner->client, &t->id, team_roles, &team_roles_len);
+    if (team_roles_len != 1) {
+        fprintf(stderr, "There should only be 1 role after creating a team but there are %zu roles.\n", team_roles_len);
+        return err;
```
gknopf-aranya, line 393: Should this return an error other than success?

examples/c/example.c
```diff
@@ -384,11 +384,29 @@ AranyaError init_team(Team* t) {
...
+    // Get list of existing roles
+    size_t team_roles_len = 1;
+    AranyaRole team_roles[team_roles_len];
+
+    err = aranya_team_roles(&owner->client, &t->id, team_roles, &team_roles_len);
+    if (team_roles_len != 1) {
+        fprintf(stderr, "There should only be 1 role after creating a team but there are %zu roles.\n", team_roles_len);
+        return err;
+    }
+    if (err != ARANYA_ERROR_SUCCESS) {
```
gknopf-aranya: I would check the error first before `team_roles_len`. If there was an error, the value of `team_roles_len` may be invalid.

examples/c/example.c
ssweetney-spok: Yea, this could potentially return `ARANYA_ERROR_SUCCESS` ~~if the policy was updated to have another role available on a newly created team.~~

Edit: I think we'd have to update the policy to have no roles available for this to return "success" in this branch

With the current policy, this should be `ARANYA_ERROR_BUFFER_TOO_SMALL` or `ARANYA_ERROR_INVALID_ARG`

examples/c/example.c
ssweetney-spok: The length should still be valid if there's an error

examples/c/example.c
gknopf-aranya: I mean if `aranya_team_roles()` returns `ARANYA_ERROR_SUCCESS` but we got > 1 role instead of the expected list only containing the owner role, I don't think we should return `err` as-is. We could instead return `ARANYA_ERROR_OTHER` to indicate an error occurred in the example even though the API call succeeded.

examples/c/example.c
gknopf-aranya: That may be the case for the current implementation of `aranya_team_roles()`. However, in general I would not rely on output args being set correctly when API methods return an error. There's no guarantee over how far the function execution got before returning the error or whether the output arg was set at all.

examples/c/example.c
ssweetney-spok: > I mean if aranya_team_roles() returns ARANYA_ERROR_SUCCESS but we got > 1 role instead of the expected list

This shouldn't be possible since we'd get an error for the buffer size being too small

examples/c/example.c
ssweetney-spok: The API method we're calling is infallible at the moment. Do you think it will return an error in the future?

ssweetney-spok: Nvm, it isn't. I can move the check up

gknopf-aranya: I don't know. It's possible it could return an error in the future. E.g. if queries no longer work after team deletion.

gknopf-aranya (approved):

jdygert-spok (approved):

gknopf-aranya (approved):

PR #610, opened 2025-11-13
client: use enum for role management permission

jdygert-spok:


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -479,20 +479,28 @@ impl Addr {
...
+#[repr(u8)]
 #[derive(Copy, Clone, Debug)]
-pub struct Permission(*const c_char);
-
-impl Permission {
-    unsafe fn as_underlying(self) -> Result<Text, imp::Error> {
-        // SAFETY: Caller must ensure the pointer is a valid C String.
-        let cstr = unsafe { CStr::from_ptr(self.0) };
-        Ok(Text::try_from(cstr)?)
+pub enum RoleManagementPermission {
```
gknopf-aranya, line 503: What about using an enum for regular permissions? May want to rebase this PR off the base custom roles PR where `add_perm_to_role()` and `remove_perm_from_role()` were added to the API so both can be updated in the same PR.

examples/c/example.c
```diff
@@ -569,11 +569,11 @@ AranyaError init_team(Team* t) {
     // assign role management permissions.
     err = aranya_assign_role_management_permission(
         &owner->client, &t->id, &operator_role_id, &admin_role_id,
-        "CanAssignRole");
+        ARANYA_ROLE_MANAGEMENT_PERMISSION_CAN_ASSIGN_ROLE);
```
gknopf-aranya, line 594: This is really long. Could be shortened to:
```suggestion
        ARANYA_PERM_CAN_ASSIGN_ROLE);
```
We would just need to make sure regular perm names don't overlap with management perm names.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -495,20 +495,111 @@ impl Addr {
...
-///
-/// E.g. "CanAssignRole"
-///
-/// Refer to the "Role Management" section of the policy for an exhaustive list.
-#[repr(transparent)]
+/// Role management permission.
+#[repr(u8)]
 #[derive(Copy, Clone, Debug)]
-pub struct Permission(*const c_char);
+pub enum RoleManagementPermission {
```
gknopf-aranya, line 503: May want to add an internal note about keeping the perm enums up-to-date with `policy.md`. Don't want that to show up in the user-facing docs though.

crates/aranya-client/src/client/team.rs
```diff
@@ -367,7 +378,7 @@ impl Team<'_> {
                 self.id,
                 role.into_api(),
                 managing_role.into_api(),
-                perm,
+                perm.as_text(),
```
gknopf-aranya, line 378: Is the best place to convert from the perm enum to `Text` in the client-daemon IPC API or right at the daemon-policy boundary? If we can push this conversion closer to the policy, then these nice perm types can be used in the IPC interface which gives additional type-safety for simple vs. management perms. I think we can push the conversion farther down since the IPC API isn't required to be a stable API.

jdygert-spok: Agreed, but that is an internal change that can be done after release, unlike this API change.

gknopf-aranya: Since the enums are in the API, does this mean new perms must be added at the end for backward compatibility? And that we cannot remove any perms from the list in the future?

gknopf-aranya (dismissed):

djohnson-spideroak (dismissed):

gknopf-aranya:
Looks like CI cancelled a bunch of jobs on this PR for no apparent reason

gknopf-aranya (dismissed):

crates/aranya-client/src/client/team.rs
jdygert-spok: IMO we don't need to keep the discriminant int values the same, and removing items has moved from a runtime error to a compile error if you were using it, which IMO is just better and not really different in terms of being a breaking change in both ways.

gknopf-aranya:
@jdygert-spok this PR has some conflicts to resolve

gknopf-aranya (approved):

PR #617, opened 2025-11-13
afc: move out of preview

### LINKED ISSUE #614: afc: move out of preview
Move AFC out of preview into stable.

jdygert-spok:
Closes #614.

### DIFF
--- a/examples/rust/aranya-example-multi-node/Cargo.toml
+++ b/examples/rust/aranya-example-multi-node/Cargo.toml
@@ -18,6 +18,7 @@ default = [
 ]
 
 afc = ["aranya-client/afc"]
+preview = ["aranya-client/preview"]
 
 # Enable debugging.
 debug = []


gknopf-aranya:
@jdygert-spok looks like `Makefile.toml` may need some updates

gknopf-aranya:
Unless we make "afc" a default feature, we'll probably want build targets added for afc

crates/aranya-client-capi/Cargo.toml
```diff
@@ -19,7 +19,7 @@ doctest = false
 
 
 [features]
-default = []
+default = ["afc"]
```
gknopf-aranya: Won't enabling "afc" in the client/daemon automatically enable it in the API? I'm not sure it needs to be a default feature here.

jdygert-spok: ?

gknopf-aranya: Sorry, I was thinking this was the IPC API. This looks right.

gknopf-aranya (dismissed):

crates/aranya-client/Cargo.toml
```diff
@@ -29,7 +32,7 @@ experimental = ["aranya-daemon-api/experimental"]
 
 
 [dependencies]
-aranya-daemon-api = { workspace = true }
+aranya-daemon-api = { workspace = true, default-features = false }
```
djohnson-spideroak, line 35: why do we need to do this? override with the default features we set?

jdygert-spok: If you don't enable the default features for aranya-client, then afc isn't enabled, so aranya-daemon-api/afc shouldn't be enabled. But aranya-daemon-api/default also has afc, so we have to not enable that.

crates/aranya-daemon-api/Cargo.toml
```diff
@@ -14,7 +14,7 @@ workspace = true
 
 
 [features]
-default = []
+default = ["afc"]
```
gknopf-aranya, line 17: I meant to leave the other comment here. I don't think we should make "afc" a default feature here. We can let the client/daemon determine whether the IPC API is built with the "afc" feature. Otherwise, we risk there being a feature mismatch if for example the client doesn't enable default features, but this API does.

jdygert-spok: This is no different from the existing issue with the "afc" feature.

gknopf-aranya: When building the client/daemon, aren't they able to specify which features the `aranya-daemon-api` crate is built with? Why would we want the API to have any default features?

crates/aranya-client/Cargo.toml
```diff
@@ -14,7 +14,10 @@ workspace = true
 
 
 [features]
-default = []
+default = [
+    "afc",
+    "aranya-daemon-api/default",
```
gknopf-aranya, line 19:
```suggestion
    "aranya-daemon-api/afc",
```

crates/aranya-daemon/Cargo.toml
```diff
@@ -14,7 +14,10 @@ workspace = true
 
 
 [features]
-default = []
+default = [
+    "afc",
+    "aranya-daemon-api/default",
```
gknopf-aranya, line 19:
```suggestion
    "aranya-daemon-api/afc",
```

crates/aranya-client/Cargo.toml
jdygert-spok: We already have `afc = ["aranya-daemon-api/afc"]` below. It currently makes no difference to forward the default feature but it seems to make sense to me in general.

gknopf-aranya: If the client/daemon do not enable default features, does that automatically forward disabling default features in the api crate as well? If so, this seems fine.

jdygert-spok: yes, they are disabled.

```toml
aranya-daemon-api = { workspace = true, default-features = false }
```

djohnson-spideroak (approved):

PR #620, opened 2025-11-14
custom roles: put new custom roles methods behind preview

gknopf-aranya:
Put role permission management APIs behind "preview" feature flag until more security auditing has been completed.

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1218,7 +1218,6 @@ pub fn revoke_role_management_permission(
 /// @param[in] new_role the ID of the role to assign to the device
 ///
 /// @relates AranyaClient.
-#[cfg(feature = "preview")]
 pub fn change_role(
     client: &Client,
     team: &TeamId,
--- a/crates/aranya-daemon-api/src/service.rs
+++ b/crates/aranya-daemon-api/src/service.rs
@@ -410,7 +410,6 @@ pub trait DaemonApi {
     /// Revoke a role from a device.
     async fn revoke_role(team: TeamId, device: DeviceId, role: RoleId) -> Result<()>;
     /// Changes the assigned role of a device.
-    #[cfg(feature = "preview")]
     async fn change_role(
         team: TeamId,
         device: DeviceId,
--- a/crates/aranya-client/src/client/device.rs
+++ b/crates/aranya-client/src/client/device.rs
@@ -127,7 +127,6 @@ impl Device<'_> {
     }
 
     /// Changes the `role` on a `device`
-    #[cfg(feature = "preview")]
     #[instrument(skip(self))]
     pub async fn change_role(&self, old_role: RoleId, new_role: RoleId) -> Result<()> {
         self.client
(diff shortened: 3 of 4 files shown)


djohnson-spideroak (dismissed):
any makefile changes needed?

crates/aranya-daemon-api/src/service.rs
```diff
@@ -402,6 +410,7 @@ pub trait DaemonApi {
     /// Revoke a role from a device.
     async fn revoke_role(team: TeamId, device: DeviceId, role: RoleId) -> Result<()>;
     /// Changes the assigned role of a device.
+    #[cfg(feature = "preview")]
     async fn change_role(
```
jdygert-spok, line 413: assign/revoke too?

jdygert-spok: Or none of this group...?

djohnson-spideroak: wouldnt we still need assign/revoke for the default roles?

gknopf-aranya: `assign_role`, `change_role` and `revoke_role` are only a risk if simple/management perms of default roles can be modified.

gknopf-aranya: I think we should keep them out from behind preview since the vector we're concerned about is broken by putting perm modifications behind "preview".

djohnson-spideroak: Yeah, I think keeping the normal role management functions we had before out of preview is fine.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1311,6 +1318,7 @@ pub fn create_role(
 /// @param[in] role the ID of the role to delete
 ///
 /// @relates AranyaClient
+#[cfg(feature = "preview")]
```
gknopf-aranya, line 1321: Debatable if we need `delete_role()` behind preview. Since `create_role()` is already behind preview, the only roles that can be deleted are default roles which isn't a risk we're concerned about.

djohnson-spideroak: I think we should keep it behind preview. We may have to make changes to its behavior before we mark it stable.

gknopf-aranya: Will the API change if we modify its behavior? From the user's perspective it may only result in a doc/behavior change and not a breaking API change.

examples/c/example.c
```diff
@@ -796,13 +800,15 @@ AranyaError run(Team* t) {
     err = aranya_sync_hello_unsubscribe(&operator->client, &t->id,
                                         sync_addrs[ADMIN]);
     EXPECT("error unsubscribing operator from admin hello notifications", err);
-    #endif
+#endif
 
     err = run_afc_example(t);
     EXPECT("error running afc example", err);
 
+#ifdef ENABLE_ARANYA_PREVIEW
```
gknopf-aranya, line 808: I put the entire custom roles example behind the preview flag here since I don't think it's worth splitting out methods that aren't behind preview when we're aiming to get a release out today.

PR #623, opened 2025-11-14
release patch: don't publish examples to crates.io

gknopf-aranya:
`cargo vet check` failed after releasing 4.0.0:
https://github.com/aranya-project/aranya/actions/runs/19378549728/job/55451916422

This occurred because the aranya example crates were inadvertently published.
Set `publish = false` to no longer publish the aranya example crates.

We've requested that DevOps yank the crates from crates.io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -424,7 +424,7 @@ dependencies = [
 
 [[package]]
 name = "aranya-example"
-version = "4.0.0"
+version = "0.0.0"
 dependencies = [
  "anyhow",
  "aranya-client",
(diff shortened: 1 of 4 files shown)


mooreoak (dismissed):

gknopf-aranya:
> I don't know. But `cargo vet check` is suddenly complaining about lots of other things now: https://github.com/aranya-project/aranya/actions/runs/19378999443/job/55453410481?pr=623

Ah, looks like the example crates are held to a harsher standard now that they are part of the workspace. These appear to be the new `age` related deps. Perhaps we can ignore that since these are only examples.

gknopf-aranya:
@djohnson-spideroak, @jdygert-spok I don't think we can prove this fixes the problem without attempting a patch release. But this should work in theory, and is low consequence if it doesn't (we'll find out during the next release anyways).

jdygert-spok:
We can also yank the release ourselves.

djohnson-spideroak (approved):

PR #628, opened 2025-11-18
sync: Parallel Sync

### LINKED ISSUE #655: syncer: Refactor Primitives
Although our syncer is pretty robust, having built it up over many disparate requirements, it's gotten a bit tangled. The goal here is to separate out various "primitives" so that they're easier to update in the future, while adding a few choice upgrades. This includes:
* A `SyncManager` that handles all client-side orchestration, scheduling sync tasks, handling cleanup, as well as more complex behaviour in the future such as debounce and dynamic sync limits depending on the state of the wider system (i.e. how similar two graphs may be, and how many nodes are in the system/accessible to this node).
* A corresponding `SyncServer` that handles all server-side orchestration, scheduling tasks to accept sync requests and handling responses, as well as similar dynamic behaviour in the future.
* Several traits to help define the boundary between the manager/server and any specific transport implementation, which will allow easier migration between different implementations of the same transport, as well as allowing for more transports in the future.

Things that are not in this PR and are future work:
* More robust syncer behaviour (debounce, dynamic concurrent limits, modifying the sync order, etc)
* Quinn/mTLS/SNI migration
* Push Sync
* Better allocation management (buffer pool, dynamic-sized buffers; this PR still allocates `MAX_SYNC_MESSAGE_SIZE`, requires changes to core to acquire the length needed)
* Updating `SyncRequester`/`SyncResponder` in aranya-core to better handle this new system
* Allowing multiple messages to be sent per exchange (i.e. a "conversation")
* Proper specs documenting the intended behaviour.

nikki-aranya:
This current PR contains the current implementation, currently housed entirely in manager.rs but planned to split out into a handful of files.

Note that this is only for the "client" portion, we'll also need a SyncServer that listens for requests sent by the SyncManager so it can respond to them, but this is more or less what I'm proposing.


crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,1219 @@
...
+                buffer,
+                msg_type,
+            }) => {
+                // Remove the finished task from the active list.
+                self.active.remove(&peer);
+
+                // Store the leftover buffer we need to own.
+                if let Some(mut entry) = self.peers.get_mut(&peer) {
+                    entry.buffer = Some(buffer);
+                }
```
nikki-aranya, line 234: note to self: could probably simplify this once we add protocol::parse_message() and store one SyncProtocol per peer, and store buffer inside of it.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,1219 @@
...
+    Subscribe { peer: SyncPeer, config: HelloConfig },
+    Unsubscribe { peer: SyncPeer },
+    HelloNotification { peer: SyncPeer, head: Address },
+}
+
+struct SyncProtocol;
+
+impl SyncProtocol {
+    fn create_message(&self, message: ProtocolMessage, buffer: &mut Vec<u8>) -> Result<()> {
+        Ok(())
```
nikki-aranya, line 557: forgot to add it, SyncProtocol is just a fancy postcard::to_allocvec/to_io wrapper that is supposed to handle all de/serialization

Cargo.toml
```diff
@@ -54,24 +54,24 @@ aranya-policy-text = { version = "0.1.2" }
...
 aranya-runtime = { version = "0.14.0", features = ["std", "libc"] }
 
-buggy = { version = "0.1.0" }
-
 aranya-client = { version = "3.0.0", path = "crates/aranya-client" }
 aranya-daemon = { version = "3.0.0", path = "crates/aranya-daemon" }
 aranya-daemon-api = { version = "3.0.0", path = "crates/aranya-daemon-api" }
 aranya-keygen = { version = "3.0.0", path = "crates/aranya-keygen" }
 aranya-util = { version = "3.0.0", path = "crates/aranya-util" }
+buggy = { version = "0.1.0" }
```
gknopf-aranya, line 58: What is the reason for moving this here?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1026,8 +1027,20 @@ pub fn sync_peer_config_build(
 /// @param[in] interval Set the interval at which syncing occurs
 ///
 /// @relates AranyaSyncPeerConfigBuilder.
-pub fn sync_peer_config_builder_set_interval(cfg: &mut SyncPeerConfigBuilder, interval: Duration) {
+pub fn sync_peer_config_builder_set_interval(
```
gknopf-aranya, line 1038: Revert. This change appears to be from the sync hello PR which has been merged. It shouldn't show up in this diff.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya, line 1: Module docs describing the design would be useful at the top of this file.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,1219 @@
...
+    Message {
+        started_at: Instant,
+        handle: AbortHandle,
+    },
+    /// Single request that expects an ack.
+    Request {
+        started_at: Instant,
+        handle: AbortHandle,
+    },
+}
```
gknopf-aranya, line 83: This may be simpler than an enum since the fields are the same. The only difference is whether they receive an ack or response:
```suggestion
struct ActiveSync {
    /// When the sync request started.
    started_at: Instant,
    /// Sync abort handle.
    handle: AbortHandle,
    /// Requires a response.
    response: bool,
}
```

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,1219 @@
...
+
+impl ActiveSync {
+    fn abort_handle(&self) -> &AbortHandle {
+        match self {
+            Self::Message { handle, .. } => handle,
+            Self::Request { handle, .. } => handle,
+        }
```
gknopf-aranya, line 90:
```suggestion
        self.handle
```

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,1219 @@
...
+    }
+
+    fn task_id(&self) -> task::Id {
+        match self {
+            Self::Message { handle, .. } => handle.id(),
+            Self::Request { handle, .. } => handle.id(),
+        }
```
gknopf-aranya, line 97:
```suggestion
        self.handle.id()
```

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,1219 @@
...
+    }
+}
+
+enum MessageType {
+    PollSync,
```
gknopf-aranya, line 102: Why is this an enum if there is only one variant? Is this expecting to add push sync later?

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,712 @@
...
+
+    /// Send a hello notification to this peer.
+    SendNotification { peer: SyncPeer, head: Address },
+    /// Received a hello notification from a peer.
+    HelloReceived { peer: SyncPeer, head: Address },
+    /// Our graph head changed, so notify all subscribed peers.
+    GraphHeadChanged { graph_id: GraphId, head: Address },
+}
+
+// TODO(nikki): not happy having to keep 3 enums synced (ActiveSync, TaskResult, ProtocolMessage).
```
gknopf-aranya, line 71: I believe this indicates they could be combined into a enum/struct with more fields. So a single object is shared and kept in sync everywhere.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: Yes, I've been trying to factor in both push sync and hello sync design work into here.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: It's mostly about if we care about the additional information of what kind of sync it is, I can definitely flatten it.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya: At first glance, I think we can group all the requests into one enum and all the responses into another enum. I don't think we need this stand-alone enum for poll sync, when it is a request/(response/ack) just like sync hello, push sync, and any other sync protocol we add in the future.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,712 @@
...
+enum MessageType {
+    PollSync,
+}
+
+enum TaskResult {
+    /// Sync message that requires a response.
+    Message {
+        peer: SyncPeer,
+        result: Result<()>,
+        buffer: Vec<u8>,
```
gknopf-aranya, line 110: Is there a reason for using a buffer here rather than the types defined in the sync protocol in aranya-core?

gknopf-aranya: I prefer to serialize/deserialize messages as close to the wire as possible. Then named types can be used in the code as much as possible.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,712 @@
...
+struct HelloSubscription {
+    config: HelloConfig,
+    queue_key: Key,
+    last_sent: Option<Instant>,
+    current_head: Option<Address>,
+}
+
+struct HelloService {
+    subscriptions: DashMap<SyncPeer, HelloSubscription>,
+    queue: DelayQueue<SyncPeer>,
```
gknopf-aranya, line 589: Why does the hello service have its own delay queue? I think all scheduled sync requests could be placed in the same delay queue.

nikki-aranya: the base DelayQueue is because the default mode is poll syncing on a schedule, and then hello sync sends its own hello notifications in two cases, if a graph head changes and periodically (i.e. this DelayQueue) so that things don't get too stale.
You could imagine that a peer only cares about sync hello but not polling, or that it wants hello notifications way more often than it does full syncs.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya: We can figure that out by looking at the `bool` field. I would use an enum when there are multiple variants that have different fields or no fields. And a struct when there are variants that have the same fields (possibly with an enum type field). In this case, there are only 2 variants: response, ack. That can be handled by a `bool` or `Option`. `Option` could be beneficial if we want to keep track of the expected response type for each request.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: the buffer is explicitly so that we only have one allocation per peer that we can reuse across multiple sync tasks, and we have to return it here because there's no other way to return the owned memory back from a spawned task afaik
the memory allocation lifetimes are very explicitly designed here

nikki-aranya: > I prefer to serialize/deserialize messages as close to the wire as possible. Then named types can be used in the code as much as possible.

That's nice in theory, but the goal in this design is that we can very easily implement Transport+Handler for any customer's weird custom transport without having to worry about basically any of the mechanisms that hand it the data to send/receive. I know what you mean though, we could probably merge some of this code into core and simplify it, that'd have to be a phase two.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya: Right, periodic poll syncing and periodic sync hello messages could be scheduled on the same delay queue. Is there a reason for having 2 delay queues when 1 will also work?

crates/aranya-daemon/src/sync/manager.rs
mooreoak: I think your original version is better using types instead of a bool.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya: Is there documentation in the code on why the memory allocation lifetimes were designed this way?

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: I was under the impression that DelayQueue worked like a map where you can only have one value per key, but I looked it up and I guess you're right, we could just have one.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+                if let Some(peer) = panicked_peer {
+                    error!("sync task for peer {peer:?} panicked: {error:?}");
+
+                    // Remove the panicked peer from our active syncs.
+                    self.active.remove(&peer);
+
+                    // If the task failed and we "lost" the buffer, we need to allocate a new one.
+                    if let Some(mut entry) = self.peers.get_mut(&peer) {
+                        if entry.buffer.is_none() {
+                            let mut buffer = Vec::with_capacity(MAX_SYNC_MESSAGE_SIZE);
```
gknopf-aranya, line 288: We probably don't want to allocate the max size unless that many bytes is used.

nikki-aranya: That's one thing I wanted to change, I had it in previous drafts. Since we're passing the Vec into Transport/Handler and using postcard::to_io, the strategy I wanted to use is just do Vec::new() and then have it grow its own high watermark (it'll only ever hold the max size it needed for any one sync), since it can resize itself with e.g. write_all afaik.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: Let me add those, basically, we have one buffer per peer, allocated when we register a new peer in manager.peers. We only allow one sync operation per peer, so we core::mem::take() the buffer (it's None in manager.peers), we serialize data into it, we hand it to the spawned sync task using `async move`, we hand it to transport+handler which use it to send the serialized data, and then write the received response into the buffer, and then we return ownership of the buffer in the TaskResult, which we then store back in manager.peers for the next task. This allows us to have a single buffer for an entire sync, because we're only doing one thing at a time, all we have to do is shuffle it around a bit due to ownership restrictions.

gknopf-aranya: So if I understand correctly, the current implementation is parallelizing syncs for different peers but not parallelizing different sync operations with the same peer (e.g. you can't perform poll sync and hello sync with the same peer). I understand not wanting to perform multiple poll sync operations in parallel with the same peer since that would be wasteful. But, is there a reason we wouldn't want to support different sync types (poll, hello, and push) with a peer in parallel?

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+    /// Holds all in-progress sync tasks for later cleanup.
+    tasks: JoinSet<TaskResult>,
+    /// Accept messages from a [`SyncHandle`] for various events.
+    messages: mpsc::Receiver<ManagerMessage>,
+    /// Keeps track of registered peers' configuration info.
+    peers: DashMap<SyncPeer, PeerEntry>,
+    /// Maintains a rolling sync queue with all registered peers.
+    queue: DelayQueue<SyncPeer>,
+    /// Handles all "hello sync" task management.
+    hello_service: HelloService,
```
gknopf-aranya, line 145: If we combine the sync hello delay queue with the delay queue in `SyncManager`, I think we can remove this `hello_service` field and instead handle graph change hello messages the same way sync_now() is handled.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+    fn create_message(&self, message: ProtocolMessage, buffer: &mut Vec<u8>) -> Result<()> {
+        Ok(())
+    }
+
+    // TODO(nikki): parse_message()
+}
+
+// TODO(nikki): just combine these two?
+#[async_trait::async_trait]
+trait Transport: Send + Clone + 'static {
```
gknopf-aranya, line 565: I couldn't find anything that implements these traits

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+    async fn run_sync(&mut self, buffer: &mut Vec<u8>, peer: SyncPeer) -> Result<()>;
+}
+
+#[async_trait::async_trait]
+trait Handler: Send + Clone + 'static {
+    async fn handle_sync(&mut self, buffer: &mut Vec<u8>, peer: SyncPeer) -> Result<()>;
+}
+
+struct HelloConfig {
```
gknopf-aranya, line 574: This will end up being `pub(crate)` once it's used:
```suggestion
pub(crate) struct HelloConfig {
```

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya, line 1: General comment to add more:
- `pub(crate)`
- type/field docs

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+    /// The [`SyncProtocol`] serialization for this SyncManager.
+    protocol: SyncProtocol,
+    /// The [`Transport`] client for this SyncManager.
+    transport: T,
+    /// The [`Handler`] server for this SyncManager.
+    handler: H,
+}
+
+impl<T, H> SyncManager<T, H> {
+    /// Creates a new [`SyncManager`] with a given [`Transport`], [`Handler`], and message buffer size.
```
gknopf-aranya, line 155: Not sure if it's worth the effort here make buffer, transport, and handler generics. As far as I know, we aren't planning to change those soon.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya, line 1: It will be easier to debug this if more tracing is added. Specifically when:
- Scheduling a task
- Starting a task
- Completing a task
- Sending a request
- Receiving a response
- There is an error
- Graph head changes
- Subscribing/unsubscriping
- Adding/removing a sync peer

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+                let _ = self.handle.send_hello(peer, head).await;
+                entry.queue_key = self.queue.insert(peer, entry.config.schedule_delay);
+            }
+        }
+    }
+
+    async fn handle_hello_received<T: Transport, H: Handler>(
+        &mut self,
+        peer: SyncPeer,
+        _head: Address,
```
gknopf-aranya, line 673: Planning to use this parameter?

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+
+            if peer.graph_id != graph_id {
+                continue;
+            }
+
+            let state = entry.value_mut();
+            state.current_head = Some(head);
+
+            let should_send = match state.last_sent {
+                Some(last) => last.elapsed() >= state.config.graph_change_delay,
```
gknopf-aranya, line 702: This will cause the sync hello to be dropped if not enough time has elapsed since `graph_change_delay`.
If we can't send the hello right away, we could schedule one to be sent on the delay queue after `state.config.graph_change_delay - last.elapsed()`. I think we would only want to schedule that if `state.config.graph_change_delay - last.elapsed()` is less than `schedule_delay`.
This is only a latency optimization since periodic sync hello will eventually send another hello.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+            self.queue.remove(&entry.queue_key);
+        }
+    }
+
+    async fn send_message(&mut self, peer: SyncPeer, message: ProtocolMessage) -> Result<()> {
+        if self.active.contains_key(&peer) {
+            return Err(SyncError::AlreadyActive);
+        }
+
+        // TODO(nikki): limit max concurrent syncs?
```
gknopf-aranya, line 367: Seems less important to limit the number of outbound concurrent sync requests than to limit the number of inbound sync requests that are processed. Outbound concurrency is already limited by the local sync configurations/subscriptions. A high number of inbound requests could be due to a DOS attack.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+            _ => unreachable!(),
+        };
+
+        let mut buffer = {
+            let mut entry = self
+                .peers
+                .get_mut(&peer)
+                .ok_or(SyncError::PeerNotRegistered)?;
+
+            entry.buffer.take().ok_or(SyncError::BufferInUse)?
```
gknopf-aranya, line 380: Does the implementation prevent the `BufferInUse` error from occurring?

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya, line 1: For the most part, our existing unit tests will confirm these changes are backward compatible with the current syncer. However, it would be beneficial to add more unit tests to cover the new functionality that is being added. Esp.:
- Syncing with multiple peers in parallel
- Confirming parallel sync still works if a connection is lost with one of the peers

crates/aranya-daemon/src/sync/manager.rs
mooreoak: > But, is there a reason we wouldn't want to support different sync types (poll, hello, and push) with a peer in parallel?

I am not sure what the implication of that would be. Until we understand *why* we would want to do that I think it is best not to allow it.

nikki-aranya: Not necessarily, other than it's a lot more complexity for a questionable throughput percentage. I can note that down as a phase 2/3 thing, if we want to pursue that.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: That's on the TODO to get this ready to merge, I need to migrate the QUIC impl to this new service, I'm mostly concerned about getting the architecture correct rn.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: It's not about our default implementation, it's about customers wanting their own fancy custom transport.

mooreoak: We want the size of the buffer to be generic, not having the size of buffers pramertized is already a issue in the code base. Also having transport and handler generic will make this code useful in other integrations. In the future this code should be moved out of the deamon and into anraya core or something.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: to my knowledge yes, I still need to think through all the lifetimes to be certain but I can probably clean up some of these error cases, and then I can throw on .assume/.expect

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya: Poll updates the client. Hello and push update the server. So allowing these to happen in parallel would improve throughput. Allowing multiple sync polls with the same peer wouldn't improve throughput.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: This will also probably inevitably change the design since I also need to build out the listener (SyncServer) that the peer uses to respond to incoming requests.

mooreoak (changes requested):
Instead of trying to puzzle out how this is supposed to work myself, please can you write a `readme.md` or doc comment that explains how this is all supposed to work? I don't think I can properly review it without a explanation of what it is supposed to do, why, and how.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya: If customers want to use their own transport for sync is it more likely they will implement their own generic transport here? Or that they would prefer an API for implementing the transport in the application layer? For AFC, we ended up providing an API that allows the application to manage its own transport. Either way, this isn't something that is on our current product roadmap and I don't know what customers would be interested in this. We have discussed wanting a QUIC implementation that uses TLS certs for sync. I think implementing these generics is out-of-scope for the initial parallel sync phase.

gknopf-aranya:
> Instead of trying to puzzle out how this is supposed to work myself, please can you write a `readme.md` or doc comment that explains how this is all supposed to work? I don't think I can properly review it without a explanation of what it is supposed to do, why, and how.

@mooreoak, @nikki-aranya I've spent some time today reading through this code in an attempt to understand it. Here are the main design decisions I see:
- `JoinSet` keeps track of all the spawned syncer tasks
- `DashMap` allows multiple threads to access maps that would normally be stored in a `BTreeMap`
- Periodic sync requests and hello messages are scheduled via a `DelayQueue`
- Each task holds a buffer for sync request/response data that can be reused later so it doesn't need to be reallocated again
- `mpsc` messages are sent to the sync manager so it can configure sync peers and sync hello subscriptions without blocking the caller

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya: The error would be caught once unit tests are added. Don't want to use `expect()` since that would panic.

nikki-aranya:
> Instead of trying to puzzle out how this is supposed to work myself, please can you write a `readme.md` or doc comment that explains how this is all supposed to work? I don't think I can properly review it without a explanation of what it is supposed to do, why, and how.

I'll see if I can add a bunch more documentation tomorrow (ideally something we can port to aranya-docs), some minor cleanup, and then I'll ping you as a follow-up.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: That would be in SyncServer, here it's only for an established connection so there's not really a way for other peers to try to talk on it when responding.

crates/aranya-daemon/src/sync/manager.rs
jdygert-spok: I'm not sure there's much need to distinguish expecting an ACK when it could just be treated like an empty response. We do however have the hello msg which I believe we decided we want to fire-and-forget, which would have to be handled differently.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+}
+
+/// Orchestrates sync tasks for all registered peers.
+///
+/// There are two kinds of syncs supported:
+/// - a `message` task that requires both a request and a response.
+/// - a `request` task that only sends a request, response is an ack.
+pub struct SyncManager<T, H> {
+    /// Keeps track of all active sync tasks and their associated peer.
+    active: DashMap<SyncPeer, ActiveSync>,
```
jdygert-spok, line 135: I think all these dashmap uses could just be regular maps, since they are only accessed with mutable access available.

crates/aranya-daemon/src/sync/manager.rs
jdygert-spok: I could see keeping it separate for organization.

crates/aranya-daemon/src/sync/manager.rs
jdygert-spok: One concern there is that you probably don't want your `sync_now` request to fail because a hello msg or something is being sent. Especially if you go to retry and keep failing for new in-progress operations since there wouldn't be any queueing. Not so sure what a `sync_now` during a sync should do.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+
+                // Ready to notify a subscribed peer, spawn its task.
+                Some(expired) = self.hello_service.queue.next() => {
+                    self.hello_service.send_scheduled_hello(expired.into_inner()).await;
+                }
+
+                // No tasks ready, sleep briefly.
+                else => {
+                    tokio::time::sleep(Duration::from_millis(100)).await;
+                }
```
jdygert-spok, line 214: I think this should only happen if nothing is in the queues/join_set and the `SyncHandle`s have been dropped, so we can just return.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -0,0 +1,714 @@
...
+        match msg {
+            ManagerMessage::AddPeer { peer, config } => {
+                self.add_peer(peer, config).await;
+            }
+            ManagerMessage::RemovePeer { peer } => {
+                self.remove_peer(peer);
+            }
+            ManagerMessage::SyncNow { peer } => {
+                let message = ProtocolMessage::SyncRequest { peer };
+                let _ = self.send_message(peer, message).await;
```
jdygert-spok, line 311: Log errors if they are being ignored.

NWPlayer123:
@jdygert-spok this can probably get closed since the majority of this landed in phase 1 and 2, I think #655 is done too, #660 is still an open question once #786 lands

Issue #636, opened 2025-11-19
bug: long blocking operations can cause daemon IPC to timeout


djohnson-spideroak:
If a long blocking operation like a sync takes longer than the tarpc timeout, tarpc will timeout and the operation may not occur. Consider changing the timeout per-request based on the possibility of executing a long-running blocking task on the daemon.

PR #637, opened 2025-11-19
daemon: schedule `add_peer`'s `sync_now`

jdygert-spok:
Updates #636.

### DIFF
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -231,6 +231,7 @@ enum AranyaError
 typedef uint32_t AranyaError;
 #endif // __cplusplus
 
+#if defined(ENABLE_ARANYA_PREVIEW)
 /**
  * Simple permission.
  */
@@ -318,7 +319,9 @@ enum AranyaPermission
 #ifndef __cplusplus
 typedef uint8_t AranyaPermission;
 #endif // __cplusplus
+#endif
 
+#if defined(ENABLE_ARANYA_PREVIEW)
 /**
  * Role management permission.
  */
@@ -346,6 +349,7 @@ enum AranyaRoleManagementPermission
 #ifndef __cplusplus
 typedef uint8_t AranyaRoleManagementPermission;
 #endif // __cplusplus
+#endif
 
 /**
  * * ─────────────────────── Extended‐error (_ext) Variants ───────────────────────
@@ -1558,6 +1562,7 @@ AranyaError aranya_sync_peer_config_builder_set_sync_later(struct AranyaSyncPeer
 AranyaError aranya_sync_peer_config_builder_set_sync_later_ext(struct AranyaSyncPeerConfigBuilder *cfg,
                                                                struct AranyaExtError *__ext_err);
 
+#if defined(ENABLE_ARANYA_PREVIEW)
 /**
  * Sets whether automatic syncing should occur when a hello message is received from this peer
  * indicating they have a head that we don't have.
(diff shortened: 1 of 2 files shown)


crates/aranya-daemon/src/sync/task.rs
```diff
@@ -310,10 +313,12 @@ pub trait SyncState: Sized {
 impl<ST> Syncer<ST> {
     /// Add a peer to the delay queue, overwriting an existing one.
     fn add_peer(&mut self, peer: SyncPeer, cfg: SyncPeerConfig) {
-        // Only insert into delay queue if interval is configured
-        let new_key = cfg
-            .interval
-            .map(|interval| self.queue.insert(peer.clone(), interval));
+        // Only insert into delay queue if interval is configured or `sync_now == true`
+        let new_key = match cfg.interval {
+            _ if cfg.sync_now => Some(self.queue.insert_at(peer.clone(), Instant::now())),
```
gknopf-aranya, line 318: The delay queue normally reinserts syncs at the interval. We should make sure sync_now operations are not reinserted.

jdygert-spok: We only reinsert when the interval is set in the config.
```rust
// Re-insert into queue if interval is still configured
*key = cfg.interval.map(|interval| self.queue.insert(peer.clone(), interval));
```

gknopf-aranya: What if `interval` and `sync_now` are both set?

jdygert-spok: Then this schedules an immediate sync, then after performing that sync it reschedules it for after `interval`.

djohnson-spideroak, line 318: what is the prefix _? and is it just jumping the queue for scheduling the next sync?

jdygert-spok: The `_` pattern is ignoring the `cfg.interval` for that arm and instead just checking `cfg.sync_now`. This could also be written like this, not sure if it would be more clear.

```rust
        let new_key = if cfg.sync_now {
            Some(self.queue.insert_at(peer.clone(), Instant::now()))
        } else {
            cfg.interval
                .map(|interval| self.queue.insert(peer.clone(), interval))
        };
```

The sync is scheduled to immediately happen, although it might come after other syncs that are also ready.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1505,11 +1505,11 @@ AranyaError aranya_sync_peer_config_builder_set_interval_ext(struct AranyaSyncPe
                                                              struct AranyaExtError *__ext_err);
 
 /**
- * Updates the config to enable immediate syncing with the peer.
+ * Updates the config so the peer will be scheduled for an immediate sync when added.
  *
  * Overrides [`aranya_sync_peer_config_builder_set_sync_later`](@ref aranya_sync_peer_config_builder_set_sync_later) if invoked afterward.
  *
- * By default, the peer is synced with immediately.
+ * By default, the peer is scheduled for an immediate sync.
```
djohnson-spideroak, line 1516: Is this true? I thought sync_now was default false.

ssweetney-spok:
```
impl Default for SyncPeerConfigBuilder {
    fn default() -> Self {
        Self {
            interval: None,
            sync_now: true,
            #[cfg(feature = "preview")]
            sync_on_hello: false,
        }
    }
}
```

In "crates/aranya-client/src/config.rs" and "crates/aranya-client-capi/src/imp/config.rs"

PR #656, opened 2025-12-10
policy: use enums for perms everywhere

gknopf-aranya:
We are currently using enums for permissions in the Aranya client Rust and C APIs.
These are converted into a `Text` string at the daemon IPC API and passed as `string` to the policy interface.
Internally, the policy converts the strings to an enum again for commands/facts, then converts back to a string before returning the permission via an effect.

There is no reason to perform all these type conversions when permissions can be represented as an enum everywhere.

### DIFF
--- a/crates/aranya-client/src/client.rs
+++ b/crates/aranya-client/src/client.rs
@@ -19,6 +19,7 @@ use aranya_daemon_api::{
     DaemonApiClient, Version, CS,
 };
 #[cfg(feature = "preview")]
+#[cfg_attr(docsrs, doc(cfg(feature = "preview")))]
 #[doc(inline)]
 pub use aranya_daemon_api::{
     RoleManagementPerm as RoleManagementPermission, SimplePerm as Permission,
--- a/crates/aranya-daemon/src/vm_policy.rs
+++ b/crates/aranya-daemon/src/vm_policy.rs
@@ -1,8 +1,8 @@
 #![allow(missing_docs)]
 
-use std::{fmt, marker::PhantomData, str::FromStr};
+use std::{fmt, marker::PhantomData};
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use aranya_afc_util::Ffi as AfcFfi;
 use aranya_crypto::{
     keystore::{fs_keystore::Store, KeyStore},
@@ -31,19 +31,6 @@ use crate::{
 /// Policy loaded from policy.md file.
 pub(crate) const POLICY_SOURCE: &str = include_str!("./policy.md");
 
-/// Converts [`ChanOp`] to string.
-impl FromStr for ChanOp {
-    type Err = anyhow::Error;
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s {
-            "ChanOp::RecvOnly" => Ok(Self::RecvOnly),
-            "ChanOp::SendOnly" => Ok(Self::SendOnly),
-            "ChanOp::SendRecv" => Ok(Self::SendRecv),
-            _ => Err(anyhow!("unknown `ChanOp`: {s}")),
-        }
-    }
-}
-
 /// Display implementation for [`ChanOp`]
 impl fmt::Display for ChanOp {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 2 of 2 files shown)


crates/aranya-daemon/src/test.rs
```diff
@@ -1173,50 +1158,14 @@ async fn test_add_perm_to_role_requires_management_delegation() -> Result<()> {
...
         .expect_err("expected add_perm_to_role without delegation to fail");
     expect_not_authorized(err);
 
     Ok(())
 }
 
-/// Rejects add_perm_to_role when the permission string is invalid.
-#[test(tokio::test(flavor = "multi_thread"))]
-#[serial]
-async fn test_add_perm_to_role_rejects_invalid_permission() -> Result<()> {
```
gknopf-aranya, line 1187: Removing this test because specifying an invalid permission will be caught at compile time rather than at runtime due to the stricter permission type.

chip-so (dismissed):

crates/aranya-client/src/client/perm.rs
```diff
@@ -18,12 +18,26 @@ pub enum RoleManagementPermission {
...
+    fn from(value: api::RoleManagementPerm) -> Self {
+        match value {
+            api::RoleManagementPerm::CanAssignRole => RoleManagementPermission::CanAssignRole,
+            api::RoleManagementPerm::CanRevokeRole => RoleManagementPermission::CanRevokeRole,
+            api::RoleManagementPerm::CanChangeRolePerms => {
+                RoleManagementPermission::CanChangeRolePerms
+            }
+        }
+    }
+}
```
jdygert-spok: We have been trying to avoid client impls that mention the daemon API types for stability reasons. Although some have slipped through and I'm honestly not sure that we need to given that we have declared the daemon API internal and unstable.

gknopf-aranya: Any changes to the perm enums in the policy will be a breaking API change regardless of whether we reference the IPC or policy types directly from the client. The client enums would need to be updated if there are any changes.

gknopf-aranya: I even considered exposing the perm enum type from the policy directly in the client APIs. The reason for creating the type wrappers in the client and IPC APIs is for backward compatibility. However, if any changes to the enums will cause a breaking change anyways, we might as well have fewer types that get redefined while breaking the API.

jdygert-spok: This is not exactly about changes to the enums but about keeping `aranya_daemon_api` out of the public API. If we aren't doing that then we might just want to `pub use aranya_daemon_api::RoleManagementPermission` instead.

gknopf-aranya: What if we define aliases for the perm enum in the IPC and client APIs? That way they don't expose a type from another crate directly because the alias appears to be its own type.

jdygert-spok: There isn't really a difference with an alias as far as rust is concerned. But we can use `#[doc(inline)]` on the re-export so it appears in the docs as if it were defined in the client crate.

gknopf-aranya: I'm considering using the same approach to re-export the policy enums from the daemon crate. As long as each crate reexports the type, we can define it differently from the underlying type if we ever need to in the future (for backward compatibility). Until then, I think it makes sense to use the same definitions for all the policy interface enums. The problem with this is that means aranya-daemon would become a dep of aranya-daemon-api which would introduce a circular dependency. I don't see a point in separating out the daemon IPC API into its own crate if there aren't plans to make it backward compatible so this could be resolved by moving the IPC API into the aranya-daemon crate. I do see a purpose in keeping the IPC API in its own crate so the client only needs to import the API as a dep. This could be resolved by keeping the aranya-daemon-api crate but changing it so it re-exports a subset of daemon types rather than defining its own types and API interface.

gknopf-aranya: This would be a larger IPC API refactor effort that I don't intend to attempt on this PR. Just some thoughts on possible improvements we could make in the future.

jdygert-spok: We could pull the interface generation out into the daemon api crate or another crate to do that. But the generated types don't have documentation, and lack more specific types like `DeviceId` instead of `BaseId`.

gknopf-aranya: I pushed a commit that re-exports the enums from the daemon ipc api in the client.

crates/aranya-daemon/src/vm_policy.rs
```diff
@@ -47,6 +51,60 @@ impl fmt::Display for ChanOp {
     }
 }
 
+/// Converts [`RoleManagementPerm`] to string.
+impl FromStr for RoleManagementPerm {
```
jdygert-spok: The 3 `FromStr` impls here are unused.

crates/aranya-client/src/client.rs
```diff
@@ -2,21 +2,27 @@
...
         txp::{self, LengthDelimitedCodec},
         PublicApiKey,
     },
     DaemonApiClient, Version, CS,
 };
+#[cfg(feature = "preview")]
+#[doc(inline)]
+pub use aranya_daemon_api::{
+    RoleManagementPerm as RoleManagementPermission, SimplePerm as Permission,
+};
```
jdygert-spok, line 26: I'm not sure how, but I checked and these do get the `preview` marker in the rust docs. I didn't think there was an automatic form of that unless `auto_doc_cfg` was enabled, and these don't have a `doc(cfg(...))` on them here or at the source :thinking:

gknopf-aranya: I put the `#[cfg(feature = "preview")]` in front of the re-exports hoping that would mean they would be put behind `preview`.

crates/aranya-daemon/src/vm_policy.rs
gknopf-aranya: Surprised that wasn't caught by the correctness checks or a linter

crates/aranya-client/src/client.rs
jdygert-spok: Right, that puts them behind the feature flag, but generally the feature flag attribute doesn't show up in the rust docs without extra work.

crates/aranya-daemon/src/vm_policy.rs
gknopf-aranya: Oh, it may have just been a warning. Maybe we want linter warnings to fail in CI.

jdygert-spok: It's because the types are exported from `aranya-daemon` so technically they could be used downstream. (Although not sure unused trait impls are detected anyways.)

gknopf-aranya: Removed these unused impls.

crates/aranya-client/src/client.rs
jdygert-spok: Looks like `doc_auto_cfg` has been merged into `doc_cfg` around September and is now enabled by default when that is enabled.

jdygert-spok: We can probably go remove the doc cfgs and double check the docs, in another PR.

jdygert-spok (approved):

crates/aranya-client/src/client.rs
jdygert-spok: Our nightly is older than that though :thinking: I'm guessing inline re-exports just behave differently due to a bug or something.

PR #662, opened 2025-12-10
ci: fix cargo deny errors/warnings

gknopf-aranya:
```
warning: 3 allowed warnings found
[cargo-make] INFO - Execute Command: "cargo" "deny" "check"
warning[license-exception-not-encountered]: license exception was not encountered
    ┌─ /Users/gknopf/scm/aranya/deny.toml:148:38
    │
148 │     { allow = ["AGPL-3.0"], crate = "aranya-example" },
    │                                      ━━━━━━━━━━━━━━ unmatched license exception

warning[license-exception-not-encountered]: license exception was not encountered
    ┌─ /Users/gknopf/scm/aranya/deny.toml:149:38
    │
149 │     { allow = ["AGPL-3.0"], crate = "aranya-example-multi-node" },
    │                                      ━━━━━━━━━━━━━━━━━━━━━━━━━ unmatched license exception

error[unmaintained]: rustls-pemfile is unmaintained
    ┌─ /Users/gknopf/scm/aranya/Cargo.lock:301:1
    │
301 │ rustls-pemfile 2.2.0 registry+https://github.com/rust-lang/crates.io-index
    │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
```

`rustls-pemfile` is no longer maintained. This can be ignored since we don't use certs and there isn't a safe upgrade available.
The license exceptions can be removed since they are causing a warning. I suspect it's because we aren't publishing the example crates.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -79,6 +79,8 @@ ignore = [
     # The `paste` crate is unmaintained. There isn't anything we
     # can do here.
     "RUSTSEC-2024-0436",
+    # The `rustls-pemfile` crate is unmaintained. We're not using certs anyways.
+    "RUSTSEC-2025-0134",
 ]
 # If this is true, then cargo deny will use the git executable to fetch advisory database.
 # If this is false, then it uses a built-in git library.
@@ -144,9 +146,6 @@ exceptions = [
     { allow = ["AGPL-3.0"], crate = "aranya-daemon-api" },
     { allow = ["AGPL-3.0"], crate = "aranya-keygen" },
     { allow = ["AGPL-3.0"], crate = "aranya-util" },
-
-    { allow = ["AGPL-3.0"], crate = "aranya-example" },
-    { allow = ["AGPL-3.0"], crate = "aranya-example-multi-node" },
 ]
 
 # Some crates don't have (easily) machine readable licensing information,
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -37,8 +37,8 @@ user-login = "epage"
 user-name = "Ed Page"
 
 [[publisher.aranya-afc-util]]
-version = "0.16.0"
-when = "2025-11-05"
+version = "0.17.0"
+when = "2025-11-13"
 user-id = 293722
 user-login = "aranya-project-bot"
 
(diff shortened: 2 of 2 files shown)


deny.toml
```diff
@@ -79,6 +79,8 @@ ignore = [
     # The `paste` crate is unmaintained. There isn't anything we
     # can do here.
     "RUSTSEC-2024-0436",
+    # The `rustls-pemfile` crate is unmaintained. We're not using certs anyways.
```
djohnson-spideroak, line 82: is rustls-pemfile only a dep for s2n?

gknopf-aranya: That's my understanding. It shouldn't be a dep of anything else since we removed AQC and AFC doesn't have a transport. The multinode example has a custom transport but we aren't using certs for that.

deny.toml
```diff
@@ -79,6 +79,8 @@ ignore = [
...
+    "RUSTSEC-2025-0134",
```
jdygert-spok, line 83: We should track this somehow so we remove this when we migrate to quinn.

gknopf-aranya: Won't there be a warning if this exception no longer applies? Doesn't quinn depend on tls certs?

gknopf-aranya: Hopefully it will be maintained again by that time. No way to predict that though.

jdygert-spok: rustls-pemfile was upstreamed and deprecated, so it won't become maintained. When we switch to quinn, we need to check that we aren't still ignoring this error but then actually using it. It looks like quinn only uses this as a dev dependency but we don't want it to sneak in somewhere. It should warn _if_ we aren't using it at that point, but if something else is using it then it wouldn't warn. We also don't really see the warnings unless we check for them.

gknopf-aranya: Then we'll want to make an issue for removing this exception and add it to the quinn migration tracking issue.

gknopf-aranya: https://github.com/aranya-project/aranya/issues/663

PR #664, opened 2025-12-10
chore: add clang-format to Makefile.toml

### LINKED ISSUE #508: example: Add `.clang-format` to build/CI
Relying on an IDE to format consistently has proven unreliable; we have a `.clang-format` file, we just need to add a new Makefile task and enforce it in CI.

jdygert-spok:
Closes #508.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,4 @@
 ---
-Language: C
 BasedOnStyle: Google
 AlignAfterOpenBracket: Align
 AlignConsecutiveAssignments: true


crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1,3 +1,6 @@
+// clang-format off
+// This header is @generated by cbindgen.
```
gknopf-aranya, line 2: Did `cbindgen` generate this part of the file or was that added manually?
```
// clang-format off
// This header is @generated by cbindgen.
```
Want to make sure the next time the file is generated this part doesn't get removed/overwritten.

.clang-format
gknopf-aranya, line 1: Are you intending to remove `.clang-format` from the C example?

Makefile.toml
```diff
@@ -183,11 +183,19 @@ run_task = { name = [
 ] }
 
 [tasks.check-fmt]
-dependencies = ["check-fmt-rust", "check-fmt-toml"]
+dependencies = [
+    "check-fmt-rust",
+    "check-fmt-toml",
+    "check-fmt-c",
```
gknopf-aranya, line 189: Confirming that this means CI will catch any formatting errors in the C code for us?

crates/aranya-client-capi/output/aranya-client.h
jdygert-spok: I added it to the cbindgen.toml so it is always generated.

.clang-format
jdygert-spok: It was moved to the repo root. clang-format searches upward from the file being formatted.

Makefile.toml
jdygert-spok: Yep.

gknopf-aranya (approved):

jdygert-spok:
This should be ready now. You can see here that it detects bad formatting:
https://github.com/aranya-project/aranya/actions/runs/20115816202/job/57724911799?pr=665

PR #651, opened 2025-12-03
custom roles: rank-based authorization

gknopf-aranya:
## Summary

Introduces a rank-based authorization system to the Aranya policy and implements it across the full stack: policy engine, daemon API, client libraries (Rust + C), tests, and examples.

### Policy changes (`policy.md`)

- **Rank system**: Every object (device, role, label) has a numerical rank. Higher-ranked objects can operate on lower-ranked objects, providing a simple, auditable authorization hierarchy.
- **Removed role ownership/management**: Replaced `OwnsRole`, `RoleOwned`, `CanAssignRole`, `CanRevokeRole`, `CanChangeRolePerms`, `CanManageLabel` facts and their associated commands with rank comparisons.
- **Unified permissions**: Combined `RoleManagementPerm` and `SimplePerm` into a single `Perm` enum. Added `ChangeRank`, `ChangeRolePerms`. Removed `ChangeRoleManagementPerms`, `ChangeRoleManagingRole`, `ChangeLabelManagingRole`.
- **Permission segmentation**: Roles can grant permissions they don't possess, enabling true separation of duties. Default roles separate `ChangeRolePerms` (Admin) from `AssignRole`/`RevokeRole` (Operator) to prevent privilege escalation.
- **Default rank constants**: Defined as `global let` constants in policy (`DEFAULT_OWNER_DEVICE_RANK`, `DEFAULT_ADMIN_ROLE_RANK`, etc.) so they are authoritative and queryable at runtime.
- **Immutable role ranks**: Role ranks are set at creation time and cannot be changed afterward.
- **Privilege escalation mitigations**: Documented scenarios with examples, security warnings for `ChangeRolePerms` + `AssignRole` combination, and permission matrix table.

### Breaking API changes

**Changed signatures:**
- `setup_default_roles` — removed `owning_role` parameter. Now checks buffer size before performing the action: returns `BufferTooSmall` with the required count if the buffer is too small, letting the caller allocate and retry.
- `add_device` / `add_device_to_team` — added `rank` parameter
- `create_label` — replaced `managing_role_id` parameter with `rank`
- `create_role` — replaced `owning_role` parameter with `rank` (was behind preview)
- `Permission` enum — added `ChangeRank`, `ChangeRolePerms`. Removed `ChangeRoleManagementPerms`, `ChangeRoleManagingRole`, `ChangeLabelManagingRole`.

**Removed:**
[...]

PR #669, opened 2025-12-12
policy: review entire custom roles rank policy

gknopf-aranya:
The PR has been created to make it easier to review the entire custom roles rank policy rather than only the diff.

Base branch `do-not-merge-delete-policy` is `main` with the `policy.md` file deleted.
`do-not-merge-custom-roles-rank-entire-policy` is the same as `650-custom-roles-rank` rebased on `do-not-merge-delete-policy`.


crates/aranya-daemon/src/policy.md
```diff
@@ -0,0 +1,3669 @@
...
+- **Devices** are the primary identity, each containing cryptographic
+  key pairs for identity, signing, and encryption.
+- **Teams** provide the organizational boundary, with exactly one
+  team per graph. Only devices added to the team can access the
+  commands published to the team's graph.
+- **Roles** control command permissions. Currently we expose a set of
+  "default roles" - `owner` (emergency access), `admin` (system
+  administration), `operator` (user management), and `member` (basic
+  usage). This policy has the capability to define custom roles as well.
+- **Rank** is an attribute associated with each object that determines whether a higher ranked object can operate on a lower ranked object.
```
gknopf-aranya, line 35: May be worth mentioning the types of objects that can have a rank: anything with an Aranya ID including devices, roles, labels. Team doesn't need a rank since operations are implicitly performed on a team.

gknopf-aranya: Moving this comment to original rank spec PR

crates/aranya-daemon/src/policy.md
```diff
@@ -0,0 +1,3669 @@
...
+```policy
+// Records the public half of the device's Signing Key.
+fact DeviceSignPubKey[device_id id]=>{key_id id, key bytes}
+```
+
+#### Device Encryption Key
+
+The Device Encryption Key is a KEM key used to securely send
+encapsulated secret keys to other devices. It is primarily used
+by AFC.
```
gknopf-aranya, line 314: The way this is worded makes it sound like the device encryption key is used directly by AFC. My understanding is that this key is used to derive other keys which are used by AFC.

gknopf-aranya:
After I reviewed the entire default policy, I pushed this commit to the rank spec with some improvements:
https://github.com/aranya-project/aranya/pull/651/commits/04fb178fb2b75476e2f9c5650946ca6df5fd0e10

gknopf-aranya:
This is not intended to be merged and is out-of-date with the latest rank spec. Closing for now.

PR #671, opened 2025-12-15
fix: replace panic with error handling for empty hello response

gknopf-aranya:
Replace `assert!(!response_buf.is_empty())` in `send_hello_request()` with proper error handling using a new `Error::EmptyResponse` error variant.

This prevents a potential DoS vulnerability where a malicious peer could crash the daemon by sending an empty response to hello requests.

### DIFF
--- a/crates/aranya-daemon/src/sync/task/hello.rs
+++ b/crates/aranya-daemon/src/sync/task/hello.rs
@@ -20,6 +20,7 @@ use tracing::{debug, instrument, trace, warn};
 use crate::{
     aranya::ClientWithState,
     sync::{
+        error::SyncError,
         task::{
             quic::{Error, Server, State},
             PeerCacheKey, SyncPeers, Syncer,
@@ -184,7 +185,9 @@ impl Syncer<State> {
         recv.read_to_end(&mut response_buf)
             .await
             .with_context(|| format!("failed to read hello {} response", operation_name))?;
-        assert!(!response_buf.is_empty());
+        if response_buf.is_empty() {
+            return Err(SyncError::EmptyResponse);
+        }
         Ok(())
     }
 
--- a/crates/aranya-daemon/src/sync/mod.rs
+++ b/crates/aranya-daemon/src/sync/mod.rs
@@ -6,7 +6,7 @@ use error::SyncError;
 /// Possible sync related errors
 pub type Result<T> = core::result::Result<T, SyncError>;
 
-mod error {
+pub(crate) mod error {
     use std::convert::Infallible;
 
     use thiserror::Error;
@@ -24,6 +24,9 @@ mod error {
         SendSyncRequest(Box<SyncError>),
         #[error("Could not receive sync response: {0}")]
         ReceiveSyncResponse(Box<SyncError>),
+        /// Peer sent an empty response
+        #[error("peer sent empty response")]
+        EmptyResponse,
         #[error(transparent)]
         Bug(#[from] buggy::Bug),
         #[error(transparent)]


crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -93,6 +93,9 @@ pub enum Error {
     /// QUIC server endpoint start error
     #[error("could not start QUIC server")]
     ServerStart(#[source] StartError),
+    /// Peer sent an empty response
+    #[error("peer sent empty response")]
+    EmptyResponse,
```
djohnson-spideroak: Is this the right place for this error type? It seems like it might be more of a sync protocol error than the other errors around here.

gknopf-aranya: I originally told claude to add the error to `SyncError`, but it decided `QuicSync` was better:
```
> Create a new SyncError variant to handle this error 

● Let me find the SyncError type definition first.

● Search(pattern: "enum.*Error|struct.*Error", path: "crates/aranya-daemon/src/sync", output_mode: "content")
  ⎿  Found 2 lines (ctrl+o to expand)

● I see the error structure. The quic::Error enum is the appropriate place for this. Let me add a new variant for empty responses and then use it in hello.rs.
```

gknopf-aranya: Going to test how well claude can make this requested change

gknopf-aranya: Looks like it was able to move the error variant to the correct error enum.

crates/aranya-daemon/src/sync/task/hello.rs
```diff
@@ -184,7 +185,9 @@ impl Syncer<State> {
         recv.read_to_end(&mut response_buf)
             .await
             .with_context(|| format!("failed to read hello {} response", operation_name))?;
-        assert!(!response_buf.is_empty());
+        if response_buf.is_empty() {
+            return Err(SyncError::EmptyResponse);
+        }
```
ssweetney-spok, line 190: I wonder if this `assert` was meant to make it into production or if it was only added to help with debugging during development. I ask because there isn't a similar check when checking the response to a sync poll message.

May want to check with @benz-spideroak

gknopf-aranya: I'm guessing this was originally for test purposes and not intended to make it into production. It was likely missed during the code review. It could be that we want to abort the sync right away if this condition occurs, but I doubt causing the entire daemon to panic on this sync error was the intent.

gknopf-aranya: While I'd prefer to check with @benz-spideroak first, if we want this change in the next release we should probably merge this PR in this week. It depends on if we plan to ship the next release out to any customers that may want this fix. cc: @djohnson-spideroak

ssweetney-spok: > I'm guessing this was originally for test purposes and not intended to make it into production. It was likely missed during the code review. It could be that we want to abort the sync right away if this condition occurs, but I doubt causing the entire daemon to panic on this sync error was the intent.

Yea, I agree. If the assert wasn't supposed to be there maybe the right approach is to just remove it rather than returning this error? That would keep the behavior similar to sending a poll sync message.

gknopf-aranya: The client is expecting a response from the server. If no response is received, that would be an error. Poll sync already has a sync request/response as part of the protocol, so doesn't need to worry about handling an empty response.

ssweetney-spok: Ah, I just looked at how we handle it in `aranya_daemon::sync::task::quic::Syncer::receive_sync_response`. An empty response will get an error with a message like "postcard unable to deserialize sync response" which doesn't seem as useful as this one. May want to use this new error variant there as well.

gknopf-aranya: That's a good idea, though much lower priority than getting a panic out of production daemon code.

PR #673, opened 2025-12-18
syncer: add jitter before sending sync hello notifications

gknopf-aranya:
Introduces a random jitter on the interval of `[0, graph_change_delay/10]` before sending each hello notification.

Jitter does not need to be introduced before connection/stream establishment because the goal is to spread out the returned sync requests after hello notifications are sent out.

TODO:
- Should the jitter be configurable in the sync config?

### DIFF
--- a/crates/aranya-daemon/src/sync/task/hello.rs
+++ b/crates/aranya-daemon/src/sync/task/hello.rs
@@ -318,6 +318,11 @@ impl Syncer<State> {
             // This helps to prevent self DoS via a bunch of subsequent sync requests at the same time from various peers.
             let rand = Rng.next_u32();
             let jitter = graph_change_delay.mul_f64(f64::from(rand) / f64::from(u32::MAX)) / 10;
+            trace!(
+                ?jitter,
+                ?peer,
+                "applying jitter before sending hello notification"
+            );
             sleep(jitter).await;
 
             let (mut recv, mut send) = stream.split();


crates/aranya-daemon/src/sync/task/hello.rs
```diff
@@ -10,10 +10,11 @@ use std::{
 };
 
 use anyhow::Context;
+use aranya_crypto::{dangerous::spideroak_crypto::csprng::rand::RngCore as _, Rng};
```
mooreoak, line 13: We could use a rng from rust `core` here if we wanted but this one is not wrong to use, it just might be slower.

crates/aranya-daemon/src/sync/task/hello.rs
```diff
@@ -307,6 +314,17 @@ impl Syncer<State> {
         // Spawn async task to send the notification
         let peer = *peer;
         self.hello_tasks.spawn(async move {
+            // Introduce jitter with random sleep in [0, graph_change_delay/10].
+            // This helps to prevent self DoS via a bunch of subsequent sync requests at the same time from various peers.
+            let rand = Rng.next_u32();
+            let jitter = graph_change_delay.mul_f64(f64::from(rand) / f64::from(u32::MAX)) / 10;
```
mooreoak, line 320: why divided by 10?

crates/aranya-daemon/src/sync/task/hello.rs
```diff
@@ -307,6 +314,17 @@ impl Syncer<State> {
...
+            trace!(
+                ?jitter,
+                ?peer,
+                "applying jitter before sending hello notification"
+            );
+            sleep(jitter).await;
```
mooreoak, line 326: would the delay be better at the call point of `send_hello_notification_to_subscriber` rather then here in the function?

crates/aranya-daemon/src/sync/task/hello.rs
gknopf-aranya: This means the jitter can be anywhere from zero to 10% of the `graph_change_delay`. If we don't divide by 10, then the jitter could be the entire `graph_change_delay` which could double the delay in the worst case.

crates/aranya-daemon/src/sync/task/hello.rs
gknopf-aranya: `send_hello_notification_to_subsriber()` is called within a loop:
```
for (peer, subscription) in subscribers {
```
This loop is not executed in a background task, so if we add a jitter there, the loop would take `jitter * num_subscribers` additional time to execute which would introduce latency on the main thread.
The jitter is intentionally added inside the `spawn` for each hello message so it doesn't block the main thread.

PR #674, opened 2025-12-19
syncer: Reorganize Syncer for Future Changes

nikki-aranya:
This PR shuffles around the existing code into new modules, as well as updating the includes, to allow for easier git diffs as we work on a new Transport trait and the Parallel Sync feature.

Note that other changes include updating the generics so we don't have to worry about those in the future. Also, "SyncPeers" has been renamed to "SyncHandle" to better convey the idea that we're sending messages to the syncer to act on. "Syncer" has been renamed to "SyncManager" as in the future it will act as more of a manager, scheduling sync tasks and handling more comprehensive sync behavior such as debounce and adaptive sync task limits. Finally, the multiple different (Addr, GraphId) types have been combined into a single SyncPeer type which has been exposed to the outer daemon which will allow for easier tracking of unique peer identities.

### DIFF
--- a/crates/aranya-daemon/src/daemon.rs
+++ b/crates/aranya-daemon/src/daemon.rs
@@ -32,7 +32,7 @@ use crate::{
     keystore::{AranyaStore, LocalStore},
     policy,
     sync::{
-        quic::{PskStore, QuicState as QuicSyncClientState, SyncParams},
+        quic::{PskStore, QuicState, SyncParams},
         SyncHandle, SyncManager,
     },
     util::{load_team_psk_pairs, SeedDir},
@@ -54,7 +54,7 @@ pub(crate) type SP = LinearStorageProvider<FileManager>;
 pub(crate) type EF = policy::Effect;
 
 pub(crate) type Client = aranya::Client<EN, SP>;
-pub(crate) type SyncServer = crate::sync::quic::QuicServer<EN, SP>;
+pub(crate) type SyncServer = crate::sync::quic::Server<EN, SP>;
 
 mod invalid_graphs {
     use std::{
@@ -121,7 +121,7 @@ impl DaemonHandle {
 #[derive(Debug)]
 pub struct Daemon {
     sync_server: SyncServer,
-    manager: SyncManager<QuicSyncClientState, EN, SP, EF>,
+    manager: SyncManager<QuicState, EN, SP, EF>,
     api: DaemonApiServer,
     span: tracing::Span,
 }
@@ -320,7 +320,7 @@ impl Daemon {
     ) -> Result<(
         Client,
         SyncServer,
-        SyncManager<QuicSyncClientState, EN, SP, EF>,
+        SyncManager<QuicState, EN, SP, EF>,
         SyncHandle,
         mpsc::Receiver<(GraphId, Vec<EF>)>,
     )> {
--- a/crates/aranya-daemon/src/sync/hello.rs
+++ b/crates/aranya-daemon/src/sync/hello.rs
@@ -18,7 +18,7 @@ use tracing::{debug, instrument, trace, warn};
 use crate::{
     aranya::ClientWithState,
     sync::{
-        transport::quic::{QuicError, QuicServer, QuicState},
+        transport::quic::{self, QuicState},
         Addr, Error, GraphId, Result, SyncHandle, SyncManager, SyncPeer,
     },
 };
@@ -148,8 +148,8 @@ where
         // Send the message
         send.send(bytes::Bytes::from(data))
             .await
-            .map_err(QuicError::from)?;
-        send.close().await.map_err(QuicError::from)?;
+            .map_err(quic::Error::from)?;
+        send.close().await.map_err(quic::Error::from)?;
 
         // Determine operation name from sync_type
         let operation_name = match &sync_type {
@@ -386,7 +386,7 @@ fn spawn_scheduled_hello_sender<EN, SP>(
     });
 }
 
-impl<EN, SP> QuicServer<EN, SP>
+impl<EN, SP> quic::Server<EN, SP>
 where
     EN: Engine + Send + 'static,
     SP: StorageProvider + Send + 'static,
--- a/crates/aranya-daemon/src/sync/mod.rs
+++ b/crates/aranya-daemon/src/sync/mod.rs
@@ -27,7 +27,7 @@ pub(crate) enum Error {
     // TODO(nikki): generalize for generic transport support.
     /// Something went wrong inside the QUIC Syncer.
     #[error(transparent)]
-    QuicSync(#[from] quic::QuicError),
+    QuicSync(#[from] quic::Error),
 
     /// Something went wrong in the Aranya Runtime.
     #[error(transparent)]
--- a/crates/aranya-daemon/src/sync/transport/quic/client.rs
+++ b/crates/aranya-daemon/src/sync/transport/quic/client.rs
@@ -25,10 +25,13 @@ use tokio::sync::mpsc;
 use tokio_util::time::DelayQueue;
 use tracing::{error, instrument, trace};
 
-use super::{PskStore, QuicError, SharedConnectionMap, SyncState, ALPN_QUIC_SYNC};
+use super::{PskStore, SharedConnectionMap, SyncState, ALPN_QUIC_SYNC};
 use crate::{
     aranya::ClientWithState,
-    sync::{Addr, Callback, Error, GraphId, Result, SyncManager, SyncPeer, SyncResponse},
+    sync::{
+        transport::quic, Addr, Callback, Error, GraphId, Result, SyncManager, SyncPeer,
+        SyncResponse,
+    },
     InvalidGraphs,
 };
 
@@ -74,7 +77,7 @@ impl QuicState {
             .with_io((client_addr.host(), client_addr.port()))
             .assume("can set quic client address")?
             .start()
-            .map_err(QuicError::ClientStart)?;
+            .map_err(quic::Error::ClientStart)?;
 
         Ok(Self {
             client,
@@ -315,8 +318,8 @@ where
 
         send.send(Bytes::from(send_buf))
             .await
-            .map_err(QuicError::from)?;
-        send.close().await.map_err(QuicError::from)?;
+            .map_err(quic::Error::from)?;
+        send.close().await.map_err(quic::Error::from)?;
         trace!("sent sync request");
 
         Ok(())
--- a/crates/aranya-daemon/src/sync/transport/quic/mod.rs
+++ b/crates/aranya-daemon/src/sync/transport/quic/mod.rs
@@ -22,20 +22,20 @@ mod server;
 pub(crate) use client::QuicState;
 pub(crate) use connections::{ConnectionUpdate, SharedConnectionMap};
 pub(crate) use psk::{PskSeed, PskStore};
-pub(crate) use server::QuicServer;
+pub(crate) use server::Server;
 
 /// ALPN protocol identifier for Aranya QUIC sync.
 const ALPN_QUIC_SYNC: &[u8] = b"quic-sync-unstable";
 
 /// Errors specific to the QUIC syncer
 #[derive(Debug, thiserror::Error)]
-pub(crate) enum QuicError {
+pub(crate) enum Error {
     /// QUIC Connection error
     #[error(transparent)]
-    QuicConnectionError(#[from] connection::Error),
+    QuicConnection(#[from] connection::Error),
     /// QUIC Stream error
     #[error(transparent)]
-    QuicStreamError(#[from] stream::Error),
+    QuicStream(#[from] stream::Error),
     /// Invalid PSK used for syncing
     #[error("invalid PSK used when attempting to sync")]
     InvalidPSK,
@@ -47,7 +47,7 @@ pub(crate) enum QuicError {
     ServerStart(#[source] StartError),
 }
 
-impl From<Infallible> for QuicError {
+impl From<Infallible> for Error {
     fn from(err: Infallible) -> Self {
         match err {}
     }
--- a/crates/aranya-daemon/src/sync/transport/quic/server.rs
+++ b/crates/aranya-daemon/src/sync/transport/quic/server.rs
@@ -30,19 +30,21 @@ use tokio::sync::mpsc;
 use tokio::sync::Mutex;
 use tracing::{error, info, info_span, instrument, trace, warn, Instrument as _};
 
-use super::{ConnectionUpdate, PskStore, QuicError, SharedConnectionMap, ALPN_QUIC_SYNC};
+use super::{ConnectionUpdate, PskStore, SharedConnectionMap, ALPN_QUIC_SYNC};
 #[cfg(feature = "preview")]
 use crate::sync::hello::HelloSubscriptions;
 use crate::{
     aranya::ClientWithState,
-    sync::{Addr, Callback, Error, GraphId, Result, SyncHandle, SyncPeer, SyncResponse},
+    sync::{
+        transport::quic, Addr, Callback, Error, GraphId, Result, SyncHandle, SyncPeer, SyncResponse,
+    },
 };
 
 /// The Aranya QUIC sync server.
 ///
 /// Used to listen for incoming `SyncRequests` and respond with `SyncResponse` when they are received.
 #[derive_where(Debug)]
-pub(crate) struct QuicServer<EN, SP> {
+pub(crate) struct Server<EN, SP> {
     /// Thread-safe Aranya client paired with caches and hello subscriptions, ensuring safe lock ordering.
     client: ClientWithState<EN, SP>,
     /// QUIC server to handle sync requests and send sync responses.
@@ -58,7 +60,7 @@ pub(crate) struct QuicServer<EN, SP> {
     local_addr: SocketAddr,
 }
 
-impl<EN, SP> QuicServer<EN, SP>
+impl<EN, SP> Server<EN, SP>
 where
     EN: Engine + Send + 'static,
     SP: StorageProvider + Send + 'static,
@@ -73,7 +75,7 @@ where
         self.local_addr
     }
 
-    /// Creates a new `QuicServer`.
+    /// Creates a new `Server`.
     ///
     /// # Panics
     ///
@@ -121,7 +123,7 @@ where
             .assume("can set sync server addr")?
             .with_congestion_controller(Bbr::default())?
             .start()
-            .map_err(QuicError::ServerStart)?;
+            .map_err(quic::Error::ServerStart)?;
 
         let local_addr = server
             .local_addr()
@@ -381,7 +383,7 @@ fn check_request(team_id: TeamId, request: &SyncRequestMessage) -> Result<GraphI
         bug!("Should be a SyncRequest")
     };
     if team_id.as_bytes() != storage_id.as_bytes() {
-        return Err(Error::QuicSync(QuicError::InvalidPSK));
+        return Err(Error::QuicSync(quic::Error::InvalidPSK));
     }
 
     Ok(*storage_id)
--- a/crates/aranya-daemon/src/sync/transport/quic/connections.rs
+++ b/crates/aranya-daemon/src/sync/transport/quic/connections.rs
@@ -81,8 +81,8 @@ impl SharedConnectionMap {
     pub(super) async fn get_or_try_insert_with(
         &mut self,
         peer: SyncPeer,
-        make_conn: impl AsyncFnOnce() -> Result<Connection, super::QuicError>,
-    ) -> Result<Handle, super::QuicError> {
+        make_conn: impl AsyncFnOnce() -> Result<Connection, super::Error>,
+    ) -> Result<Handle, super::Error> {
         let (handle, maybe_acceptor) = match self.handles.lock().await.entry(peer) {
             Entry::Occupied(mut entry) => {
                 debug!("existing QUIC connection found");
--- a/crates/aranya-daemon/src/sync/transport/quic/psk.rs
+++ b/crates/aranya-daemon/src/sync/transport/quic/psk.rs
@@ -89,7 +89,7 @@ fn psk_to_rustls(psk: aranya_crypto::tls::Psk<CS>) -> Result<PresharedKey> {
 }
 
 /// PSK store that's shared between [`super::SyncManager`]
-/// and [`super::QuicServer`]
+/// and [`super::Server`]
 #[derive(Debug)]
 pub(crate) struct PskStore {
     inner: SyncMutex<PskStoreInner>,
--- a/crates/aranya-daemon/src/test.rs
+++ b/crates/aranya-daemon/src/test.rs
@@ -61,7 +61,7 @@ type TestState = sync::quic::QuicState;
 type TestSyncer = sync::SyncManager<TestState, crate::EN, crate::SP, crate::EF>;
 
 // Aranya sync server for testing.
-type TestServer = sync::quic::QuicServer<crate::EN, crate::SP>;
+type TestServer = sync::quic::Server<crate::EN, crate::SP>;
 
 struct TestDevice {
     /// Aranya sync client.


crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+use aranya_runtime::{Engine, Sink};
+
+#[cfg(feature = "preview")]
+use super::GraphId;
+use super::{Result, SyncManager, SyncPeer};
+
+pub mod quic;
+
+/// Types that contain additional data that are part of a [`SyncManager`] object.
+pub trait SyncState: Sized {
```
gknopf-aranya: This trait appears to be defining the operations that can be performed on a generic syncer rather than keeping track of state so `SyncApi` or `SyncOps` could be a better name:
```suggestion
pub trait SyncApi: Sized {
```

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -0,0 +1,382 @@
...
+    /// Address -> Connection map to lookup existing connections before creating a new connection.
+    conns: SharedConnectionMap,
+    /// PSK store shared between the daemon API server and QUIC syncer client and server.
+    /// This store is modified by [`crate::api::DaemonApiServer`].
+    store: Arc<PskStore>,
+}
+
+impl QuicState {
+    /// Get a reference to the PSK store
+    #[cfg(feature = "preview")]
```
gknopf-aranya, line 52: The hello sync feature is behind `preview` but I don't think the PSK store should be behind preview.

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya, line 1: We may want a different module for hello sync under `transport/quic/hello.rs` so the entire module can be put behind a preview flag in one shot rather than keeping track of which imports/methods go behind the flag here.

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    /// Establishes a QUIC connection to a peer and opens a bidirectional stream.
+    ///
+    /// This method first checks if there's an existing connection to the peer.
+    /// If not, it creates a new QUIC connection. Then it opens a bidirectional
+    /// stream for sending sync requests and receiving responses.
+    ///
+    /// # Arguments
+    /// * `peer` - The network address of the peer to connect to
+    /// * `id` - The graph ID for the team/graph to sync with
```
gknopf-aranya: These have been combined into `SyncPeer`

crates/aranya-daemon/src/sync/transport/quic/client.rs
nikki-aranya: The goal is a) to not change stuff we don't have to b) aren't we removing PSKs in the near future anyways. It compiles and passes tests and this is how it is in the og code so eh

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: I could but it's getting replaced in the next PR with the transport trait so eh

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Since the client and server both have their respective hello sync implementations, we could try this organization to avoid having a single `hello.rs` for both the client and server:
```
transport/quic/client
    client.rs
    hello.rs
transport/quic/server
    server.rs
    hello.rs
```

crates/aranya-daemon/src/sync/transport/quic/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+pub(crate) use connections::{ConnectionUpdate, SharedConnectionMap};
+pub(crate) use psk::{PskSeed, PskStore};
+pub(crate) use server::QuicServer;
+
+/// ALPN protocol identifier for Aranya QUIC sync.
+const ALPN_QUIC_SYNC: &[u8] = b"quic-sync-unstable";
+
+/// Errors specific to the QUIC syncer
+#[derive(Debug, thiserror::Error)]
+pub enum QuicError {
```
gknopf-aranya: I don't think this error is needed outside the daemon crate:
```suggestion
pub(crate) enum QuicError {
```

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Oh, was this already behind the preview flag before the refactor? If so, seems fine as is.

crates/aranya-daemon/src/api.rs
```diff
@@ -363,8 +363,8 @@ struct ApiInner {
     local_addr: SocketAddr,
     /// Public keys of current device.
     pk: std::sync::Mutex<PublicKeys<CS>>,
-    /// Aranya sync peers,
-    peers: SyncPeers,
+    /// Handle to talk with the syncer.
+    handle: SyncHandle,
```
gknopf-aranya: From the perspective of the API, we could call this field a `syncer`. The `SyncHandle` type makes it clear that it is a handle to the syncer. But calling the field `handle` isn't ideal because it is not clear what it is a handle to without looking at the field definition.
```suggestion
    syncer: SyncHandle,
```

crates/aranya-daemon/src/sync/types.rs
```diff
@@ -0,0 +1,29 @@
+/// A sync peer.
+///
+/// Contains the information needed to sync with a single peer:
+/// - network address
+/// - Aranya graph id
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SyncPeer {
```
gknopf-aranya: Unless this type is needed by the daemon's IPC API interface, it can be `pub(crate)`:
```suggestion
pub(crate) struct SyncPeer {
```

crates/aranya-daemon/src/api/quic_sync.rs
gknopf-aranya, line 1: This module appears to contain QUIC and PSK code that I would expect to be refactored into `transport/quic` based on the other changes in the PR.

gknopf-aranya: Maybe not though if this is only defining the API for the QUIC syncer in the IPC API.

crates/aranya-daemon/src/sync/handle.rs
```diff
@@ -0,0 +1,130 @@
...
+        peer: SyncPeer,
+    },
+    #[cfg(feature = "preview")]
+    BroadcastHello {
+        graph_id: GraphId,
+        head: Address,
+    },
+}
+
+/// Handles adding and removing sync peers.
```
gknopf-aranya: And sync hello subscriptions. Wonder if we should change the doc to be something more generic like "Handles sending the syncer commands to perform particular operations such as adding/removing sync peers and subscribing to hello sync updates".

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -0,0 +1,54 @@
+//! Holds all transport traits, as well as any transport-specific implementations.
```
gknopf-aranya:
```suggestion
//! This module contains all generic syncer transport traits, as well as any transport-specific syncer implementations.
```

crates/aranya-daemon/src/sync/mod.rs
```diff
@@ -1,59 +1,74 @@
 //! Aranya syncer for syncing Aranya graph commands.
-pub mod task;
+pub mod handle;
+#[cfg(feature = "preview")]
+pub mod hello;
+pub mod manager;
+pub mod transport;
```
gknopf-aranya: Based on the design, I would have expected that nothing other than the `sync` module could directly access the `transport` so I would guess the `pub` could be removed here:
```suggestion
mod transport;
```
I haven't confirmed that the transport methods aren't invoked from outside the `sync` module, that's just my intuition.

crates/aranya-daemon/src/sync/types.rs
nikki-aranya: It's pub because it's included in rustdoc unless I stub it out

crates/aranya-daemon/src/sync/mod.rs
nikki-aranya: We access the quic stuff directly in testing, I could make it pub(crate)

crates/aranya-daemon/src/sync/hello.rs
```diff
@@ -44,25 +35,16 @@ pub struct HelloSubscription {
...
-impl Syncer<State> {
+impl SyncManager<QuicState> {
     /// Broadcast hello notifications to all subscribers of a graph.
     #[instrument(skip_all)]
     pub async fn broadcast_hello_notifications(
         &mut self,
         graph_id: GraphId,
         head: Address,
-    ) -> SyncResult<()> {
+    ) -> Result<()> {
```
gknopf-aranya, line 53: This makes it appear as though `anyhow::Result` is being used even though it's not. May be more clear to import as `crate::sync::Result as SyncResult`.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya, line 1: Does this PR migrate sending of hello notifications into the delay queue? Or will that be a future PR?

gknopf-aranya (dismissed):
Mostly lgtm, left a few comments. Would like to get a few more eyes on it including @jdygert-spok, @benz-spideroak, @djohnson-spideroak.

crates/aranya-daemon/src/sync/hello.rs
nikki-aranya: How often are people using anyhow::Result, I've never assumed that as a given, using a Result<T> just means we're enforcing a specific E in the entire module

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: I have not changed any functionality, that'll probably be PR 3

crates/aranya-daemon/src/sync/hello.rs
gknopf-aranya: We shouldn't really be using it in our codebase other than testing or examples. I assumed `anyhow::Result` when looking at `Result` until I looked up the import to see that it was something else.

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -0,0 +1,382 @@
...
+    /// Sends a sync request to a peer over an established QUIC stream.
+    ///
+    /// This method uses the SyncRequester to generate the sync request data,
+    /// serializes it, and sends it over the provided QUIC send stream.
+    ///
+    /// # Arguments
+    /// * `send` - The QUIC send stream to use for sending the request
+    /// * `syncer` - The SyncRequester instance that generates the sync request
+    /// * `id` - The graph ID for the team/graph to sync
+    /// * `peer` - The network address of the peer
```
chip-so: Ditto https://github.com/aranya-project/aranya/pull/674/changes#r2635856257

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -0,0 +1,378 @@
...
+            SyncType::Push { .. } => {
+                bug!("Push messages are not implemented")
+            }
+            SyncType::Hello(_hello_msg) => {
+                #[cfg(feature = "preview")]
+                {
+                    Self::process_hello_message(_hello_msg, client, &active_team, _handle).await;
+                    // Hello messages are fire-and-forget, return empty response
+                    // Note: returning empty response which will be ignored by client
+                    return Ok(Box::new([]));
```
chip-so, line 330: I'm sure this is something that wasn't decided in this PR, but why send an empty response at all?

gknopf-aranya: I believe it's being used as an application level ACK. Even if protocol/network level ACK succeeds, there isn't a guarantee the application saw the data if there was some application level error or even an OS level error that prevented the application from reading the data.

chip-so: Oh wait Quic is connection-oriented, right? So this is just sending nothing on an existing connection.

gknopf-aranya: Yes, on an existing connection where a stream was opened for this particular hello notification and application level ACK.

crates/aranya-daemon/src/sync/hello.rs
chip-so: IMO _any_ time you're using something called `Result` that isn't `std::result::Result`, you should qualify it in a way that makes it clear it's not the standard `Result`.

gknopf-aranya: Yeah, generally using the same name as something in the standard library can be confusing. `-> sync::Result` is also acceptable.

crates/aranya-daemon/src/sync/types.rs
gknopf-aranya: Noting here that I found some docs indicating adding this to the `Cargo.toml` could allow private items to be documented without adding `pub` to them:
```
[build]
rustdocflags = ["--document-private-items"]
```

chip-so: Yes, but be aware `--document-private-items` has no granularity. It includes _everything_ private, whether it has docstrings or not.

gknopf-aranya: Do you know of a way to not make a type `pub` but still include it in the Rust docs without that drawback?

chip-so: I don't know of any way, no.

gknopf-aranya: Then it's probably safe to leave this `pub` until we have a better alternative.

chip-so (dismissed):
This looks alright to me.

crates/aranya-daemon/src/sync/transport/quic/client.rs
ssweetney-spok: Need to move the "hello sync" related methods out of the `SyncState` trait in order to do this for the client.

ssweetney-spok (dismissed):

crates/aranya-daemon/src/sync/hello.rs
jdygert-spok: It is `std::result::Result`, just with a defaulted second parameter. This is a pretty common idiom. See `std::io` which does the same, and also just refers to it as `Result`.

I think it is fairly clear to use it unqualified when you are in the same "place" as it is defined. IOW within the `sync` module, `Result` refers to the sync result. But I wouldn't do e.g. `use rkyv::Result` in our code, except _maybe_ in a module that is only handling rkyv impls.

crates/aranya-daemon/src/sync/mod.rs
```diff
@@ -1,59 +1,74 @@
...
+mod types;
 
-use error::SyncError;
+pub(crate) use aranya_runtime::GraphId;
+pub(crate) use aranya_util::Addr;
+pub(crate) use handle::{Request, SyncHandle};
+pub(crate) use manager::SyncManager;
+pub(crate) use types::SyncPeer;
+pub(super) use types::{Client, EffectSender, SyncResponse};
```
jdygert-spok: nit: group into one `use self::{...}` (I wish rustfmt could convert these for us...)

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -0,0 +1,378 @@
...
+    /// Handles sync poll requests and generates sync responses.
+    #[instrument(skip_all)]
+    async fn process_poll_message(
+        request_msg: SyncRequestMessage,
+        client: Client,
+        local_addr: Addr,
+        peer_server_addr: Addr,
+        active_team: &TeamId,
+    ) -> Result<Box<[u8]>> {
+        let mut resp = SyncResponder::new(local_addr);
```
jdygert-spok, line 349: Looks like the old code might have been passing in the peer's address instead of our own? Good fix.

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -0,0 +1,378 @@
...
+            client,
+            server,
+            server_keys,
+            conns: conns.clone(),
+            conn_rx: server_conn_rx,
+            handle: sync_peers.clone(),
+            local_addr: local_addr.into(),
+        };
+
+        Ok((server_instance, sync_peers, conns, syncer_recv, local_addr))
```
jdygert-spok: Could remove local_addr from return and add getter on server.

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -48,28 +50,31 @@ use tracing::{error, info, instrument, warn};
...
 /// Receives added/removed peers from [`SyncHandle`] via mpsc channels.
 ///
 /// [`SyncHandle`]: super::SyncHandle
-#[derive(Debug)]
-pub struct SyncManager<ST> {
+#[derive_where(Debug)]
+pub struct SyncManager<ST, EN, SP, EF>
+where
+    ST: fmt::Debug,
+{
```
jdygert-spok: Could `EF` param with `EN: Engine` bound and `EN::Effect` instead.

Use `#[derive_where(Debug; ST)]` to avoid the unconditional `ST: fmt::Debug` bound.

jdygert-spok: I see now the effects are different :thinking:

crates/aranya-daemon/src/sync/transport/quic/client.rs
nikki-aranya: Yeah, the goal is to simplify this in a future PR, probably PR 3

crates/aranya-daemon/src/sync/transport/quic/server.rs
nikki-aranya: Yeah it seemed a little weird

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: yeah VmEffect vs policy::Effect weirdness

crates/aranya-daemon/src/aranya.rs
```diff
@@ -51,7 +51,7 @@ impl<EN, SP> Deref for Client<EN, SP> {
 ///
 /// For a given peer, there should only be one cache. If separate caches are used
 /// for the server and state it will reduce the efficiency of the syncer.
-pub(crate) type PeerCacheMap = Arc<Mutex<BTreeMap<PeerCacheKey, PeerCache>>>;
+pub(crate) type PeerCacheMap = Arc<Mutex<BTreeMap<SyncPeer, PeerCache>>>;
```
gknopf-aranya, line 54: This is a cache key though. Sync peers are defined in the application API as a peer we are configured to sync with (e.g. via `AddSyncPeer()` or `SyncNow()`). Calling both a sync peer could be confusing.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -50,7 +54,7 @@ pub(crate) type SP = LinearStorageProvider<FileManager>;
 pub(crate) type EF = policy::Effect;
 
 pub(crate) type Client = aranya::Client<EN, SP>;
-pub(crate) type SyncServer = crate::sync::task::quic::Server<EN, SP>;
+pub(crate) type SyncServer = crate::sync::quic::QuicServer<EN, SP>;
```
gknopf-aranya: We intentionally named the server `quic::Server`. The import path already has the context that it is a QUIC server. If it is important to have that context in code we can import it as `quic::Server as QuicServer` wherever it's used or require the path with `quic::Server`. This isn't incorrect, just seems a bit redundant.

crates/aranya-daemon/src/sync/hello.rs
```diff
@@ -71,15 +59,15 @@ impl Syncer<State> {
...
             // Remove expired subscriptions and collect valid ones
             let mut valid_subscribers = Vec::new();
-            subscriptions.retain(|(sub_graph_id, addr), subscription| {
-                if *sub_graph_id == graph_id {
+            subscriptions.retain(|peer, subscription| {
+                if peer.graph_id == graph_id {
                     if now >= subscription.expires_at {
                         // Subscription has expired, remove it
-                        debug!(?addr, ?graph_id, "Removed expired subscription");
+                        debug!(?peer, "Removed expired subscription");
```
gknopf-aranya, line 66: Is `peer` the right abstraction for `Addr` and `GraphId`? We could sync multiple different graph IDs with the same peer.

gknopf-aranya: `subscriber` may make more sense in this context. Since a subscriber is a peer that subscribed to hello notifications for a specific graph.

crates/aranya-daemon/src/aranya.rs
nikki-aranya: A SyncPeer is just an identifier for a unique endpoint, we don't need multiple types mapped to (Addr, GraphId)

crates/aranya-daemon/src/daemon.rs
nikki-aranya: I could, I'm not sure I won't need to disambiguate when I add a SyncServer type to go along with the SyncManager but

crates/aranya-daemon/src/sync/hello.rs
nikki-aranya: peer is an (Addr, GraphId) so this is reporting what we were previously, if y'all have an issue then that's a separate problem to fix after this PR

crates/aranya-daemon/src/daemon.rs
```diff
@@ -347,7 +349,7 @@ impl Daemon {
             #[cfg(feature = "preview")]
             Arc::clone(&hello_subscriptions),
         );
-        let (server, peers, conns, syncer_recv, server_addr) = SyncServer::new(
+        let (server, peers, conns, syncer_recv) = SyncServer::new(
```
gknopf-aranya, line 352: Since type was renamed to `Handle`:
```suggestion
        let (server, sync_handle, conns, syncer_recv) = SyncServer::new(
```

crates/aranya-daemon/src/daemon.rs
```diff
@@ -10,26 +10,30 @@ use aranya_crypto::{
...
+    api::{self, ApiKey, DaemonApiServer, DaemonApiServerArgs, QSData},
     aranya::{self, ClientWithState, PeerCacheMap},
     config::{Config, Toggle},
     keystore::{AranyaStore, LocalStore},
     policy,
-    sync::task::{
-        quic::{PskStore, State as QuicSyncClientState, SyncParams},
-        SyncPeers, Syncer,
+    sync::{
+        quic::{PskStore, QuicState as QuicSyncClientState, SyncParams},
```
gknopf-aranya: Do we still need this type alias `QuicState as QuicSyncClientState` now that `State` is changed to `QuicState`?

crates/aranya-daemon/src/sync/hello.rs
gknopf-aranya: A follow-up for this seems fine since it's pre-existing. Would be interested in hearing if anyone else thinks it would be a good change.

jdygert-spok: I wouldn't change this to subscriber, but if we can think of a better name for `(Addr, GraphId)`, we could consider changing that everywhere.

One downside to using peer instead of addr/graph in tracing is that it is more verbose and harder to filter since it is less "structured". IOW instead of having two separate fields with simple values, it is now one field with a struct-like debug repr. If only tracing could ever stabilize `Valuable`, then we could easily fix that...

gknopf-aranya (dismissed):

gknopf-aranya (dismissed):

crates/aranya-daemon/src/sync/mod.rs
```diff
@@ -1,59 +1,73 @@
...
+pub(crate) use self::{
+    handle::SyncHandle,
+    manager::SyncManager,
+    transport::{quic, SyncState},
+    types::SyncPeer,
+};
+
+/// The error type which is returned from syncing with peers.
+#[derive(Debug, thiserror::Error)]
+#[allow(
```
gknopf-aranya: How are we going to remember to remove this allow when hello sync is moved out from behind the feature flag? Seems like we should reference an issue number here.

PR #678, opened 2026-01-06
chore: claude code settings

djohnson-spideroak:
Whitelist the domains for the webfetch tool, and allow some common tools that are unlikely to be destructive, including cargo build.


.claude/settings.json
```diff
@@ -0,0 +1,20 @@
...
+      "Bash(cargo check)",
+      "Bash(cargo build)",
+      "Bash(cargo make)",
+      "Bash(find)",
+      "Bash(cat)",
+      "Bash(head)",
+      "Bash(tail)"
+    ],
+    "deny": [
+      "WebFetch"
```
gknopf-aranya: Does this mean Claude wouldn't be able to download content from the Internet? I may have possibly relied on this for downloading the policy docs for context when having Claude generate policy code.

.claude/settings.json
```diff
@@ -0,0 +1,20 @@
+{
+  "permissions": {
+    "allow": [
```
gknopf-aranya, line 3: May want to add these in advance for the ctest suite `cmake`, `ctest`

.claude/settings.json
djohnson-spideroak: Yes, it disables WebFetching. WebFetch in general is a huge prompt injection risk, so any context from the internet should be manually downloaded and placed in a temp file. 

Eventually, we are going to want to build a policy docs skill.

.claude/settings.json
```diff
@@ -0,0 +1,20 @@
...
+      "Bash(cargo check)",
+      "Bash(cargo build)",
+      "Bash(cargo make)",
+      "Bash(find)",
+      "Bash(cat)",
+      "Bash(head)",
+      "Bash(tail)"
```
djohnson-spideroak:
```suggestion
      "Bash(find)",
      "Bash(cat)",
      "Bash(head)",
      "Bash(cmake)",
      "Bash(ctest)",
      "Bash(tail)"
```

gknopf-aranya (dismissed):

.claude/settings.json
gknopf-aranya: Is this something we can manually allow claude to do from within the cli or would we run commands/scripts outside of claude to download things from the Internet?

.claude/settings.json
```diff
@@ -0,0 +1,22 @@
...
+      "Bash(cargo build)",
+      "Bash(cargo make)",
+      "Bash(find)",
+      "Bash(cat)",
+      "Bash(head)",
+      "Bash(cmake)",
+      "Bash(ctest)",
+      "Bash(tail)"
+    ],
+    "deny": [
```
gknopf-aranya, line 14: Should we block `git` commands such as `git commit`, `git push`, or other write operations? May not want to block `git pull` since pulling from repositories is convenient.

djohnson-spideroak: I dont think we want to block them, we just dont want them on the allow list. This will still prompt for git commands. Blocking git push might be a good idea though.

.claude/settings.json
djohnson-spideroak: We would manually run scripts or just download it ourselves.

.claude/settings.json
gknopf-aranya: Yeah, we can easily open another terminal on the claude machine to manually run `git push` after verifying changes suggested by claude look good.

ssweetney-spok:
> ... including cargo make/test/build.

`cargo test` isn't in the "allow" list but the agent should still be able to run the "test" task using `cargo make`

Edit: I'm assuming this is the case but idk for sure

ssweetney-spok (dismissed):

.claude/settings.json
```diff
@@ -0,0 +1,22 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(cargo check)",
+      "Bash(cargo build)",
+      "Bash(cargo make)",
```
djohnson-spideroak:
```suggestion
      "Bash(cargo build)",
      "Bash(cargo make)",
      "Bash(cargo test)",
```

.claude/settings.json
```diff
@@ -0,0 +1,23 @@
...
+      "Bash(cargo test)",
+      "Bash(find)",
+      "Bash(cat)",
+      "Bash(head)",
+      "Bash(cmake)",
+      "Bash(ctest)",
+      "Bash(tail)"
+    ],
+    "deny": [
+      "WebFetch"
```
djohnson-spideroak:
```suggestion
      "WebFetch",
      "Bash(git push)"
```

djohnson-spideroak:
> An interesting topic to discuss could be whether we want claude to be able to vet deps. It seems beneficial to allow claude to assist with reviewing changes, but I think we'll want to keep a human in the loop for now.

Remember, this is just allowing claude to run these commands without a permission prompt, or denying commands directly. Claude can still run `cargo vet` using bash.

.claude/settings.json
```diff
@@ -0,0 +1,24 @@
+{
+  "permissions": {
+    "allow": [
```
gknopf-aranya, line 3: claude seems to use `grep` frequently.

djohnson-spideroak: Yeah, it does also have an allowed-by-default search tool as well, so if we see a lot of grep requests we can add it in.

ssweetney-spok (approved):

.claude/settings.json
```diff
@@ -0,0 +1,24 @@
...
+      "Bash(cargo check)",
+      "Bash(cargo build)",
+      "Bash(cargo make)",
```
jdygert-spok: IDK what the workflow looks like for these tools. Would you have to approve edit permissions to get changes in, and then it could edit the makefile to run whatever scripts it wants? Not sure if this is just meant to be best-effort anyways.

gknopf-aranya: Claude generally prompts the user before running certain commands. The user can then allow that command to be run once, or for the rest of the session, or tell claude to do something else. I believe enabling these will allow claude to run these commands without prompting the user first.

jdygert-spok: Right, I'm saying that permission to run `cargo make` is a "vulnerability" if you'd also generally give it permission to edit files, since cargo make can run whatever is in `Makefile.toml`. (assuming run permissions aren't "deep")

gknopf-aranya: More generally, giving claude write+execute permissions could be considered a vulnerability if we can't trust it.
A development workflow often requires both to happen in a workspace.

djohnson-spideroak: With this change, it could write rust code and then run it directly without inbetween approval. I think this is generally safer if we have webfetch disabled, as the primary vectors for attack would be dependencies or dependency docs at that point.

I will leave these approve-only if others think we should keep it that way. Enabling the agents to run the tests and binaries directly is very useful. I generally think the risk is pretty low, as long as we are not running this on a machine with sensitive data. 

@mooreoak what do you think?

djohnson-spideroak: I removed the test and run commands. We should have further discussion on code execution, but I would like re-review on the current state.

mooreoak: Given that we are working on a open source code base, and that our practice is to run in sandboxes, I am not too worried about attacks. They are still possible but the simple ones are not going to be effective. For an attack to work the attacker would have to take over the container and then escape from it to other systems or compartments.

.claude/settings.json
```diff
@@ -0,0 +1,28 @@
...
+    "allow": [
+      "WebFetch(domain:aranya-project.github.io)",
+      "WebFetch(domain:github.com/aranya-project/aranya-docs/)",
+      "WebFetch(domain:raw.githubusercontent.com/aranya-project/aranya-docs/)",
+      "WebFetch(domain:github.com/aranya-project/aranya-core/)",
+      "WebFetch(domain:raw.githubusercontent.com/aranya-project/aranya-core/)",
+      "WebFetch(domain:github.com/aranya-project/aranya/)",
+      "WebFetch(domain:raw.githubusercontent.com/aranya-project/aranya/)",
+      "Bash(cargo check *)",
+      "Bash(cargo build *)",
```
gknopf-aranya, line 8: Was `cargo test` removed on purpose?
`Bash(cargo test *)`

djohnson-spideroak: Yes, it was removed related to the code execution concerns above. I am aware we also have build.rs executing on build, so if we need to be really safe we need to remove that too.

.claude/settings.json
djohnson-spideroak: I think the larger risk that Dygert is concerned about is gaining code execution on our dev machines via prompt injection to write a rust backdoor.

gknopf-aranya (dismissed):

.claude/settings.json
mooreoak: All the code should go though code review before it gets downloaded and built right? Also the dev who uses claud should review the commit changes too even before that.

mooreoak: I suppose there is some risk if the rust language server could trigger the attack and someone tried to review the code locally in a IDE.

mooreoak: One question is if this is a bigger risk then building third party deps from crates.io. We use audits but I expect those happen last and it would be too late.

gknopf-aranya: What if we gate our builds on `cargo vet`? Then audits would need to be completed first before introducing a new dep to the build.

gknopf-aranya: Currently `cargo vet` issues are caught as a last step by CI before a PR can be merged.

djohnson-spideroak: > All the code should go though code review before it gets downloaded and built right? Also the dev who uses claud should review the commit changes too even before that.

This change is in relation to adding "cargo make" and similar commands to the allowed commands for claude. This would allow claude to run them without asking for permission. The primary risk is that a prompt injection causes claude to generate malicious code, and then immediately run it. 

If we want to mitigate this completely, we would need to not allow any cargo build/test commands without approval.

gknopf-aranya: I think we should prevent claude from running `cargo build` and `cargo test` directly. Instead, it can run the same commands via `cargo make build ...` `cargo make test ...` where execution is gated on cargo vet checks passing to ensure no unaudited deps are run.

gknopf-aranya: The downside of this is that claude often tries to build/test specific creates/modules after making changes. So we need a way to get it to use the new commands.

djohnson-spideroak: We would have to deny the direct build commands. The problem with gating on cargo vet is that it still doesn't prevent a malicious change being run if no dependencies were added. 

We have to accept that claude will be able to write and run code, or we have to remove its ability to iterate without approval. Or we need to set up infrastructure to allow us to run these with tighter sandboxes.

gknopf-aranya: Manually approving each iteration seems like a huge productivity hit. If possible, I'd prefer to make the host running claude more isolated/ephemeral to protect against malicious code.

djohnson-spideroak: I think the current settings are fine. It does give access to cargo build, but we could see build.rs being modified. I dont think this adds much risk. We can look at setting up tighter sandboxes.

PR #681, opened 2026-01-08
certgen

gknopf-aranya:
A CLI tool and lib for generating PEM formatted root CA certificates and signed certificates using P-256 ECDSA keys.

### DIFF
--- a/crates/aranya-certgen/README.md
+++ b/crates/aranya-certgen/README.md
@@ -24,7 +24,7 @@ aranya-certgen ca --cn "My Company CA"
 ### Create a Signed Certificate
 
 ```bash
-aranya-certgen signed ca --cn "webserver"
+aranya-certgen signed ca --cn webserver
 ```
 
 ## Commands
@@ -63,7 +63,7 @@ $ aranya-certgen ca --cn "My Company CA"
 Generating root CA certificate...
   Certificate: ./ca.crt.pem
 
-$ aranya-certgen signed ca --cn "webserver"
+$ aranya-certgen signed ca --cn webserver
 Generating certificate 'webserver'...
   Certificate: ./cert.crt.pem
 ```
@@ -74,7 +74,7 @@ To renew or reissue a device certificate, generate a new certificate using the s
 
 ```bash
 # Load existing CA and generate a new certificate
-aranya-certgen signed ca --cn "webserver" --days 365 -f
+aranya-certgen signed ca --cn webserver --days 365 -f
 ```
 
 The `-f` (force) flag overwrites the existing certificate files. The new certificate will:
@@ -93,8 +93,8 @@ CA certificates can also be renewed, but this requires all device certificates t
 aranya-certgen ca --cn "My Company CA" --days 365 -f
 
 # Reissue all device certificates
-aranya-certgen signed ca --cn "webserver" --days 365 -f
-aranya-certgen signed ca --cn "device1" -o device1 --days 365 -f
+aranya-certgen signed ca --cn webserver --days 365 -f
+aranya-certgen signed ca --cn device1 -o device1 --days 365 -f
 # ... repeat for all devices
 ```
 
@@ -106,7 +106,7 @@ certgen:
 ```
 aranya-certgen ca --cn "My Company CA" --days 365
 
-aranya-certgen signed ca --cn "webserver" --days 365
+aranya-certgen signed ca --cn webserver --days 365
 ```
 
 openssl:
--- a/crates/aranya-certgen/src/main.rs
+++ b/crates/aranya-certgen/src/main.rs
@@ -9,7 +9,7 @@
 //! aranya-certgen ca --cn "My CA"
 //!
 //! # Create a signed certificate with P-256 ECDSA key (creates ./cert.crt.pem and ./cert.key.pem)
-//! aranya-certgen signed ca --cn "server"
+//! aranya-certgen signed ca --cn server
 //! ```
 
 use std::path::PathBuf;


crates/aranya-certgen/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "aranya-certgen"
+description = "CLI tool for generating root CA certificates and signed certificates using P-256 ECDSA"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+publish = false
```
gknopf-aranya: We may want to publish this tool. I'm uploading it as an artifact to the release because it seems useful to have a compiled tool for integration and testing.

gknopf-aranya:
We should make this a library so it can be used from the Rust example and by unit tests for mTLS cert testing

crates/aranya-certgen/Cargo.toml
gknopf-aranya: Now that we're making `certgen` both a CLI and a library, it should be published.

crates/aranya-certgen/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[[bin]]
+name = "aranya-certgen"
+path = "src/main.rs"
+
+[dependencies]
+clap = { workspace = true }
+thiserror = { workspace = true }
+
+rcgen = { workspace = true }
```
gknopf-aranya:
```suggestion
rcgen = { workspace = true }

```

.github/workflows/build.yml
```diff
@@ -85,3 +85,32 @@ jobs:
...
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Environment
+        uses: ./.github/actions/setup
+
+      - name: Build aranya-certgen
+        run: |
+          cargo build --release --package=aranya-certgen
```
gknopf-aranya: Need a task in `Makefile.toml` for this

crates/aranya-certgen/README.md
```diff
@@ -0,0 +1,92 @@
...
+## Comparison with OpenSSL
+
+This tool simplifies certificate generation compared to OpenSSL:
+
+| Aspect | aranya-certgen | OpenSSL |
+|--------|---------|---------|
+| Commands for CA + 1 cert | 2 | 6+ |
+| Intermediate files | None | CSR, serial |
+| SAN syntax | `--dns x --ip y` | `-extfile <(echo "...")` |
+| Key algorithm | P-256 ECDSA (built-in) | Must specify manually |
```
jdygert-spok: I do think there's some value in the simplicity/consistency of this tool, but I don't think this comparison is fair. Here's the openssl commands corresponding to the example commands above.

```sh
openssl req -x509 \
    -newkey ec -pkeyopt ec_paramgen_curve:secp384r1 -keyout ca.key -nodes \
    -days 365 \
    -subj "/CN=My Company CA" \
    -out ca.pem

openssl req -x509 \
    -newkey ec -pkeyopt ec_paramgen_curve:secp384r1 -keyout server.key -nodes \
    -CA ca.pem -CAkey ca.key \
    -days 365 \
    -subj "/CN=webserver" \
    -addext "subjectAltName=DNS.1:example.com,DNS.2:www.example.com,IP:192.168.1.10" \
    -out server.pem
```

It's two commands and doesn't generate any intermediate files either.

gknopf-aranya: Deciding whether to make this `certgen` tool boils down to whether we'd prefer to teach users how to run the right `openssl` commands, or provide them with a tool that generates them the right way every time. Even though the above `openssl` commands are only slightly more complex than `certgen` in terms of correct usage, there are more ways to provide incorrect input such as different curves.

chip-so: I would say this could probably be a shell script that takes a few args, except the extra syntax for SANs is kind of gnarly. I guess the question is, do we expect to have to use this with a lot of different parameters, or is this part of a setup that can be kept in a few scripts that rarely change?

gknopf-aranya: I suspect that every integration with Aranya will need to generate a root cert and a signed cert for each device. We don't know what the SANs will be ahead of time.

gknopf-aranya: Another thing to consider is that I'm planning to use `certgen` as a lib in the Rust example and unit tests. Having a consistent way to generate certs is preferred over different methods including `openssl`, a bash script for cert gen, and code using the `rcgen` lib for unit tests.

chip-so: Yeah, a simplified library for testing is a good case (though I wonder how it would compare to using `rcgen` directly).

gknopf-aranya: Leveraging claude to help analyze the tradeoffs:
```
❯ Is the certgen tool an improvement over running openssl commands directly? 

● It depends on the use case. Here's an objective comparison:

  Advantages of aranya-certgen:
  - Simpler syntax - Commands are shorter and more readable
  - Guarantees correct key type - Always uses P-256 ECDSA; OpenSSL requires verbose -newkey ec -pkeyopt ec_paramgen_curve:prime256v1 which is error-prone (we just fixed a typo where the README had the wrong curve)
  - Cleaner SAN handling - --dns example.com vs OpenSSL's -addext "subjectAltName=DNS.1:example.com,..."
  - Purpose-built - Designed for Aranya's specific certificate requirements

  Advantages of OpenSSL:
  - Already installed - Most systems have OpenSSL; aranya-certgen requires building
  - More flexible - Supports any key type, curve, or certificate extension
  - Better documented - Extensive community knowledge and examples
  - No Rust toolchain needed - No build step required

  Verdict: For Aranya users who need P-256 ECDSA certificates with SANs, aranya-certgen reduces errors and simplifies the workflow. For general certificate generation or one-off tasks where OpenSSL is already available, the benefit is marginal. The tool is most valuable when consistency matters (e.g., ensuring all team members generate compatible certificates).

❯ Is the certgen tool written in Rust better than a comparable tool written in a bash script? What are the tradeoffs? Consider   
  that we need this tool to work in Rust examples, C examples, Rust unit tests, integrating Aranya into customer deployments     
  with external PKI, etc.                                                                                                        

● Bash script pros:
  - Simpler to write and modify
  - No build step for the script itself
  - Transparent - can see exactly what OpenSSL commands run
  - Easy for non-Rust developers to customize

  Bash script cons:
  - Requires OpenSSL as runtime dependency
  - Windows compatibility is poor (needs WSL/Git Bash)
  - Can't be used as a library - always requires subprocess + temp files
  - Shell differences (bash/sh/zsh) cause subtle bugs
  - Error handling is crude
  - Harder to unit test

  Rust tool pros:
  - Library + binary - Can be called directly from Rust code without subprocess
  - Cross-platform once compiled
  - No runtime dependencies (crypto is statically linked)
  - Proper error handling and type safety
  - Easy to unit test

  Rust tool cons:
  - Requires Rust toolchain to build
  - More complex to maintain
  - Harder for non-Rust developers to modify

  For your specific use cases:
  ┌─────────────────┬────────────────────────────────────┬──────────────────────────┐
  │    Use Case     │            Bash Script             │       Rust Library       │
  ├─────────────────┼────────────────────────────────────┼──────────────────────────┤
  │ Rust examples   │ Spawn process, temp files          │ Direct API call          │
  ├─────────────────┼────────────────────────────────────┼──────────────────────────┤
  │ Rust unit tests │ Spawn process, temp files, cleanup │ Direct API call, no I/O  │
  ├─────────────────┼────────────────────────────────────┼──────────────────────────┤
  │ C examples      │ Spawn process                      │ Spawn process (equal)    │
  ├─────────────────┼────────────────────────────────────┼──────────────────────────┤
  │ Customer PKI    │ They use their own tools           │ They use their own tools │
  └─────────────────┴────────────────────────────────────┴──────────────────────────┘
  Verdict: The Rust library approach is better for Rust-heavy workflows because it eliminates subprocess spawning and temp file management in tests and examples. For customer deployments with external PKI, neither tool matters - they'll use their existing infrastructure. The main cost is build complexity, but since this project already requires a Rust toolchain, that's not additional burden.

❯ btw, we'll upload a pre-built executable of the certgen tool to CI. So building the tool isn't a con. 

● That changes the calculus significantly. With pre-built binaries:

  Rust tool cons (revised):
  - ~~Requires Rust toolchain to build~~ — Users download pre-built binary
  - More complex to maintain (but already done)

  Bash script cons (unchanged):
  - Requires OpenSSL as runtime dependency
  - Windows compatibility issues
  - Can't be used as a library
  - Subprocess + temp files in tests

  With pre-built binaries available, the Rust tool has essentially no disadvantages over a bash script for CLI usage, while retaining the library benefits for Rust code. The only remaining argument for a bash script would be if you wanted something users could trivially inspect and modify — but that's not a typical requirement for a certificate generation tool.
```

mooreoak: What we need is an easy way for people to get started. I don't really care if it is a bash script or a rust too/lib. This dose not need to replace what people already have or could build but a way to make it easy to get up and running for testing and development.

gknopf-aranya: > Yeah, a simplified library for testing is a good case (though I wonder how it would compare to using `rcgen` directly).

The `certgen` lib is nearly identical to the `rcgen` code we will implement in mTLS unit tests. The `certgen` CLI is a thin wrapper around that lib. If we only needed a bash script, I think it's debatable whether a new tool is needed vs. simply using `openssl`. However, since we are going to implement Rust code to generate certs with `rcgen` anyways in unit tests, we may as well use that code elsewhere as it improves: maintainability, code-reuse, usability, and is less likely to be misused.

gknopf-aranya:
Decision made today that we are going to use this tool to generate certs for Aranya since it has safe defaults for the cipher suite that reduce the chance of someone misconfiguring certs in production.

supply-chain/audits.toml
```diff
@@ -16,6 +16,24 @@ who = "gknopf <gknopf@spideroak.com>"
 criteria = "safe-to-deploy"
 delta = "1.0.94 -> 1.0.97"
 
+[[audits.asn1-rs]]
+who = "gknopf <gknopf@spideroak.com>"
+criteria = "safe-to-deploy"
+version = "0.7.1"
+notes = "Pure Rust ASN.1/BER/DER library. Uses #![forbid(unsafe_code)]. No file system or network access. Defensive programming with recursion limits."
```
jdygert-spok: Did you add the notes on these yourself?

gknopf-aranya: No. I used claude to help analyze the deps, summarize the changes, and write these notes.

jdygert-spok: Seems dubious to add an audit based on AI.

chip-so: I agree. If Claude can do the audits, what am I doing here looking at pull requests? 😆

gknopf-aranya: While we can't rely on AI to complete the audits, it seems like a valuable tool to help with analyzing the deps.

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Generates a self-signed CA certificate and P-256 ECDSA private key that
+    /// can be used to sign other certificates.
+    Ca {
+        /// Path for the CA certificate file (PEM format).
+        #[arg(long)]
+        cert: PathBuf,
+
+        /// Path for the CA P-256 ECDSA private key file (PEM format).
+        #[arg(long)]
+        key: PathBuf,
```
jdygert-spok: If we are aiming for convenience, maybe we could do something like this:

```suggestion
        /// Output directory
        #[arg(short, long, default = ".")]
        out: PathBuf,
        
        /// Output base file name
        #[arg(short, long, default = "ca")]
        name: String,
```

So then the default would produce `./ca.key.pem` and `./ca.crt.pem`.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+            }
+        }
+        fs::write(cert_path, &self.cert_pem).map_err(|e| CertGenError::io(cert_path, e))?;
+
+        if let Some(parent) = key_path.parent() {
+            if !parent.as_os_str().is_empty() {
+                fs::create_dir_all(parent).map_err(|e| CertGenError::io(key_path, e))?;
+            }
+        }
+        fs::write(key_path, self.key.serialize_pem()).map_err(|e| CertGenError::io(key_path, e))?;
```
jdygert-spok: We might not want to create parent directories or overwrite existing files by default.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+
+    /// Adds an IP address.
+    pub fn with_ip(mut self, ip: IpAddr) -> Self {
+        self.ip_addresses.push(ip);
+        self
+    }
+}
+
+/// A PEM-encoded string.
+type PemString = String;
```
chip-so: This seems fairly useless.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+    /// # Example
+    ///
+    /// ```no_run
+    /// use aranya_certgen::CertGen;
+    ///
+    /// let ca = CertGen::ca("My Root CA", 365)?;
+    /// ca.save("ca.pem", "ca.key")?;
+    /// # Ok::<(), aranya_certgen::CertGenError>(())
+    /// ```
+    pub fn ca(cn: &str, days: u32) -> Result<Self, CertGenError> {
```
chip-so: If the function docs say "Creates a new X", where X is the type of the `impl`, then IMO the function should be called `new`.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+        if days == 0 {
+            return Err(CertGenError::InvalidDays);
+        }
+        let (cert, key) = generate_root_ca(cn, days)?;
+        let cert_pem = cert.pem();
+
+        let issuer_key = KeyPair::from_pem(&key.serialize_pem())?;
```
chip-so: ~~Serialization round-trip for no apparent reason? `key` is already a `KeyPair`.~~

Christ on a bike. This is a workaround for `KeyPair` not being cloneable. :[ But I'm kind of guessing that's by design as it contains a private key.

This is done because `CertGen` owns the `Issuer`, which as used here owns a copy of the key (it can take a reference, but that's not possible here because `CertGen` also owns the key and it hasn't been moved yet). Which also means two copies of the private key in memory. Fun.

I think the way this _should_ work is that the `Issuer` shouldn't be stored, and should be created as needed with a reference to the owned cert and key. But I'm not sure since there aren't a lot of examples in `rcgen`.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+        let issuer = Issuer::from_ca_cert_pem(&cert_pem, issuer_key)?;
+
+        Ok(Self {
+            cert_pem,
+            key,
+            issuer,
+        })
```
chip-so: This pattern of creating a leaf certificate as a `CertGen` is frankly kind of odd. It effectively unifies the type between CAs and leaf certs and that feels like bad design. I think it would make sense if you could create further descendant leaf certificates, but I'm pretty sure the lack of "is CA" on these certs prevents that.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+            if !parent.as_os_str().is_empty() {
+                fs::create_dir_all(parent).map_err(|e| CertGenError::io(key_path, e))?;
+            }
+        }
+        fs::write(key_path, self.key.serialize_pem()).map_err(|e| CertGenError::io(key_path, e))?;
+
+        // Set restrictive permissions on private key (Unix only)
+        #[cfg(unix)]
+        fs::set_permissions(key_path, fs::Permissions::from_mode(0o600))
+            .map_err(|e| CertGenError::io(key_path, e))?;
```
chip-so: To prevent a race where a file could be read by others, the permissions should be set on open using e.g. [`OpenOptionsExt::mode()`](https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#tymethod.mode).

I don't think this is a big deal for this tool in particular, but it still looks sloppy.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+    pub fn key_pem(&self) -> String {
+        self.key.serialize_pem()
+    }
+}
+
+impl std::fmt::Debug for CertGen {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("CertGen")
+            .field("cert_pem", &"<PEM data>")
+            .field("key", &"<private key>")
```
chip-so: The private key should probably not be part of the debugging output.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+            .finish_non_exhaustive()
+    }
+}
+
+impl PartialEq for CertGen {
```
chip-so: This seems to only be used by `assert_eq!()` in tests, so maybe worth putting under `#[cfg(test)]`.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        self.cert_pem == other.cert_pem && self.key.serialize_pem() == other.key.serialize_pem()
+    }
+}
+
+impl Eq for CertGen {}
```
chip-so: This, on the other hand, doesn't seem to be used at all and I assume is only included for completeness.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+
+    // Build Subject Alternative Names
+    let mut sans: Vec<SanType> = Vec::new();
+    for dns in &san.dns_names {
+        sans.push(SanType::DnsName(dns.clone().try_into()?));
+    }
+    for ip in &san.ip_addresses {
+        sans.push(SanType::IpAddress(*ip));
+    }
+    // If no SANs provided, use the CN as a default DNS name
```
chip-so: So is the requirement for SANs specific to our daemon? I'm mostly familiar with certificates as used in the common internet certificate infrastructure, and typically they'll be fine just using the CommonName.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,565 @@
...
+    ///     .with_ip("192.168.1.10".parse().unwrap());
+    /// let signed = ca.generate("server", 365, &sans)?;
+    /// signed.save("server.pem", "server.key")?;
+    /// # Ok::<(), aranya_certgen::CertGenError>(())
+    /// ```
+    pub fn generate(
+        &self,
+        cn: &str,
+        days: u32,
+        sans: &SubjectAltNames,
```
chip-so: I think it would make more sense if this just moved the value instead of taking the reference. A `SubjectAltNames` is probably not useful after it's been used once.

crates/aranya-certgen/src/lib.rs
chip-so, line 1: I think this should probably be split up into at least a library and a test module. I would probably separate out the error types as well.

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        /// Common Name (CN) for the root CA.
+        #[arg(long, default_value = "My Root CA")]
+        ca_name: String,
```
chip-so: Both this and the `--cn` argument below set the same thing, so they should have the same name. I'd recommend having `--cn` for both.

In fact, this and `--days` could probably be factored out into a common structure used by both commands.

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,169 @@
...
+            key,
+            ca_name,
+            days,
+        } => {
+            println!("Generating root CA certificate...");
+            let cert_gen = CertGen::ca(&ca_name, days)?;
+            cert_gen.save(&cert, &key)?;
+
+            println!("  Root CA certificate: {}", cert.display());
+            println!("  Root CA private key: {}", key.display());
```
chip-so: Probably also should not print this out.

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,169 @@
...
+            let ca = CertGen::load(&ca_cert, &ca_key)?;
+
+            println!("Generating certificate '{}'...", cn);
+            let signed = ca.generate(&cn, days, &sans.clone().into())?;
+            signed.save(&cert, &key)?;
+
+            println!("  Certificate: {}", cert.display());
+            println!("  Private key: {}", key.display());
+
+            if !sans.dns_names.is_empty() || !sans.ip_addresses.is_empty() {
```
chip-so: If no SANs were provided on the command-line, it warns that one will be added. But this summary output pulls directly from the command line arguments, so it will not print the added SANs. I am not sure that was intended behavior.

crates/aranya-certgen/README.md
```diff
@@ -0,0 +1,113 @@
+# aranya-certgen
+
+A CLI tool for generating root CA certificates and signed certificates.
+
+All generated keys use **P-256 ECDSA** (NIST P-256 / secp256r1 curve with ECDSA signatures).
+
+## Building
+
+```bash
+cargo make build-certgen
```
chip-so: This requires `cargo-make`, but the README doesn't say that you need it.

I don't think we _should_ require `cargo-make` here. `cargo build --release` is sufficient.

crates/aranya-certgen/src/lib.rs
jdygert-spok: Looks like we could also just access the key from the issuer with `.key()` instead of storing it separately.

crates/aranya-certgen/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[dependencies]
+clap = { workspace = true }
+rcgen = { workspace = true }
+thiserror = { workspace = true }
+
+time = "0.3"
+
+[dev-dependencies]
+tempfile = { workspace = true }
+x509-parser = "0.18"
```
chip-so, line 37: `rcgen` is only used in this crate and is added as a workspace dependency, but `x509-parser` is not.

Cargo.toml
```diff
@@ -89,6 +89,7 @@ tarpc = { version = "0.35.0", features = ["unix", "serde-transport", "serde-tran
 tempfile = { version = "3.17.1" }
 test-log = { version = "0.2.14", default-features = false, features = ["trace"] }
 thiserror = { version = "2.0" }
+rcgen = { version = "0.14", default-features = false, features = ["crypto", "pem", "x509-parser", "ring"] }
```
chip-so: This could just be `rcgen = { version = "0.14", features = ["x509-parser"] }`, as `crypto`, `pem`, and `ring` are all default features.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: For testing purposes, I don't think we need to specify cert SANs. If someone wants to use our `certgen` tool to generate certs for nodes on a production network with different hostnames and IP addresses, I'd expect them to want to configure SANs. For simplicity, I'm leaning towards removing SANs support from this tool until we have a reason to add it in.

crates/aranya-certgen/src/main.rs
gknopf-aranya: That is easier to use. Refactoring.

crates/aranya-certgen/src/main.rs
gknopf-aranya: Refactored shared args into a shared type for both commands. The exception is `--name` where I want to default to `ca` and `cert` depending on which command is running.

crates/aranya-certgen/src/main.rs
gknopf-aranya: Good catch. Removed this from output.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: The debug output redacted the private key to `<private key>` already. But I could skip displaying the field altogether.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: This was used by `PartialEq` since we can't derive `Eq` on `Certgen` directly. Looking into this more, I realized the equality checks were only needed for unit tests and were comparing both the cert and private key. The tests have been updated to only compare certs so `Eq` and `PartialEq` are not needed.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: This has been removed.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: I found the `String` type to not clearly document that a PEM cert is contained in it. As long as field names are good, this should be okay.

supply-chain/audits.toml
gknopf-aranya: Reverted the audits until I have a chance to review diffs manually.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Not having to add a `mkdir -p ./dir` before running the CLI tool is convenient. What if we add a `--p` arg to the tool to create them automatically?

jdygert-spok: I think a `-p/--parent` and `-f/--force` are probably fine, just shouldn't do it by default.

gknopf-aranya: Yep, about to push a commit for that.

gknopf-aranya: Done

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Good catch. Pushed a fix for this.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: This ties in to https://github.com/aranya-project/aranya/pull/681#discussion_r2692082964.
Because `CertGen` can create either CA certs, or certs signed by a CA cert, it doesn't make sense to have a single `new()` method (which type of cert would it refer to?). If the certs are broken out into their own types, then they could each have a new method on their respective types.

crates/aranya-certgen/src/lib.rs
chip-so: Ah, I misunderstood how this works, then. I suppose that means this doesn't print _any_ useful data about the `CertGen` instance, which makes it somewhat useless as a `Debug` impl.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Removed this extra round-trip in [d4af0f3](https://github.com/aranya-project/aranya/pull/681/commits/d4af0f390bce259eab54d1266ab617d15a5eea01)

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Removed SANs support until there is a reason to add it in.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Comment no longer applies after removing SANs support from the tool.

crates/aranya-certgen/src/main.rs
gknopf-aranya: Comment no longer applies after removing SANs support from the tool.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: I've split out `CaCert` and `SignedCert` types to enforce that only CA certs can be used to sign leaf certs. This allows the `CaCert` type to have a `new()` method.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: I've split out `CaCert` and `SignedCert` types to enforce that only CA certs can be used to sign leaf certs. This allows the `CaCert` type to have a `new()` method.

crates/aranya-certgen/Cargo.toml
gknopf-aranya: Updated

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Refactored code into different modules

crates/aranya-certgen/src/lib.rs
gknopf-aranya: @chip-so, @jdygert-spok removing the SANs may have been a mistake. Here's what claude is claiming after asking it to check the certgen tool for compatibility with quinn:
```
  The certificates will NOT work with rustls/quinn for server authentication.

  rustls follows RFC 6125 and requires Subject Alternative Name (SAN) for hostname verification. The Common Name (CN) alone is insufficient. Without SAN, clients will fail with hostname verification errors.
```

jdygert-spok: RFC 6125 says you MAY check the CN if no SANs are present in the cert. From what I can see elsewhere it seems to be generally discouraged/deprecated to rely on the CN.

gknopf-aranya:
```
❯ Doesn't rustls check the CN if SANs is not present? 

● No, rustls does not fall back to CN. This is intentional.

  rustls uses rustls-webpki (formerly webpki) for certificate verification, which only checks SAN and ignores CN entirely. This follows:

  1. CA/Browser Forum Baseline Requirements - Public CAs haven't issued CN-only certs since 2015
  2. RFC 6125 recommends SAN over CN
  3. Modern browsers also ignore CN

  This is different from OpenSSL, which does fall back to CN when SAN is absent.

  So for rustls/quinn compatibility, you have two options:

  1. Re-add SAN support with --dns and --ip flags
  2. Auto-add CN as DNS SAN (simpler - what the code did before):
```

gknopf-aranya: I think the simplest approach for now is to add the CN as a DNS hostname in the SANs. Since we don't currently have a requirement to make the SANs configurable.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+//!
+//! // Create a new CA
+//! let ca = CaCert::new("My Root CA", 365).unwrap();
+//!
+//! // Generate a signed certificate
+//! let signed = ca.generate("my-server", 365).unwrap();
+//!
+//! // Save CA and signed certificates to files
+//! // Creates ./ca.crt.pem and ./ca.key.pem
+//! ca.save(".", "ca", None).unwrap();
```
gknopf-aranya: The lib API could be simplified with path prefixes. E.g. `ca.save("./ca", None)`.

crates/aranya-certgen/src/lib.rs
chip-so: I'm looking at my LetsEncrypt certificates and that seems to be what they're doing.

gknopf-aranya:
Added notes on future work to allow us to land this tool as-is to unblock mTLS integration.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Ugh, I need to unresolve this conversation. Once I integrated the `certgen` tool without SANs support into the mTLS branch, certificate verification is failing for localhost IP addresses. There are tests that use 127.0.0.1 as an IP address. An IP address is not allowed to be configured as a DNS SANs.

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        let cert_pem =
+            fs::read_to_string(&cert_path).map_err(|e| CertGenError::io(&cert_path, e))?;
+
+        // Wrap in Zeroizing to ensure the key PEM is zeroized on drop, even if parsing fails
+        let key_pem = Zeroizing::new(
+            fs::read_to_string(&key_path).map_err(|e| CertGenError::io(&key_path, e))?,
+        );
+
+        let key = KeyPair::from_pem(&key_pem).map_err(|e| CertGenError::parse_key(&key_path, e))?;
```
chip-so: I'm unclear on the memory semantics here, but I think this should be zeroed on drop too, since it gets moved into the newly created `CaCert`. I don't think there's any guarantee that the memory doesn't get copied to do this (though I imagine it will effectively get optimized to construct it in place).

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,388 @@
...
+    /// Returns the private key as a PEM-encoded string.
+    pub fn key_pem(&self) -> String {
+        self.key.serialize_pem()
+    }
+}
+
+/// Zeroizes the private key when dropped to prevent key material from lingering in memory.
+impl Drop for CaCert {
+    fn drop(&mut self) {
+        self.key.zeroize();
```
chip-so: This is kind of six of one, half a dozen of the other, but it might be cleaner to just wrap `key` in `Zeroize`.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: It looks like I could workaround this by making the hostname "localhost". However, that means that we'd be unable to use the certgen tool on networks that use IP addresses without DNS support.

gknopf-aranya: I think the most general approach that will work for various use-cases is to support both DNS and IP SANs with this tool.

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+            let ca_path = ca.to_str().expect("invalid CA path");
+            let ca = CaCert::load(ca_path)?;
+
+            println!("Generating certificate '{}'...", args.cn);
+            let signed = ca.generate(&args.cn, args.days)?;
+
+            let output_str = output.to_str().expect("invalid path");
+            signed.save(output_str, args.save_options())?;
+
+            let cert_path = output.with_extension("crt.pem");
```
chip-so: I don't love that this effectively does this separately in two locations. I think it would be better if all the path calculations were done ahead of generation and passed in.

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+fn main() -> Result<(), CertGenError> {
+    let cli = CliArgs::parse();
+
+    match cli.command {
+        Commands::Ca { args, output } => {
+            println!("Generating root CA certificate...");
+            let ca = CaCert::new(&args.cn, args.days)?;
+
+            let output_str = output.to_str().expect("invalid path");
+            ca.save(output_str, args.save_options())?;
```
chip-so: This are some strange type manipulations on the path. Clap parses it into a `PathBuf`, it's passed into `save()` here as a `&str`, which is passed into an `AsRef<Path>` in `save_cert_and_key()`. I think `save()` and `load()` should probably just take `&Path` or `AsRef<Path>`.

(Since any IO function I can think of will take a `&Path`, any time you see `Path::to_str()` being used there's probably something weird going on.)

crates/aranya-certgen/src/lib.rs
gknopf-aranya: After discussing this with @chip-so, we're going to auto-detect whether the CN is hostname or IPv4/IPv6 address in order to set the appropriate SANs type.

crates/aranya-certgen/Cargo.toml
chip-so: I see that `rcgen` is no longer using `default-features = false`, but this comment was not addressed.

crates/aranya-certgen/src/main.rs
gknopf-aranya: I'm using path prefixes to simplify API ergonomics. E.g. `./ca` can be used to save certs to `./ca.crt.pem` `./ca.key.pem` instead of specifying the path and prefix separately with `./` and `ca`. I haven't validated if path prefixes are compatible with `Path` yet.

crates/aranya-certgen/Cargo.toml
gknopf-aranya: If `rcgen` and `x509-parser` are only used by `certgen`, they don't need to be defined as workspace deps. I'll move any that are also used by the daemon into a workspace dep on the `mtls` branch.

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,391 @@
...
+
+    /// Returns the certificate as a PEM-encoded string.
+    pub fn cert_pem(&self) -> &str {
+        &self.cert_pem
+    }
+
+    /// Returns the private key as a PEM-encoded string.
+    pub fn key_pem(&self) -> String {
+        self.key.serialize_pem()
+    }
```
benz-spideroak: Is there a reason cert_pem returns &str and key_pem returns String?

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,391 @@
...
+        save_cert_and_key(path, &self.cert_pem, &self.key.serialize_pem(), options)
+    }
+
+    /// Returns the certificate as a PEM-encoded string.
+    pub fn cert_pem(&self) -> &str {
+        &self.cert_pem
+    }
+
+    /// Returns the private key as a PEM-encoded string.
+    pub fn key_pem(&self) -> String {
```
benz-spideroak: Consider returning a `Zeroizing<String>` to ensure it gets zeroized.

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,391 @@
...
+    ///   - `"ca"` → loads `./ca.crt.pem` and `./ca.key.pem`
+    ///   - `"./certs/ca"` → loads `./certs/ca.crt.pem` and `./certs/ca.key.pem`
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - The files cannot be read ([`CertGenError::Io`])
+    /// - The certificate cannot be parsed ([`CertGenError::ParseCert`])
+    /// - The private key cannot be parsed ([`CertGenError::ParseKey`])
+    pub fn load(path: &str) -> Result<Self, CertGenError> {
```
benz-spideroak: What happens if you try to load a non CA cert? It appears that it will be loaded without error and then fail when you try to use it.

crates/aranya-certgen/src/main.rs
chip-so: I think it should work fine. Using `Path` for stems that get extensions added is a pretty common pattern.

Alternatively, if you don't want it to be a path to start with, then the `output` fields in the args should be `String`, not `PathBuf`.

crates/aranya-certgen/src/cert.rs
chip-so: I would imagine it's because the cert PEM is a field that can be referenced, and the key needs to be serialized and thus generates new data.

(Though the more I look at this, the weirder it is that the serialized form of the cert is stored in the object, even though it makes no functional difference)

Makefile.toml
```diff
@@ -52,6 +52,18 @@ category = "build"
...
+[tasks.build-certgen]
+category = "build"
+description = "Build aranya-certgen CLI tool"
+command = "cargo"
+args = ["build", "--verbose", "--release", "--package=aranya-certgen", "${@}"]
+
+[tasks.tgz-certgen]
+category = "build"
+description = "Archive aranya-certgen"
+script = "tar -czf aranya-certgen.tar.gz -C target/release aranya-certgen"
```
djohnson-spideroak, line 66: I think this needs to depend on the build task

crates/aranya-certgen/README.md
djohnson-spideroak, line 1: Please add a section on: how does one renew/reissue a device certificate?

crates/aranya-certgen/src/cert.rs
ssweetney-spok: > I don't think there's any guarantee that the memory doesn't get copied to do this (though I imagine it will effectively get optimized to construct it in place).

Yea, I believe this is correct. The [docs](https://docs.rs/zeroize/latest/zeroize/index.html#stackheap-zeroing-notes) seem to indicate this (moves may leave copies) too.

`CaCert::key` could just be `Zeroizing<KeyPair>` like you suggested [here](https://github.com/aranya-project/aranya/pull/681/files#r2700269734). The same could be done for `SignedCert::key`.

crates/aranya-certgen/src/cert.rs
ssweetney-spok: I believe the call to `Issuer::from_ca_cert_pem` down below would fail

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+        let mut opts = SaveOptions::default();
+        if self.create_parents {
+            opts = opts.create_parents();
+        }
+        if self.force {
+            opts = opts.force();
+        }
+        Some(opts)
+    }
```
ssweetney-spok: This could just return `SaveOptions`. `Option::unwrap_or_default` is called later when trying to use the save options when `SaveOptions::default()` could just be returned here instead of `None`.

crates/aranya-certgen/src/cert.rs
gknopf-aranya: After looking into this, it appears that the key is allocated internally on the heap with `Vec`. So move only copies a reference to the key, not the entire key vector. `Zeroizing` would be inefficient in this case.

gknopf-aranya: Also worth noting that the library implements zeroize for `KeyPair` already:
```
  #[cfg(feature = "zeroize")]
  impl zeroize::Zeroize for KeyPair {
      fn zeroize(&mut self) {
          self.serialized_der.zeroize();
      }
  }
```

gknopf-aranya: Wrapping the key field in `Zeroizing<KeyPair>` side-steps this by requiring the key to be wrapped on initialization rather than depending on the `Drop` impl. So it's a cleaner solution.

crates/aranya-certgen/src/cert.rs
gknopf-aranya: Updated to `key: Zeroizing<KeyPair>`

crates/aranya-certgen/src/main.rs
gknopf-aranya: Introducing a `CertPaths` struct that allows paths to be calculated uniformly ahead of cert generation.

crates/aranya-certgen/src/main.rs
gknopf-aranya: This feedback was addressed by the CertPaths refactoring. The current code:
```
let paths = CertPaths::new(&output);  // output is PathBuf from clap
ca.save(&paths, args.save_options())?;
```
No more .to_str() calls - CertPaths::new() takes impl AsRef<Path> and works directly with PathBuf.

crates/aranya-certgen/src/cert.rs
gknopf-aranya: Here's a rationale provided by claude:
```
The asymmetry is intentional for security:                                                                                         
                                                                                                                                     
  - cert_pem: Not sensitive, stored as String, returns &str                                                                          
  - key_pem: Sensitive, stored as Zeroizing<KeyPair>, serialized on demand to minimize time key material exists as a String          
                                                                                                                                     
  Options:                                                                                                                           
  1. Document the asymmetry - add a comment explaining why                                                                           
  2. Make both return String - serialize cert on demand too (unnecessary allocation but consistent API)                              
  3. Keep as-is - the current behavior is correct, just inconsistent naming                                                          
                                                                                                                                     
  I'd lean toward option 1. Want me to add a brief doc comment explaining the security rationale?   
```

crates/aranya-certgen/src/cert.rs
gknopf-aranya: Good point

crates/aranya-certgen/src/cert.rs
gknopf-aranya:
```
● The validation exists at line 205 - Issuer::from_ca_cert_pem() will fail if the cert isn't a CA. But the comment and error could be
   clearer. Options:                                                                                                                 
                                                                                                                                     
  1. Update the comment - Change "Validate that the cert can be parsed" to "Validate that the cert is a valid CA cert"               
  2. Add specific error variant - Add NotCaCert error for clearer user feedback                                                      
  3. Both - Clearer comment + better error                                                                                           
```
3. sounds good

crates/aranya-certgen/README.md
gknopf-aranya: I'm adding a section to the README.md for this. Note that in order for the daemon to load the renewed/reissued CA/device certs it would need to be restarted since we don't have a way to change cert configs at runtime yet.

crates/aranya-certgen/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+name = "aranya_certgen"
+path = "src/lib.rs"
+
+[[bin]]
+name = "aranya-certgen"
+path = "src/main.rs"
+
+[dependencies]
+clap = { workspace = true }
+rcgen = { version = "0.14", features = ["x509-parser", "zeroize"] }
```
gknopf-aranya, line 30: Removing the `zeroize` feature would result in a security bug. The problem is that if the feature was removed, the code would still compile which could allow that bug to be introduced silently in the future. This seems more likely to happen if the `rcgen` dep is moved to a workspace dep where context of the requirements for the `certgen` tool are less obvious. Is there a way to have the `certgen` crate require the `zeroize` feature to be enabled in order to compile? Or should we add a doc here indicating what the consequence of removing the feature would be.

jdygert-spok: Wouldn't removing the feature make e.g. `Zeroizing<Key>` not compile since it requires `Key: Zeroize`?

gknopf-aranya: Yep, verified the code does not compile without this feature. Marking as resolved.

crates/aranya-certgen/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+thiserror = { workspace = true }
+time = "0.3"
+zeroize = "1"
+
+[dev-dependencies]
+tempfile = { workspace = true }
+x509-parser = "0.18"
+
+[lints]
+workspace = true
```
jdygert-spok: nit: move this up under package like the other crates have

crates/aranya-certgen/README.md
```diff
@@ -0,0 +1,123 @@
+# aranya-certgen
+
+A CLI tool for generating root CA certificates and signed certificates.
+
+All generated keys use **P-256 ECDSA** (NIST P-256 / secp256r1 curve with ECDSA signatures).
```
jdygert-spok: Do we want to commit to this or should we note that the algorithm is an implementation detail and will be updated as per whatever latest crypto guidelines?

crates/aranya-certgen/README.md
```diff
@@ -0,0 +1,123 @@
...
+| `-p` | Create parent directories if they don't exist | — |
+| `-f, --force` | Overwrite existing files | — |
+
+### `aranya-certgen signed`
+
+Create a new certificate signed by an existing root CA with a P-256 ECDSA private key.
+
+| Option | Description | Default |
+|--------|-------------|---------|
+| `--cn <NAME>` | Common Name (CN) for the certificate | required |
```
jdygert-spok, line 53: If this is the only required arg we could just make it positional.

crates/aranya-certgen/README.md
```diff
@@ -0,0 +1,123 @@
...
+- Be signed by the same CA, so clients trusting the CA will accept the new certificate
+
+**Note:** If you need to preserve the existing private key (not recommended), this tool does not currently support that workflow. You would need to use OpenSSL or similar tools to create a CSR with the existing key.
+
+### Renewing the CA Certificate
+
+CA certificates can also be renewed, but this requires all device certificates to be reissued since they reference the CA:
+
+```bash
+# Generate new CA (will invalidate all existing device certs!)
```
jdygert-spok: This is... true-ish. It doesn't immediately invalidate all the existing certs, but if you replace the ca cert that those devices are using to validate the certs then they will no longer validate. Not really sure what we want to say here but I don't want to be misleading.

supply-chain/audits.toml
```diff
@@ -16,6 +16,24 @@ who = "gknopf <gknopf@spideroak.com>"
 criteria = "safe-to-deploy"
 delta = "1.0.94 -> 1.0.97"
 
+[[audits.asn1-rs]]
+who = "gknopf <gknopf@spideroak.com>"
+criteria = "safe-to-deploy"
+version = "0.7.1"
+notes = "Parser/encoder for ASN.1 BER/DER data. Unsafe code is forbidden. No networking code. Reader/writer interface provided for serializing/deserializing types to/from files."
```
jdygert-spok, line 23: Have you reviewed these yourself now?

crates/aranya-certgen/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "aranya-certgen"
+description = "Library and CLI tool for generating root CA certificates and signed certificates using P-256 ECDSA"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
```
jdygert-spok, line 10: Note that I think this won't publish until we next bump the workspace version.

For publishing this we will need an entry in `supply-chain/config.toml` and we should add it to `Status.md` too.

Makefile.toml
jdygert-spok: We could handle it either way but tgz-daemon and tgz-certgen should probably match.

supply-chain/audits.toml
mooreoak: I think we need to have more clarity on audit files and how we accept them. I think we might want some kind of review process one audits? I don't think we need to audit fully every dependency we add but we should make sure the project looks maintained at a minimum. We should also audit diffs for updates of dependencies to make sure they are not obviously malicious.

jdygert-spok: I think the best way to do this with the current tooling is to add new dependencies as exemptions, but still do a quick review on those new exemptions.

This will still require us to audit the diffs for updates.

(Obviously if you do a full review of the dependency that is better and you can add it as a real audit.)

gknopf-aranya: Yes, I've reviewed all these myself. I followed the recommendations provided by the `cargo vet` CLI tool to look for unsafe, networking, and file I/O code. I then had claude spot check my review to check for anything I may have missed.

gknopf-aranya: The cargo vet process is cumbersome esp. when adding new crates/deps to the repo. It would help with the review process if there was a checklist that could be followed. It would also be ideal if reviewers could easily verify that changes were reviewed rather than taking the PR author's word for it.

gknopf-aranya: Would be neat if we could create a cargo vet claude skill to assist with dep reviews.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,69 @@
+//! Library for generating root CA certificates and signed certificates.
```
jdygert-spok, line 1: Are we committing to stability for this library?

gknopf-aranya: I don't think it's wise to commit to stability on the initial PR. Not until after we land mTLS and have some track time with it.

gknopf-aranya: There are several TODOs documented in this lib for changes we know we want to make in the future. And likely other changes that aren't anticipated yet.

jdygert-spok: We need at least a comment here about the unstability, like aranya-daemon has.

crates/aranya-certgen/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[lib]
+name = "aranya_certgen"
+path = "src/lib.rs"
+
+[[bin]]
+name = "aranya-certgen"
+path = "src/main.rs"
+
+[dependencies]
+clap = { workspace = true }
```
jdygert-spok: Would be nice to make `clap` not required for the library. Either through a `cli` feature or moving the bin to a separate crate `aranya-certgen-cli`.

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,421 @@
...
+    if !options.force {
+        if paths.cert.exists() {
+            return Err(CertGenError::FileExists(paths.cert.display().to_string()));
+        }
+        if paths.key.exists() {
+            return Err(CertGenError::FileExists(paths.key.display().to_string()));
+        }
+    }
+
+    fs::write(&paths.cert, cert_pem).map_err(|e| CertGenError::io(&paths.cert, e))?;
```
jdygert-spok: TOCTOU, if a file is created after the check then it will be overwritten here. Instead use different `OpenOptions` flags based on `force`.

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,421 @@
...
+    // Check/create parent directory
+    if let Some(dir) = paths.cert.parent() {
+        if !dir.as_os_str().is_empty() && !dir.exists() {
+            if options.create_parents {
+                fs::create_dir_all(dir).map_err(|e| CertGenError::io(dir, e))?;
+            } else {
+                return Err(CertGenError::DirNotFound(dir.display().to_string()));
+            }
+        }
+    }
```
jdygert-spok, line 315: The key and cert could be in different directories but I only see this for the cert.

crates/aranya-certgen/src/lib.rs
jdygert-spok: Decided in rust call to not publish bin or lib until later.

crates/aranya-certgen/src/cert.rs
```diff
@@ -40,39 +40,87 @@ impl SaveOptions {
...
 pub struct CaCert {
     cert_pem: String,
-    key: KeyPair,
+    key: Zeroizing<KeyPair>,
 }
 
 impl CaCert {
+    /// Returns a reference to the inner key for use with rcgen APIs.
+    fn key_ref(&self) -> &KeyPair {
+        &self.key
```
chip-so: I guess this is a way to coerce `&Zeroizing<KeyPair>` to `&KeyPair`? You could just do `self.key.deref()` (you do have to import `std::ops::Deref`).

crates/aranya-certgen/README.md
chip-so: I'd prefer it to be an option for consistency. Though "required option" is definitely an oxymoron. I'd call it a "flag" but \*shrug\*.

supply-chain/audits.toml
gknopf-aranya: Resolving this since we discussed adjustments to the cargo vet review process today and this isn't related to the current PR.

crates/aranya-certgen/README.md
gknopf-aranya: Documenting this in a README isn't committing to anything since it isn't part of a stable API. If the crypto implementation changes, the README can be updated later.

crates/aranya-certgen/README.md
gknopf-aranya: I'll keep this as an option for now. Positional flags are less flexible and I anticipate the usage changing until we understand what this tool needs to be able to do.

crates/aranya-certgen/README.md
gknopf-aranya: I'll mention that invalidating the certs requires the daemon to reload the new CA cert.

crates/aranya-certgen/Cargo.toml
gknopf-aranya: Does not apply if crate is not being published yet.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Setting `publish = false`

crates/aranya-certgen/Cargo.toml
gknopf-aranya: Prefer to use a `cli` feature over creating another crate.

crates/aranya-certgen/src/cert.rs
gknopf-aranya: Sort of. That's not possible via CLI usage since it requires the same path prefix for both cert and key and only possible with the lib if someone manually constructs a `CertPaths` struct (which they can do because it is `pub`). But, that is not part of the expected usage of the library. Doesn't hurt to be defensive here though.
```
● Looking at CertPaths:                                                                                                              
  - The new() function creates both paths from the same prefix, so they're always in the same directory                              
  - But the fields are pub, so users could manually construct a CertPaths with different directories                                 
                                                                                                                                     
  CertPaths {                                                                                                                        
      cert: PathBuf::from("/dir1/cert.pem"),                                                                                         
      key: PathBuf::from("/dir2/key.pem"),  // different dir                                                                         
  }                                                                                                                                  
                                                                                                                                     
  So technically yes, but in practice via CertPaths::new(), no. Do you want to support the edge case or document that they must be in
   the same directory?   
```

gknopf-aranya: To be safe, let's make fields of `CertPaths` not `pub` and provide getter methods for them so the struct cannot be constructed manually and requires `CertPaths::new()` to be created. This will provide read-only access to the cert and key path fields and guarantee the cert/key are in the same directory as expected based on the path prefix.

crates/aranya-certgen/README.md
jdygert-spok: Observable behavior _can_ be part of the stability contract. The crate mentions it 20 times across the README and docs. If I used a tool that was _that_ adamant about telling you what the algorithm is, I'd be at least a little surprised when it changes.

djohnson-spideroak: Yeah, if in the future we are planning on adding other algorithms or updating based on regulatory guidance, we should mention that.

crates/aranya-certgen/README.md
```diff
@@ -0,0 +1,123 @@
...
+| `-f, --force` | Overwrite existing files | — |
+
+## Example Output
+
+```
+$ aranya-certgen ca --cn "My Company CA"
+Generating root CA certificate...
+  Certificate: ./ca.crt.pem
+
+$ aranya-certgen signed --cn webserver
```
djohnson-spideroak:
```suggestion
$ aranya-certgen signed --cn "webserver"
```

djohnson-spideroak: I also think we shouldnt really allow implicit usage of the CA for signing the cert. Seems like it could be easy to sign with the wrong CA. Generating a signed certificate should require an explicit path to the CA.

crates/aranya-certgen/README.md
gknopf-aranya: The library docs mention future work here in lib.rs:
```
//! # Future Work
//!
//! The following features may be added in the future:
//!
//! - **DER/BER format support**: Currently only PEM format is supported for saving/loading
//!   certificates. DER (Distinguished Encoding Rules) and BER (Basic Encoding Rules)
//!   binary format support could be added for environments where base64 encoding
//!   overhead is undesirable.
//!
//! - **Additional key types**: Support for other key algorithms such as:
//!   - P-384 ECDSA
//!   - Ed25519
//!   - HPKE (Hybrid Public Key Encryption) keys for use with MLS or other protocols
//!   - Post-quantum algorithms (e.g., ML-KEM, ML-DSA)
//!
//! - **Subject Alternative Names (SANs)**: Currently, only the Common Name (CN) is added
//!   as a DNS SAN. Explicit SAN support could be added via `--dns` and `--ip` flags to
//!   allow multiple DNS names and IP addresses in a single certificate.
//!
//! - **Certificate extensions**: Support for additional X.509 extensions such as
//!   custom OIDs and extension values.
//!
//! - **CSR support**: Certificate Signing Request (CSR) generation and signing.
```
I'd prefer not to duplicate this in the README.md. Although mentioning that it could change seems good.

jdygert-spok: IMO we should hardly mention the algorithm at all given that it is an implementation detail that will change.

gknopf-aranya: It may be an implementation detail of the certgen tool. However, this tool is used to generate certs which are then used by the Aranya daemon. I think it's important for users to understand what type of certs have been generated. They may wish to generate certs differently if they have different security requirements.

crates/aranya-certgen/README.md
gknopf-aranya: Making the CA a required position argument to `Signed`

crates/aranya-certgen/README.md
gknopf-aranya: The algorithm is relevant for understanding the security properties of our mTLS implementation that relies on this. I've intentionally included it in the README.md and code docs for transparency.

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+//!   overhead is undesirable.
+//!
+//! - **Additional key types**: Support for other key algorithms such as:
+//!   - P-384 ECDSA
+//!   - Ed25519
+//!   - HPKE (Hybrid Public Key Encryption) keys for use with MLS or other protocols
+//!   - Post-quantum algorithms (e.g., ML-KEM, ML-DSA)
+//!
+//! - **Subject Alternative Names (SANs)**: Currently, only the Common Name (CN) is added
+//!   as a DNS SAN. Explicit SAN support could be added via `--dns` and `--ip` flags to
```
djohnson-spideroak: I am not quite sure what "Currently, only the Common Name (CN) is added as a DNS SAN" means in this context. It seems like it implies the CN needs to be resolveable via DNS?

crates/aranya-certgen/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+//!   custom OIDs and extension values.
+//!
+//! - **CSR support**: Certificate Signing Request (CSR) generation and signing.
+//!
+//! # Example
+//!
+//! ```no_run
+//! use aranya_certgen::{CaCert, CertPaths, SaveOptions};
+//!
+//! // Create a new CA and save
```
djohnson-spideroak, line 35: any reason these are not rustdoc tests?

crates/aranya-certgen/src/lib.rs
gknopf-aranya: It means that the CN is set to either a DNS or IP SAN based on whether it is a hostname or IP address. We don't support specifying general SANs by adding them with `--dns` or `--ip` options. The wording could be more clear. This is left over from a previous iteration where the CN was always set to the DNS SANs.

crates/aranya-certgen/src/lib.rs
gknopf-aranya: They are meant to be simple examples. The `no_run` ensures these examples compile without running. If we want them to run, then a tmp dir would be more appropriate than generating certs on the user's file system. That would add complexity that isn't relevant to an example of how to use the library. We have unit tests that are intended to verify the functionality of the library.

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,129 @@
+//! CLI tool for generating root CA certificates and signed certificates.
+//!
+//! All generated keys use **P-256 ECDSA** (NIST P-256 / secp256r1 curve with ECDSA signatures).
```
djohnson-spideroak:
```suggestion
//! All generated keys currently use **P-256 ECDSA** (NIST P-256 / secp256r1 curve with ECDSA signatures).
```

crates/aranya-certgen/src/lib.rs
gknopf-aranya: Updated the wording

crates/aranya-certgen/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    #[arg(short = 'p')]
+    create_parents: bool,
+
+    /// Overwrite existing files.
+    #[arg(short, long)]
+    force: bool,
+}
+
+impl OutputArgs {
+    fn save_options(&self) -> SaveOptions {
```
djohnson-spideroak: nit: we may want to rename this, I understand it is implicitly "get_save_options" but given that save is a verb that would be used in a similar way, its confusing on first read.

crates/aranya-certgen/tests/cert_tests.rs
djohnson-spideroak, line 1: Maybe not on this PR, but we might want a test to confirm we can load these certs in our TLS library.

gknopf-aranya: I added a test to confirm that on the `mtls` branch. It seemed more fitting to add the test in a place where the TLS lib is already a dep than to add that as a dev dependency here.

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    /// Returns the certificate as a PEM-encoded string.
+    pub fn cert_pem(&self) -> &str {
+        &self.cert_pem
+    }
+
+    /// Returns the private key as a PEM-encoded string.
+    ///
+    /// Returns [`Zeroizing<String>`] to ensure key material is zeroized when dropped.
+    pub fn key_pem(&self) -> Zeroizing<String> {
```
djohnson-spideroak: does this need to be pub?

.github/workflows/publish.yml
```diff
@@ -115,6 +115,37 @@ jobs:
...
+        run: ls -R -al
+      - name: Upload certgen to tagged release
+        uses: actions/upload-release-asset@v1.0.2
+        env:
+            GITHUB_TOKEN: ${{ github.token }}
+        with:
+          upload_url: ${{ needs.release.outputs.upload-url }}
+          asset_path: aranya-certgen-${{ matrix.target }}/aranya-certgen.tar.gz
+          asset_name: aranya-certgen-${{ inputs.tag }}-${{ matrix.target }}.tar.gz
+          asset_content_type: application/gzip
```
jdygert-spok: Remove since we aren't publishing it yet?

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,436 @@
...
+        if !dir.as_os_str().is_empty() && !dir.exists() {
+            if options.create_parents {
+                fs::create_dir_all(dir).map_err(|e| CertGenError::io(dir, e))?;
+            } else {
+                return Err(CertGenError::DirNotFound(dir.display().to_string()));
+            }
+        }
+    }
+
+    // Write certificate file using OpenOptions to avoid TOCTOU race condition.
```
djohnson-spideroak, line 317: What is "TOCTOU race condition"?

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,436 @@
...
+    // permissions are set.
+    let mut key_options = OpenOptions::new();
+    key_options.write(true);
+    if options.force {
+        key_options.create(true).truncate(true);
+    } else {
+        key_options.create_new(true);
+    }
+    #[cfg(unix)]
+    key_options.mode(0o600);
```
djohnson-spideroak, line 345: Maybe add docs that the permissions are only set with #[cfg(unix)]

crates/aranya-certgen/README.md
chip-so: Adding quotes around a single word is absolutely unnecessary on a shell command line. It's fully equivalent with or without them. If you want to show that more complex names with spaces can be used, the example should do that.

crates/aranya-certgen/src/cert.rs
gknopf-aranya: https://github.com/aranya-project/aranya/pull/681#discussion_r2718193860
https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

.github/workflows/publish.yml
gknopf-aranya: We may still want to use it for integration testing. Would be convenient to have a pre-built binary available for this.

djohnson-spideroak: We shouldnt publish a binary if we dont intend this to be used by others, especially because it will be next to our main binaries in the release.

crates/aranya-certgen/src/cert.rs
gknopf-aranya: It's unused. Removing.

.github/workflows/publish.yml
gknopf-aranya: Removing it

crates/aranya-certgen/src/cert.rs
gknopf-aranya: Added docs explaining why we don't set file permissions for Windows.

crates/aranya-certgen/README.md
gknopf-aranya: Removing the quotes. A terminal user should know to put quotes around multiple words. That's not something a CLI tool is responsible for documenting.

crates/aranya-certgen/src/cert.rs
```diff
@@ -354,6 +340,7 @@ fn save_cert_and_key(
     } else {
         key_options.create_new(true);
     }
+    // Set restrictive permissions on Unix. Windows uses ACLs which require different APIs.
```
chip-so, line 343: I thought we didn't support Windows anyway.

crates/aranya-certgen/src/cert.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// - [`CertGenError::FileExists`] if files exist and `force` is false
+    /// - [`CertGenError::Io`] if writing files fails
+    pub fn save(&self, paths: &CertPaths, options: SaveOptions) -> Result<(), CertGenError> {
+        save_cert_and_key(paths, &self.cert_pem, &self.key.serialize_pem(), options)
+    }
+
+    /// Returns the certificate as a PEM-encoded string.
+    pub fn cert_pem(&self) -> &str {
+        &self.cert_pem
+    }
```
djohnson-spideroak, line 292: is this used, and if so, should it be public?

gknopf-aranya: Some of the `pub` methods aren't used here but they were needed by the `mtls` branch. Perhaps I can wait to add them until that PR so it's more clear which ones are used. Example:
```
      262 -                device_cert: device_paths.cert.clone(),                                                                   
      263 -                device_key: device_paths.key.clone(),                                                                     
      262 +                device_cert: device_paths.cert().to_path_buf(),                                                           
      263 +                device_key: device_paths.key().to_path_buf(),    
```

crates/aranya-certgen/src/cert.rs
gknopf-aranya: Right, we don't support it.

chip-so: So... why do we even care to mention it?

gknopf-aranya: Because it looks strange to set file permissions for `unix` with no explanation for why nothing is done for non-unix.

chip-so: I suppose. But this comment looks strange to me because it seems to imply that we do support windows, we just don't care to implement it properly.

PR #682, opened 2026-01-08
client: expose `aranya-id` based IDs

### LINKED ISSUE #680: afc: make `ChannelId` more useful
`Hash, Eq, Ord` and to/from bytes?

jdygert-spok:
Closes #680.


crates/aranya-client/src/client/device.rs
```diff
@@ -37,6 +38,7 @@ custom_id! {
     /// Uniquely identifies a device.
     pub struct DeviceId;
 }
+impl ApiId<api::DeviceId> for DeviceId {}
```
gknopf-aranya, line 41: What is the purpose of having an empty impl here?

crates/aranya-client/src/util.rs
```diff
@@ -1,65 +1,25 @@
-macro_rules! custom_id {
```
gknopf-aranya, line 1: Is this using the macro from `aranya-id` instead?

crates/aranya-client/src/client/device.rs
jdygert-spok: This is used by the blanket `ApiConv` impl to add conversions to/from the corresponding API type without adding a conversion from e.g. `DeviceId` to `api::RoleId` which could introduce bugs.

crates/aranya-client/src/util.rs
jdygert-spok: Yes

crates/aranya-client/src/client/device.rs
gknopf-aranya: Wouldn't this typically be done with a derived trait?

jdygert-spok: We need to associate the api type with this type somehow. We could add back the util custom_id macro and hide it in there but I like that this makes the association more explicit.

Issue #663, opened 2025-12-10
ci: remove RUSTSEC-2025-0134 exception during quinn migration


gknopf-aranya:
The `rustls-pemfile` crate is no longer maintained.
We've added an exception for this to `deny.toml` since we aren't currently using certs anywhere:
`RUSTSEC-2025-0134`

After/during the quinn migration, we should attempt to remove this exception to validate that nothing depends on this unmaintained crate.

cc: @jdygert-spok, @djohnson-spideroak, @benz-spideroak

Issue #648, opened 2025-12-02
quinn: implement quinn transport for syncer


gknopf-aranya:
Implement the quinn transport for the syncer.
Suggest skipping TLS cert verification for the initial prototype then adding mTLS certs in a different issue.
mTLS certs are required before landing the quinn syncer feature.

cc: @djohnson-spideroak, @jdygert-spok, @benz-spideroak, @nikki-aranya

Issue #647, opened 2025-12-02
quinn: implement mTLS certs


gknopf-aranya:
Implement mTLS certs for quinn based on the mTLS spec.

cc: @djohnson-spideroak, @jdygert-spok, @benz-spideroak, @nikki-aranya

PR #684, opened 2026-01-09
mtls

gknopf-aranya:
Migrate from (s2n_quic, PSKs) -> (quinn, mTLS)

# Summary

Migrates QUIC sync authentication from PSK (Pre-Shared Key) to mTLS (mutual TLS), providing stronger identity verification between sync peers.

# Key changes:
- Use aranya-certgen crate for X.509 certificate generation
- Replace PSK-based QUIC authentication with mTLS using per-device certificates
- Deprecate CreateTeamQuicSyncConfig, AddTeamQuicSyncConfig, and related PSK types (kept for backward compatibility)
- Remove psk.rs module, add certs.rs module for certificate handling
- Update all examples (Rust and C) to generate and use mTLS certificates

# Why mTLS over PSK?

- Stronger identity verification: Each device presents a certificate derived from its cryptographic identity
- Simplified key management: No need to distribute PSK seeds to new team members
- Standard TLS infrastructure: Leverages well-audited TLS certificate verification and supports external PKI

# Test plan

- cargo make test passes
- cargo make run-rust-example works with mTLS certificates
- cargo make run-capi-example works with mTLS certificates
- cargo make run-rust-example-multi-node works with mTLS certificates
- Verify deprecation warnings appear when using old PSK APIs

🤖 Generated with https://claude.com/claude-code

TODOs:
- [x] Deprecation warnings in C API with `cbindgen`

# Solves

- https://github.com/aranya-project/aranya/issues/663
- https://github.com/aranya-project/aranya/issues/647
- https://github.com/aranya-project/aranya/issues/648

### DIFF
--- a/crates/aranya-client/tests/common/mod.rs
+++ b/crates/aranya-client/tests/common/mod.rs
@@ -5,8 +5,8 @@ use std::{
 use anyhow::{anyhow, Context, Result};
 use aranya_certgen::{CaCert, CertPaths, SaveOptions};
 use aranya_client::{
-    client::{Client, DeviceId, PublicKeyBundle, Role, RoleManagementPermission, TeamId},
-    Addr, Permission, Rank, SyncPeerConfig,
+    client::{Client, DeviceId, PublicKeyBundle, Role, TeamId},
+    Addr, Rank, SyncPeerConfig,
 };
 // These deprecated types are only used in create_and_add_team for backward compatibility testing.
 #[expect(deprecated)]
--- a/crates/aranya-client/tests/tests.rs
+++ b/crates/aranya-client/tests/tests.rs
@@ -16,7 +16,7 @@ use std::{ptr, time::Duration};
 
 use anyhow::{bail, Context, Result};
 use aranya_client::{
-    client::{ChanOp, Permission, RoleId, RoleManagementPermission},
+    client::{ChanOp, Permission, Rank},
     config::{AddTeamConfig, CreateTeamConfig, HelloSubscriptionConfig, SyncPeerConfig},
 };
 use aranya_daemon_api::text;


crates/aranya-daemon/src/config/toggle.rs
```diff
@@ -63,7 +63,7 @@ mod imp {
     pub struct False;
     impl<'de> Deserialize<'de> for False {
         fn deserialize<D: de::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-            if bool::deserialize(deserializer)? {
+            if !bool::deserialize(deserializer)? {
```
gknopf-aranya: claude claims this was a bug. It caused the required cert fields to not be required even though the QUIC syncer was enabled in the config.

crates/aranya-daemon/src/sync/task/hello.rs
```diff
@@ -155,21 +155,20 @@ impl Syncer<State> {
     async fn send_hello_request(
         &mut self,
         peer: &Addr,
-        id: GraphId,
+        _id: GraphId, // GraphId is kept in API for future use but not needed for mTLS connection
```
gknopf-aranya: This arg can be removed

crates/aranya-daemon/src/sync/task/hello.rs
```diff
@@ -181,8 +180,8 @@ impl Syncer<State> {
             _ => "unknown",
         };
         // Read the response to avoid race condition with server
-        let mut response_buf = Vec::new();
-        recv.read_to_end(&mut response_buf)
+        let response_buf = recv
+            .read_to_end(64 * 1024) // 64KB limit for hello response
```
gknopf-aranya: Magic number. Why was 64KB selected?

crates/aranya-client/src/config/team/quic_sync.rs
```diff
@@ -98,13 +122,9 @@ impl AddTeamQuicSyncConfigBuilder {
...
     ///
     /// Overwrites [`Self::seed_ikm`].
-    pub fn wrapped_seed(mut self, wrapped_seed: &[u8]) -> Result<Self> {
-        let wrapped = postcard::from_bytes(wrapped_seed).map_err(|err| {
-            error!(error = %err, "could not deserialize wrapped_seed");
-            ConfigError::InvalidArg(InvalidArg::new("wrapped_seed", "could not deserialize"))
-        })?;
-        self.mode = Some(AddSeedMode::Wrapped(wrapped));
-        Ok(self)
+    pub fn wrapped_seed(mut self, wrapped_seed: &[u8]) -> Self {
```
gknopf-aranya: Looks like this broke backward compatibility. Revert.

crates/aranya-client/src/config/team/quic_sync.rs
```diff
@@ -5,15 +5,38 @@
...
+//! These types exist for backward compatibility but are ignored internally.
 
 use crate::{error::InvalidArg, ConfigError, Result};
 
+/// Size of the seed IKM (Input Keying Material) in bytes.
+pub const SEED_IKM_SIZE: usize = 32;
+
+/// Mode for creating a PSK seed when creating a team.
+#[derive(Clone, Debug, Default)]
+pub enum CreateSeedMode {
```
gknopf-aranya: If this wasn't in the API before these types should be `pub(crate)`

crates/aranya-client/tests/tests.rs
```diff
@@ -787,6 +668,7 @@ async fn test_remove_team() -> Result<()> {
 }
 
 /// Tests that a device can create multiple teams and receive sync requests for each team.
+/// With mTLS, devices can sync with any team they have the ID for - no add_team call needed.
```
```suggestion
/// With mTLS, devices can sync with any team they have the team ID for - no add_team call needed.
```

crates/aranya-client/src/config/team/quic_sync.rs
```diff
@@ -5,15 +5,38 @@
...
-use aranya_daemon_api::{AddSeedMode, CreateSeedMode, SEED_IKM_SIZE};
-use tracing::error;
+//!
+//! Note: With mTLS authentication, PSK seeds are no longer used.
+//! These types exist for backward compatibility but are ignored internally.
 
 use crate::{error::InvalidArg, ConfigError, Result};
 
+/// Size of the seed IKM (Input Keying Material) in bytes.
+pub const SEED_IKM_SIZE: usize = 32;
```
gknopf-aranya: Need more docs for deprecated types/methods

crates/aranya-client/tests/tests.rs
```diff
@@ -996,9 +796,9 @@ async fn test_hello_subscription() -> Result<()> {
         .sync_hello_subscribe(
             admin_addr,
             Duration::from_millis(100),
-            Duration::from_millis(1000),
+            Duration::from_secs(30), // Duration must be long enough for label creation + hello + sync
```
gknopf-aranya: This increased duration seems excessive

examples/rust/aranya-example/src/main.rs
```diff
@@ -91,6 +129,11 @@ impl ClientCtx {
                     .with_context(|| format!("unable to create directory: {}", dir.display()))?;
             }
 
+            // Generate device certificate for this daemon
+            let device_cert = config_dir.join("device.pem");
+            let device_key = config_dir.join("device-key.pem");
```
gknopf-aranya:
```suggestion
            let device_key = config_dir.join("device.key");
```

crates/aranya-daemon/src/sync/task/hello.rs
gknopf-aranya: @benz-spideroak it appears that the GraphId is passed around the sync hello implementation but not used in the actual sync hello protocol. Does this mean we subscribe to sync hello messages for all graphs rather than for specific graphs? Or was the sync hello implementation deriving the graph ID from the connection which no longer applies for the mTLS implementation?

gknopf-aranya: From what I can tell, the GraphID was looked up based on the PSK for the connection. This worked since we had a different PSK per connection/team. With the initial mTLS implementation, we have the same root/device certs for all teams, which means we cannot look up the GraphID based on which cert is used. I think it would be best if we include the GraphId in the sync protocol to allow syncing multiple teams over the same connection. This will be more flexible as we may change transport/crypto implementations in the future.

Cargo.toml
```diff
@@ -80,7 +81,10 @@ metrics-util = { version = "0.20" }
 postcard = { version = "1", default-features = false, features = ["use-std", "heapless", "experimental-derive"] }
 pretty_assertions = { version = "1.4" }
 rustix = { version = "1.1.2", features = ["shm"] }
-s2n-quic = { version = "1.58.0-psk.2", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"], package = "aranya-internal-s2n-quic" }
+quinn = { version = "0.11", default-features = false, features = ["runtime-tokio", "rustls"] }
+rcgen = { version = "0.14", default-features = false, features = ["crypto", "pem", "x509-parser", "ring"] }
+rustls = { version = "0.23", default-features = false, features = ["std", "tls12"] }
+rustls-pemfile = { version = "2" }
```
jdygert-spok: rustls-pemfile is deprecated, #663.

https://github.com/rustls/pemfile?tab=readme-ov-file#see-also-rustls-pki-types

crates/aranya-daemon/src/api.rs
```diff
@@ -587,18 +541,7 @@ impl DaemonApi for Api {
...
-            }
-            None => {
-                warn!("Missing QUIC sync config");
-
-                let seed = qs::PskSeed::new(&mut Rng, team_id);
-                self.add_seed(team_id, seed).await?;
-            }
-        }
-
+        // With mTLS, no PSK setup is required - authentication is handled by certificates
```
gknopf-aranya:
```suggestion
```

crates/aranya-daemon/src/sync/task/hello.rs
gknopf-aranya: Opened an issue for this here:
https://github.com/aranya-project/aranya-core/issues/538
I'm updating the TODOs on this PR to point to that issue so when it's resolved, we don't need to look up the graph ID from the cache of a previous sync.

crates/aranya-daemon/src/sync/task/quic.rs
```diff
@@ -511,38 +500,47 @@ where
...
-            .with_no_client_auth()
-            .with_cert_resolver(Arc::new(NoCertResolver::default()));
-        server_config.alpn_protocols = vec![ALPN_QUIC_SYNC.to_vec()]; // Set field directly
-        server_config.preshared_keys =
-            PresharedKeySelection::Required(Arc::clone(&server_keys) as _);
-
-        let tls_server_provider = rustls_provider::Server::new(server_config);
-
-        let addr = tokio::net::lookup_host(addr.to_socket_addrs())
+        // Load certificates
```
gknopf-aranya: Code for loading certs and keeping connections alive is nearly identical for client/server. Looks like it can be refactored.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,287 @@
...
+///
+/// # Arguments
+/// * `dir` - Path to the directory containing root CA certificate PEM files
+///
+/// # Returns
+/// A `RootCertStore` containing all loaded certificates
+///
+/// # Errors
+/// Returns an error if the directory cannot be read or if any certificate fails to parse.
+pub fn load_root_certs(dir: &Path) -> Result<rustls::RootCertStore> {
```
gknopf-aranya, line 104: Should define an error enum rather than `anyhow::Result`.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,361 @@
...
+    PrivateKeyDer<'static>,
+)> {
+    let root_store = load_root_certs(&config.root_certs_dir)?;
+    let (device_certs, device_key) = load_device_cert(&config.device_cert, &config.device_key)?;
+    Ok((root_store, device_certs, device_key))
+}
+
+#[cfg(test)]
+#[allow(clippy::expect_used, clippy::unwrap_used)]
+mod tests {
```
gknopf-aranya, line 296: May want to add a test to verify we can load certs generated by the `certgen` tool here. Although as long as the certs are PEM format, it should work.

crates/aranya-daemon/src/util.rs
gknopf-aranya, line 1: Questioning whether we still want this `util.rs` file when most of it was related to PSKs which are being removed.

crates/aranya-metrics/src/main.rs
```diff
@@ -302,21 +285,11 @@ async fn run_demo_body(ctx: DemoContext) -> Result<()> {
...
-            .team_id(team_id)
-            .build()?
-    };
-
-    // TODO: Delegate to admin and operator.
-    let _admin = ctx.admin.client.add_team(add_team_cfg.clone()).await?;
-    let _operator = ctx.operator.client.add_team(add_team_cfg.clone()).await?;
-    let membera = ctx.membera.client.add_team(add_team_cfg.clone()).await?;
-    let memberb = ctx.memberb.client.add_team(add_team_cfg).await?;
+    // With mTLS, no add_team call is required - just get team references
```
gknopf-aranya: Would expect cert generation to be added to metrics

examples/c/example.c
```diff
@@ -476,109 +414,7 @@ AranyaError init_team(Team *t) {
...
-
-        Client *client = &t->clients_arr[i];
-        err = aranya_add_team(&client->client, &cfg);
-        if (err != ARANYA_ERROR_SUCCESS) {
-            fprintf(stderr, "unable to add_team() for client: %s\n",
-                    client_names[i]);
-            return err;
-        }
-    }
+    // With mTLS, no add_team() call is required for non-owner devices
```
gknopf-aranya: This doc is more useful in the deprecated add_team() API than here. No point documenting why methods aren't called in the example.

examples/c/run.bash
```diff
@@ -16,8 +16,5 @@ cleanup() {
 trap 'cleanup' EXIT
 trap 'trap - SIGTERM && cleanup && kill -- -$$ || true' SIGINT SIGTERM EXIT
 
-# Run example with randomly generated QUIC syncer PSK seed
+# Run example (mTLS handles authentication - no PSK seed modes needed)
```
gknopf-aranya:
```suggestion
# Run example
```

examples/rust/aranya-example/src/main.rs
```diff
@@ -225,11 +288,11 @@ async fn main() -> Result<()> {
     let membera_addr = membera.aranya_local_addr().await?;
     let memberb_addr = memberb.aranya_local_addr().await?;
 
-    // Create a team.
+    // Create a team (with mTLS, config is no longer required - authentication is handled by certificates)
```
gknopf-aranya:
```suggestion
    // Create a team.
```

examples/rust/aranya-example-multi-node/src/admin/main.rs
```diff
@@ -53,18 +53,9 @@ async fn main() -> Result<()> {
...
-            .seed_ikm(team_info.seed_ikm)
-            .build()?;
-        AddTeamConfig::builder()
-            .quic_sync(qs_cfg)
-            .team_id(team_info.team_id)
-            .build()?
-    };
-    let team = client.add_team(add_team_cfg.clone()).await?;
-    info!("admin: added team");
+    // With mTLS, no add_team call is required - just get team reference
```
gknopf-aranya:
```suggestion
```

examples/rust/aranya-example-multi-node/src/owner/main.rs
```diff
@@ -53,21 +53,10 @@ async fn main() -> Result<()> {
     .expect("expected to initialize client");
     info!("owner: initialized client");
 
-    // Create team.
+    // Create team (mTLS handles authentication - config is accepted for backward compatibility).
```
gknopf-aranya:
```suggestion
    // Create team.
```

examples/rust/aranya-example/src/main.rs
```diff
@@ -262,20 +325,11 @@ async fn main() -> Result<()> {
...
-            .quic_sync(qs_cfg)
-            .team_id(team_id)
-            .build()?
-    };
-
-    let admin_team = admin.client.add_team(add_team_cfg.clone()).await?;
-    let operator_team = operator.client.add_team(add_team_cfg.clone()).await?;
-    let membera_team = membera.client.add_team(add_team_cfg.clone()).await?;
-    let memberb_team = memberb.client.add_team(add_team_cfg).await?;
+    // With mTLS, no add_team call is required - just get team references
```
gknopf-aranya:
```suggestion
```

examples/rust/aranya-example-multi-node/src/owner/main.rs
```diff
@@ -95,11 +84,9 @@ async fn main() -> Result<()> {
             continue;
         }
 
-        // Send team ID to device.
+        // Send team ID to device (mTLS handles auth - no PSK seed needed).
```
gknopf-aranya:
```suggestion
        // Send team ID to device.
```

crates/aranya-client/tests/common/mod.rs
```diff
@@ -121,43 +169,18 @@ impl DevicesCtx {
...
-                .quic_sync(qs_cfg)
-                .build()?
-        };
-
-        // Owner has the team added due to calling `create_team`, now we assign it to all other peers
-        self.admin.client.add_team(cfg.clone()).await?;
-        self.operator.client.add_team(cfg.clone()).await?;
-        self.membera.client.add_team(cfg.clone()).await?;
-        self.memberb.client.add_team(cfg).await?;
+        // With mTLS, no add_team call is required - devices can sync any team they have the ID for
```
gknopf-aranya:
```suggestion
```

crates/aranya-daemon/src/config/toggle.rs
gknopf-aranya: @jdygert-spok could you confirm whether you agree this is actually a bug?

Cargo.toml
gknopf-aranya: Removing this dep in favor of rustls-pki-types which is supported and recommended to be used instead of rustls-pemfile.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya, line 1: Reminder to use `cbindgen` to warn application if using deprecated types/methods.

gknopf-aranya:
Pulled in recent changes from `main` and `certgen`. The unit tests are failing due to recent SANs changes with the certgen tool.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Updated `cbindgen.toml` and `defs.rs` with code needed for deprecation warnings.

crates/aranya-daemon/src/config/toggle.rs
jdygert-spok: Agreed 🙃

jdygert-spok: This seems to manifest as a `enabled = true` with missing fields being parsed as `Toggle::Disabled`. And `enabled = false` as not a valid parse.

crates/aranya-daemon/src/aranya.rs
```diff
@@ -3,112 +3,124 @@
...
-    use std::{collections::HashSet, sync::RwLock};
+impl<EN, SP> Client<EN, SP> {
+    /// Creates a new Client
+    pub fn new(aranya: Arc<Mutex<ClientState<EN, SP>>>) -> Self {
+        Client { aranya }
+    }
 
-    use aranya_runtime::GraphId;
+    /// Locks the Aranya client state.
+    pub async fn lock_aranya(&self) -> MutexGuard<'_, ClientState<EN, SP>> {
```
gknopf-aranya: This shouldn't be showing up in the diff

examples/rust/aranya-example-multi-node/src/config.rs
```diff
@@ -1,14 +1,75 @@
...
+        Ok(Self { ca, root_certs_dir })
+    }
+
+    /// Returns the path to the root certificates directory.
+    pub fn root_certs_dir(&self) -> &Path {
+        &self.root_certs_dir
+    }
+
+    /// Generates a signed certificate.
+    pub fn generate_signed_cert(&self, _name: &str, work_dir: &Path) -> Result<(PathBuf, PathBuf)> {
```
gknopf-aranya: Could return `CertPaths` here.

crates/aranya-daemon/src/aranya.rs
gknopf-aranya: It is because `main` introduced a significant architectural change where resolving the merge conflicts is not straightforward.

jdygert-spok: Yeah IDK what you did, the merge didn't actually pull in any of the changes.

gknopf-aranya: I used claude to merge in changes from main but it made a bad call to keep some of the code from this branch instead of the new code from main. Once I told it what was wrong, it helped to resolve the merge conflicts correctly:
```
  Refactoring: Merge ClientWithState into Client                                                                                     
                                                                                                                                     
  The changes consolidate the architecture by:                                                                                       
                                                                                                                                     
  1. aranya.rs: Client<EN, SP> now directly contains:                                                                                
    - aranya (the ClientState)                                                                                                       
    - caches (the PeerCacheMap)                                                                                                      
    - hello_subscriptions (for the preview feature)                                                                                  
    - invalid_graphs (moved from daemon.rs)                                                                                          
                                                                                                                                     
  The separate ClientWithState wrapper was eliminated.                                                                               
  2. daemon.rs: Removed InvalidGraphs definition (moved to aranya.rs) and simplified the initialization code - no longer need to     
  create separate caches or InvalidGraphs instances.                                                                                 
  3. api.rs: Uses client.invalid_graphs() and client.lock_aranya() instead of separate invalid field.                                
  4. sync/task.rs, sync/task/quic.rs, sync/task/hello.rs: Updated to use Client instead of ClientWithState.                          
  5. test.rs: Simplified test setup, removed unused imports.  
```

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -100,32 +70,27 @@ where
...
-
-        let stream = syncer
+        let (send, recv) = syncer
             .connect(peer)
             .await
             .inspect_err(|e| error!(error = %e.report(), "Could not create connection"))?;
-        // TODO: spawn a task for send/recv?
-        let (mut recv, mut send) = stream.split();
+
+        let mut send = send;
```
gknopf-aranya: Make these mutable to begin with

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -1,88 +1,90 @@
...
+    endpoint: Endpoint,
+    /// Connection map shared with [`SyncManager`]
     conns: SharedConnectionMap,
     /// Receives updates for connections inserted into the [connection map][`Self::conns`].
     conn_rx: mpsc::Receiver<ConnectionUpdate>,
     /// Interface to trigger sync operations
-    handle: SyncHandle,
-    /// The address we're currently serving on.
-    local_addr: SocketAddr,
+    _sync_peers: SyncHandle,
```
gknopf-aranya: Unused?

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -366,15 +392,55 @@ where
...
-fn check_request(team_id: TeamId, request: &SyncRequestMessage) -> Result<GraphId> {
-    let SyncRequestMessage::SyncRequest { storage_id, .. } = request else {
-        bug!("Should be a SyncRequest")
-    };
-    if team_id.as_bytes() != storage_id.as_bytes() {
-        return Err(Error::QuicSync(quic::Error::InvalidPSK));
+    /// Processes a hello message for the preview feature.
+    #[cfg(feature = "preview")]
+    #[instrument(skip_all)]
+    async fn process_hello_message(
```
gknopf-aranya: I thought this was in a hello.rs file

crates/aranya-daemon/src/sync/transport/quic/server.rs
gknopf-aranya: Used by sync hello feature. Clarifying this.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya, line 1: Only the PSK stuff should be deprecated, we should keep the QUIC syncer config in case it's needed later.

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
```diff
@@ -142,12 +145,7 @@ impl AddTeamQuicSyncConfigBuilder {
     ///
     /// This method will be removed soon since certificates will be used instead of PSKs in the future.
```
gknopf-aranya: Comment out-of-date

crates/aranya-client/src/client/team.rs
```diff
@@ -55,22 +52,16 @@ impl Team<'_> {
...
-            .encrypt_psk_seed_for_peer(context::current(), self.id, peer_enc_pk)
-            .await
-            .map_err(IpcError::new)?
-            .map_err(aranya_error)?;
-        let wrapped = postcard::to_allocvec(&wrapped).assume("can serialize")?;
-        Ok(wrapped)
+    pub async fn encrypt_psk_seed_for_peer(&self, _peer_enc_pk: &[u8]) -> Result<Vec<u8>> {
+        // With mTLS authentication, PSK seeds are no longer used.
+        // Return empty vector for backward compatibility.
+        Ok(Vec::new())
```
gknopf-aranya, line 55: Returning an empty vec would cause a security bug if the deprecated PSK methods were used. We could:
1. Write random data into this vec
2. Always return an error if this deprecated method is used
3. Continue encrypting the PSK seed as before

If the goal is to not break applications already using the existing API, I think we should continue encrypting the PSK seed the same way we were before. So 3. seems like the best approach to take here.

crates/aranya-client/tests/tests.rs
```diff
@@ -996,9 +820,9 @@ async fn test_hello_subscription() -> Result<()> {
         .sync_hello_subscribe(
             admin_addr,
             Duration::from_millis(100),
-            Duration::from_millis(1000),
-            Duration::from_secs(60), // Schedule delay for periodic sends
-        ) // Short delay for faster testing
+            Duration::from_secs(120), // Duration long enough for test to complete
```
gknopf-aranya: These sync hello flake fixes could be moved onto their own PR

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -797,6 +806,8 @@ pub type CreateTeamQuicSyncConfigBuilder = Safe<imp::CreateTeamQuicSyncConfigBui
 /// This method will be removed soon since certificates will be used instead of PSKs in the future.
 ///
 /// @relates AranyaCreateTeamQuicSyncConfigBuilder.
+#[deprecated(note = "PSK seeds are no longer used with mTLS authentication")]
+#[allow(deprecated)]
 pub fn create_team_quic_sync_config_generate(
```
gknopf-aranya, line 897: This shouldn't be deprecated

crates/aranya-client/src/config/team.rs
```diff
@@ -82,8 +83,9 @@ impl AddTeamConfigBuilder {
 }
 
 /// Configuration for creating a new team.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct CreateTeamConfig {
+    #[allow(dead_code)]
```
gknopf-aranya: Explain that this is due to deprecated PSKs

crates/aranya-client/src/config/team.rs
```diff
@@ -94,18 +96,11 @@ impl CreateTeamConfig {
...
-        }
-    }
-}
-
 /// Configuration for joining an existing team.
 #[derive(Clone, Debug)]
 pub struct AddTeamConfig {
-    id: TeamId,
+    pub(crate) id: TeamId,
+    #[allow(dead_code)]
```
gknopf-aranya: Explain that this is due to deprecated PSKs.

examples/c/run.bash
```diff
@@ -16,8 +16,5 @@ cleanup() {
 trap 'cleanup' EXIT
 trap 'trap - SIGTERM && cleanup && kill -- -$$ || true' SIGINT SIGTERM EXIT
 
-# Run example with randomly generated QUIC syncer PSK seed
+# Run example
 ./example.bash
```
gknopf-aranya: Double-check that argument parsing code is removed from example.bash and example.c.

examples/c/example.c
```diff
@@ -1211,26 +1121,12 @@ typedef struct {
 } channel_context_t;
 
 int main(int argc, char *argv[]) {
+    (void)argc;
+    (void)argv;
```
gknopf-aranya: These don't appear to be needed.

examples/c/example.c
```diff
@@ -1211,26 +1121,12 @@ typedef struct {
 } channel_context_t;
 
 int main(int argc, char *argv[]) {
```
gknopf-aranya: `argc` and `argv` are unused.

examples/rust/aranya-example/src/main.rs
```diff
@@ -262,20 +322,10 @@ async fn main() -> Result<()> {
...
-            .quic_sync(qs_cfg)
-            .team_id(team_id)
-            .build()?
-    };
-
-    let admin_team = admin.client.add_team(add_team_cfg.clone()).await?;
-    let operator_team = operator.client.add_team(add_team_cfg.clone()).await?;
-    let membera_team = membera.client.add_team(add_team_cfg.clone()).await?;
-    let memberb_team = memberb.client.add_team(add_team_cfg).await?;
+    let admin_team = admin.client.team(team_id);
```
gknopf-aranya: Still need to call `add_team()` to be able to sync

crates/aranya-client/src/client/team.rs
gknopf-aranya: Then again, the new mTLS implementation will ignore these PSKs anyways since it won't be possible to configure the daemon to sync with PSKs anymore. So there isn't a risk of someone accidentally configuring an empty PSK with the syncer once they have upgraded the client and daemon versions. A new daemon with an old client would fail because it no longer has the `encrypt_psk_seed_for_peer()` IPC API. A new client with an old daemon could theoretically configure an empty PSK to be used by the syncer. Up to this point, we've been assuming that the client & daemon versions match each other so that should be okay. Stating it here in case that assumption is no longer valid.

crates/aranya-client/src/client.rs
```diff
@@ -254,14 +254,7 @@ impl Client {
 
     /// Add a team to local device storage.
     pub async fn add_team(&self, cfg: AddTeamConfig) -> Result<Team<'_>> {
```
gknopf-aranya, line 273: Deprecated now

crates/aranya-daemon/src/daemon.rs
```diff
@@ -100,17 +99,13 @@ impl Daemon {
...
-                Some(v) => v,
-            };
 
             Self::setup_env(&cfg).await?;
             let mut aranya_store = Self::load_aranya_keystore(&cfg).await?;
             let mut eng = Self::load_crypto_engine(&cfg).await?;
             let pks = Self::load_or_gen_public_keys(&cfg, &mut eng, &mut aranya_store).await?;
 
-            let mut local_store = Self::load_local_keystore(&cfg).await?;
+            let _local_store = Self::load_local_keystore(&cfg).await?;
```
gknopf-aranya: Why is this unused?

examples/c/run.bash
gknopf-aranya: We don't need a run.bash and example.bash anymore. They can be unified into a single script.

crates/aranya-client/tests/tests.rs
```diff
@@ -627,101 +627,6 @@ async fn test_query_functions() -> Result<()> {
     Ok(())
 }
 
-/// Tests add_team() by demonstrating that syncing can only occur after
-/// a peer calls the add_team() API
-#[test(tokio::test(flavor = "multi_thread"))]
-async fn test_add_team() -> Result<()> {
```
gknopf-aranya, line 759: We should keep this test but ignore the deprecation warnings if we want to claim backward compatibility for the deprecated add_team() API.

crates/aranya-client/tests/common/mod.rs
```diff
@@ -1,24 +1,73 @@
...
+use std::{
+    collections::HashMap, iter, net::Ipv4Addr, path::PathBuf, ptr, sync::Arc, time::Duration,
+};
 
 use anyhow::{anyhow, Context, Result};
+use aranya_certgen::{CaCert, CertPaths, SaveOptions};
 use aranya_client::{
     client::{Client, DeviceId, KeyBundle, Role, RoleManagementPermission, TeamId},
-    config::CreateTeamConfig,
+    config::{CreateTeamConfig, SEED_IKM_SIZE},
```
gknopf-aranya: Aren't these types deprecated? Why are they allowed to be imported here without `allow(deprecated)`?

gknopf-aranya: They aren't, I just noticed it before CI did:
```
       7  use aranya_client::{                                                                                                       
       8      client::{Client, DeviceId, KeyBundle, Role, RoleManagementPermission, TeamId},                                         
       9 +    Addr, SyncPeerConfig,                                                                                                  
      10 +};                                                                                                                         
      11 +// These deprecated types are only used in create_and_add_team for backward compatibility testing.                         
      12 +#[allow(deprecated)]                                                                                                       
      13 +use aranya_client::{                                                                                                       
      14      config::{CreateTeamConfig, SEED_IKM_SIZE},                                                                             
      10 -    AddTeamConfig, AddTeamQuicSyncConfig, Addr, CreateTeamQuicSyncConfig, SyncPeerConfig,                                  
      15 +    AddTeamConfig, AddTeamQuicSyncConfig, CreateTeamQuicSyncConfig,                                                        
      16  };   
```

crates/aranya-client/src/client/team.rs
jdygert-spok: IIRC empty PSKs won't be accepted by rustls anyways.

gknopf-aranya: That would make sense

crates/aranya-client/tests/tests.rs
```diff
@@ -251,26 +251,7 @@ async fn test_role_create_assign_revoke() -> Result<()> {
...
-                    AddTeamQuicSyncConfig::builder()
-                        .wrapped_seed(&admin_seed)?
-                        .build()?,
-                )
-                .build()?
-        })
-        .await?;
-
-    // Admin sync with owner.
+    // Admin sync with owner (with mTLS, no add_team call needed).
```
gknopf-aranya: The no add_team docs aren't needed since add_team is deprecated.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Opened a new PR for this to clean up the diff noise on this PR:
https://github.com/aranya-project/aranya/pull/699

crates/aranya-client-capi/src/api/mod.rs
```diff
@@ -12,6 +12,7 @@ pub mod defs;
 #[allow(unused_unsafe)]
 #[allow(clippy::undocumented_unsafe_blocks)]
 #[allow(rustdoc::broken_intra_doc_links)]
+#[allow(deprecated)]
```
djohnson-spideroak: Why is this here?

gknopf-aranya: The PSK related types are deprecated by this PR which means types/methods that depend on them need `#[allow(deprecated)]`.

gknopf-aranya: Worth mentioning that I've also deprecated `add_team()` function and the CreateTeam, AddTeam configs on this PR. While we may want to use those functions/types in the future, they aren't currently used without the underlying QUIC sync PSK types so marking them as deprecated indicates to users that they shouldn't be using those associated types/methods anymore.

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
```diff
@@ -124,30 +122,21 @@ pub struct AddTeamQuicSyncConfigBuilder {
...
-    /// This method will be removed soon since certificates will be used instead of PSKs in the future.
+    #[deprecated(note = "PSK seeds are no longer used with mTLS authentication")]
     pub fn wrapped_seed(&mut self, encap_seed: &[u8]) -> Result<(), Error> {
-        let wrapped = postcard::from_bytes(encap_seed).map_err(|err| {
-            error!(error = %err, "could not deserialize wrapped_seed");
-            InvalidArg::new("wrapped_seed", "could not deserialize")
-        })?;
-        self.mode = Some(AddSeedMode::Wrapped(wrapped));
-
+        self.mode = Some(AddSeedMode::Wrapped(encap_seed.to_vec()));
```
djohnson-spideroak: Why was the error handling removed?

crates/aranya-client/src/client/team.rs
djohnson-spideroak: Why are we keeping functions/endpoints around as deprecated when we are removing their functionality?

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
gknopf-aranya: The PSK seed no longer needs to be deserialized since it is no longer used by the mTLS implementation. We are keeping the PSK related methods and types in the API for backward compatibility but they are ignored/unused.

crates/aranya-client/src/client/team.rs
gknopf-aranya: For API backward compatibility. This topic was discussed at a previous Rust call where I asked if the unused PSK types/methods could be removed from the API because they are no longer needed by the mTLS implementation. The consensus was that we need to keep them in the API but mark them as deprecated in order to preserve backward compatibility in the API.

gknopf-aranya: If this PR is allowed to break backward compatibility in the API, I agree that it would be much simpler to remove all the unused methods and the added complexity of the deprecation code.

Cargo.toml
```diff
@@ -79,8 +80,9 @@ metrics = { version = "0.24" }
 metrics-util = { version = "0.20" }
 postcard = { version = "1", default-features = false, features = ["use-std", "heapless", "experimental-derive"] }
 pretty_assertions = { version = "1.4" }
+quinn = { version = "0.11", default-features = false, features = ["rustls", "runtime-tokio", "log"] }
 rustix = { version = "1.1.2", features = ["shm"] }
-s2n-quic = { version = "1.58.0-psk.2", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"], package = "aranya-internal-s2n-quic" }
+rustls = { version = "0.23", default-features = false, features = ["std", "tls12"] }
```
djohnson-spideroak: double check the version is current. If this was added by claude, it probably isnt.

crates/aranya-client/src/client/team.rs
djohnson-spideroak: I think we should really just consider eating the change and fully removing the now useless functions. We can discuss it more in at the product call.

gknopf-aranya: We'll want to be clear on exactly what is being removed. Is it just the QUIC PSK stuff? Is it also the now unused add_team() and CreateTeam, AddTeam config objects (which only exist to send PSK info at the moment)?

gknopf-aranya: The argument was made that we may want to keep the AddTeam, CreateTeam config objects in case they are needed in the future. But we don't know what those might be needed for in the future at this time.

Cargo.toml
gknopf-aranya: Latest version is 0.23.36:
https://docs.rs/rustls/latest/rustls/

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -100,32 +68,24 @@ where
...
-        let (mut recv, mut send) = stream.split();
 
         let mut sync_requester = SyncRequester::new(peer.graph_id, &mut Rng, syncer.server_addr);
 
         // send sync request.
         syncer
             .send_sync_request(&mut send, &mut sync_requester, peer)
             .await
-            .map_err(|e| Error::SendSyncRequest(e.into()))?;
+            .map_err(|e| crate::sync::Error::SendSyncRequest(e.into()))?;
```
gknopf-aranya: Import the `Error` type to simplify this.

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -1,88 +1,56 @@
...
-
-    /// Creates a new instance
-    fn new(
-        psk_store: Arc<PskStore>,
+    /// Creates a new instance using a shared endpoint.
+    ///
+    /// The endpoint is created by [`Server::new()`] and shared with the SyncManager.
+    /// Using a single endpoint ensures that outbound connections use the server's
+    /// bound address as the source, enabling bidirectional connection reuse.
+    pub(super) fn new(
```
gknopf-aranya, line 69: This constructor is so lightweight that it could be a struct initialization.

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -313,22 +285,26 @@ where
...
         trace!("sent sync request");
 
         Ok(())
     }
 
-    #[instrument(skip_all)]
     /// Receives and processes a sync response from the server.
     ///
     /// Returns the number of commands that were received and successfully processed.
+    #[instrument(skip_all)]
```
djohnson-spideroak: I dont think we want to skip all here

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: The encapsulation is nice in case we need more initialization in the future. Let's keep this as-is.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
```diff
@@ -1,157 +1,193 @@
-//! This modules contains a map for storing persistent QUIC connections between pairs of sync peers.
-//! The [connection map][SharedConnectionMap] allows the sync client and server to share existing connections by providing
-//! the client access to a mutable connection [handle][Handle]. The corresponding [stream acceptor][StreamAcceptor],
-//! that's used by the sync server, is sent over a channel when new connections are inserted in the map.
+//! This module contains a map for storing persistent QUIC connections between pairs of sync peers.
+//!
+//! The [connection map][SharedConnectionMap] allows the sync client and server to share existing
+//! connections. With mTLS authentication, a single connection can be used for syncing any team
+//! since authentication is based on device certificates rather than team-specific PSKs.
```
gknopf-aranya: Don't need to mention PSKs since they are being removed.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
```diff
@@ -1,157 +1,193 @@
...
-        match self.handles.lock().await.entry(peer) {
+    pub(super) async fn remove(&mut self, key: ConnectionKey, conn: Connection) {
+        match self.connections.lock().await.entry(key) {
             Entry::Vacant(_) => {}
             Entry::Occupied(entry) => {
-                if entry.get().id() == handle.id() {
+                if entry.get().stable_id() == conn.stable_id() {
                     entry.remove();
-                    handle.close(AppError::UNKNOWN);
+                    conn.close(0u32.into(), b"connection closed");
```
gknopf-aranya, line 75: Would be simpler to return early if entry.get().stable_id() != conn.stable_id()

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,414 @@
...
+/// # Arguments
+/// * `cert_path` - Path to the device certificate PEM file (may contain certificate chain)
+/// * `key_path` - Path to the private key PEM file (PKCS#8 or SEC1/EC format)
+///
+/// # Returns
+/// A tuple of (certificate chain, private key)
+///
+/// # Errors
+/// Returns an error if the files cannot be read or parsed.
+pub fn load_device_cert(
```
djohnson-spideroak, line 173: Might be good to have a function to load multiple device certs. In the future I imagine we are going to have devices that need to join a bunch of teams with different device certs.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
gknopf-aranya, line 1: Looking at this code more closely, there appears to be a possibility that multiple connections with the same peer could be opened then discarded if duplicate connections are detected.
```
❯ In connections.rs, it seems like there are several instances where multiple connections with the same peer could be opened         
  where one is closed if there is a duplicate connection. Is there a way to avoid opening duplicate connections in the first         
  place?                                                                                                                             
                                                                                                                                     
● Read(crates/aranya-daemon/src/sync/transport/quic/connections.rs)                                                                  
  ⎿  Read 194 lines                                                                                                                  
                                                                                                                                     
● Looking at the code, the race condition occurs in get_or_try_insert_with:                                                          
                                                                                                                                     
  1. Lock is held while checking for existing connection (lines 99-117)                                                              
  2. Lock is dropped (line 118)                                                                                                      
  3. New connection is created without lock (line 122) - this is slow/async                                                          
  4. Lock is re-acquired and race is detected (lines 125-140)                                                                        
                                                                                                                                     
  The gap between steps 2 and 4 allows multiple tasks to simultaneously decide "no connection exists" and all create connections,    
  wasting resources.                                                                                                                 
                                                                                                                                     
  Options to avoid duplicate connections:                                                                                            
                                                                                                                                     
  1. Pending state pattern: Store an enum like enum ConnState { Pending(watch::Receiver), Connected(Connection) }. When a task needs 
  to connect, it inserts Pending first (while holding the lock), then creates the connection. Other tasks see Pending and wait on the
   receiver instead of creating their own.                                                                                           
  2. Per-peer OnceCell: Store Arc<OnceCell<Connection>> per peer. The first task to call get_or_init creates the connection; others  
  wait on the same future.                                                                                                           
  3. Hold lock during connection (not recommended): Would serialize all connection attempts and hurt performance.                    
                                                                                                                                     
  The pending state pattern is probably the cleanest - it ensures only one task creates a connection per peer while allowing         
  concurrent connections to different peers. The trade-off is added complexity in handling the case where the pending connection     
  fails.                                                                                                                             
                                                                                                                                     
  Want me to implement option 1 or 2?
```
Could be a future improvement to consider on a different PR. Documenting this here in case anyone would like to discuss this more.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,414 @@
...
+    // Load private key
+    let key_file = File::open(key_path).map_err(|e| CertError::OpenFile {
+        path: key_path.to_path_buf(),
+        source: e,
+    })?;
+
+    let key = PrivateKeyDer::from_pem_reader(key_file).map_err(|e| CertError::ParseKey {
+        path: key_path.to_path_buf(),
+        source: e,
+    })?;
```
djohnson-spideroak: Zeroize needed here?

crates/aranya-daemon/src/sync/transport/quic/connections.rs
gknopf-aranya: Also worth noting that this behavior was present in the previous syncer and is not introduced on this PR. I merely noticed it while reviewing this file diff.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,414 @@
...
+#[allow(clippy::expect_used, clippy::unwrap_used)]
+mod tests {
+    use std::io::Write;
+
+    use tempfile::TempDir;
+
+    use super::*;
+
+    // Helper to create a test PEM file with dummy content
+    fn create_pem_file(dir: &Path, name: &str, content: &str) -> std::io::Result<()> {
```
djohnson-spideroak: rename to "create_file"

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,414 @@
...
+    fn test_load_root_certs_no_pem_files() {
+        let temp_dir = TempDir::new().expect("failed to create temp dir");
+        create_pem_file(temp_dir.path(), "not-a-cert.txt", "hello world")
+            .expect("failed to create file");
+        let result = load_root_certs(temp_dir.path());
+        assert!(result.is_err());
+        assert!(result
+            .unwrap_err()
+            .to_string()
+            .contains("no root CA certificates found"));
```
djohnson-spideroak: Why not check the error type instead of the string? Same for the other tests.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,414 @@
...
+        let temp_dir = TempDir::new().expect("failed to create temp dir");
+        create_pem_file(temp_dir.path(), "invalid.pem", "not a valid certificate")
+            .expect("failed to create file");
+        let result = load_root_certs(temp_dir.path());
+        // Invalid PEM content results in no certs parsed, which is now an error
+        assert!(result.is_err());
+        assert!(result
+            .unwrap_err()
+            .to_string()
+            .contains("no root CA certificates found"));
```
djohnson-spideroak: Shouldnt this result in a parse error?

crates/aranya-daemon/src/sync/transport/quic/certs.rs
djohnson-spideroak, line 1: We may want to have the ability to pass certs/keys into the API directly, we probably wont always be working with a filesystem. I know thats difficult to do in C without the content potentially sticking around in memory, but via rust I think we could do it.

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -332,20 +350,29 @@ where
...
         peer_server_addr: Addr,
-        active_team: &TeamId,
     ) -> Result<Box<[u8]>> {
-        let mut resp = SyncResponder::new(local_addr);
-        let storage_id = check_request(*active_team, &request_msg)?;
-        let peer = SyncPeer::new(peer_server_addr, storage_id);
+        trace!("server responding to sync request");
+
+        // Extract the storage_id (GraphId) from the request
+        let SyncRequestMessage::SyncRequest { storage_id, .. } = &request_msg else {
```
gknopf-aranya: `GraphID` has been added to the sync protocol

crates/aranya-daemon/src/sync/transport/quic/server.rs
gknopf-aranya, line 1: Reminder to remove any code that relies on the peer's server address in the sync protocol since it has been removed from the protocol.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: I discussed this with @mooreoak when writing the mtls spec but we decided against it for now in favor of a daemon config for certs. Passing in binary cert buffers isn't as easy as loading them from files.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: We didn't include loading of multiple device certs in the mtls spec. Is this something we want on this PR or a future enhancement to mTLS?

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: That's a good idea

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: It also requires the application to manage the secret keys securely which may not be desired. If the `certgen` tool is used, we have more control over how certs and keys are managed.

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: This wasn't changed on the PR, it was moved below the function doc. Since the other fields are complex, `peer` is the only one I'd add to trace output.

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -222,57 +180,72 @@ where
...
+            .map_err(|e| Error::EndpointError(format!("unable to get local address: {e}")))?;
+        let local_is_ipv4 = local_addr.is_ipv4();
 
         let addr = tokio::net::lookup_host(peer.addr.to_socket_addrs())
             .await
-            .context("DNS lookup on for peer address")?
-            .next()
-            .context("could not resolve peer address")?;
+            .context("DNS lookup for peer address")?
+            .find(|addr| addr.is_ipv4() == local_is_ipv4)
```
djohnson-spideroak: nit: I dont really like this find, Its hard to understand what is happening with v4 vs v6.

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -222,57 +180,72 @@ where
...
-                    .await?;
-                conn.keep_alive(true)?;
-                Ok(conn)
+            .get_or_try_insert_with(key, async || {
+                let connecting = endpoint
+                    .connect_with(client_config, addr, &peer_host)
+                    .map_err(Error::from)?;
+
+                // Add timeout to connection attempt to avoid hanging on failed TLS handshakes
+                let conn = tokio::time::timeout(Duration::from_secs(5), connecting)
```
djohnson-spideroak: Does quinn not provide this for us?

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Added `peer` to trace

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: ~It does via `max_handshake_duration`~
`s2n_quic` has `max_handshake_duration` but the `quinn` `TransportConfig` struct does not:
https://docs.rs/quinn/latest/quinn/struct.TransportConfig.html
The only timeout that is configurable with `quinn` is the [max_idle_timeout](https://docs.rs/quinn/latest/quinn/struct.TransportConfig.html#method.max_idle_timeout)

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Don't really like this code either. This was added because it appears both peers must have matching IP versions for the TLS handshake to succeed.

djohnson-spideroak: My problem is mostly just that its hard to reason about the logic in the find call by just reading it. It might be better stated as a two-clause if statement.

crates/aranya-daemon/src/sync/transport/quic/client.rs
djohnson-spideroak: We should rm the extra code then and use the builtin

gknopf-aranya: There is an open issue for this that is marked as unplanned in the `quinn` project:
https://github.com/quinn-rs/quinn/issues/2298
The current code is the correct way to handle the timeout in the application layer rather than within the library since the library does not support it.

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: I'll simplify this code and move it to a helper method with better docs.

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Documenting the rationale for this along with the open issue # in a code comment.

jdygert-spok: AFAICT it will still time out when the connection is idle for too long during the handshake. Do we really need a separate timeout for this? "avoid hanging on failed TLS handshakes" seems dubious, I wouldn't expect it to hang.

gknopf-aranya: The issue I linked on this discussion mentioned that the idle timeout doesn't cover this case. When I added a test that intentionally tries to connect with invalid certificates, the TLS handshake hung indefinitely until this TLS handshake timeout was manually added.

jdygert-spok: I read the linked issue as saying that it will time out for idle, but if you want a separate timeout for the handshake to do it that way. I see we have set an unlimited idle timeout, maybe that is why it hangs? I think we want to use keep-alive rather than unlimited idle timeout.

crates/aranya-daemon/src/sync/transport/quic/server.rs
gknopf-aranya: On a similar note, we can derive the local server address directly from the `Endpoint` now, which means code can be refactored so there are fewer `Addr` parameters to pass around with the server address.

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: I set the unlimited idle timeout so QUIC connections could be reused. As long as the timeout is high and there is a keep-alive that should work.

gknopf-aranya: Even if we use a longer idle timeout and the keep-alive, we'll still want this separate TLS timeout. For example, if the idle timeout is 30 seconds, we don't want to wait 30 seconds to find out if the TLS handshake failed. I'd like to know if the TLS handshake failed in more like 5-10 seconds.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: Yeah, adding a `NoCertsInFile` error variant.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: Yep, pattern matching on errors would be better

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: This is creating a file and writing a string to it. While it isn't really PEM format specific, it also does more than just creating a file. How about `create_file_from_str()`?

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: When enabling the `alloc` feature with `rustls-pki-types`, `PrivateKeyDer` will automatically zeroize for us due to `Zeroizing<Vec<u8>>`. I need to upgrade that crate to v1.14 to enable the `alloc` feature.

gknopf-aranya: Enabled the `alloc` feature and documented the behavior here.

crates/aranya-client-capi/src/api/mod.rs
gknopf-aranya: Marking this as resolved since we've started a more detailed discussion on deprecation and API backward compatibility in another thread:
https://github.com/aranya-project/aranya/pull/684#discussion_r2723168026

crates/aranya-daemon/src/sync/transport/quic/server.rs
gknopf-aranya: Removed any remnants of sending the server address in the sync protocol and refactored code to pull the local addr directly from the `Endpoint` rather than passing it around as an `Addr` param in a bunch of places.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
gknopf-aranya: @ssweetney-spok interested in your thoughts on this since you've helped write the current QUIC connection code.

crates/aranya-daemon/src/sync/transport/quic/server.rs
gknopf-aranya: This is resolved by merging in the latest changes from `main` which have the newer sync protocol from aranya-core and removing a few TODOs that were waiting on the lib change.

crates/aranya-daemon/src/sync/manager.rs
gknopf-aranya, line 1: Revert these changes. Bad merge.

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya, line 1: Revert these changes. Bad merge.

crates/aranya-daemon/src/vm_policy.rs
gknopf-aranya, line 1: Revert these changes. Bad merge.

supply-chain/config.toml
```diff
@@ -788,12 +768,8 @@ criteria = "safe-to-deploy"
 version = "0.8.1"
 criteria = "safe-to-deploy"
 
-[[exemptions.rustls-pemfile]]
-version = "2.2.0"
-criteria = "safe-to-deploy"
-
 [[exemptions.rustls-pki-types]]
-version = "1.11.0"
+version = "1.14.0"
```
gknopf-aranya, line 776: I need to review 1.11.0 -> 1.14.0

crates/aranya-daemon/src/sync/transport/quic/connections.rs
ssweetney-spok: The rationale for option 3 is why I avoided holding the lock during the connection creation attempt so that makes sense to me.

I didn't realize `tokio::sync::OnceCell` existed... it seems like a good solution to this problem. Option 2 handles the scenario where [something goes wrong during connection creation](https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html#method.get_or_try_init) with less complexity from what I can tell:

> If the provided operation returns an error, is cancelled or panics, the initialization attempt is cancelled. If there are other tasks waiting for the value to be initialized, one of them will start another attempt at initializing the value.

If that behavior is OK then option 2 works. I wonder why Claude prefers option 1 🤔?

gknopf-aranya: Option 2 seems more intuitive to me and seems to achieve a similar result as option 1.

supply-chain/config.toml
gknopf-aranya: Reviewed this and certified it.

crates/aranya-daemon/src/actions.rs
```diff
@@ -33,9 +33,9 @@ pub(crate) struct SessionData {
 }
 
 /// Functions related to Aranya actions
-impl<PS, SP, CE> Client<PS, SP>
+impl<EN, SP, CE> Client<EN, SP>
```
gknopf-aranya: This shouldn't show up in the diff

crates/aranya-daemon/src/actions.rs
```diff
@@ -75,6 +75,7 @@ where
 
     /// Create new ephemeral Session.
     /// Once the Session has been created, call `session_receive` to add an ephemeral command to the Session.
+    #[cfg(feature = "afc")]
```
gknopf-aranya, line 79: Not sure why these need to be added here when they weren't needed on `main`

gknopf-aranya: Oh, it's because the `Session` import was put behind the `afc` feature which causes a compilation error if these methods are not also behind the feature flag. I think we should leave this change in since `afc` is the only feature that uses session commands. This allows sessions to be optimized out of the build if they are not used anywhere.

crates/aranya-daemon/src/sync/transport/quic/client.rs
jdygert-spok: Actually even with an unlimited idle timeout it should fail immediately when the handshake fails.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
gknopf-aranya: Decision made today to include this in future work

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: @jdygert-spok, @djohnson-spideroak I'm adding some unit tests closer to the sync module in `mod.rs` to allow for more direct testing of the sync server, client, and cert verification. These tests show that the TLS handshake hangs if different client and server CA certs are used. So we will need to wrap a timeout around the TLS handshake so it doesn't hang on invalid certs in production.

crates/aranya-client/src/client/team.rs
gknopf-aranya: We decided to keep the deprecated APIs with no-ops for backward compatibility.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
gknopf-aranya: Adding a TODO for this

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Turns out the client TLS handshake only hangs if the server doesn't call `accept()`. So we don't need the TLS handshake timeout for the original reason it was added. Cert failures will return early. We should probably keep the timeout in case the server is unresponsive and does not call `accept()` in a timely manner. I've added docs to explain why I'm keeping the timeout for the handshake.

crates/aranya-client-capi/CMakeLists.txt
```diff
@@ -31,6 +31,17 @@ corrosion_import_crate(
...
+# ==== Certgen executable (for test certificate generation) ====
+
+corrosion_import_crate(
+    MANIFEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../aranya-certgen/Cargo.toml
+    CRATES aranya-certgen
+    CRATE_TYPE bin
+    FEATURES cli
+)
+
+add_executable(Aranya::Certgen ALIAS aranya-certgen)
```
jdygert-spok, line 43: Move to tests

crates/aranya-daemon/src/sync/transport/quic/mod.rs
```diff
@@ -1,50 +1,71 @@
...
 
+/// Creates a QUIC transport configuration with no idle timeout.
+///
+/// Connections are kept open indefinitely without keep-alive pings.
+/// This is appropriate for daemon-to-daemon connections where resources
+/// are not constrained and connection reuse is desired.
+fn transport_config() -> Arc<TransportConfig> {
+    let mut transport_config = TransportConfig::default();
+    // Disable idle timeout - connections stay open forever.
+    transport_config.max_idle_timeout(None);
```
jdygert-spok, line 34: > WARNING: If a peer or its network path malfunctions or acts maliciously, an infinite idle timeout can result in permanently hung futures!

We should look into what this entails.

PR #685, opened 2026-01-12
(client) - IPC timeout handling

### LINKED ISSUE #636: bug: long blocking operations can cause daemon IPC to timeout

ssweetney-spok:
Closes #636

~~Bump the tarpc request deadline for `sync_now` calls from the default, 10 seconds, to 5 minutes.~~

~~`sync_now` takes a "timeout" parameter. [See here](https://github.com/aranya-project/aranya/pull/685#discussion_r2688058724)~~

- The C and Rust client APIs have the IPC timeout bumped to 1 year

### DIFF
--- a/crates/aranya-client/src/client.rs
+++ b/crates/aranya-client/src/client.rs
@@ -5,7 +5,12 @@ mod label;
 mod role;
 mod team;
 
-use std::{fmt::Debug, io, path::Path, time::Instant};
+use std::{
+    fmt::Debug,
+    io,
+    path::Path,
+    time::{Duration, Instant},
+};
 
 use anyhow::Context as _;
 use aranya_crypto::{Csprng, Rng};
@@ -42,10 +47,15 @@ pub use crate::client::{
     team::{Team, TeamId},
 };
 use crate::{
-    config::{AddTeamConfig, CreateTeamConfig, MAX_SYNC_INTERVAL},
+    config::{AddTeamConfig, CreateTeamConfig},
     error::{self, aranya_error, InvalidArg, IpcError, Result},
     util::ApiConv as _,
 };
+
+/// IPC timeout of 1 year (365 days).
+// A large value helps resolve IPC calls timing out when there are long-running
+// operations happening in the daemon.
+const IPC_TIMEOUT: Duration = Duration::from_secs(365 * 24 * 60 * 60);
 
 /// Builds a [`Client`].
 #[derive(Debug, Default)]
@@ -286,10 +296,10 @@ impl Client {
 }
 
 /// Returns the current [`Context`](context::Context) with a deadline set to the current time
-/// plus [`MAX_SYNC_INTERVAL`].
+/// plus [`IPC_TIMEOUT`].
 pub(crate) fn create_ctx() -> context::Context {
     let mut ctx = context::current();
-    ctx.deadline = Instant::now() + MAX_SYNC_INTERVAL;
+    ctx.deadline = Instant::now() + IPC_TIMEOUT;
 
     ctx
 }


crates/aranya-client/src/client/team.rs
```diff
@@ -52,6 +51,8 @@ impl Team<'_> {
 }
 
 impl Team<'_> {
+    const SYNC_NOW_REQUEST_DEADLINE: Duration = Duration::from_secs(60 * 5);
```
gknopf-aranya: Do we know what the current QUIC connection timeout is? If I recall, it was something like 30s. If a connection timeout causes the sync request to return with an error, then I don't think the IPC deadline needs to be any longer than the QUIC connection timeout.

ssweetney-spok: IIRC the QUIC connections do not timeout but instead are held indefinitely since we set the "keep_alive" attribute to true. If you mean the default TLS handshake duration, then I believe that is 10 seconds.

gknopf-aranya (approved):

gknopf-aranya:
Thinking about whether we would want to add a test to verify a long sync no longer times out the IPC. That seems optional because this code is increasing the timeout properly according to the TARPC docs.

ssweetney-spok:
> Thinking about whether we would want to add a test to verify a long sync no longer times out the IPC. That seems optional because this code is increasing the timeout properly according to the TARPC docs.

Yea, I wondered about that too. I suppose we could add a long sleep to simulate a long sync but it seems like we'd really be testing tarpc

crates/aranya-client/src/client/team.rs
jdygert-spok: A slow network could take an arbitrary amount of time to sync. This is definitely better but not a "perfect" solution.

gknopf-aranya: Rather than attempting to predict how long a sync could take on an arbitrary user's network, is this timeout something we should allow to be configured via the API?

jdygert-spok: I'm not sold on the way tarpc does timeouts. I think it'd be more convenient for the rust API to wait indefinitely, which allows the caller to stop polling the future whenever they want.

But for C we'd probably still need timeouts, so I guess we'd add them to that API.

ssweetney-spok: Adding a timeout parameter to the API seems better 👍

gknopf-aranya: I interpreted this as having sync_now() return a future in the Rust API, while taking a timeout parameter in the C API. So the recent change that added the timeout parameter to the Rust API doesn't match @jdygert-spok's suggestion.

ssweetney-spok: I read that as something that'd be nice to have. Is it possible to pass futures across different processes?

jdygert-spok: > Is it possible to pass futures across different processes?

No, but I'm not sure why you'd need to.

To be clear, the rust client's methods are already async (returning futures), they just also time out internally.

ssweetney-spok: > No, but I'm not sure why you'd need to.

I thought the suggestion was to have `sync_now`'s result return a future so it'd look something like:

```rust
pub async(...) -> Result<impl Future<..>>
```

> To be clear, the rust client's methods are already async (returning futures), they just also time out internally.

Yea, that's my understanding of things.

What's the solution for the Rust API that doesn't have the timeout parameter?

ssweetney-spok: Currently as of this [commit](https://github.com/aranya-project/aranya/pull/685/commits/410ae8fdfb5de7dcaead65e0b941c77f5c563149), the timeout duration for `sync_now` defaults to 10 seconds (the tarpc default) if `None` is passed in. Can we leave the timeout parameter and fall back to 1 year if the user doesn't pass in a timeout parameter?

cc: @djohnson-spideroak

ssweetney-spok: > I think it'd be more convenient for the rust API to wait indefinitely, which allows the caller to stop polling the future whenever they want.

> But for C we'd probably still need timeouts, so I guess we'd add them to that API.

Or do we want the above instead? We'd set the IPC timeout to 1 year (not indefinite but close enough) and remove the timeout parameter from the Rust API. Users of the Rust API can do what they want with the returned future

gknopf-aranya: 1yr is effectively indefinite, so this seems close enough.

gknopf-aranya: I suppose a counter-argument could be that it could result in untestable edge-cases that fail a year after deploying something that are difficult to debug.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -2765,15 +2765,19 @@ AranyaError aranya_sync_hello_unsubscribe_ext(const struct AranyaClient *client,
...
+ * The default timeout, 1 year, will be used if `timeout` is `NULL`.
+ *
  * @relates AranyaClient.
  */
 AranyaError aranya_sync_now(const struct AranyaClient *client,
                             const struct AranyaTeamId *team,
                             AranyaAddr addr,
-                            const struct AranyaSyncPeerConfig *config);
+                            const struct AranyaSyncPeerConfig *config,
+                            const AranyaDuration *timeout);
```
jdygert-spok: We'll want to expose this in a backwards compatible way. Some options to consider:
1. `sync_now_with_timeout` variant
2. varargs for optional trailing arg
3. field in `SyncPeerConfig`
4. config on client

crates/aranya-client/src/client/team.rs
```diff
@@ -88,9 +87,12 @@ impl Team<'_> {
     /// be used.
     #[instrument(skip(self))]
     pub async fn sync_now(&self, addr: Addr, cfg: Option<SyncPeerConfig>) -> Result<()> {
+        let mut req_ctx = context::current();
+        req_ctx.deadline = Instant::now() + MAX_SYNC_INTERVAL;
+
         self.client
             .daemon
-            .sync_now(context::current(), addr, self.id, cfg.map(Into::into))
+            .sync_now(req_ctx, addr, self.id, cfg.map(Into::into))
```
jdygert-spok: I am thinking we will want the "unlimited" timeout for the whole Rust API.

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: If we want to the configurable timeout to apply to the whole API, then number 4 seems like a good option

djohnson-spideroak: I think adding sync_now_with_timeout and keeping sync_now as it was is a good option. Most operations should correctly have IPC timeout after 10 seconds, but sync does need the ability to bypass that.

ssweetney-spok: Yea, I came to the same conclusion after playing around with the config idea. It seemed to make the API more cumbersome to use

jdygert-spok: My concern is that we want a timeout on every function, otherwise it is awkward since you might time out due to another long running operation, then you'll have to add a retry loop to every function that you call.

ssweetney-spok: Isn't `sync_now` the only "long running" operation at the moment? The other ones should be able to complete in ~10 seconds but there's no way to guarantee that though

jdygert-spok: `add_sync_peer` will time out if you have a long sync going, but that should get fixed as part of parallel sync or sooner.

With a slow filesystem (e.g. network share), slow processor, or a very large graph, you could make any operation take a while...

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1939,8 +1969,7 @@ pub unsafe fn sync_hello_unsubscribe(
 /// @param[in] team the team's ID
 /// @param[in] addr the peer's Aranya network address
 /// @param[in] config configuration values for syncing with a peer.
-///
-/// Default values for a sync config will be used if `config` is `NULL`
+/// @param[in] timeout the maximum amount of time this request can take.
```
ssweetney-spok: revert

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: For the C API, there's a new field on the Client Config now. There isn't a function to update the timeout field in the config (do we want that?) so users have to create a new client to have a different timeout value.

See crates/aranya-client-capi/ctest/test_timeout.c

.github/workflows/tests.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         uses: ./.github/actions/setup
 
       - name: Unit Tests
-        run: cargo make ctest
+        run: cargo make ctest-verbose
```
gknopf-aranya: Questioning whether we want to always run this in verbose mode in CI or only when debugging the C example.

crates/aranya-client-capi/ctest/include/utils.h
```diff
@@ -132,6 +132,17 @@ static inline AranyaError get_role_id_by_name(const AranyaRole *role_list,
         }                                                                      \
     } while (0)
 
+// Macro for printing an error message when to stderr.
```
gknopf-aranya: This sentence doesn't make sense

crates/aranya-client-capi/ctest/test_timeout.c
```diff
@@ -0,0 +1,101 @@
...
+                  aranya_client_config_builder_init(&owner_builder));
+
+    CLIENT_EXPECT("Failed to set owner daemon path", "",
+                  aranya_client_config_builder_set_daemon_uds_path(
+                      &owner_builder, g_owner_uds));
+
+    // Set a timeout of 10 seconds
+    CLIENT_EXPECT("Failed to set IPC timeout", "",
+                  aranya_client_config_builder_set_ipc_timeout(
+                      &owner_builder, ARANYA_DURATION_SECONDS * (uint64_t)10));
```
gknopf-aranya: This appears to be reusing the original builder from the first short timeout step without cleaning up the previous builder.

gknopf-aranya: This may be okay if `init` zeroizes everything so that it's essentially a brand new object. If not, there's a risk some of the values from the previous step could get carried over accidentally.

crates/aranya-client-capi/ctest/test_timeout.c
```diff
@@ -0,0 +1,101 @@
...
+    // Create team as owner
+    CLIENT_EXPECT("Expected team creation to succeed", "",
+                  create_team(&owner_client, &team_id));
+
+    printf("Team creation succeeded \n");
+
+    result = EXIT_SUCCESS;
+
+exit:
+    aranya_client_cleanup(&owner_client);
```
gknopf-aranya: Should we clean up the config objects/builders here?

crates/aranya-client-capi/ctest/test_timeout.c
```diff
@@ -0,0 +1,101 @@
...
+    AranyaClientConfig owner_config;
+    CLIENT_EXPECT("Failed to build owner config", "",
+                  aranya_client_config_build(&owner_builder, &owner_config));
+
+    CLIENT_EXPECT("Failed to init owner client", "",
+                  aranya_client_init(&owner_client, &owner_config));
+
+    // Create team as owner
+    AranyaTeamId team_id;
+    CLIENT_EXPECT_ERR("Expected team creation to time out",
```
gknopf-aranya: Is it worth checking for a specific error instead of any error here?

crates/aranya-client-capi/ctest/test_timeout.c
```diff
@@ -0,0 +1,101 @@
...
+
+    CLIENT_EXPECT("Failed to build owner config", "",
+                  aranya_client_config_build(&owner_builder, &owner_config));
+
+    CLIENT_EXPECT("Failed to init owner client", "",
+                  aranya_client_init(&owner_client, &owner_config));
+
+    // Create team as owner
+    CLIENT_EXPECT("Expected team creation to succeed", "",
+                  create_team(&owner_client, &team_id));
```
gknopf-aranya: If `create_team()` times out, we don't know if there is any state change (e.g. in the client/daemon) that occurred before the operation timed out. I suppose it's unlikely anything changed if the timeout is 0. Worth considering if a different owner_client object should be initialized for proper isolation of the two tests.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -14,25 +14,36 @@ pub(crate) use team::*;
...
 
 impl ClientConfigBuilder {
     /// Set the address for the daemon
     pub fn daemon_addr(&mut self, addr: *const c_char) {
         self.daemon_addr = addr;
     }
+
+    /// Set the timeout for IPC calls
+    pub fn ipc_timeout(&mut self, duration: Duration) {
+        self.ipc_timeout = Some(duration);
```
gknopf-aranya: Is this setting a timeout for the entire IPC or just sync_now? I thought we had a discussion at some point about only needing this timeout for sync_now. The rest of the IPC calls should return quickly without blocking.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -49,6 +60,7 @@ impl Builder for ClientConfigBuilder {
 
         let cfg = ClientConfig {
             daemon_addr: self.daemon_addr,
+            ipc_timeout: self.ipc_timeout.map(Into::into),
```
gknopf-aranya: Could alternatively implement `From` to convert between the two config types.

crates/aranya-client/src/client/team.rs
```diff
@@ -46,7 +46,7 @@ impl Team<'_> {
     pub async fn close_team(&self) -> Result<()> {
         self.client
             .daemon
-            .close_team(context::current(), self.id)
+            .close_team(create_ctx(), self.id)
```
gknopf-aranya, line 46: Or use a default:
```suggestion
            .close_team(Context::default(), self.id)
```
But for sync_now():
```suggestion
            .close_team(Context::default().with_timeout(cfg.sync_now_timeout), self.id)
```

crates/aranya-client/src/client.rs
```diff
@@ -284,3 +284,12 @@ impl Client {
         AfcChannels::new(self.daemon.clone(), self.afc_keys.clone())
     }
 }
+
+/// Returns the current [`Context`](context::Context) with a deadline set to the current time
+/// plus [`MAX_SYNC_INTERVAL`].
+pub(crate) fn create_ctx() -> context::Context {
```
gknopf-aranya, line 300: Could be moved to impl of a `Context` object.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -17,6 +17,21 @@ use tracing::{debug, error};
...
 use crate::imp;
 
+/// Uses the client config to determine whether `fut` should be timed.
+macro_rules! timed_future {
+    ($client:expr, $fut:expr) => {{
+        let client_cfg: &imp::ClientConfig = &$client.cfg;
+
+        match client_cfg.ipc_timeout() {
+            Some(timeout) => {
+                let timed_fut = async { tokio::time::timeout(timeout, $fut).await };
```
gknopf-aranya: This will execute each time a future is needed. Would it be possible to initialize `timed_fut` once on client init and reuse it later by cloning? Could that improve performance at all?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -252,7 +267,12 @@ pub unsafe fn client_init(
             .connect()
     })?;
 
-    Client::init(client, imp::Client { rt, inner });
+    let cfg = {
```
gknopf-aranya: This is a bit messy. Could we use an `.into()` conversion here instead?

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: It may make more sense to pass the timeout into the C API via a `Context` object in a similar way that the IPC does. Then we could support changing the timeout of that context object. That way the timeout is kept in a context object that is separate from the client config.

crates/aranya-client-capi/ctest/test_timeout.c
ssweetney-spok: Yea, that can be added.

Edit: Added

crates/aranya-client-capi/output/aranya-client.h
gknopf-aranya: If most of our API calls need a timeout, I'm questioning whether we want to support this via a config/param for each method or a flag so the cbindgen tool can generate functions that time out in a more general way. Autogenerating code that could timeout from a `#[timeout]` flag seems cleaner in defs.rs and could automatically add timeout parameters to each method.

gknopf-aranya: If we don't want to require the user to input the timeout each time, we could generate both `api_call()` and `api_call_with_timeout()` methods.

crates/aranya-client-capi/ctest/test_timeout.c
ssweetney-spok: The `init` functions initialize the object with default values instead of zeroizing

crates/aranya-client-capi/ctest/test_timeout.c
ssweetney-spok: Yea, I'm not sure there will be state changes if the initial timeout is 0. I could add a second daemon but I don't think it's necessary

crates/aranya-client-capi/ctest/test_timeout.c
ssweetney-spok: I believe the `AddressSanitizer` check would've failed if they needed to be. That's the issue I was running into when trying to re-use the `owner_client` variable without calling the cleanup function.

The config builders don't hold any pointers so that's probably why they don't need to be

crates/aranya-client-capi/ctest/test_timeout.c
jdygert-spok: This is something I've been thinking about that makes me kinda wary to have timeouts at all :/

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Generally async timers expire at a certain instant and are inserted in some timer wheel structure so you can't really precompute them.

crates/aranya-client-capi/ctest/test_timeout.c
jdygert-spok: But I guess the timeouts aren't really the issue but just one way that things can get interrupted.

crates/aranya-client-capi/output/aranya-client.h
jdygert-spok: Maybe what we need for the C api are blocking/nonblocking modes. The blocking mode would just block on the call until it finishes. The nonblocking mode would immediately return a handle which can be used to check the progress of the call or cancel it.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: `From`/`Into` are not the right traits to use here. I want to call the `Deref` impls on `Opaque` and `Safe`.

I could change this to:

`let cfg = (**config).clone()`

crates/aranya-client-capi/src/imp/config.rs
ssweetney-spok: For the functions that use it. See usages of `timed_future!`

[Additional context
](https://github.com/aranya-project/aranya/pull/685#discussion_r2722119759)

crates/aranya-client/src/client/team.rs
ssweetney-spok: `tarpc::context::Context` is a foreign type.

crates/aranya-client-capi/src/api/defs.rs
ssweetney-spok: As an aside...

I may be misremembering but I remember code like `let cfg = (**config).clone()` failing to compile in the past. Maybe the upgrade to 1.90 changed things 🤔

crates/aranya-client/src/client.rs
ssweetney-spok: Rust doesn't allow crates to define `impl`s on foreign types.

Or do you mean create a new type to wrap `tarpc::context::Context`?

crates/aranya-client-capi/src/imp/config.rs
ssweetney-spok: There's only 1 config type here. The builder is building a config

.github/workflows/tests.yml
ssweetney-spok: I can revert it. I thought it'd be useful for errors like the ones from the `AddressSanitizer` that can't be reproduced on our dev machines.

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: > The blocking mode would just block on the call until it finishes. The nonblocking mode would immediately return a handle which can be used to check the progress of the call or cancel it.

That'd make the C API behave more like the Rust API. Maybe that's better

But I suppose we'd have to have 2 versions for each function

jdygert-spok: What do we think of this approach?
1. Remove timeouts from both APIs
2. Later add nonblocking mode to the C API if it has demand

gknopf-aranya: 2 versions of each function isn't bad if it's autogenerated by the cbindgen tool. We generate the `_cleanup()` methods already.

.github/workflows/tests.yml
gknopf-aranya: If there is an AddressSanitizer issue that happens frequently, I imagine we would increase verbosity to resolve that. Too much output makes it hard to read through normal success/failure modes.

ssweetney-spok: Reverted

crates/aranya-client-capi/output/aranya-client.h
ssweetney-spok: I've removed the timeouts from both APIs

crates/aranya-client/src/client.rs
gknopf-aranya: I mean wrapping it in a new type

gknopf-aranya (dismissed):

crates/aranya-client/src/client.rs
jdygert-spok: Free function is fine

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Could be related to when we made the client shared. Before that, if you used DerefMut to access a field it borrows the whole item so then you can't access another field.

djohnson-spideroak (dismissed):

crates/aranya-client/src/client.rs
```diff
@@ -284,3 +284,12 @@ impl Client {
...
+    let mut ctx = context::current();
+    ctx.deadline = Instant::now() + MAX_SYNC_INTERVAL;
```
jdygert-spok: nit: weird to use MAX_SYNC_INTERVAL since this isn't sync related

jdygert-spok (dismissed):

chip-so (dismissed):

crates/aranya-client/src/client.rs
gknopf-aranya: Yeah, there should be a different const defined for this. If they are the same value, that's just a coincidence.

crates/aranya-client/src/client.rs
gknopf-aranya: Resolving this since it's an optional suggestion

crates/aranya-client/src/client.rs
jdygert-spok: They do both serve the same purpose of being a long time that can safely be used in a delay queue. (tarpc also uses `DelayQueue` internally for the timeouts)

ssweetney-spok: [Added a new constant](https://github.com/aranya-project/aranya/pull/685/changes/0743c3cb73490a0a5629c815468eb83efb7ccfd3)

PR #687, opened 2026-01-14
Simplify ctest setup

jdygert-spok:


crates/aranya-client-capi/ctest/CMakeLists.txt
```diff
@@ -94,83 +94,22 @@ if(NOT APPLE AND NOT WINDOWS)
     )
 endif()
 
-# --- Paths ---
-set(RUST_TARGET_DIR "$ENV{ARANYA_TARGET_DIR}")
```
gknopf-aranya, line 98: What allows this whole section to be removed? Is it because more conventional paths are used that don't need to be configured?

crates/aranya-client-capi/ctest/README.md
```diff
@@ -77,7 +79,6 @@ Add your test to the "Test Executables" section in `CMakeLists.txt`:
 ```cmake
 # --- Test Executables ---
 add_capi_test(TestOnboarding test_onboarding.c REQUIRES_DAEMON DAEMON_NAMES "owner,member")
-add_capi_test(TestSimple test_simple.c)
```
gknopf-aranya, line 80: This is removed here but I don't see a `test_simple.c` file removed from the PR

crates/aranya-client-capi/ctest/README.md
```diff
@@ -5,19 +5,21 @@ This document describes how to build and run the C API integration tests for Ara
...
 
+Start from the workspace/repo root.
+
 ### Run the Full Test Suite using ctest
 
 1. **Configure the build directory:**
   ```sh
-  cd [workspace_path]/crates/aranya-client-capi/ctest/build
-  cmake ..
+  cmake -B build -S .
```
gknopf-aranya, line 14: Since we've made the repo a `cmake` dep, I wonder if we should be able to run these build/test commands from the workspace root?

crates/aranya-client-capi/ctest/CMakeLists.txt
jdygert-spok: `corrosion` manages building the rust crates entirely, and creates cmake targets for them. Then you can just use `target_link_libraries` to link to the library or `$<TARGET_FILE>` to refer to the executable path for the daemon.

crates/aranya-client-capi/ctest/README.md
jdygert-spok: It's removed on the base PR

crates/aranya-client-capi/ctest/README.md
jdygert-spok: Yes, that's how it's set up on main and after this PR.

gknopf-aranya: Should some of the info in this README.md be moved up to the root README.md then?

jdygert-spok: I'd rather not clutter the main README. I don't think instructions on running our C test suite are relevant to most users.

jnewell-spok (approved):
Cleans it up quite a bit. looks good

PR #691, opened 2026-01-16

jdygert-spok:
- Use renamed policy engine and rename `E` to `PS`, `CE`, `Eff`
- Send return address over first stream on a connection to account for removal from sync protocol
- Check new graph ID in sync protocol against active graph ID

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,3 +85,32 @@ jobs:
           name: ${{ matrix.artifact }}-${{ matrix.features }}
           path: aranya-client-lib.tar.gz
           if-no-files-found: error
+
+  build-certgen:
+    strategy:
+      matrix:
+        runner: ["ubuntu2404-arm64-4core", "ubuntu-latest"]
+        include:
+          - runner: "ubuntu2404-arm64-4core"
+            artifact: "aranya-certgen-aarch64-unknown-linux-gnu"
+          - runner: "ubuntu-latest"
+            artifact: "aranya-certgen-x86_64-unknown-linux-gnu"
+    name: build certgen ${{ matrix.artifact }}
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Environment
+        uses: ./.github/actions/setup
+
+      - name: Build aranya-certgen
+        run: |
+          cargo make build-certgen
+          cargo make tgz-certgen
+
+      - name: Publish aranya-certgen
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.artifact }}
+          path: aranya-certgen.tar.gz
+          if-no-files-found: error
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -244,6 +244,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "aranya-certgen"
+version = "4.1.0"
+dependencies = [
+ "clap",
+ "rcgen",
+ "tempfile",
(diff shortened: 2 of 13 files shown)


Cargo.toml
djohnson-spideroak, line 1: Did you want to do a core release to avoid the patches here?

jdygert-spok: Yes, we will need to.

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -314,7 +311,18 @@ where
             SyncType::Hello(_hello_msg) => {
                 #[cfg(feature = "preview")]
                 {
-                    Self::process_hello_message(_hello_msg, client, &active_team, _handle).await;
+                    if let Err(error) = Self::process_hello_message(
+                        _hello_msg,
```
mooreoak, line 315: Why do we have "\_" vars being used as arguments? I thought the convention is that symbols that start with "\_" are unused.

djohnson-spideroak: Because of the feature flag above, without preview enabled the variable is unused and would cause lints.

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -113,7 +112,8 @@ where
         // TODO: spawn a task for send/recv?
         let (mut recv, mut send) = stream.split();
 
-        let mut sync_requester = SyncRequester::new(peer.graph_id, &mut Rng, syncer.server_addr);
+        let mut sync_requester = SyncRequester::new(peer.graph_id, &mut Rng);
+        // TODO: send syncer.server_addr
```
djohnson-spideroak: Will this be done in this PR or a future one?

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -246,6 +241,10 @@ where
                     .connect(Connect::new(addr).with_server_name(addr.ip().to_string()))
                     .await?;
                 conn.keep_alive(true)?;
+                conn.open_send_stream()
```
djohnson-spideroak, line 243: does this send stream need to be closed? or is it implicit in the drop?

crates/aranya-daemon/src/sync/transport/quic/client.rs
jdygert-spok: Oops, left that in. It is already sent in `fn connect` below

crates/aranya-daemon/src/sync/transport/quic/client.rs
jdygert-spok: It'll close on drop.

nikki-aranya: We close the send streams after we send all the data we're going to send in each exchange which sends a FIN flag, plus finish() is implicit in the drop yes, I'm planning to make this a lot easier to reason about in my next PR

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -314,7 +311,18 @@ where
...
+                        client,
+                        &active_team,
+                        _handle,
+                        peer_server_addr,
+                    )
+                    .await
+                    {
+                        // TODO: Respond with error or don't respond?
```
gknopf-aranya, line 323: Since hello messages are best-effort or fire-and-forget, it doesn't seem like we need to return errors.

PR #693, opened 2026-01-21
Handle removal of return address in sync protocol

jdygert-spok:
See aranya-project/aranya-core#541.

### DIFF
--- a/crates/aranya-daemon/src/sync/transport/quic/server.rs
+++ b/crates/aranya-daemon/src/sync/transport/quic/server.rs
@@ -182,12 +182,17 @@ where
                 .server_keys
                 .get_team_for_identity(&identity)
                 .context("no active team for accepted connection")?;
-            let peer = conn
-                .remote_addr()
-                .context("unable to get peer address from connection")?;
+            let peer_addr: Addr = {
+                let mut recv = conn
+                    .accept_receive_stream()
+                    .await?
+                    .context("no stream for peer address")?;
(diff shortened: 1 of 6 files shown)


crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -182,12 +182,17 @@ where
...
                 .context("no active team for accepted connection")?;
-            let peer = conn
-                .remote_addr()
-                .context("unable to get peer address from connection")?;
+            let peer_addr: Addr = {
+                let mut recv = conn
+                    .accept_receive_stream()
+                    .await?
+                    .context("no stream for peer address")?;
+                let bytes = recv.receive().await?.context("no peer address sent")?;
```
gknopf-aranya, line 190: I believe the peer address can be obtained from the QUIC connection rather than requiring peers to send it over the stream.

jdygert-spok: No, we want the peer's server address, not their client address.

gknopf-aranya: Wonder if this would be better to send via the `ConnectionContext` rather than in each stream.
https://docs.rs/s2n-quic/latest/s2n_quic/provider/event/trait.Subscriber.html#associatedtype.ConnectionContext

jdygert-spok: I think that context is local-only. Also I only send it once per connection.

Issue #491, opened 2025-09-22
aranya-daemon: decide how to handle feature flagged fields in config


ssweetney-spok:
Should we continue to gate those fields behind features or allow them in the config unconditionally?

Currently this would be `afc` in `aranya_daemon::config::Config`.

See [this thread](https://github.com/aranya-project/aranya/pull/463#discussion_r2364546390)

PR #695, opened 2026-01-22
daemon: fix config toggles

jdygert-spok:


crates/aranya-daemon/src/config/toggle.rs
```diff
@@ -117,3 +117,74 @@ mod imp {
...
+        // TODO(jdygert): Improve error.
+        assert_eq!(
+            err,
+            serde::de::Error::custom("data did not match any variant of untagged enum E")
+        );
+    }
+
+    #[test]
+    fn test_enabled_with_unknown_field() {
+        // TODO(jdygert): Deny unknown fields?
```
gknopf-aranya, line 164: Or warn

djohnson-spideroak: I think deny is probably best

crates/aranya-daemon/src/config/toggle.rs
```diff
@@ -63,7 +63,7 @@ mod imp {
     pub struct False;
     impl<'de> Deserialize<'de> for False {
         fn deserialize<D: de::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-            if bool::deserialize(deserializer)? {
+            if !bool::deserialize(deserializer)? {
```
djohnson-spideroak: Can you elaborate on the change here?

jdygert-spok: We want to deserialize only `false` successfully, so we deserialize a bool and then check if it is false. I think `== true` and `== false` would be a bit more clear here, although it's usually bad style and will fire a lint...

crates/aranya-daemon/src/config/toggle.rs
jdygert-spok: We need to decide generally how we want to handle unknown fields in the configs. They can cause pains with feature flags and possibly with upgrades. See #491.

crates/aranya-daemon/src/config/toggle.rs
djohnson-spideroak: imo this is a justified place to ignore that lint and have `== false` and a comment.

jdygert-spok: Switched to `match` to be more explicit.

PR #698, opened 2026-01-22
chore: update to Rust 1.90

jdygert-spok:


crates/aranya-client/src/client/device.rs
```diff
@@ -64,12 +64,6 @@ pub struct IterDevices<'a>(slice::Iter<'a, DeviceId>);
 
 impl_slice_iter_wrapper!(IterDevices<'a> for DeviceId);
 
-/// An owning iterator over [`DeviceId`]s.
-#[derive(Clone, Debug)]
-pub struct IntoIterDevices(vec::IntoIter<DeviceId>);
-
-impl_vec_into_iter_wrapper!(IntoIterDevices for DeviceId);
```
djohnson-spideroak, line 71: Any info why this is being removed?

jdygert-spok: It's unused and not exported.

PR #699, opened 2026-01-26
sync: fix hello sync flake and implement builder pattern

### LINKED ISSUE #700: hello sync: init subscriptions with config instead of duration params
Passing in a config to initialize sync hello subscriptions is more self-documenting than passing in a list of `Duration` parameters. For example, the following code shows clearly what each parameter means:
```
      1276 -            Duration::from_secs(60),   // graph_change_delay                                                             
      1277 -            Duration::from_secs(120),  // duration                                                                       
      1278 -            Duration::from_millis(10), // schedule_delay - low                                                           
      1276 +            HelloSubscriptionConfig::builder()                                                                           
      1277 +                .graph_change_delay(Duration::from_secs(60))                                                             
      1278 +                .expiration(Duration::from_secs(120))                                                                    
      1279 +                .periodic_interval(Duration::from_millis(10))                                                            
      1280 +                .build()?, 
```

Use a builder pattern for the config type so it can be modified in a backward compatible way.

https://github.com/aranya-project/aranya/pull/699#discussion_r2729012263

cc: @djohnson-spideroak, @benz-spideroak, @jdygert-spok

gknopf-aranya:
Fixes test flakes in test_hello_subscription caused by subscriptions expiring before tests completed on slower machines running parallel tests.                                                                                                                    

Replaces individual duration parameters in sync_hello_subscribe with a HelloSubscriptionConfig struct that provides:               

- Builder pattern for configuration                                                                                                
- Sensible defaults (100ms graph change delay, 10s periodic interval, 1 year expiration)                                           
- Clearer parameter names (periodic_interval instead of schedule_delay)                                                            

The long default expiration (1 year) prevents the subscription timeout flakes, while callers can still customize via the builder.

Closes #700

### DIFF
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -1644,45 +1644,6 @@ AranyaError aranya_hello_subscription_config_builder_cleanup(struct AranyaHelloS
 
 #if defined(ENABLE_ARANYA_PREVIEW)
 /**
- * Returns a [`AranyaHelloSubscriptionConfig`](@ref AranyaHelloSubscriptionConfig) with sensible defaults.
- *
- * This is a convenience function for C API consumers who want defaults
- * without using the full builder pattern.
- *
- * Defaults:
- * - graph_change_delay: 100ms
- * - expiration: 1 year
- * - periodic_interval: 10 seconds
- *
- * @param[out] out a pointer to write the config to
- *
- * @relates AranyaHelloSubscriptionConfig.
- */
-AranyaError aranya_hello_subscription_config_default(struct AranyaHelloSubscriptionConfig *out);
-#endif
-
-#if defined(ENABLE_ARANYA_PREVIEW)
-/**
- * Returns a [`AranyaHelloSubscriptionConfig`](@ref AranyaHelloSubscriptionConfig) with sensible defaults.
- *
- * This is a convenience function for C API consumers who want defaults
- * without using the full builder pattern.
- *
- * Defaults:
- * - graph_change_delay: 100ms
- * - expiration: 1 year
- * - periodic_interval: 10 seconds
- *
- * @param[out] out a pointer to write the config to
- *
- * @relates AranyaHelloSubscriptionConfig.
- */
-AranyaError aranya_hello_subscription_config_default_ext(struct AranyaHelloSubscriptionConfig *out,
-                                                         struct AranyaExtError *__ext_err);
-#endif
-
-#if defined(ENABLE_ARANYA_PREVIEW)
-/**
  * Attempts to build a [`AranyaHelloSubscriptionConfig`](@ref AranyaHelloSubscriptionConfig).
  *
  * This function consumes and releases any resources associated
@@ -1716,33 +1677,35 @@ AranyaError aranya_hello_subscription_config_build_ext(struct AranyaHelloSubscri
 
 #if defined(ENABLE_ARANYA_PREVIEW)
 /**
- * Sets the minimum delay after a graph change before sending a hello notification.
+ * Sets the debounce interval for hello notifications after graph changes.
  *
- * This helps batch rapid changes into fewer notifications. Default is 100ms.
+ * After sending a hello notification, no further notification will be sent
+ * for graph changes within this interval. Default is 100ms.
  *
  * @param[in,out] cfg a pointer to the builder for a hello subscription config
- * @param[in] delay the minimum delay after a graph change
+ * @param[in] duration the debounce interval
  *
  * @relates AranyaHelloSubscriptionConfigBuilder.
  */
-AranyaError aranya_hello_subscription_config_builder_set_graph_change_delay(struct AranyaHelloSubscriptionConfigBuilder *cfg,
-                                                                            AranyaDuration delay);
+AranyaError aranya_hello_subscription_config_builder_set_graph_change_debounce(struct AranyaHelloSubscriptionConfigBuilder *cfg,
+                                                                               AranyaDuration duration);
 #endif
 
 #if defined(ENABLE_ARANYA_PREVIEW)
 /**
- * Sets the minimum delay after a graph change before sending a hello notification.
+ * Sets the debounce interval for hello notifications after graph changes.
  *
- * This helps batch rapid changes into fewer notifications. Default is 100ms.
+ * After sending a hello notification, no further notification will be sent
+ * for graph changes within this interval. Default is 100ms.
  *
  * @param[in,out] cfg a pointer to the builder for a hello subscription config
- * @param[in] delay the minimum delay after a graph change
+ * @param[in] duration the debounce interval
  *
  * @relates AranyaHelloSubscriptionConfigBuilder.
  */
-AranyaError aranya_hello_subscription_config_builder_set_graph_change_delay_ext(struct AranyaHelloSubscriptionConfigBuilder *cfg,
-                                                                                AranyaDuration delay,
-                                                                                struct AranyaExtError *__ext_err);
+AranyaError aranya_hello_subscription_config_builder_set_graph_change_debounce_ext(struct AranyaHelloSubscriptionConfigBuilder *cfg,
+                                                                                   AranyaDuration duration,
+                                                                                   struct AranyaExtError *__ext_err);
 #endif
 
 #if defined(ENABLE_ARANYA_PREVIEW)
@@ -2897,7 +2860,7 @@ AranyaError aranya_remove_sync_peer_ext(const struct AranyaClient *client,
  * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
  * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
  * @param[in] peer the peer's Aranya network address [`AranyaAddr`](@ref AranyaAddr).
- * @param[in] config the hello subscription configuration [`AranyaHelloSubscriptionConfig`](@ref AranyaHelloSubscriptionConfig).
+ * @param[in] config the hello subscription configuration [`AranyaHelloSubscriptionConfig`](@ref AranyaHelloSubscriptionConfig), or null to use defaults.
  *
  * @relates AranyaClient.
  */
@@ -2916,7 +2879,7 @@ AranyaError aranya_sync_hello_subscribe(const struct AranyaClient *client,
  * @param[in] client the Aranya Client [`AranyaClient`](@ref AranyaClient).
  * @param[in] team the team's ID [`AranyaTeamId`](@ref AranyaTeamId).
  * @param[in] peer the peer's Aranya network address [`AranyaAddr`](@ref AranyaAddr).
- * @param[in] config the hello subscription configuration [`AranyaHelloSubscriptionConfig`](@ref AranyaHelloSubscriptionConfig).
+ * @param[in] config the hello subscription configuration [`AranyaHelloSubscriptionConfig`](@ref AranyaHelloSubscriptionConfig), or null to use defaults.
  *
  * @relates AranyaClient.
  */
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1120,28 +1120,6 @@ pub type HelloSubscriptionConfig = Safe<imp::HelloSubscriptionConfig>;
 #[aranya_capi_core::opaque(size = 40, align = 8)]
 pub type HelloSubscriptionConfigBuilder = Safe<imp::HelloSubscriptionConfigBuilder>;
 
-/// Returns a [`HelloSubscriptionConfig`] with sensible defaults.
-///
-/// This is a convenience function for C API consumers who want defaults
-/// without using the full builder pattern.
-///
-/// Defaults:
-/// - graph_change_delay: 100ms
-/// - expiration: 1 year
-/// - periodic_interval: 10 seconds
-///
-/// @param[out] out a pointer to write the config to
-///
-/// @relates AranyaHelloSubscriptionConfig.
-#[cfg(feature = "preview")]
-pub fn hello_subscription_config_default(
-    out: &mut MaybeUninit<HelloSubscriptionConfig>,
-) -> Result<(), imp::Error> {
-    let builder = imp::HelloSubscriptionConfigBuilder::default();
-    // SAFETY: No special considerations.
-    unsafe { builder.build(out) }
-}
-
 /// Attempts to build a [`HelloSubscriptionConfig`].
 ///
 /// This function consumes and releases any resources associated
@@ -1161,20 +1139,21 @@ pub fn hello_subscription_config_build(
     Ok(())
 }
 
-/// Sets the minimum delay after a graph change before sending a hello notification.
+/// Sets the debounce interval for hello notifications after graph changes.
 ///
-/// This helps batch rapid changes into fewer notifications. Default is 100ms.
+/// After sending a hello notification, no further notification will be sent
+/// for graph changes within this interval. Default is 100ms.
 ///
 /// @param[in,out] cfg a pointer to the builder for a hello subscription config
-/// @param[in] delay the minimum delay after a graph change
+/// @param[in] duration the debounce interval
 ///
 /// @relates AranyaHelloSubscriptionConfigBuilder.
 #[cfg(feature = "preview")]
-pub fn hello_subscription_config_builder_set_graph_change_delay(
+pub fn hello_subscription_config_builder_set_graph_change_debounce(
     cfg: &mut HelloSubscriptionConfigBuilder,
-    delay: Duration,
+    duration: Duration,
 ) {
-    cfg.graph_change_delay(delay);
+    cfg.graph_change_debounce(duration);
 }
 
 /// Sets how long the subscription remains active before expiring.
@@ -1986,7 +1965,7 @@ pub unsafe fn remove_sync_peer(
 /// @param[in] client the Aranya Client [`Client`].
 /// @param[in] team the team's ID [`TeamId`].
 /// @param[in] peer the peer's Aranya network address [`Addr`].
-/// @param[in] config the hello subscription configuration [`HelloSubscriptionConfig`].
+/// @param[in] config the hello subscription configuration [`HelloSubscriptionConfig`], or null to use defaults.
 ///
 /// @relates AranyaClient.
 #[cfg(feature = "preview")]
@@ -1994,15 +1973,19 @@ pub unsafe fn sync_hello_subscribe(
     client: &Client,
     team: &TeamId,
     peer: Addr,
-    config: &HelloSubscriptionConfig,
+    config: Option<&HelloSubscriptionConfig>,
 ) -> Result<(), imp::Error> {
     // SAFETY: Caller must ensure `addr` is a valid C String.
     let addr = unsafe { peer.as_underlying() }?;
+    let config: aranya_client::HelloSubscriptionConfig = match config {
+        Some(cfg) => (*cfg).clone().into(),
+        None => aranya_client::HelloSubscriptionConfig::default(),
+    };
     client.rt.block_on(
         client
             .inner
             .team(team.into())
-            .sync_hello_subscribe(addr, (*config).clone().into()),
+            .sync_hello_subscribe(addr, config),
     )?;
     Ok(())
 }
--- a/crates/aranya-client-capi/src/imp/config.rs
+++ b/crates/aranya-client-capi/src/imp/config.rs
@@ -159,8 +159,8 @@ impl Default for SyncPeerConfigBuilder {
 #[cfg(feature = "preview")]
 #[derive(Clone, Debug)]
 pub struct HelloSubscriptionConfig {
-    /// Minimum delay after a graph change before sending a hello notification.
-    pub graph_change_delay: Duration,
+    /// Debounce interval for hello notifications after sending one.
+    pub graph_change_debounce: Duration,
     /// How long the subscription remains active before expiring.
     pub expiration: Duration,
     /// Interval between periodic hello messages.
@@ -171,7 +171,7 @@ pub struct HelloSubscriptionConfig {
 impl From<aranya_client::HelloSubscriptionConfig> for HelloSubscriptionConfig {
     fn from(value: aranya_client::HelloSubscriptionConfig) -> Self {
         Self {
-            graph_change_delay: value.graph_change_delay().into(),
+            graph_change_debounce: value.graph_change_debounce().into(),
             expiration: value.expiration().into(),
             periodic_interval: value.periodic_interval().into(),
         }
@@ -182,7 +182,7 @@ impl From<aranya_client::HelloSubscriptionConfig> for HelloSubscriptionConfig {
 impl From<HelloSubscriptionConfig> for aranya_client::HelloSubscriptionConfig {
     fn from(value: HelloSubscriptionConfig) -> Self {
         aranya_client::HelloSubscriptionConfig::builder()
-            .graph_change_delay(value.graph_change_delay.into())
+            .graph_change_debounce(value.graph_change_debounce.into())
             .expiration(value.expiration.into())
             .periodic_interval(value.periodic_interval.into())
             .build()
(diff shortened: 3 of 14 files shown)


crates/aranya-client/tests/tests.rs
```diff
@@ -996,9 +996,9 @@ async fn test_hello_subscription() -> Result<()> {
         .sync_hello_subscribe(
             admin_addr,
             Duration::from_millis(100),
-            Duration::from_millis(1000),
-            Duration::from_secs(60), // Schedule delay for periodic sends
-        ) // Short delay for faster testing
+            Duration::from_secs(120), // Duration long enough for test to complete
```
gknopf-aranya: It's difficult to tell what each of these Durations are without looking up the definition of `sync_hello_subscribe()`. Seems like it would be clearer if we passed in a config struct with named fields. That way it's clear that this is setting the sync hello subscription expiration time.

gknopf-aranya: Planning to open a follow-up PR for this since it's out of scope for the flake fix.

gknopf-aranya: https://github.com/aranya-project/aranya/issues/700

crates/aranya-client/src/client/team.rs
```diff
@@ -174,19 +169,17 @@ impl Team<'_> {
...
         self.client
             .daemon
             .sync_hello_subscribe(
                 context::current(),
                 peer,
                 self.id,
-                graph_change_delay,
-                duration,
-                schedule_delay,
+                config.graph_change_delay,
```
gknopf-aranya: Eventually, we could pass the config type directly into the daemon IPC and internal daemon implementation. However, the goal of this PR is to improve clarity/usability of the application API.

crates/aranya-client/tests/tests.rs
```diff
@@ -995,10 +995,10 @@ async fn test_hello_subscription() -> Result<()> {
     membera_team
         .sync_hello_subscribe(
             admin_addr,
-            Duration::from_millis(100),
-            Duration::from_millis(1000),
-            Duration::from_secs(60), // Schedule delay for periodic sends
-        ) // Short delay for faster testing
+            HelloSubscriptionConfig::builder()
+                .graph_change_delay(Duration::from_millis(100))
```
gknopf-aranya: What happened to the rest of the durations that were configured here previously?

gknopf-aranya: Oh, the defaults were being used. It would be more clear to make them explicit in tests.

.github/workflows/correctness.yml
```diff
@@ -32,10 +32,9 @@ jobs:
 
       - name: Install newer clang-format
         run: |
-          sudo apt-get install -y lsb-release wget software-properties-common gnupg
-          sudo bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
-          sudo apt-get install -y clang-format-20
-          sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-20 200
+          sudo apt-get update
```
gknopf-aranya: Revert

examples/c/example.c
```diff
@@ -260,7 +260,7 @@ AranyaError init_team(Team *t) {
 
     Client *owner = &t->clients.owner;
     Client *admin = &t->clients.admin;
-    Client *operator = &t->clients.operator;
+    Client *operator= & t->clients.operator;
```
gknopf-aranya: Revert

examples/rust/aranya-example/src/main.rs
```diff
@@ -293,17 +293,28 @@ async fn main() -> Result<()> {
...
         info!("admin subscribing to hello notifications from owner");
-        let duration = Duration::from_secs(30);
-        let graph_change_delay = Duration::from_millis(1000);
-        let schedule_delay = Duration::from_secs(5);
         admin_team
-            .sync_hello_subscribe(owner_addr, graph_change_delay * 2, duration, schedule_delay)
+            .sync_hello_subscribe(
+                owner_addr,
+                HelloSubscriptionConfig::builder()
+                    .graph_change_delay(Duration::from_millis(2000))
```
gknopf-aranya: Why was graph change delay increased?

examples/c/example.c
```diff
@@ -719,19 +719,38 @@ AranyaError run(Team *t) {
...
     printf("admin subscribing to hello notifications from owner\n");
-    AranyaDuration graph_change_delay = ARANYA_DURATION_SECONDS * 1ULL;
-    AranyaDuration duration = ARANYA_DURATION_SECONDS * 30ULL;
-    AranyaDuration schedule_delay = ARANYA_DURATION_SECONDS * 5ULL;
+    AranyaHelloSubscriptionConfigBuilder hello_cfg_builder;
+    AranyaHelloSubscriptionConfig hello_cfg;
+    err = aranya_hello_subscription_config_builder_init(&hello_cfg_builder);
+    EXPECT("error initializing hello subscription config builder", err);
+    aranya_hello_subscription_config_builder_set_graph_change_delay(
+        &hello_cfg_builder, ARANYA_DURATION_SECONDS * 2ULL);
```
gknopf-aranya: Why was graph change delay increased?

examples/c/example.c
```diff
@@ -719,19 +719,38 @@ AranyaError run(Team *t) {
...
-                                      schedule_delay);
+                                      &hello_cfg);
     EXPECT("error subscribing admin to owner hello notifications", err);
 
     // Operator subscribes to hello notifications from Admin with 1-second delay
     printf("operator subscribing to hello notifications from admin\n");
+    err = aranya_hello_subscription_config_builder_init(&hello_cfg_builder);
+    EXPECT("error initializing hello subscription config builder", err);
+    aranya_hello_subscription_config_builder_set_graph_change_delay(
+        &hello_cfg_builder, ARANYA_DURATION_SECONDS * 1ULL);
```
gknopf-aranya: Graph change delay used to be `graph_change_delay * 2` which would be `2ULL`.

examples/rust/aranya-example/src/main.rs
gknopf-aranya: Oh, this was `graph_change_delay * 2` so it's correct.

crates/aranya-client/src/client/team.rs
gknopf-aranya: Added a TODO comment for this.

crates/aranya-client/src/client/team.rs
```diff
@@ -174,19 +169,19 @@ impl Team<'_> {
     pub async fn sync_hello_subscribe(
         &self,
         peer: Addr,
-        graph_change_delay: Duration,
-        duration: Duration,
-        schedule_delay: Duration,
+        config: crate::config::HelloSubscriptionConfig,
```
ssweetney-spok, line 172: Shouldn't this go in a new method like `sync_hello_subscribe_with_config` in order to maintain backwards compatibility?

crates/aranya-client/src/config.rs
```diff
@@ -119,3 +119,100 @@ impl Default for SyncPeerConfigBuilder {
...
+
+/// Configuration for hello subscription.
+///
+/// This configures how a device subscribes to hello notifications from a sync peer.
+/// Hello notifications inform subscribers when the peer's graph has changed,
+/// enabling reactive syncing.
+#[cfg(feature = "preview")]
+#[cfg_attr(docsrs, doc(cfg(feature = "preview")))]
+#[derive(Clone, Debug)]
+pub struct HelloSubscriptionConfig {
```
ssweetney-spok, line 131: If these fields are gonna be public then this type should be marked with `#[non_exhaustive]` or add a private field so that adding new fields won't break code.

The other config types we define, like `aranya_client::config::SyncPeerConfig`, have don't have public fields. Why is this one different?

crates/aranya-client/src/client/team.rs
gknopf-aranya: Sync hello is still behind `preview`, so we can break backward compatibility.

crates/aranya-client/src/config.rs
gknopf-aranya: Good point. This should be implemented in a similar way as the other public types so future changes can be made in a backward compatible way. Even though we're allowed to break the hello sync API for now since it's behind the `preview` flag, we want to be mindful about designing in the API in a way where future changes can be made in a backward compatible way. Using the builder pattern for the config gets us closer to that goal by not hard-coding subscription params in a function.

gknopf-aranya: I've updated this to private fields with getter methods so the type cannot be externally constructed. `HelloSubscriptionConfig` is a config type like `SyncPeerConfig`, so private fields without #[non_exhaustive] is consistent with the existing pattern. The current change follows the established convention.

examples/c/example.c
```diff
@@ -719,19 +719,38 @@ AranyaError run(Team *t) {
...
+    EXPECT("error initializing hello subscription config builder", err);
+    aranya_hello_subscription_config_builder_set_graph_change_delay(
+        &hello_cfg_builder, ARANYA_DURATION_SECONDS * 2ULL);
+    aranya_hello_subscription_config_builder_set_expiration(
+        &hello_cfg_builder, ARANYA_DURATION_SECONDS * 30ULL);
+    aranya_hello_subscription_config_builder_set_periodic_interval(
+        &hello_cfg_builder, ARANYA_DURATION_SECONDS * 5ULL);
+    err =
+        aranya_hello_subscription_config_build(&hello_cfg_builder, &hello_cfg);
+    EXPECT("error building hello subscription config", err);
```
jdygert-spok: This is so verbose, and we aren't even handling the errors from the setters :/

crates/aranya-client/src/config.rs
```diff
@@ -119,3 +119,111 @@ impl Default for SyncPeerConfigBuilder {
...
+    }
+
+    /// Sets how long the subscription remains active before expiring.
+    ///
+    /// After this duration, the subscription will automatically end and no more
+    /// hello notifications will be sent. Default is 120 seconds.
+    pub fn expiration(mut self, duration: Duration) -> Self {
+        self.expiration = duration;
+        self
+    }
```
jdygert-spok, line 216: Are there even many situations you want your subscription to time out? This should probably default to none/MAX.

examples/c/example.c
gknopf-aranya: The example could use a `_default()` method to set up the config. If a user wants to configure something other than the defaults, they can set those values case-by-case. That would help to shorten this example and keep it focused on setting up the subscription.

crates/aranya-client/src/config.rs
gknopf-aranya: That is a question that was discussed during the initial hello sync design. We intentionally made the sync subscription time a required parameter to ensure subscriptions eventually expire. The goal of this PR isn't to change that behavior but to switch from fixed parameters to a builder pattern for the sync hello config.

gknopf-aranya: MAX seems like a decent default given that shorter times are arbitrary.

.clang-format
```diff
@@ -1,4 +1,8 @@
 ---
+Language: Cpp
```
gknopf-aranya: This resolves the issue where `operator` is considered a C++ keyword when formatting the example.c file:
```
  + xargs -0 -- clang-format --Werror --dry-run                                                                                      
  examples/c/example.c:263:21: error: code should be clang-formatted [-Wclang-format-violations]                                     
  Client *operator= & t->clients.operator;                                                                                           
  ^   
```

jdygert-spok: I don't get what you mean. That error is correctly complaining that the code is not formatted right. If I run clang-format on it with the original .clang-format config it formats it as `Client *operator = &t->client.operator;` as it should.

crates/aranya-client-capi/src/imp/config.rs
```diff
@@ -154,3 +154,110 @@ impl Default for SyncPeerConfigBuilder {
...
+
+#[cfg(feature = "preview")]
+impl From<HelloSubscriptionConfig> for aranya_client::HelloSubscriptionConfig {
+    fn from(value: HelloSubscriptionConfig) -> Self {
+        aranya_client::HelloSubscriptionConfig::builder()
+            .graph_change_delay(value.graph_change_delay.into())
+            .expiration(value.expiration.into())
+            .periodic_interval(value.periodic_interval.into())
+            .build()
+            .expect("All values are set")
```
gknopf-aranya, line 189: Shouldn't panic here

.clang-format
gknopf-aranya: When I ran the formatting it didn't match what is in CI until after I made this change.

crates/aranya-client/tests/tests.rs
gknopf-aranya, line 1: Tests can be simplified to use defaults except for configurations where specific values matter. E.g. default 1yr subscription that doesn't expire.

.clang-format
jdygert-spok: This change shouldn't have any effect. With no language specified, the config applies to all languages.

gknopf-aranya: It does for me. I haven't been able to get my local dev machine or sp1 to format the same way as CI for some time. This finally worked.

crates/aranya-client-capi/src/imp/config.rs
gknopf-aranya: It's infallible and matches our existing conventions for builder patterns.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Updated tests to use defaults when possible. Only sets specific values where needed.

crates/aranya-client/tests/tests.rs
```diff
@@ -1168,9 +1149,11 @@ async fn test_hello_subscription_schedule_delay() -> Result<()> {
     membera_team
         .sync_hello_subscribe(
             admin_addr,
-            Duration::from_secs(60),  // graph_change_delay
-            Duration::from_secs(120), // duration
-            Duration::from_secs(60),  // schedule_delay - high
+            HelloSubscriptionConfig::builder()
+                .graph_change_delay(Duration::from_secs(60))
+                .expiration(Duration::from_secs(120))
```
gknopf-aranya: Use default

examples/rust/aranya-example/src/main.rs
```diff
@@ -293,17 +293,28 @@ async fn main() -> Result<()> {
 
         // Admin subscribes to hello notifications from Owner with 2-second delay
         info!("admin subscribing to hello notifications from owner");
-        let duration = Duration::from_secs(30);
-        let graph_change_delay = Duration::from_millis(1000);
-        let schedule_delay = Duration::from_secs(5);
         admin_team
-            .sync_hello_subscribe(owner_addr, graph_change_delay * 2, duration, schedule_delay)
+            .sync_hello_subscribe(
```
gknopf-aranya: Why not use defaults here?

gknopf-aranya: Updated to use defaults

.clang-format
jdygert-spok: Weird. What's your version? Mine:

```
> which -a clang-format
/opt/homebrew/bin/clang-format
> clang-format --version
clang-format version 21.1.7
```

gknopf-aranya: A little older than yours:
```
gknopf@sp1:~/scm/aranya$ clang-format --version
Ubuntu clang-format version 20.1.8 (++20250708083436+6fb913d3e2ec-1~exp1~20250708203453.133)
```
```
gknopf@geoffreys-macbook-pro aranya % clang-format --version
clang-format version 18.1.6
```

jdygert-spok: clang-format-18 errors out with this change since it doesn't support `C` for the language field.
clang-format-20 seems to format it just fine to me. IDK what's making our tools behave differently.

gknopf-aranya: idk, but I'm tired of the fmt being different and this finally helped.

gknopf-aranya: Except now the code analysis is failing in CI...

jdygert-spok: The code QL stuff is unrelated, I had fiddled with that and thought it was working but now it's trying to autobuild again even though the default is supposed to be the mode that doesn't build at all...

jdygert-spok: Well, it was working in a previous run on this PR and I haven't touched it since then :thinking:

jdygert-spok: I guess it never switched off from autobuild, but the autobuild is just flaky :/

jdygert-spok: Turned the C CodeQL back off since it's just causing issues for little gain.

jdygert-spok: I only get formatting issues with clang-format-18, which there is just no way to avoid AFAICT. Does it not format for you correctly with clang-format-20 on main?

gknopf-aranya: Don't know. I just upgraded to the latest which is 21.

gknopf-aranya: Same issue after upgrading:
```
gknopf@geoffreys-macbook-pro aranya % cargo make fmt
gknopf@geoffreys-macbook-pro aranya % git diff
diff --git a/examples/c/example.c b/examples/c/example.c
index d6bdcba7..05b7e54e 100644
--- a/examples/c/example.c
+++ b/examples/c/example.c
@@ -260,7 +260,7 @@ AranyaError init_team(Team *t) {
 
     Client *owner = &t->clients.owner;
     Client *admin = &t->clients.admin;
-    Client *operator = &t->clients.operator;
+    Client *operator= & t->clients.operator;
     Client *membera = &t->clients.membera;
     Client *memberb = &t->clients.memberb;
 
@@ -443,8 +443,9 @@ AranyaError init_team(Team *t) {
     }
 
     // add operator to team.
-    err = aranya_add_device_to_team(&owner->client, &t->id, operator->pk,
-                                    operator->pk_len, &operator_role_id);
+    err = aranya_add_device_to_team(&owner->client,
+                                    &t->id, operator->pk, operator->pk_len,
+                                    &operator_role_id);
         fprintf(stderr, "unable to add operator to team\n");
         return err;
@@ -671,7 +672,7 @@ AranyaError run(Team *t) {
     AranyaDeviceId *devices = NULL;
 
     Client *admin = &t->clients.admin;
-    Client *operator = &t->clients.operator;
+    Client *operator= & t->clients.operator;
     Client *memberb = &t->clients.memberb;
 
     // initialize logging.
```

gknopf-aranya: Then I checkout this branch and get:
```
[cargo-make] INFO - Running Task: fmt-c
/Users/gknopf/scm/aranya/.clang-format:6:11: error: unknown enumerated scalar
Language: C
```

gknopf-aranya: It doesn't look like `Language: C` is valid:
```
Fixed. The issue was that Language: C is not a valid clang-format option - C code is handled by Language: Cpp. I removed the       
  redundant section since both had identical settings.

❯ Will this work for clang-format 20 as well?                                                                                        
                                                                                                                                     
⏺ Yes. Language: Cpp has always been the standard way to format C code in clang-format - it covers C, C++, Objective-C, and          
  Objective-C++. This hasn't changed and there's no indication it will. Language: C has never been a valid option in any clang-format
   version.   
```

jdygert-spok: I had to look this up when I set up clang-format initially for CI. `Language: C` was added in clang-format-20. Compare the documentation for the two versions. This means you must still be running an older version somehow?

https://releases.llvm.org/20.1.0/tools/clang/docs/ClangFormatStyleOptions.html#language
https://releases.llvm.org/18.1.1/tools/clang/docs/ClangFormatStyleOptions.html#language

gknopf-aranya: The problem is happening because `operator` is a reserved C++ variable. I think we should rename the variable if we want to have consistent formatting and not use a reserved word.

.github/workflows/correctness.yml
```diff
@@ -34,8 +34,8 @@ jobs:
         run: |
           sudo apt-get install -y lsb-release wget software-properties-common gnupg
           sudo bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
-          sudo apt-get install -y clang-format-20
-          sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-20 200
+          sudo apt-get install -y clang-format-21
```
gknopf-aranya: Revert this

.clang-format
gknopf-aranya: It was using the old version even though I installed the new one. symlinks and paths needed to be updated. Appears to work now.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1875,30 +1986,24 @@ pub unsafe fn remove_sync_peer(
...
 /// @relates AranyaClient.
 #[cfg(feature = "preview")]
 pub unsafe fn sync_hello_subscribe(
     client: &Client,
     team: &TeamId,
     peer: Addr,
-    graph_change_delay: Duration,
-    duration: Duration,
-    schedule_delay: Duration,
+    config: &HelloSubscriptionConfig,
```
jdygert-spok: If we expect using the defaults to be reasonable, let's make this optional so you can pass null to use the defaults.

crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -1589,6 +1619,196 @@ AranyaError aranya_sync_peer_config_builder_set_sync_on_hello_ext(struct AranyaS
...
+ * - graph_change_delay: 100ms
+ * - expiration: 1 year
+ * - periodic_interval: 10 seconds
+ *
+ * @param[out] out a pointer to write the config to
+ *
+ * @relates AranyaHelloSubscriptionConfig.
+ */
+AranyaError aranya_hello_subscription_config_default(struct AranyaHelloSubscriptionConfig *out);
+#endif
```
jdygert-spok: We can remove this convenience function if we make the param optional.

crates/aranya-client/src/client/team.rs
```diff
@@ -174,19 +169,19 @@ impl Team<'_> {
...
     ) -> Result<()> {
+        // TODO: Pass the config type directly into the daemon IPC and internal
+        // daemon implementation instead of extracting individual fields here.
```
djohnson-spideroak: This PR or another?

crates/aranya-client/src/config.rs
```diff
@@ -119,3 +119,115 @@ impl Default for SyncPeerConfigBuilder {
...
+impl HelloSubscriptionConfig {
+    /// Creates a default [`HelloSubscriptionConfigBuilder`].
+    pub fn builder() -> HelloSubscriptionConfigBuilder {
+        Default::default()
+    }
+
+    /// Minimum delay after a graph change before sending a hello notification.
+    pub fn graph_change_delay(&self) -> Duration {
+        self.graph_change_delay
+    }
```
jdygert-spok: I know the name and possibly description were pre-existing, but they are off from how we are currently using it. Currently this is the amount of time after sending a notification in which no notification will be sent upon graph change.

I think we want this to ideally behave like a debounce so we could just call and describe it as that.

crates/aranya-client/src/client/team.rs
gknopf-aranya: A different PR. This PR has already increased in scope from the original goal of fixing a test flake. I want to nail down the API now in case this feature is moved out from behind the preview flag since this is an API improvement. Internal changes like this can happen later.

crates/aranya-client/src/config.rs
gknopf-aranya: Renaming seems out of scope of the current PR. This is focused on solving a flake and making API improvements. A rename would affect naming throughout the repo and seems better suited for a stand-alone PR.

crates/aranya-client/src/client/team.rs
djohnson-spideroak: Lets have an issue to track the changes.

crates/aranya-client/src/config.rs
djohnson-spideroak: I am fine doing the rename in a followup as long as we have an issue for it.

jdygert-spok: The name wasn't part of the public API before this change. We could just change it in this new builder method.

crates/aranya-client/src/client/team.rs
gknopf-aranya: Here's an issue for it:
https://github.com/aranya-project/aranya/issues/709

crates/aranya-client/src/config.rs
gknopf-aranya: That would work. Mostly want to stabilize whatever part of the API that cannot change after we move this out from behind a feature flag.

gknopf-aranya: Is the suggestion to call this `graph_change_debounce`?

gknopf-aranya: Renaming everywhere requires a change to the sync protocol. I can rename in client/daemon code here. Mostly a find-replace operation.

jdygert-spok: Yeah something like that. That sounds fine.

jdygert-spok: It's fine if you want to limit the change to the public API of the rust and C builder.

gknopf-aranya: claude made it a little faster, so I'll tackle it here

gknopf-aranya:
@jdygert-spok ready for another look

PR #702, opened 2026-01-30
fix(syncer): replce return address with just port

### LINKED ISSUE #701: bug(syncer): return address change handles meta addresses poorly
If a node is configured to serve on `0.0.0.0:1234` then they will tell peers they are reachable at that address, which is wrong for a device on another machine, so this causes various issues.

The previous behavior was to use the peer's remote address, which was different from the server address, which prevented proper connection reuse and peer cache duplication, but still behaved "correctly" in a downgraded way.

jdygert-spok:
Fixes #701.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,9 @@ cargo make build
 cargo make build-capi
 
 cargo make test
-cargo test
+
+# Run tests in memory to avoid slow fsync
+TMPDIR=/dev/shm cargo test
 
 cargo make correctness   # fmt + clippy + feature checks
 cargo make fmt


jdygert-spok:
This is unfortunately not yet covered by tests since it needs a true multi-node setup such as #385.

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -172,17 +172,19 @@ where
                 .server_keys
                 .get_team_for_identity(&identity)
                 .context("no active team for accepted connection")?;
-            let peer_addr: Addr = {
+            let peer_port: u16 = {
```
gknopf-aranya: Could be confused with the client addr/port. `peer_server_port` or `peer_return_port`

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -172,17 +172,19 @@ where
...
                 let mut recv = conn
                     .accept_receive_stream()
                     .await?
                     .context("no stream for peer address")?;
                 let bytes = recv.receive().await?.context("no peer address sent")?;
-                postcard::from_bytes(&bytes).context("bad peer address")?
+                u16::from_be_bytes(bytes.as_ref().try_into().context("bad peer address")?)
             };
+            let mut peer_addr = conn.remote_addr()?;
+            peer_addr.set_port(peer_port);
```
gknopf-aranya: Or construct with something like `Addr::new(ip, port)`. Setting the port after construction seems like something someone could easily forget to do when refactoring since it will still compile.

PR #703, opened 2026-02-03
update `bytes` to resolve security vulnerability warning

ssweetney-spok:
See [here](https://rustsec.org/advisories/RUSTSEC-2026-0007)

PR #705, opened 2026-02-03
release: 4.1.1

gknopf-aranya:
Patch release for https://rustsec.org/advisories/RUSTSEC-2026-0007

Cherry-picks: eccd7a0d Update `bytes` to resolve security vulnerability warning (#703)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,6 @@ permissions:
 on:
   push:
     branches:
-      - main
       - release-4.1.0-base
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -10,6 +10,7 @@ on:
   push:
     branches:
       - main
```
jdygert-spok:
```suggestion
```

To be safe against a later merge or something. Probably isn't an issue but doesn't hurt to remove.

PR #710, opened 2026-02-04
Rename `KeyBundle` to `PubKeyBundle`

### LINKED ISSUE #574: Rename `KeyBundle` to `PubKeyBundle`
Just clarifying usage by changing a name. Should be changed in rust and policy code.

jdygert-spok:
Closes #574.

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -2495,14 +2495,18 @@ pub unsafe fn afc_channel_seal(
         );
     }
 
-    if *dst_len < (plaintext.len() + ARANYA_AFC_CHANNEL_OVERHEAD) {
+    let sealed_len = plaintext
+        .len()
+        .checked_add(ARANYA_AFC_CHANNEL_OVERHEAD)
+        .ok_or(imp::Error::BufferTooSmall)?;
+    if *dst_len < sealed_len {
         return Err(imp::Error::BufferTooSmall);
     }
 
     // SAFETY: the user is responsible for giving us a valid pointer.
     let dst = aranya_capi_core::try_as_mut_slice!(dst, *dst_len);
     channel.seal(dst, plaintext)?;
-    *dst_len = plaintext.len() + ARANYA_AFC_CHANNEL_OVERHEAD;
+    *dst_len = sealed_len;
 
     Ok(())
 }
@@ -2530,7 +2534,11 @@ pub unsafe fn afc_channel_open(
         );
     }
 
-    if *dst_len < (ciphertext.len() - ARANYA_AFC_CHANNEL_OVERHEAD) {
+    let opened_len = ciphertext
+        .len()
+        .checked_sub(ARANYA_AFC_CHANNEL_OVERHEAD)
+        .ok_or(imp::Error::BufferTooSmall)?;
+    if *dst_len < opened_len {
         return Err(imp::Error::BufferTooSmall);
     }
 
@@ -2540,7 +2548,7 @@ pub unsafe fn afc_channel_open(
 
     AfcSeq::init(seq, seq_raw);
     // Do our best to set a max bound, even if we can't know if they pass in a larger ciphertext than needed.
-    *dst_len = ciphertext.len() - ARANYA_AFC_CHANNEL_OVERHEAD;
+    *dst_len = opened_len;
 
     Ok(())
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ unused_lifetimes = "warn"
 unused_qualifications = "warn"
 
 [workspace.lints.clippy]
+arithmetic_side_effects = "warn"
 cast_lossless = "warn"
 cast_possible_wrap = "warn"
 cast_precision_loss = "warn"
--- a/crates/aranya-certgen/src/cert.rs
+++ b/crates/aranya-certgen/src/cert.rs
@@ -374,7 +374,9 @@ fn generate_root_ca(cn: &str, days: u32) -> Result<(Certificate, KeyPair), CertG
 
     let now = OffsetDateTime::now_utc();
     params.not_before = now;
-    params.not_after = now + Duration::days(i64::from(days));
+    params.not_after = now
+        .checked_add(Duration::days(i64::from(days)))
+        .ok_or(CertGenError::InvalidDays)?;
 
     let key_pair = KeyPair::generate()?;
     let cert = params.self_signed(&key_pair)?;
@@ -414,7 +416,9 @@ fn generate_signed_cert(
 
     let now = OffsetDateTime::now_utc();
     params.not_before = now;
-    params.not_after = now + Duration::days(i64::from(days));
+    params.not_after = now
+        .checked_add(Duration::days(i64::from(days)))
+        .ok_or(CertGenError::InvalidDays)?;
 
     let key_pair = KeyPair::generate()?;
     let cert = params.signed_by(&key_pair, issuer)?;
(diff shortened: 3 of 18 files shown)


crates/aranya-client-capi/output/aranya-client.h
```diff
@@ -905,8 +905,8 @@ AranyaError aranya_init_logging_ext(struct AranyaExtError *__ext_err);
  * Gets the public key bundle for this device.
  *
  * @param[in] client the Aranya Client
- * @param[out] keybundle keybundle byte buffer `KeyBundle`.
- * @param[in,out] keybundle_len returns the length of the serialized keybundle.
+ * @param[out] keybundle key bundle byte buffer
```
chip-so: Removing the type name seems like a documentation regression. Or is this already clear somehow else?

chip-so (approved):

crates/aranya-client-capi/output/aranya-client.h
jdygert-spok: We previously removed types ala #523 but in this case it's also not even the type of the parameter (it is just an opaque byte buffer).

I am realizing it's not ideal that we handle this differently from `AfcCtrl` :/

gknopf-aranya: Not directly related to this thread, but I generally prefer to comment on defs.rs instead of aranya-client.h since the header is autogenerated. Unless I'm commenting about something related to how cbindgen may have generated output incorrectly.

chip-so: > We previously removed types ala https://github.com/aranya-project/aranya/issues/523 but in this case it's also not even the type of the parameter (it is just an opaque byte buffer).

I see. That's fine, then.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -657,8 +657,8 @@ pub fn init_logging() -> Result<(), imp::Error> {
 /// Gets the public key bundle for this device.
 ///
 /// @param[in] client the Aranya Client
-/// @param[out] keybundle keybundle byte buffer `KeyBundle`.
-/// @param[in,out] keybundle_len returns the length of the serialized keybundle.
+/// @param[out] keybundle key bundle byte buffer
```
gknopf-aranya: If we're renaming the type, should we also change the function, doc, param, etc. wording?
```suggestion
/// @param[out] pubkeybundle public key bundle byte buffer
```

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -657,8 +657,8 @@ pub fn init_logging() -> Result<(), imp::Error> {
...
+/// @param[in,out] keybundle_len returns the length of the serialized key bundle.
 ///
 /// @relates AranyaClient.
 pub unsafe fn get_key_bundle(
```
```suggestion
pub unsafe fn get_public_key_bundle(
```

jdygert-spok: There's about a dozen methods we could rename. If we do want to, that's a lot more effort since we need to make a deprecated alias for each.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: IMO it's fine to keep in at least some of these places as essentially a shorthand, we don't have to always spell out the full meaning of something every time we mention it.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -414,15 +414,15 @@ pub trait DaemonApi {
...
         team: TeamId,
-        keys: KeyBundle,
+        keys: PubKeyBundle,
         initial_role: Option<RoleId>,
     ) -> Result<()>;
     /// Remove device from the team.
     async fn remove_device_from_team(team: TeamId, device: DeviceId) -> Result<()>;
     /// Returns all the devices on the team.
     async fn devices_on_team(team: TeamId) -> Result<Box<[DeviceId]>>;
     /// Returns the device's key bundle.
```
```suggestion
    /// Returns the device's public key bundle.
```

crates/aranya-daemon-api/src/service.rs
```diff
@@ -414,15 +414,15 @@ pub trait DaemonApi {
...
-    async fn device_keybundle(team: TeamId, device: DeviceId) -> Result<KeyBundle>;
+    async fn device_keybundle(team: TeamId, device: DeviceId) -> Result<PubKeyBundle>;
```
gknopf-aranya: IPC API doesn't need to be backward compatible so we can update the method name:
```suggestion
    async fn device_public_keybundle(team: TeamId, device: DeviceId) -> Result<PubKeyBundle>;
```

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: True, it seems more important in API docs than internal methods. I do think we should be as explicit as possible in the C API docs if we're not able to update the function/param names due to backward compatibility.

gknopf-aranya (approved):
lgtm pending more clarity on this being a public key bundle in the C API docs

jdygert-spok:
Now that I think about it this should probably be spelled out `PublicKeyBundle`

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: I renamed the Rust client methods but I doing the C ones is a bit tricker right now. To mark them deprecated we either need some updates to the capi tooling or to manually add them all in to the cbindgen.toml.

djohnson-spideroak: In my opinion, we dont need to do the breaking renames. I think as long as the docs are clear and we mention it in the appropriate places, we can keep the functions as is.

chip-so (approved):

gknopf-aranya (approved):

gknopf-aranya:
@djohnson-spideroak, @ssweetney-spok just need 1 more approval on this PR

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: @jdygert-spok, @djohnson-spideroak can this discussion be resolved?

jdygert-spok: We could do the C deprecated renames now that aranya-project/aranya-core#559 was merged :shrug:

gknopf-aranya: Up to you. Renaming the methods would be an improvement but doesn't seem required.

gknopf-aranya: Looks like the methods have already been deprecated on this PR. Since the work is completed, there doesn't appear to be a reason to revert that. @djohnson-spideroak, @chip-so please let me know if you disagree. I'll mark this discussion as resolved soon since we want to include this PR in today's release.

jdygert-spok: Just waiting on #726 to push update with deprecated C.

gknopf-aranya: Could base this PR on that branch to push the changes earlier.

gknopf-aranya: That PR was merged so this is unblocked now

crates/aranya-client-capi/ctest/test_onboarding.c
```diff
@@ -69,17 +69,18 @@ static int test_create_team_and_onboard_member(const char *tmpdir) {
...
         member_pk = realloc(member_pk, member_pk_len);
         if (!member_pk) {
             fprintf(stderr,
                     "Failed to reallocate memory for member key bundle\n");
             goto exit;
         }
-        CLIENT_EXPECT(
-            "Failed to get member key bundle", "",
-            aranya_get_key_bundle(&member_client, member_pk, &member_pk_len));
+        CLIENT_EXPECT("Failed to get member key bundle", "",
```
gknopf-aranya: nit:
```suggestion
        CLIENT_EXPECT("Failed to get member public key bundle", "",
```

PR #711, opened 2026-02-06
tests: larger scale tests

djohnson-spideroak:
add in scale testing, ring tests from 10..70, custom topology, and test harness

### DIFF
--- a/examples/rust/aranya-example/src/main.rs
+++ b/examples/rust/aranya-example/src/main.rs
@@ -392,30 +392,46 @@ async fn main() -> Result<()> {
     membera_team.sync_now(owner_addr, None).await?;
     memberb_team.sync_now(owner_addr, None).await?;
 
-    // add membera to team.
-    info!("adding membera to team");
-    let member_role_rank = owner_team.query_rank(member_role.id).await?;
-    owner_team
+    // Admin adds membera and memberb to the team (without roles, since
+    // the admin does not have AssignRole permission).
+    info!("admin adding membera to team");
+    let member_role_rank = admin_team.query_rank(member_role.id).await?;
+    admin_team
         .add_device(
             membera.pk.clone(),
-            Some(member_role.id),
+            None,
             Rank::new(member_role_rank.value().saturating_sub(1)),
         )
         .await?;
 
-    // add memberb to team.
-    info!("adding memberb to team");
-    owner_team
+    info!("admin adding memberb to team");
+    admin_team
         .add_device(
             memberb.pk.clone(),
-            Some(member_role.id),
+            None,
             Rank::new(member_role_rank.value().saturating_sub(1)),
         )
         .await?;
 
-    // Sync membera and memberb so they see their team membership and roles.
-    membera_team.sync_now(owner_addr, None).await?;
-    memberb_team.sync_now(owner_addr, None).await?;
+    // Operator assigns the member role to membera and memberb.
+    // Sync with admin to pick up the AddDevice commands.
+    operator_team.sync_now(admin_addr, None).await?;
+
+    info!("operator assigning member role to membera");
+    operator_team
+        .device(membera.id)
+        .assign_role(member_role.id)
+        .await?;
+    info!("operator assigning member role to memberb");
+    operator_team
+        .device(memberb.id)
+        .assign_role(member_role.id)
+        .await?;
+
+    // Sync membera and memberb with the operator so they see their
+    // team membership (added by admin) and role assignments (by operator).
+    membera_team.sync_now(operator_addr, None).await?;
+    memberb_team.sync_now(operator_addr, None).await?;
 
     // Demo custom roles.
     info!("demo custom roles functionality");
@@ -508,26 +524,31 @@ async fn main() -> Result<()> {
     let keybundle = owner_device.public_key_bundle().await?;
     info!("owner keybundle: {:?}", keybundle);
 
-    info!("creating label");
-    let owner_device_rank = owner_team.query_rank(owner.id).await?;
-    // Label rank must be lower than the owner's device rank so the owner can operate on it.
-    let label_rank = Rank::new(owner_device_rank.value().saturating_sub(1));
-    let label3 = owner_team.create_label(text!("label3"), label_rank).await?;
+    // Admin creates a label. The label rank must be lower than the
+    // operator's device rank so the operator can assign it to devices.
+    info!("admin creating label");
+    let operator_device_rank = admin_team.query_rank(operator.id).await?;
+    let label_rank = Rank::new(operator_device_rank.value().saturating_sub(1));
+    let label3 = admin_team.create_label(text!("label3"), label_rank).await?;
+
+    // Operator assigns the label to membera and memberb.
     let op = ChanOp::SendRecv;
-    info!("assigning label to membera");
-    owner_team
+    operator_team.sync_now(admin_addr, None).await?;
+
+    info!("operator assigning label to membera");
+    operator_team
         .device(membera.id)
         .assign_label(label3, op)
         .await?;
-    info!("assigning label to memberb");
-    owner_team
+    info!("operator assigning label to memberb");
+    operator_team
         .device(memberb.id)
         .assign_label(label3, op)
         .await?;
 
-    // Sync membera and memberb so they see the label assignments.
-    membera_team.sync_now(owner_addr, None).await?;
-    memberb_team.sync_now(owner_addr, None).await?;
+    // Sync membera and memberb with the operator so they see the label assignments.
+    membera_team.sync_now(operator_addr, None).await?;
+    memberb_team.sync_now(operator_addr, None).await?;
 
     // Demo AFC.
     info!("demo afc functionality");


Cargo.toml
```diff
@@ -83,6 +83,7 @@ rustix = { version = "1.1.2", features = ["shm"] }
 s2n-quic = { version = "1.58.0-psk.2", default-features = false, features = ["provider-address-token-default", "provider-tls-rustls"], package = "aranya-internal-s2n-quic" }
 semver = { version = "1", features = ["serde"] }
 serde = "1"
+serde_json = "1"
```
djohnson-spideroak: unused

.duvet/config.toml
```diff
@@ -0,0 +1,30 @@
...
+type = "implementation"
+
+[[source]]
+pattern = "crates/aranya-client/tests/ring_convergence.rs"
+type = "test"
+
+# Specification document
+[[specification]]
+source = "docs/multi-daemon-convergence-test.md"
+format = "markdown"
```
djohnson-spideroak, line 15: TODO: update to remote URL to aranya-docs

.duvet/config.toml
```diff
@@ -0,0 +1,30 @@
...
+[[specification]]
+source = "https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md"
+format = "markdown"
+
+# Report configuration
+[report.html]
+enabled = true
+path = ".duvet/reports/report.html"
+blob-link = "https://github.com/aranya-project/aranya/blob/main"
+issue-link = "https://github.com/aranya-project/aranya/issues"
```
djohnson-spideroak, line 22: check these

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// for identification and AFC shared memory path generation.
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#init-001
+    //# Each node MUST be initialized with a unique daemon instance.
+    #[instrument(skip(work_dir), fields(node_index = index))]
+    pub(crate) async fn new(index: usize, work_dir: PathBuf, team_name: &str) -> Result<Self> {
+        let addr_any = Addr::from((Ipv4Addr::LOCALHOST, 0));
+
+        // Generate unique AFC shm path per node
+        //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#init-002
+        //# Each node MUST have its own cryptographic keys.
```
djohnson-spideroak: not really in the right place, will probably move below to getting keybundle.

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,214 @@
...
+            .context("unable to start daemon")?;
+
+        // Connect client with retry
+        let client = (|| Client::builder().with_daemon_uds_path(&uds_path).connect())
+            .retry(ExponentialBuilder::default())
+            .await
+            .context("unable to init client")?;
+
+        //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#init-003
+        //# All nodes MUST have unique device IDs.
```
djohnson-spideroak, line 94: move one line lower

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,214 @@
...
+                config.node_count,
+                nodes.len()
+            );
+        }
+
+        // Verify all nodes have unique device IDs
+        let unique_ids: HashSet<_> = nodes.iter().map(|n| n.id).collect();
+        if unique_ids.len() != nodes.len() {
+            bail!("Duplicate device IDs detected among nodes");
+        }
```
djohnson-spideroak: probably rm

crates/aranya-client/tests/scale/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+            .team(team_id)
+            .remove_sync_peer(to_addr)
+            .await
+            .with_context(|| format!("node {} unable to remove sync peer {}", from, to))?;
+
+        if matches!(self.sync_mode, SyncMode::Hello { .. }) {
+            self.nodes[from.0]
+                .client
+                .team(team_id)
+                .sync_hello_unsubscribe(to_addr)
```
djohnson-spideroak, line 507: Do we need to unsubscribe if we remove the peer?

crates/aranya-client/tests/scale/ring.rs
```diff
@@ -0,0 +1,215 @@
...
+            debug!(
+                node = i,
+                clockwise, counter_clockwise, "Configured sync peers"
+            );
+        }
+
+        //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#sync-004
+        //# Sync peer configuration MUST complete before the convergence test phase.
+        // Give a small delay for sync configuration to settle
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
djohnson-spideroak: not a great place for this annotation, and unneeded delay.

crates/aranya-client/tests/scale/ring.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// 1. Each node has exactly 2 peers
+    /// 2. Peers are bidirectional (if A peers with B, B peers with A)
+    /// 3. The graph forms a single connected component
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#ring-001
+    //# In the ring topology, each node MUST connect to exactly two other nodes: its clockwise neighbor and its counter-clockwise neighbor.
+
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#ring-002
+    //# The ring topology MUST form a single connected ring (each node's two peers link to form one cycle covering all nodes).
+    #[instrument(skip(self))]
+    pub fn verify_ring_topology(&self) -> Result<()> {
```
djohnson-spideroak: can probably remove this, and move it to a test or just careful review of the ring construction is fine.

examples/c/example.c
djohnson-spideroak, line 1: revert

crates/aranya-client/tests/scale/team.rs
```diff
@@ -0,0 +1,214 @@
...
+            }
+        }
+
+        info!("Team propagation verified on all nodes");
+        Ok(())
+    }
+
+    /// Performs a manual sync from all nodes to ensure team configuration propagates.
+    ///
+    /// This is used during setup before the ring topology is configured.
```
djohnson-spideroak:
```suggestion
    /// This is used during setup before the topology is configured.
```

crates/aranya-client/tests/ring_convergence.rs
```diff
@@ -0,0 +1,686 @@
...
+
+/// Tests ring convergence with 10 nodes using poll sync mode.
+///
+/// This is a smaller test suitable for CI with reasonable execution time.
+//= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#conf-001
+//= type=test
+//# The test MUST support configuring the number of nodes.
+#[test(tokio::test(flavor = "multi_thread"))]
+#[serial]
+async fn test_ring_convergence_10_nodes() -> Result<()> {
```
djohnson-spideroak: use helper

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,266 @@
...
+            .collect();
+
+        if times.is_empty() {
+            return None;
+        }
+
+        times.sort();
+
+        let min_time = *times.first().unwrap();
+        let max_time = *times.last().unwrap();
```
djohnson-spideroak: thoughts on unwrap here? Its test code, so imo we are probably fine with the panics.

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,266 @@
...
+        Duration::from_millis(mode_bucket as u64)
+    }
+
+    /// Calculates the percentile using nearest-rank method.
+    /// `times` must be sorted.
+    #[allow(
+        clippy::cast_sign_loss,
+        clippy::cast_precision_loss,
+        clippy::cast_possible_truncation
+    )]
```
djohnson-spideroak, line 139: this is test code, so I felt comfortable keeping these allows.

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,266 @@
...
+                println!("Min convergence time:    {:?}", metrics.min_time);
+                println!("Max convergence time:    {:?}", metrics.max_time);
+                println!("Mean convergence time:   {:?}", metrics.mean_time);
+                println!("Median convergence time: {:?}", metrics.median_time);
+                println!("Mode convergence time:   {:?}", metrics.mode_time);
+                println!("p95 convergence time:    {:?}", metrics.p95_time);
+                println!("p99 convergence time:    {:?}", metrics.p99_time);
+                println!("Std deviation:           {:?}", metrics.std_dev);
+                if let Some(total) = metrics.total_convergence_time {
+                    println!("Total convergence time:  {:?}", total);
```
djohnson-spideroak, line 185: Do we want to allow println like this? I dont see any other test code in aranya that uses println, so I guess I coudl switch this over to tracing.

gknopf-aranya: Tracing is good. More flexible and allows output to stdout just like println if we want.

crates/aranya-client/tests/scale/metrics.rs
gknopf-aranya: Or `.expect("error msg")`. Panicking is okay in test code.

crates/aranya-client/tests/scale/metrics.rs
jdygert-spok: It's a bit of a gray area IMO. If the test is mostly meant as pass/fail, and sometimes you want to look at these metrics, then I'd lean toward tracing. If the primary point of the test is to see the metrics, then I'd probably print them.

djohnson-spideroak: The metrics are a big part of the value of the test. The pass/fail is helpful to tell when things have gone wrong with syncing or onboarding, but really we set up tests like this to look at the performance and pass/fail.

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,266 @@
...
+            Ok(()) => info!(path = %csv_path, "CSV convergence data exported"),
+            Err(e) => warn!(path = %csv_path, error = %e, "Failed to export CSV"),
+        }
+    }
+}
+
+/// Gets the current process memory usage in KB (Linux only).
+#[cfg(target_os = "linux")]
+fn get_process_memory_kb() -> Option<u64> {
+    use std::fs;
```
djohnson-spideroak: I might remove this. I dont think the way it is calculated is very useful for this. An outlier could greatly skew the memory stats.

Makefile.toml
```diff
@@ -338,6 +338,35 @@ description = "Run Rust example application on multiple nodes"
...
+# auto-detection sees a leading '#' and chooses the markdown parser.
+find .duvet/specifications -name '*.txt' -print0 | while IFS= read -r -d '' f; do
+  if head -1 "$f" | grep -q '^---$'; then
+    awk 'BEGIN{n=0} /^---$/{n++; next} n>=2{print}' "$f" > "$f.tmp"
+    mv "$f.tmp" "$f"
+  fi
+done
+
+duvet report --config-path .duvet/config.toml
+'''
```
djohnson-spideroak: moving to its own file in scripts/ folder. duvet uses a pretty poor way to determining if a file is markdown.

scripts/duvet-report.bash
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Run extract to download and cache the remote spec (uses config format).
+duvet extract --config-path .duvet/config.toml \
+  -f markdown \
+  "https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md" \
+  2>/dev/null || true
```
djohnson-spideroak, line 8: for now, we only have this spec with duvet annotations, we would want to update this in the future to get the full list of specs.

.duvet/.gitignore
gknopf-aranya, line 1: Doesn't look like there's a need to ignore any of this yet. Assuming this is standard for `duvet`?

.duvet/config.toml
```diff
@@ -0,0 +1,30 @@
+'$schema' = "https://awslabs.github.io/duvet/config/v0.4.0.json"
```
gknopf-aranya, line 1: Should we copy this file in case it isn't available at this URL?

.duvet/config.toml
```diff
@@ -0,0 +1,30 @@
...
+
+# Source files to scan for annotations
+[[source]]
+pattern = "crates/aranya-client/tests/scale/**/*.rs"
```
gknopf-aranya, line 5: We'll update this as duvet is added to more of the codebase?

scripts/duvet-report.bash
gknopf-aranya, line 1: Are `duvet` reports something we'll want to run in CI eventually?

scripts/duvet-report.bash
```diff
@@ -0,0 +1,19 @@
...
+
+# Strip YAML front matter from cached .txt specs so duvet's
```
gknopf-aranya, line 10: Why is there YAML front matter on the specs?

scripts/duvet-report.bash
gknopf-aranya: I think we'll at least want a `cargo make duvet` task for this in `Makefile.toml` even if we're not ready to add this to CI yet. Will make it easier for devs to run and easier to add to CI in the future.

gknopf-aranya: Oh, I see that this was already added to `Makefile.toml`. Just needed to read more of the PR.

.duvet/config.toml
gknopf-aranya: While it's inefficient to scan the entire codebase for duvet requirements when they've only been added for scale testing, I think we'll need to at some point once duvet is something that is more integrated into our code. Worth adding a TODO here.

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,203 @@
...
+impl NodeCtx {
+    /// Creates a new node context.
+    ///
+    /// This follows the `DeviceCtx::new()` pattern but uses a unique index
+    /// for identification and AFC shared memory path generation.
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#init-001
+    //# Each node MUST be initialized with a unique daemon instance.
+    #[instrument(skip(work_dir), fields(node_index = index))]
+    pub(crate) async fn new(index: usize, work_dir: PathBuf, team_name: &str) -> Result<Self> {
+        let addr_any = Addr::from((Ipv4Addr::LOCALHOST, 0));
```
gknopf-aranya, line 28: Will scale testing always run all the nodes on a single machine? 1k node tests could be tricky to perform this way if the machine doesn't have enough resources even if Aranya may scale okay with that many nodes in a distributed way.

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,203 @@
...
+            let path: Box<shm::Path> = path
+                .as_str()
+                .try_into()
+                .context("unable to parse AFC shared memory path")?;
+            let _ = shm::unlink(&path);
+            path
+        };
+
+        // Setup daemon config
+        let cfg = Config {
```
gknopf-aranya, line 42: We setup these configs very similarly in several places (e.g. Rust examples) due to the subdirectories having similar names. Noting this as an opportunity for better code reuse, not something needed on this PR.

crates/aranya-client/tests/scale/init.rs
gknopf-aranya, line 1: Could put `NodeCtx` in a `node.rs` module. Would `TestCtx` go in a `test.rs` module? `init.rs` doesn't seem like the right abstraction.

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl TestCtx {
+    /// Creates a new test context with all nodes initialized.
+    ///
+    /// Nodes are initialized in parallel batches to avoid resource exhaustion
+    /// while still providing reasonable startup performance.
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#init-004
+    //# Node initialization MUST occur in parallel batches to avoid resource exhaustion.
```
gknopf-aranya, line 117: Parallelization doesn't prevent the resource exhaustion. Limiting the number of concurrent node initializations does. Wording could be more clear.

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,203 @@
...
+    #[instrument(skip(config), fields(node_count = config.node_count))]
+    pub async fn new(config: TestConfig, topology: Option<Vec<Topology>>) -> Result<Self> {
+        config.validate()?;
+
+        let work_dir = TempDir::new().context("unable to create temp dir")?;
+        let mut nodes = Vec::with_capacity(config.node_count);
+        let team_name = "scale_test";
```
gknopf-aranya: Could make the `team_name` part of the test config instead of hard-coding it.

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,203 @@
...
+
+        info!(
+            node_count = config.node_count,
+            batch_size = config.init_batch_size,
+            "Initializing nodes"
+        );
+
+        // Initialize nodes in batches
+        for batch_start in (0..config.node_count).step_by(config.init_batch_size) {
```
gknopf-aranya, line 132: A thread pool seems like the idiomatic way to perform tasks in parallel on a limited number of threads.

crates/aranya-client/tests/scale/init.rs
```diff
@@ -0,0 +1,203 @@
...
+
+        info!(
+            node_count = nodes.len(),
+            "All nodes initialized successfully"
+        );
+
+        // Generate shared seed IKM for QUIC sync
+        let seed_ikm = {
+            let mut buf = [0u8; SEED_IKM_SIZE];
+            nodes[0].client.rand(&mut buf).await;
```
gknopf-aranya, line 188: `nodes.len()` must be at least 1 for this to work. Better to handle that error rather than a runtime exception. Surprised the linter didn't complain about this.

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub p95_time: Duration,
+    /// 99th percentile convergence time.
+    pub p99_time: Duration,
+    /// Standard deviation of convergence times.
+    pub std_dev: Duration,
+    /// Total time from command issuance to full convergence.
+    pub total_convergence_time: Option<Duration>,
+}
+
+impl TestCtx {
```
gknopf-aranya: May want a `Metrics` trait defined here with these methods so the `TestCtx` implementation can stay with the rest of the `TestCtx` implementation in a `test.rs` module while implementing the `Metrics` trait.

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,240 @@
...
+            .collect();
+
+        if times.is_empty() {
+            return None;
+        }
+
+        times.sort();
+
+        let min_time = *times.first().expect("times must not be empty");
+        let max_time = *times.last().expect("times must not be empty");
```
gknopf-aranya, line 73: These errors are true, but this may have more meaning in the context of the operation being performed:
```suggestion
        let min_time = *times.first().expect("unable to get min time");
        let max_time = *times.last().expect("unable to get max time");
```

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        let sum: Duration = times.iter().sum();
+        let mean_time = sum / times.len() as u32;
```
gknopf-aranya, line 76: Could there be any issues with casting to a `u32` here if the number is too large to fit?

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        // Mode: bucket times into 100ms intervals to produce a meaningful mode
+        let mode_time = Self::calculate_mode(&times);
+
+        // Percentiles (using nearest-rank method)
+        let p95_time = Self::percentile(&times, 95.0);
+        let p99_time = Self::percentile(&times, 99.0);
+
+        // Standard deviation
+        let n = u32::try_from(times.len()).expect("node count fits in u32");
```
gknopf-aranya: I'd consider moving the standard deviation calculation into its own method. May want to consider importing a stat library for calculating some of these.

crates/aranya-client/tests/scale/convergence.rs
```diff
@@ -0,0 +1,176 @@
...
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#conv-001
+    //# The test MUST assign a label to the source node's graph to mark the start of convergence testing.
+
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#conv-002
+    //# The default source node for label assignment MUST be node 0.
+    #[instrument(skip(self), fields(source_node = %source_node))]
+    pub async fn issue_test_command(&mut self, source_node: NodeIndex) -> Result<()> {
+        let team_id = self.team_id.context("Team not created")?;
+
+        if source_node.0 >= self.nodes.len() {
```
gknopf-aranya: You could derive some stuff like `Eq`, `PartialEq`, `Ord`, `PartialOrd` on `NodeIndex` to compare node indices directly without accessing the `.0` field.

crates/aranya-client/tests/scale/convergence.rs
```diff
@@ -0,0 +1,176 @@
...
+            bail!(
+                "Source node {} out of range (max: {})",
+                source_node,
+                self.nodes.len() - 1
+            );
+        }
+
+        // Generate unique label name for this test run using Aranya's CSPRNG
```
gknopf-aranya, line 33: A CSPRNG is overkill for generating a unique test name that doesn't need to be cryptographically secure. The random generator from the standard library is sufficient.

crates/aranya-client/tests/scale/convergence.rs
```diff
@@ -0,0 +1,176 @@
...
+        let rand_hex = {
+            let mut buf = [0u8; 16];
+            self.nodes[source_node.0].client.rand(&mut buf).await;
```
gknopf-aranya: There are enough places where `source_node.0` is accessed that I think a getter would be more clear: `source_node.value()`

crates/aranya-client/tests/scale/convergence.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Issues a test command from the specified source node.
+    ///
+    /// Creates a label as the observable command that will propagate through the ring.
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#conv-001
+    //# The test MUST assign a label to the source node's graph to mark the start of convergence testing.
+
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#conv-002
+    //# The default source node for label assignment MUST be node 0.
+    #[instrument(skip(self), fields(source_node = %source_node))]
+    pub async fn issue_test_command(&mut self, source_node: NodeIndex) -> Result<()> {
```
gknopf-aranya, line 22:
```suggestion
    pub async fn test_create_label_command(&mut self, source_node: NodeIndex) -> Result<()> {
```

crates/aranya-client/tests/scale/convergence.rs
```diff
@@ -0,0 +1,176 @@
...
+        let team_id = self.team_id.context("Team not created")?;
+        let start = Instant::now();
+
+        info!(
+            timeout = ?self.config.max_duration,
+            poll_interval = ?self.config.poll_interval,
+            "Waiting for convergence"
+        );
+
+        loop {
```
gknopf-aranya, line 95: Would be more idiomatic to `join_all` on multiple futures.

crates/aranya-client/tests/scale/topology.rs
```diff
@@ -0,0 +1,55 @@
+//! Topology configuration dispatcher for scale convergence tests.
+
+use anyhow::Result;
+use tracing::info;
+
+use crate::scale::{NodeIndex, TestCtx, Topology};
+
+impl TestCtx {
```
gknopf-aranya, line 95: Was expecting to see a `Topology` type in this module.

crates/aranya-client/tests/scale/ring.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl TestCtx {
+    /// Configures the bidirectional ring topology.
+    ///
+    /// Each node is connected to its clockwise and counter-clockwise neighbors,
+    /// forming a ring where data can propagate in both directions.
+    //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#ring-001
+    //# In the ring topology, each node MUST connect to exactly two other nodes: its clockwise neighbor and its counter-clockwise neighbor.
+    #[instrument(skip(self))]
+    pub async fn configure_ring_topology(&mut self) -> Result<()> {
```
gknopf-aranya, line 17: Configuring a ring topology seems like something to include in the implementation of `Topology` in the `topology.rs` module. Or possibly on a `Ring` type which implements a `Topology` trait.

crates/aranya-client/tests/scale/team.rs
```diff
@@ -0,0 +1,212 @@
...
+use aranya_client::{
+    client::TeamId, config::CreateTeamConfig, AddTeamConfig, AddTeamQuicSyncConfig,
+    CreateTeamQuicSyncConfig,
+};
+use tokio::time::sleep;
+use tracing::{info, instrument, warn};
+
+use crate::scale::TestCtx;
+
+impl TestCtx {
```
gknopf-aranya, line 15: Or implement this on a `TeamCtx` type. If we're going to modularize the implementations of `TestCtx` into a bunch of different files I think it makes sense to abstract each of those module into their own types.

crates/aranya-client/tests/ring_convergence.rs
```diff
@@ -0,0 +1,756 @@
...
+    let mut ctx = TestCtx::new(config, Some(vec![topology])).await?;
+
+    ctx.setup_team().await?;
+    ctx.sync_team_from_owner().await?;
+    ctx.configure_topology().await?;
+    ctx.verify_team_propagation().await?;
+
+    ctx.issue_test_command(NodeIndex(0)).await?;
+    ctx.wait_for_convergence().await?;
+    ctx.report_metrics();
```
gknopf-aranya: This will only report metrics if convergence succeeds. Is there a use-case for reporting metrics if there is an error or the test hangs?

crates/aranya-client/tests/ring_convergence.rs
```diff
@@ -0,0 +1,756 @@
...
+async fn run_convergence(config: TestConfig, topology: Topology) -> Result<()> {
+    let mut ctx = TestCtx::new(config, Some(vec![topology])).await?;
+
+    ctx.setup_team().await?;
+    ctx.sync_team_from_owner().await?;
+    ctx.configure_topology().await?;
+    ctx.verify_team_propagation().await?;
+
+    ctx.issue_test_command(NodeIndex(0)).await?;
+    ctx.wait_for_convergence().await?;
```
gknopf-aranya: May want to put a timeout here in case the test never converges.

crates/aranya-client/tests/ring_convergence.rs
```diff
@@ -0,0 +1,756 @@
...
+}
+
+/// Tests convergence with 10 nodes arranged in two 5-node rings connected
+/// by a single bidirectional bridge, using poll sync mode.
+#[test(tokio::test(flavor = "multi_thread"))]
+#[serial]
+async fn test_dual_ring_bridge_10_nodes() -> Result<()> {
+    let config = TestConfig::builder()
+        .test_name("10-node dual ring bridge (poll)")
+        .node_count(10)
```
gknopf-aranya, line 53: Is there a default node count if this isn't configured?

crates/aranya-client/tests/ring_convergence.rs
```diff
@@ -0,0 +1,756 @@
...
+        .max_duration(Duration::from_secs(120))
+        .build()?;
+
+    info!(
+        node_count = config.node_count,
+        "Starting 10-node dual ring bridge test (poll mode)"
+    );
+    run_convergence(
+        config,
+        Topology::Custom {
```
gknopf-aranya: Why is the topology a separate arg vs. passing it in as part of the config to `run_convergence()`?

gknopf-aranya: I can see a use-case where you want to test the same config with different topologies. But you can achieve that by modifying the config before passing it into each `run_convergence()` call.

scripts/duvet-report.bash
gknopf-aranya: Is that related to the table at the top of the specs? I didn't realize we had YAML in the specs.

djohnson-spideroak: We put frontmatter in specs in aranya-docs.

crates/aranya-client/tests/scale/init.rs
djohnson-spideroak: Yes, this version is always single machine. This harness will not scale to 1000 nodes with full daemons unless you have a very large machine.

crates/aranya-client/tests/scale/metrics.rs
djohnson-spideroak: Given that times.len() is getting cast, I think we would run out of memory if the array had u32::max before hitting this panic.

crates/aranya-client/tests/scale/metrics.rs
djohnson-spideroak: I will move it to its own method, but standard deviation is simple and standard enough we do not need to introduce a dependency.

crates/aranya-client/tests/ring_convergence.rs
```diff
@@ -0,0 +1,756 @@
...
+    run_convergence(config, Topology::Ring).await?;
+    info!("60-node ring convergence test (poll mode) completed successfully");
+    Ok(())
+}
+
+/// Tests ring convergence with 60 nodes using hello sync mode.
+#[test(tokio::test(flavor = "multi_thread"))]
+#[serial]
+#[ignore = "Long-running test - run with: cargo test --test ring_convergence test_ring_convergence_60_nodes_hello -- --ignored"]
+async fn test_ring_convergence_60_nodes_hello() -> Result<()> {
```
benz-spideroak: There's a lot of repeated code in these tests. What about parameterized tests or a macro? That makes it easier to add new sizes and makes it easier to fix issues in the tests.

crates/aranya-client/tests/scale/metrics.rs
```diff
@@ -0,0 +1,240 @@
...
+            _ => return,
+        };
+
+        let command_issued = self.tracker.timestamps.command_issued;
+
+        //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#perf-006
+        //# The CSV output MUST include one row per node with the following columns: node index, label assignment time (T0), node convergence time, and convergence duration (time from T0 to node convergence).
+        let mut csv = String::from("node_index,t0_secs,convergence_time_secs,duration_secs\n");
+
+        let t0 = 0.0_f64; // T0 is the reference point (0)
```
benz-spideroak: What's the purpose of this?

crates/aranya-client/tests/ring_convergence.rs
djohnson-spideroak: wait_for_convergence has its own timeout internal to the function.

crates/aranya-client/tests/scale/init.rs
djohnson-spideroak: unreachable with 0 nodes, the function bails above if not all nodes started successfully.

crates/aranya-client/tests/scale/topology.rs
```diff
@@ -0,0 +1,142 @@
...
+///                   |
+///               bridge (4 <-> 5)
+///                   |
+///           +-------+-------+
+///           |               |
+///   Ring B: 5 - 6 - 7 - 8 - 9
+/// ```
+///
+/// Requires `n` to be even and `n >= 6` (each ring needs at least 3 nodes).
+pub fn dual_ring_bridge_topology(n: usize) -> Vec<Vec<NodeIndex>> {
```
gknopf-aranya, line 60: Could refactor this so it's more object-oriented:
```suggestion
pub enum TopologyType {
    ...
}

struct Topology {
    type: TopologyType,
    topology: Vec<Vec<NodeIndex>>,
}

impl Topology {
    pub fn dual_ring_bridge(n: usize) -> Self {
}
```

crates/aranya-client/tests/ring_convergence.rs
```diff
@@ -0,0 +1,300 @@
...
+// ---------------------------------------------------------------------------
+// Configuration validation
+// ---------------------------------------------------------------------------
+
+/// Tests that invalid configurations are rejected.
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_ring_config_validation() -> Result<()> {
+    // Test that < 3 nodes is rejected
+    let result = TestConfig::builder().node_count(2).build();
+    assert!(result.is_err(), "Should reject node_count < 3");
```
gknopf-aranya, line 281:
```suggestion
    TestConfig::builder().node_count(2).build().expect_err("Should reject node_count < 3");
```

crates/aranya-client/tests/ring_convergence.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    let result = TestConfig::builder().node_count(1).build();
+    assert!(result.is_err(), "Should reject node_count < 3");
+
+    // Test that >= 3 nodes is accepted
+    let result = TestConfig::builder().node_count(3).build();
+    assert!(result.is_ok(), "Should accept node_count >= 3");
```
gknopf-aranya, line 288:
```suggestion
    let result = TestConfig::builder().node_count(3).build().expect("Should accept node_count >= 3");
```

crates/aranya-client/tests/scale/topology.rs
djohnson-spideroak: I decided not to do this, because then we have no way to demonstrate a custom topology. I also really dont want to add in the complexity of creating a struct and such for topology when we have no need for it to live very long. The topology enums are consumed and not used again past the setup.

benz-spideroak (dismissed):

gknopf-aranya (dismissed):
lgtm, none of the remaining feedback I left seems worth blocking on

crates/aranya-client/tests/scale/team.rs
```diff
@@ -0,0 +1,212 @@
...
+        info!(
+            node_count = self.nodes.len() - 1,
+            "Adding all nodes to team as members"
+        );
+
+        //= https://raw.githubusercontent.com/aranya-project/aranya-docs/refs/heads/main/docs/multi-daemon-convergence-test.md#team-004
+        //# Each non-owner node MUST be added as a team member by the owner.
+        for node in &self.nodes[1..] {
+            owner_team
+                .add_device(node.pk.clone(), None, Rank::new(0))
```
gknopf-aranya, line 106: Adding the device with rank zero means it will only be able to operate on other objects with negative rank. If you're not planning to have the devices perform any operations, this should be okay.

PR #714, opened 2026-02-09
docs: update AGENTS.md with more focused guidance

djohnson-spideroak:
Refactor AGENTS.md to be more focused. Ensure policy is correctly described, and that common issues are directly addressed in the file by establishing a correct model of the system. 

Key additions:
- "What is Aranya" conceptual overview with control/data plane distinction
- Crate map table covering all 8 crates
- Detailed policy section: vocabulary, execution model (action runs once on author vs command policy runs on every node), ordering/recall via the weave algorithm, and comprehensive language restrictions including bounded execution guarantees
- Clear warnings about never putting plaintext secrets in command fields

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,24 @@
+{
+  "permissions": {
+    "allow": [
+      "WebFetch(domain:aranya-project.github.io)",
+      "WebFetch(domain:github.com)",
+      "WebFetch(domain:raw.githubusercontent.com)",
+      "Bash(cargo check *)",
+      "Bash(cargo build *)",
+      "Bash(find *)",
+      "Bash(cat *)",
+      "Bash(head *)",
+      "Bash(tail *)"
+    ],
+    "deny": [
+      "Bash(git push *)",
+      "Bash(curl *)",
+      "Bash(wget *)"
+    ]
+  },
+  "env": {
+  },
+  "companyAnnouncements": [
+  ]
+}


AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+8. The `policy` block runs again -- same checks, same finish logic. The receiving device must be able to fully evaluate the command with only the command fields and its local FactDB. This is why command fields cannot contain plaintext secrets.
+
+### Ordering and recall
 
-## Build & Test Commands
+- The DAG stores commands with parent pointers forming a causal graph. A deterministic **weave** algorithm produces a total ordering of all commands, respecting causality and using command **priority** to break ties between concurrent (non-causal) commands.
+- A command can be **recalled** if it was accepted in one ordering but fails when a new command arrives and changes the weave. The `recall` block (if defined) handles this by emitting compensating effects and/or adjusting facts. Check failures fall to recall; runtime exceptions use a default handler.
+- Policy writers must design commands so that concurrent mutations to the same facts don't create unresolvable conflicts.
+
+### Language restrictions
```
mooreoak, line 79: Dose it help if we link here to the documetation?

djohnson-spideroak: I am not sure. I do think it would help, but it would require that we keep WebFetch enabled, and I would like to disable WebFetch. It would be much better if we had docs in the same repo as everything else. 

Or maybe we can add instructions on cloning the docs to a tempdir?

AGENTS.md
```diff
@@ -1,21 +1,102 @@
 # AGENTS.md
 
-This document follows the [agents.md](https://agents.md) guidance and applies to any AI pair programmer or code assistant working in this repository.
```
gknopf-aranya, line 3: This note that was removed contains useful information

djohnson-spideroak: I disagree, I dont see anything useful on that page aside from general advice and that AGENTS.md is cross-program compatible. Maybe the stuff about nested AGENTS.md but we only have a single document, and that behavior is likely not consistent across coding agents.

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+## What is Aranya
 
-## Repository Purpose
+Aranya is a zero-trust security framework for decentralized applications.
 
-Aranya is a zero-trust security framework for decentralized applications. Devices join a shared CRDT-backed graph, and every command they can publish is mediated by the policy defined in `crates/aranya-daemon/src/policy.md`. That policy is a literate markdown file written in Aranya Policy Language v2 and compiled (via `policy-ifgen`) into the daemon during build.
+- Devices form a **team**, sharing a CRDT-backed DAG (directed acyclic graph).
+- Every mutation is a **command**. Commands are synced to all devices and each device independently enforces the policy before accepting a command into its local graph (endpoint enforcement).
+- The policy defines RBAC roles (owner / admin / operator / member), device lifecycle, label-based access control, and AFC (Aranya Fast Channels) permissions.
+- There is no central authority -- every device stores the full graph and enforces the same policy. Devices converge via CRDT sync.
```
gknopf-aranya: "full graph" isn't quite true. Every device has a graph of all the commands they have generated or synced up to that point. If all the devices sync without generating any new commands they would each have the full graph.

AGENTS.md
gknopf-aranya: You're right. I assumed this link was providing useful guidelines on how to write a good AGENTS.md file. Referencing guidelines on how to write a good AGENTS.md file seems useful for maintaining this document. While I understand that the README.md is for human consumption and AGENTS.md is for AI, that doesn't provide any actionable suggestions on how to write them differently to optimize for each.

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+
+On **every other device** (runs on sync):
+7. The command envelope arrives via sync. The `open` block deserializes it.
+8. The `policy` block runs again -- same checks, same finish logic. The receiving device must be able to fully evaluate the command with only the command fields and its local FactDB. This is why command fields cannot contain plaintext secrets.
+
+### Ordering and recall
 
-## Build & Test Commands
+- The DAG stores commands with parent pointers forming a causal graph. A deterministic **weave** algorithm produces a total ordering of all commands, respecting causality and using command **priority** to break ties between concurrent (non-causal) commands.
+- A command can be **recalled** if it was accepted in one ordering but fails when a new command arrives and changes the weave. The `recall` block (if defined) handles this by emitting compensating effects and/or adjusting facts. Check failures fall to recall; runtime exceptions use a default handler.
```
ssweetney-spok:
```suggestion
- The DAG stores commands with parent pointers forming a causal graph. A deterministic **braid** algorithm produces a total ordering of all commands, respecting causality and using command **priority** to break ties between concurrent (non-causal) commands.
- A command can be **recalled** if it was accepted in one ordering but fails when a new command arrives and changes the braid. The `recall` block (if defined) handles this by emitting compensating effects and/or adjusting facts. Check failures fall to recall; runtime exceptions use a default handler.
```

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+- One daemon per device; it owns crypto, policy enforcement, and graph sync.
+- Client libraries (Rust + C bindings) issue RPCs to the local daemon.
+- Sync uses QUIC transport; CRDT semantics ensure all devices converge.
+- Keys are managed by the daemon's keystore (`crates/aranya-daemon/src/keystore.rs`).
+
+## Crate Map
+
+| Crate | Purpose |
+|---|---|
+| `aranya-daemon` | Long-running daemon: policy, crypto, sync, device/team state |
```
gknopf-aranya, line 34: Primary purpose of the daemon is to sync in the background so the application doesn't need to

AGENTS.md
gknopf-aranya, line 1: How often should we update this file to keep it up-to-date with the repo? Updating it on each PR seems too frequent. Once a quarter/release seems reasonable. I wonder if there's a way we tell claude we want it to update the file to reflect the current state of the repo as we make changes to reduce the manual overhead of maintaining this file.

djohnson-spideroak: Next on my list is to create an "update docs" or "sync docs" skill to handle this.

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+| `aranya-client` | Rust client library (tarpc RPC to daemon) |
+| `aranya-client-capi` | C bindings (cdylib) for aranya-client |
+| `aranya-daemon-api` | Shared RPC/service definitions |
+| `aranya-keygen` | CLI for generating device keys |
+| `aranya-certgen` | CLI for generating root CA / signed P-256 certs |
+| `aranya-util` | Shared utilities (QUIC helpers, async) |
+| `aranya-metrics` | Observability: Datadog/Prometheus/TCP exporters |
+
+Examples live in `examples/rust/` and `examples/c/`.
```
gknopf-aranya, line 43:
```suggestion
Rust example applications live in `/examples/rust/` while C example applications live in `/examples/c/`.
```

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+
+## Policy
+
+The policy is the heart of Aranya. It lives at `crates/aranya-daemon/src/policy.md` as a literate markdown file (>100 KB). Only code inside ` ```policy ``` ` fences is compiled (via `policy-ifgen` at build time). The surrounding prose documents invariants -- always update prose when changing code.
```
gknopf-aranya, line 47: If there are users interested in building applications directly on top of the policy without the daemon, I wonder if we may want to move the policy to its own crate eventually.

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+| `aranya-client-capi` | C bindings (cdylib) for aranya-client |
+| `aranya-daemon-api` | Shared RPC/service definitions |
+| `aranya-keygen` | CLI for generating device keys |
+| `aranya-certgen` | CLI for generating root CA / signed P-256 certs |
+| `aranya-util` | Shared utilities (QUIC helpers, async) |
+| `aranya-metrics` | Observability: Datadog/Prometheus/TCP exporters |
+
+Examples live in `examples/rust/` and `examples/c/`.
+
+## Policy
```
gknopf-aranya, line 45: I would point claude to our full policy docs website so it has built-in context on the policy before reading the policy.md file.

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+1. Application calls an **action** via the client library.
+2. The action runs locally: it can do sensitive work, iterate facts, and `publish` zero or more **commands**.
+3. For each command, the `seal` block serializes it into an envelope.
+4. The `policy` block runs: it queries facts, runs `check` statements, and terminates with a `finish` block.
+5. The `finish` block performs fact mutations (`create`/`update`/`delete`) and `emit`s effects.
+6. If all commands succeed, the action is committed atomically. If any command fails, the entire action is rolled back.
+
+On **every other device** (runs on sync):
+7. The command envelope arrives via sync. The `open` block deserializes it.
+8. The `policy` block runs again -- same checks, same finish logic. The receiving device must be able to fully evaluate the command with only the command fields and its local FactDB. This is why command fields cannot contain plaintext secrets.
```
gknopf-aranya: There are actions, commands, and effects. From what I can tell, actions/commands are mentioned in these docs but not effects.

AGENTS.md
gknopf-aranya: Is there a way to allow WebFetch for our own organization docs/repos/websites?

djohnson-spideroak: I think we might be able to only allow specific command prefixes, but I am not sure. It might be easiest to just have it clone the repo.

AGENTS.md
djohnson-spideroak: I agree, but I dont want to add that to the context for every request. We can have those guidelines in the skill for updating this file.

AGENTS.md
djohnson-spideroak: I dont think we need the extra prose, it seems obvious based on the path IMO.

AGENTS.md
djohnson-spideroak: If we arent allowing webfetch, we will end up with the same solution we are talking about above.  https://github.com/aranya-project/aranya/pull/714#discussion_r2783868305

AGENTS.md
gknopf-aranya, line 1: This mentions the policy, DAG, and syncing. But there are core features such as AFC, custom roles, PSK/mTLS auth that don't have an overview.

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+- Devices form a **team**, sharing a CRDT-backed DAG (directed acyclic graph).
+- Every mutation is a **command**. Commands are synced to all devices and each device independently enforces the policy before accepting a command into its local graph (endpoint enforcement).
+- The policy defines RBAC roles (owner / admin / operator / member), device lifecycle, label-based access control, and AFC (Aranya Fast Channels) permissions.
+- There is no central authority -- every device stores the full graph and enforces the same policy. Devices converge via CRDT sync.
+- Separation of duties is enforced: no device can self-promote or bypass the role hierarchy.
+- The system has two planes:
+  - **Control plane (on-graph):** commands stored in the DAG, broadcast to all devices via sync. Used for access-control operations (add device, assign role, manage labels). Low throughput (~100s msgs/sec), high resilience.
+  - **Data plane (off-graph/AFC):** encrypted point-to-point channels between devices, governed by labels in policy. High throughput, low latency. Keys are negotiated on-graph but data flows off-graph.
```
gknopf-aranya, line 14: Can we reference an AFC spec here?

AGENTS.md
gknopf-aranya: Yeah, it should be able to clone the repo since it is public if it is allowed to run `git clone`. Do we want to prevent claude from cloning public repos that aren't ours? Cloning repos seems like a similar attack vector as WebFetch.

AGENTS.md
gknopf-aranya: Was about to suggest "Examples live in `/examples/`" for that reason. But, it's useful to know that there are both C and Rust examples without looking inside that folder first.

AGENTS.md
gknopf-aranya: I like the scope of "update docs". Periodically checks the repo for out-of-date docs and updates them based on recent changes. It's low-risk to use claude for this since it isn't changing any functionality. If we can find a way to incorporate that into release cycles before releases that would be an awesome way to avoid doc patch releases.

AGENTS.md
```diff
@@ -1,21 +1,102 @@
...
+### Ordering and recall
 
-## Build & Test Commands
+- The DAG stores commands with parent pointers forming a causal graph. A deterministic **weave** algorithm produces a total ordering of all commands, respecting causality and using command **priority** to break ties between concurrent (non-causal) commands.
+- A command can be **recalled** if it was accepted in one ordering but fails when a new command arrives and changes the weave. The `recall` block (if defined) handles this by emitting compensating effects and/or adjusting facts. Check failures fall to recall; runtime exceptions use a default handler.
+- Policy writers must design commands so that concurrent mutations to the same facts don't create unresolvable conflicts.
+
+### Language restrictions
+
+Policy Language v2. The language guarantees **bounded execution** -- there are no arrays, no general-purpose loops, and no recursion. Every policy evaluation is guaranteed to terminate. The only loop-like construct is `map` (iterate over matching facts), and it is restricted to actions only. Key restrictions:
```
ssweetney-spok, line 81: > there are no arrays, no general-purpose loops, and no **recursion**

This is also present in the [language docs](https://aranya-project.github.io/policy-book/reference/expressions/functions.html?highlight=recursion#functions) but you can have recursion currently. I guess this is not intentional

cc: @chip-so

ssweetney-spok:
```
// Add a test case like this in `crates/aranya-policy-vm/tests/vm.rs`

action foo(val int) {
    check val == bar(0)
}

function bar(x int) int {
    if x > 42 {
        return 0
    }

    if x == 42 {
        return x
    }

    let b = saturating_add(x,1)
    return bar(b)
}
```

AGENTS.md
ssweetney-spok: > Or maybe we can add instructions on cloning the docs to a tempdir?

That seems like a good idea. I'm not sure that we should duplicate the language docs in this file since it's very likely for it to be out of sync

djohnson-spideroak: Given we will be able to restrict webfetch (first to just github, then to specific paths with hooks), I will add in links to the docs. The links will also include the repo, so it can clone if needed.

AGENTS.md
chip-so: Yeah, recursion is definitely not intentional. We just haven't blocked it in the compiler yet. Please don't implement fibonacci in policy lang. :D

mooreoak: I believe it was innately impossible due to the way the compiler was built but we improved the compiler and accidentally introduced it.

djohnson-spideroak:
> From what I can tell `CLAUDE.md` is supposed to be for general context while `AGENTS.md` is supposed to provide more specific context for particular sub-agents/tasks that may not need as much context to perform more specific operations. Unless we have particular sub-agents/tasks in mind I believe the preference is to include information like this in the `CLAUDE.md` file. Maintaining both files with overlapping information could create more maintenance overhead for us unless we have a specific reason for doing so.

CLAUDE.md is a symlink to AGENTS.md. I also dont share that assumption, to me AGENTS.md and CLAUDE.md serve the same purpose. The only difference is that AGENTS.md works across many harnesses (codex, claude code, opencode, etc) while CLAUDE.md is claude only. I would end up with the same info in both files if we split them IMO.

AGENTS.md
```diff
@@ -1,21 +1,115 @@
 # AGENTS.md
 
-This document follows the [agents.md](https://agents.md) guidance and applies to any AI pair programmer or code assistant working in this repository.
+## What is Aranya
 
-## Repository Purpose
+Aranya is a zero-trust security framework for decentralized applications.
 
-Aranya is a zero-trust security framework for decentralized applications. Devices join a shared CRDT-backed graph, and every command they can publish is mediated by the policy defined in `crates/aranya-daemon/src/policy.md`. That policy is a literate markdown file written in Aranya Policy Language v2 and compiled (via `policy-ifgen`) into the daemon during build.
+- Devices form a **team**, sharing a CRDT-backed DAG (directed acyclic graph).
```
jnewell-spok, line 7: Would it be beneficial to expand out all acronyms- atleast once in the document? I see that DAG was expanded, but there are some others that aren't.

AGENTS.md
```diff
@@ -1,21 +1,115 @@
...
+This spins up daemons and exercises the integration suite against the new policy.
+
+### AFC (Aranya Fast Channels)
+
+AFC provides **high-throughput, low-latency encrypted channels** between devices on the data plane. Key concepts:
+
+- Channels are **unidirectional** (one sender, one receiver). Created via the `create_afc_uni_channel(receiver_id, label_id)` ephemeral action.
+- Access is governed by **labels**. A label is a named topic with a unique `label_id`; each label has one or more managing roles. Devices are assigned a `ChanOp` per label: `SendOnly`, `RecvOnly`, or `SendRecv`.
+- Channel creation is **ephemeral** -- it does not persist to the DAG or mutate the FactDB. The sender generates a key (RFC 9180 KEM), encapsulates it for the receiver, and both sides store the key in shared memory.
+- Both peers validate the channel: the label must exist, sender must have send permission, receiver must have receive permission, and both must have `CanUseAfc`.
```
jnewell-spok, line 109: should the ChanOp permissions be specifically stated here? (sender must have send permission...)

AGENTS.md
djohnson-spideroak: I think most of them are obvious in context, especially for an LLM. Aside from CRDT, most of the other are common (FFI, RBAC, etc.). I will expand CRDT, but the others I think are unambiguous. Are there specific ones you think we should expand?

AGENTS.md
djohnson-spideroak: I think its fine, we mention ChanOp 2 points above and it still mentions that they need send/recv permissions.

Issue #650, opened 2025-12-02
custom roles: rank


gknopf-aranya:
Implement a rank based approach to custom roles RBAC.

Each object in system has a rank associated with its Aranya ID (role, device, label, etc.).

Permissions grant objects permission to perform certain operations in the system.
Rank determines which objects can operate on other objects in the system.
An object must have higher rank than the objects it is operating on and permission to perform the operation in order to perform an operation.

cc: @djohnson-spideroak, @mooreoak, @jdygert-spok, @chip-so

Issue #638, opened 2025-11-19
custom roles: move out from behind preview flag


gknopf-aranya:
Once we're happy with the custom roles API, we can move it out from behind the preview flag.

cc: @djohnson-spideroak, @chip-so, @jdygert-spok

PR #715, opened 2026-02-10
custom roles: implement rank changes according to the policy

gknopf-aranya:
Implement custom roles rank changes:
- APIs (Rust, C, daemon IPC)
- Unit tests
- Rust/C examples

Solves:
#650, #638, #734

### DIFF
--- a/examples/rust/aranya-example-multi-node/src/admin/main.rs
+++ b/examples/rust/aranya-example-multi-node/src/admin/main.rs
@@ -4,7 +4,7 @@ use std::path::PathBuf;
 
 use anyhow::Result;
 use aranya_client::{
-    text, AddTeamConfig, AddTeamQuicSyncConfig, Client, ObjectId, Rank, SyncPeerConfig,
+    text, AddTeamConfig, AddTeamQuicSyncConfig, Client, ObjectId, Rank, Role, SyncPeerConfig,
 };
 use aranya_example_multi_node::{
     env::EnvVars,
@@ -113,9 +113,15 @@ async fn main() -> Result<()> {
 
     // Create label.
     info!("admin: creating label");
-    let admin_device_rank = team.query_rank(ObjectId::transmute(device_id)).await?;
-    // Label rank must be lower than the admin's device rank so the admin can operate on it.
-    let label_rank = Rank::new(admin_device_rank.value().saturating_sub(1));
+    let member_role = team
+        .roles()
+        .await?
+        .into_iter()
+        .find(|r: &Role| r.name == "member")
+        .ok_or_else(|| anyhow::anyhow!("member role not found"))?;
+    let member_role_rank = team.query_rank(ObjectId::transmute(member_role.id)).await?;
+    // Label rank must be lower than the member role rank so all team members can operate on it.
+    let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     team.create_label_with_rank(text!("label1"), label_rank)
         .await?;
     info!("admin: created label");
--- a/crates/aranya-client-capi/ctest/test_onboarding.c
+++ b/crates/aranya-client-capi/ctest/test_onboarding.c
@@ -86,10 +86,22 @@ static int test_create_team_and_onboard_member(const char *tmpdir) {
 
     printf("Got member key bundle (%zu bytes)\n", member_pk_len);
 
-    /* Add member to team */
+    /* Query owner's device rank to derive the new device's rank */
+    AranyaDeviceId owner_device_id;
+    CLIENT_EXPECT("Failed to get owner device id", "",
+                  aranya_get_device_id(&owner_client, &owner_device_id));
+
+    AranyaObjectId owner_object_id = {.id = owner_device_id.id};
+    int64_t owner_device_rank = 0;
+    CLIENT_EXPECT("Failed to query owner device rank", "",
+                  aranya_query_rank(&owner_client, &team_id, &owner_object_id,
+                                    &owner_device_rank));
+
+    /* Add member to team with rank below owner's device rank */
     CLIENT_EXPECT("Failed to add member to team", "",
-                  aranya_add_device_to_team(&owner_client, &team_id, member_pk,
-                                            member_pk_len, NULL));
+                  aranya_add_device_to_team_with_rank(
+                      &owner_client, &team_id, member_pk, member_pk_len, NULL,
+                      owner_device_rank - 1));
 
     printf("Member added to team successfully\n");
 


crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -458,6 +458,49 @@ impl From<&LabelId> for aranya_client::LabelId {
     }
 }
 
+/// A generic object ID that can refer to a device, role, or label.
```
gknopf-aranya: This could refer to any object defined in the Aranya policy. More object types could be added in the future.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -458,6 +458,49 @@ impl From<&LabelId> for aranya_client::LabelId {
...
+///
+/// Used with [`change_rank`] and [`query_rank`].
```
gknopf-aranya: Don't see a reason to document which methods use ObjectId here. That could change later and this would need to be updated.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -458,6 +458,49 @@ impl From<&LabelId> for aranya_client::LabelId {
...
+    }
+}
+
+impl From<&ObjectId> for aranya_client::ObjectId {
+    fn from(value: &ObjectId) -> Self {
+        value.id.bytes.into()
+    }
+}
+
+impl From<&DeviceId> for ObjectId {
```
gknopf-aranya: Or should we convert ObjectId -> BaseId -> DeviceId, ObjectId -> BaseId -> RoleId, etc. so ObjectId doesn't need conversions to/from all the different object types?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -504,35 +546,7 @@ impl Addr {
...
-            RoleManagementPermission::CanChangeRolePerms => Self::CanChangeRolePerms,
-        }
-    }
-}
-
 /// Simple permission.
-#[cfg(feature = "preview")]
 #[repr(u8)]
 #[derive(Copy, Clone, Debug)]
 pub enum Permission {
```
gknopf-aranya, line 535: May not want to take `Permission` out of preview since the permission list seems like something likely to change frequently in the future. Should we consider exposing strings instead of an enum for them so they can change more easily in the future without breaking things?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1204,13 +1210,17 @@ pub fn hello_subscription_config_builder_set_periodic_interval(
...
 /// @param[in] client the Aranya Client
 /// @param[in] team the team's ID
-/// @param[in] owning_role the ID of the owning role
-/// @param[in] roles_out returns a list of roles that own `role`
+/// @param[in] owning_role (ignored) formerly the owning role
+/// @param[in] roles_out returns a list of default roles
 /// @param[in,out] roles_len the number of roles written to the buffer.
 ///
 /// @relates AranyaClient.
+#[allow(deprecated)]
```
gknopf-aranya: Deprecation of methods in the C API should follow same convention as the mTLS PR

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1244,69 +1249,19 @@ pub unsafe fn setup_default_roles(
...
-
-    Ok(())
-}
-
-/// Returns the roles that own `role`.
-///
-/// Returns an `AranyaBufferTooSmall` error if the output buffer is too small to hold the roles.
-/// Writes the number of roles that would have been returned to `roles_len`.
-/// The application can use `roles_len` to allocate a larger buffer.
+/// Role ownership has been replaced by rank-based authorization.
```
gknopf-aranya: Don't need to mention role ownership here if that is being removed from the API.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1244,69 +1249,19 @@ pub unsafe fn setup_default_roles(
...
+/// This function always returns an empty list.
 ///
 /// @param[in] client the Aranya Client
 /// @param[in] team the team's ID
 /// @param[in] role the ID of the subject role
-/// @param[in] roles_out returns a list of roles that own `role`
+/// @param[out] roles_out returns a list of roles that own `role`
```
gknopf-aranya: We should document that roles_out should be ignored and now returns an empty list. We can no longer return a list of role owners since the concept of role owners has been removed. This could be a breaking change for any application code relying on a non-empty list of role owners to be returned.

gknopf-aranya: A safer approach could be to return a list that always contains the team owner role as the only entry. If that role is used later by another method, it will be ignored. This approach decreases the chance of breaking an application that may rely on the previous behavior of this method returning a non-empty list of owning roles. This will not solve the case where an application relies on there being multiple owning roles. If we want to make the API completely backward compatible, then I believe we would need a counter to keep track of how many dummy owning roles there are in order to return a list that is the expected size.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I don't think `role_owners()` should be part of the daemon API anymore since we've removed the concept of owning roles.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1468,12 +1369,11 @@ pub unsafe fn team_roles(
 /// @param[out] role_out the newly created role
 ///
 /// @relates AranyaClient
-#[cfg(feature = "preview")]
 pub fn create_role(
     client: &Client,
     team: &TeamId,
     role_name: RoleName,
-    owning_role: &RoleId,
+    rank: i64,
```
gknopf-aranya, line 1415: Still deciding whether to create a `Rank` type for this or not. Passing in an integer for the rank is very convenient and it is unlikely the type will change since all numbers in the policy are `i64`.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1661,81 +1590,105 @@ pub fn delete_label(client: &Client, team: &TeamId, label_id: &LabelId) -> Resul
     Ok(())
 }
 
-/// Assign a label to a device so that it can be used for a channel.
+/// Change the rank of an object (device, role, or label).
+///
+/// The caller must know the current rank of the object. The caller's
```
gknopf-aranya: Explain that the reason why the caller must know the current rank of the object is to prevent a race where another device could change the rank at the same time. By confirming that the current rank is expected, the device's intent to change the rank under certain conditions is preserved. It's possible they wouldn't want to change the rank if it was anything other than what was expected.

crates/aranya-client/src/client/label.rs
```diff
@@ -23,6 +23,12 @@ pub struct Label {
     pub id: LabelId,
     /// The human-readable label name.
     pub name: Text,
+    /// The label's rank in the authorization hierarchy.
+    ///
+    /// Populated when available (e.g., on creation). Use
+    /// [`Team::query_rank`](crate::client::team::Team::query_rank)
+    /// to retrieve the rank for labels returned by queries.
+    pub rank: Option<Rank>,
```
gknopf-aranya: A label should always have a rank at the time of creation. While it is possible to change the rank, it cannot be unset. So this field doesn't need to be optional.

gknopf-aranya: Questioning whether rank needs to be a field of `Label`. If someone wants to know the rank, they can query for it.

examples/rust/aranya-example-multi-node/src/owner/main.rs
```diff
@@ -81,7 +81,7 @@ async fn main() -> Result<()> {
             panic!("unexpected roles");
         }
         assert_eq!(owner_role.name, "owner");
-        team.setup_default_roles(owner_role.id)
+        team.setup_default_roles_no_owner()
```
gknopf-aranya: There is still an owner role, the change is that there is no longer an owning role of the default roles:
```suggestion
        team.setup_default_roles_no_owning_role()
```

examples/rust/aranya-example-multi-node/src/admin/main.rs
```diff
@@ -111,23 +111,9 @@ async fn main() -> Result<()> {
 
     // Create label.
     info!("admin: creating label");
-    let label1 = team.create_label(text!("label1"), admin_role.id).await?;
+    let _label1 = team.create_label_with_rank(text!("label1"), aranya_client::Rank::new(100)).await?;
```
gknopf-aranya: If `_label1` is unused, there is no reason to define it:
```suggestion
   team.create_label_with_rank(text!("label1"), aranya_client::Rank::new(100)).await?;
```

examples/rust/aranya-example/src/main.rs
```diff
@@ -237,15 +237,15 @@ async fn main() -> Result<()> {
 
     // Create default roles
     info!("creating default roles");
-    let owner_role = owner
+    let _owner_role = owner
```
gknopf-aranya:
```suggestion
    owner
```

examples/rust/aranya-example/src/main.rs
```diff
@@ -376,83 +376,76 @@ async fn main() -> Result<()> {
...
 
-        // Remove `CanUseAfc` permission from the custom role.
-        info!("removing CanUseAfc permission from custom role");
-        owner_team
-            .remove_perm_from_role(custom_role.id, Permission::CanUseAfc)
-            .await?;
+    // Demo change_rank/query_rank: change the custom role's rank from 50 to 75.
+    let role_bytes: [u8; 32] = custom_role.id.into();
+    let role_object_id: ObjectId = role_bytes.into();
+    let current_rank = owner_team.query_rank(role_object_id).await?;
```
gknopf-aranya: We should have a cleaner `.into()` conversion from `RoleId` into `ObjectId`.

examples/rust/aranya-example/src/main.rs
```diff
@@ -376,83 +376,76 @@ async fn main() -> Result<()> {
...
+    info!("custom role rank before change: {}", current_rank);
+    assert_eq!(current_rank, Rank::new(50));
```
gknopf-aranya: How do we know rank 50 is the expected rank for this test?

examples/c/example.bash
```diff
@@ -98,7 +98,7 @@ for device in "${devices[@]}"; do
         --config "${cfg_path}" &
 done
 # give the daemons time to startup
-sleep 1
+sleep 5
```
gknopf-aranya: Try 2-3 seconds instead. The purpose of increasing is for slower machines but it shouldn't be that much slower.

crates/aranya-metrics/src/main.rs
```diff
@@ -279,13 +279,13 @@ async fn run_demo_body(ctx: DemoContext) -> Result<()> {
 
     // Create default roles
     info!("creating default roles");
-    let owner_role = owner
+    let _owner_role = owner
```
gknopf-aranya:
```suggestion
    owner
```

crates/aranya-daemon/src/policy.md
gknopf-aranya, line 1: These policy fixes should be moved to the base `650-custom-roles-rank` branch. They are fixes identified while working on the implementation that should be included as part of the custom roles rank spec.

crates/aranya-daemon/src/policy.md
```diff
@@ -269,19 +269,16 @@ function valid_device_invariants(device_id id) bool {
         check !exists DeviceEncPubKey[device_id: device_id]
     } else {
         // The device DOES exist, so the device keys MUST also
-        // exist and match the key IDs in `Device`.
-        let dev = unwrap device
+        // exist and their key IDs must match the derived IDs.
```
gknopf-aranya: Why was this comment changed?

crates/aranya-client/src/client/team.rs
```diff
@@ -109,15 +108,48 @@ impl Team<'_> {
...
-    pub async fn add_device(&self, keys: KeyBundle, initial_role: Option<RoleId>) -> Result<()> {
+    pub async fn add_device_with_rank(
+        &self,
+        keys: KeyBundle,
+        initial_role: Option<RoleId>,
+        rank: Rank,
+    ) -> Result<()> {
         self.client
             .daemon
             .add_device_to_team(
```
gknopf-aranya: Should call `add_device_to_team_with_rank()` here. `add_device_to_team()` is deprecated and may be removed in the future.

crates/aranya-client/src/client/team.rs
```diff
@@ -422,56 +371,88 @@ impl Team<'_> {
             .collect();
         Ok(Roles { roles })
     }
+
+    /// Returns the roles that own the given role.
+    ///
+    /// Role ownership has been replaced by rank-based authorization.
+    /// This method always returns an empty list. Use
+    /// [`Self::query_rank`] instead.
+    #[deprecated(note = "role ownership replaced by rank-based authorization; use `query_rank`")]
```
gknopf-aranya: `query_rank()` is not a replacement for `role_owners()`. There is no method that should be used instead, it shouldn't be used anymore but we're keeping it for now to not break the API.

crates/aranya-client/src/client/team.rs
```diff
@@ -422,56 +371,88 @@ impl Team<'_> {
...
 
 impl Team<'_> {
     /// Create a label.
+    ///
+    /// The `managing_role_id` parameter is accepted for backward
+    /// compatibility but is ignored in the rank-based authorization
+    /// model. The label is created with a default rank of the command
+    /// author's rank minus one.
+    ///
+    /// Use [`Self::create_label_with_rank`] (preview) to specify an
```
gknopf-aranya: `create_label_with_rank()` shouldn't be behind preview

crates/aranya-client/src/client/team.rs
gknopf-aranya: Nevermind, the IPC API doesn't need to maintain backward compatibility so `add_device_to_team()` is the more ergonomic API.

crates/aranya-client/src/client/team.rs
```diff
@@ -422,56 +371,88 @@ impl Team<'_> {
...
+            .map(LabelId::from_api)
     }
 
-    /// Add label managing role
-    #[instrument(skip(self))]
+    /// Adds a managing role to a label.
+    ///
+    /// Label managing roles have been replaced by rank-based
+    /// authorization. This method is a no-op. Use
+    /// [`Self::create_label_with_rank`] instead.
```
gknopf-aranya: `create_label_with_rank()` isn't a replacement for `add_label_managing_role()`.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -105,6 +105,35 @@ custom_id! {
     pub struct RoleId;
 }
 
+custom_id! {
+    /// A generic object ID that can refer to a device, role, or label.
```
gknopf-aranya: And any other objects with a unique Aranya ID defined in the Aranya policy.md

crates/aranya-daemon-api/src/service.rs
```diff
@@ -105,6 +105,35 @@ custom_id! {
...
+        Self(value)
+    }
+
+    /// Returns the raw rank value.
+    pub const fn value(self) -> i64 {
+        self.0
+    }
+}
+
+impl fmt::Display for Rank {
```
gknopf-aranya, line 131: Why not use the default `Display` derive?

crates/aranya-daemon-api/src/service.rs
```diff
@@ -113,6 +142,11 @@ pub struct Role {
     pub name: Text,
     /// The author of the role.
     pub author_id: DeviceId,
+    /// The role's rank in the authorization hierarchy.
+    ///
+    /// Populated when available (e.g., on creation). Use `query_rank`
+    /// to retrieve the rank for roles returned by queries.
+    pub rank: Option<Rank>,
```
gknopf-aranya: Remove the rank field from `Role`. Require user to query for rank.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -155,6 +189,11 @@ pub struct CreateTeamConfig {
 pub struct Label {
     pub id: LabelId,
     pub name: Text,
+    /// The label's rank in the authorization hierarchy.
+    ///
+    /// Populated when available (e.g., on creation). Use `query_rank`
+    /// to retrieve the rank for labels returned by queries.
+    pub rank: Option<Rank>,
```
gknopf-aranya: Remove rank field from `Label`. Require user to query for rank.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -285,10 +310,14 @@ pub enum SimplePerm {
     AddDevice,
     // The role can remove a device from the team.
     RemoveDevice,
-    // The role can terminate the team. This causes all team
-    // commands to fail until a new team is created.
+    // The role can terminate the team.
```
gknopf-aranya, line 322: Why was "This causes all team commands to fail until a new team is created" removed?

crates/aranya-daemon-api/src/service.rs
```diff
@@ -299,39 +328,25 @@ pub enum SimplePerm {
...
-    // with the `ChangeLabelManagingRole` permission above.
+    // The role can revoke a label from a device.
     RevokeLabel,
 
     // # AFC
     //
-    // The role can use AFC. This controls the ability to
-    // create or receive a unidirectional AFC channels.
+    // The role can use AFC.
     CanUseAfc,
```
gknopf-aranya, line 359: Why was "This controls the ability to create or receive a unidirectional AFC channels." removed?

crates/aranya-daemon-api/src/service.rs
```diff
@@ -411,11 +426,15 @@ pub trait DaemonApi {
...
-    /// Adds a device to the team with optional initial roles.
+    /// Adds a device to the team with optional initial role and rank.
+    ///
+    /// If `rank` is `None`, the daemon will use the command author's
+    /// rank minus one as the default.
     async fn add_device_to_team(
         team: TeamId,
         keys: KeyBundle,
         initial_role: Option<RoleId>,
+        rank: Option<Rank>,
```
gknopf-aranya: If someone calls add_device_to_team_with_rank() from an application API they should be able to call a method called `add_device_to_team_with_rank()` in this daemon IPC API that does not have an optional rank. Having an optional rank param could cause the rank to accidentally not be set.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -499,24 +499,17 @@ pub trait DaemonApi {
     //
 
     /// Create a label.
-    async fn create_label(team: TeamId, name: Text, managing_role_id: RoleId) -> Result<LabelId>;
+    ///
+    /// If `rank` is `None`, the daemon will use the command author's
+    /// rank minus one as the default.
+    async fn create_label(team: TeamId, name: Text, rank: Option<Rank>) -> Result<LabelId>;
```
gknopf-aranya: Should require rank to be specified here. Perform the rank arithmetic in the Rust API before calling this method.

crates/aranya-daemon/src/api.rs
```diff
@@ -653,13 +652,38 @@ impl DaemonApi for Api {
         team: api::TeamId,
         keys: api::KeyBundle,
         initial_role: Option<api::RoleId>,
+        rank: Option<api::Rank>,
     ) -> api::Result<()> {
         let graph = self.check_team_valid(team).await?;
+        // When rank is not specified and an initial_role is provided,
+        // default to role_rank - 1 so the policy check
```
gknopf-aranya: Move this logic into the client so daemon IPC API is closer to how we want it once deprecated custom roles methods are removed.

crates/aranya-client/tests/tests.rs
```diff
@@ -98,13 +98,13 @@ async fn test_sync_now() -> Result<()> {
         .await
         .context("owner unable to assign admin role")?;
 
-    // Now, we try to assign a role using the admin, which is expected to fail.
+    // Now, we try to create a label using the admin, which is expected to fail
```
gknopf-aranya: Why was this test changed from assigning a role to creating a label?

crates/aranya-client/tests/tests.rs
```diff
@@ -1629,521 +1638,798 @@ async fn test_owner_cannot_revoke_owner_role() -> Result<()> {
...
+    devices.add_all_device_roles(team_id, &roles).await?;
+
+    let owner_team = devices.owner.client.team(team_id);
+
+    let r = owner_team
+        .device(devices.membera.id)
+        .assign_role(roles.operator().id)
+        .await;
+
+    assert!(matches!(r, Err(aranya_client::Error::Aranya(_))));
```
gknopf-aranya: Check whether this matches our error checking conventions in other tests.

crates/aranya-client/tests/tests.rs
```diff
@@ -1629,521 +1638,798 @@ async fn test_owner_cannot_revoke_owner_role() -> Result<()> {
...
+// Rank-based tests
+// ========================================================================
+
+/// Tests that a role can be created with a specific rank and that the rank
+/// can be queried back.
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_create_role_with_rank() -> Result<()> {
+    let mut devices = DevicesCtx::new("test_create_role_with_rank").await?;
+    let team_id = devices.create_and_add_team().await?;
+    let _roles = devices.setup_default_roles(team_id).await?;
```
gknopf-aranya: If `_roles` is unused, don't define it. Fix here and elsewhere.
```suggestion
    devices.setup_default_roles(team_id).await?;
```

crates/aranya-client/tests/tests.rs
```diff
@@ -1629,521 +1638,798 @@ async fn test_owner_cannot_revoke_owner_role() -> Result<()> {
...
+
+    let owner_team = devices.owner.client.team(team_id);
+    let role = owner_team.create_role(text!("mutable_role"), Rank::new(50)).await?;
+
+    let role_bytes: [u8; 32] = role.id.into();
+    let role_obj: ObjectId = role_bytes.into();
+
+    owner_team.change_rank(role_obj, Rank::new(50), Rank::new(75)).await?;
```
gknopf-aranya: We need a test to show that changing rank will fail if `old_rank` is incorrect.

crates/aranya-client/tests/tests.rs
gknopf-aranya, line 1: Create a new test to demonstrate that the role rank can be set lower than the device rank. Explain that this should normally not be done, but there isn't an easy way to mitigate against this. Refer to policy.md for more details.

examples/c/example.c
```diff
@@ -1042,19 +984,6 @@ AranyaError run_custom_roles_example(Team *t) {
         goto exit;
     }
 
-    // Get the role assigned to 'admin'
-    AranyaRole admin_role;
-    printf("getting the role assigned to 'admin'.\n");
-    err = aranya_team_device_role(&admin->client, &t->id, &admin->id,
```
gknopf-aranya, line 1048: Is `aranya_team_device_role()` still covered by this example anywhere? Even if we aren't using it in the same way because of the role management changes if it is still part of the API we want to show an example of it. While checking this, confirm that the example covers any methods that are not preview/deprecated and flag anywhere there is a lack of coverage.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Looks like the C example is currently converting between ID types with `memcpy`. @jdygert-spok is this acceptable or should we use explicit type conversions in the C API for IDs?

jdygert-spok: I believe you can just assign the ID field in C, e.g. `my_object_id.id = my_device_id.id`

examples/c/example.c
```diff
@@ -1177,6 +1029,38 @@ AranyaError run_custom_roles_example(Team *t) {
     EXPECT("unable to add 'CanUseAfc' permission to 'buddy' role", err);
     printf("Assigned 'buddy' the 'CanUseAfc' permission\n");
 
+    // Demo change_rank/query_rank: change the buddy role's rank from 50 to 75.
+    AranyaObjectId buddy_object_id;
+    memcpy(buddy_object_id.id.bytes, buddy_role_id.id.bytes, ARANYA_ID_LEN);
```
jdygert-spok:
```suggestion
    AranyaObjectId buddy_object_id = { .id = buddy_role_id.id };
```

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Or construct it: https://github.com/aranya-project/aranya/pull/715/changes#r2790532038

jdygert-spok: We probably don't need these conversion impls in Rust.

gknopf-aranya: I've removed the conversions locally, haven't pushed the commit yet

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: For now I'll keep this as returning en empty list unless we decide to do something differently. I'll add better docs to make this more clear.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: @jdygert-spok, @chip-so any objections to using a `i64` for rank in the C API?

chip-so: No, I think that makes sense. We're treating them as numbers directly, so a numerical type is fine.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Reverting since this change isn't needed and increases diff noise.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Looks like the convention is `match` but `matches!` is cleaner:
```
      1667 -    let r = owner_team                                                                                                 
      1667 +    match owner_team                                                                                            
      1668          .device(devices.membera.id)                                                                                    
      1669          .assign_role(roles.operator().id)                                                                              
      1670 -        .await;                                                                                                 
      1670 +        .await                                                                                                  
      1671 +    {                                                                                                           
      1672 +        Ok(_) => bail!("Expected role assignment to fail"),                                                     
      1673 +        Err(aranya_client::Error::Aranya(_)) => {}                                                              
      1674 +        Err(_) => bail!("Unexpected error"),                                                                    
      1675 +    }                                                                                                           
      1676  
      1672 -    assert!(matches!(r, Err(aranya_client::Error::Aranya(_))));                                                 
      1673 -                                                                                                                
      1677      Ok(())
      1678  }
```
I'd prefer not to increase diff noise on this PR so we can possibly update the convention first in the rest of the tests on another PR.

gknopf-aranya: https://github.com/aranya-project/aranya/issues/716

crates/aranya-client/tests/tests.rs
gknopf-aranya: `test_change_rank_stale_old_rank_rejected` shows this

examples/c/example.c
gknopf-aranya: Yes, it's covered.

crates/aranya-client-capi/src/api/defs.rs
chip-so: I guess it depends on how we expect it to change. I don't think it should break stability to add new permissions, but if we're not sure we will continue to support this set, it should probably remain in preview.

gknopf-aranya: As long as we only add new permissions at the end this should remain backward compatible. Removing permissions or changing their name would break backward compatibility. We could print out a `warn!` if a deprecated permission is used.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1244,146 +1235,76 @@ pub unsafe fn setup_default_roles(
...
-///
-/// @relates AranyaClient.
-#[cfg(feature = "preview")]
-pub fn revoke_role_management_permission(
-    client: &Client,
-    team: &TeamId,
-    role: &RoleId,
-    managing_role: &RoleId,
-    perm: RoleManagementPermission,
+#[deprecated(note = "role_owners is deprecated")]
```
chip-so, line 1314: Seems like this is more than deprecated. It's entirely disabled.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: I like the warn/deprecation strategy because it allows for use of a stronger type while allowing for future changes without breaking backward compatibility. This could mean keeping deprecated permissions in the policy so enum conversions between APIs and the policy continue to work without some complicated type conversion code.

chip-so: I prefer that approach as well.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1204,13 +1190,18 @@ pub fn hello_subscription_config_builder_set_periodic_interval(
...
 /// @param[in] client the Aranya Client
 /// @param[in] team the team's ID
-/// @param[in] owning_role the ID of the owning role
-/// @param[in] roles_out returns a list of roles that own `role`
+/// @param[in] owning_role (ignored) formerly the owning role
+/// @param[in] roles_out returns a list of default roles
 /// @param[in,out] roles_len the number of roles written to the buffer.
 ///
 /// @relates AranyaClient.
+#[deprecated(note = "use `setup_default_roles_no_owning_role` instead")]
```
gknopf-aranya: Not a huge fan of referencing deprecated owning roles in a non-deprecated function. But, I haven't thought of another way to differentiate the new method from the original `setup_default_roles()` method.

gknopf-aranya: `setup_default_roles_v2()`?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1244,146 +1235,76 @@ pub unsafe fn setup_default_roles(
...
-            .team(team.into())
-            .remove_role_owner(role.into(), owning_role.into()),
-    )?;
-
-    Ok(())
-}
-
-/// Returns the roles that own `role`.
+/// This sets up the following roles with default permissions as
+/// defined in Aranya's default policy:
```
gknopf-aranya, line 1256: Should probably mention that the owner role was already set up when creating the team so doesn't need to be setup here.

crates/aranya-client/src/client/team.rs
```diff
@@ -109,15 +108,59 @@ impl Team<'_> {
...
+    #[deprecated(note = "use `add_device_with_rank` to specify an explicit rank")]
     #[instrument(skip(self))]
     pub async fn add_device(&self, keys: KeyBundle, initial_role: Option<RoleId>) -> Result<()> {
+        // Default to role_rank - 1 when an initial_role is provided,
+        // otherwise fall back to author_rank - 1.
+        let rank = match &initial_role {
+            Some(role_id) => {
+                let role_obj: ObjectId = (*role_id).into();
+                let role_rank = self.query_rank(role_obj).await?;
+                Rank::new(role_rank.value() - 1)
```
chip-so: Do we not have the arithmetic lints on this crate? This isn't necessarily wrong (well okay, it could underflow if the role already had `i64::MIN` and I can't recall if we guard against that elsewhere), I was just kind of surprised to see it expressed in the usual way.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1244,146 +1235,76 @@ pub unsafe fn setup_default_roles(
...
+            client
+                .inner
+                .team(team.into())
+                .setup_default_roles_no_owning_role(),
+        )?
         .__into_data();
 
-    if *roles_len < owning_roles.len() {
-        *roles_len = owning_roles.len();
+    debug_assert_eq!(DEFAULT_ROLES_LEN, default_roles.len());
```
gknopf-aranya: Should we return an error to the user here if the length isn't correct?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1244,146 +1235,76 @@ pub unsafe fn setup_default_roles(
...
 
-/// Assigns a role management permission to a managing role.
+/// Deprecated: always returns an empty list.
 ///
 /// @param[in] client the Aranya Client
 /// @param[in] team the team's ID
 /// @param[in] role the ID of the subject role
-/// @param[in] managing_role the ID of the managing role
-/// @param[in] perm the management permission to assign
+/// @param[out] roles_out unused, always returns an empty list
```
gknopf-aranya, line 1310: We can't guarantee that roles_out is unused, but we can recommend that the application ignore it.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1244,146 +1235,76 @@ pub unsafe fn setup_default_roles(
...
-/// @relates AranyaClient.
-#[cfg(feature = "preview")]
-pub fn revoke_role_management_permission(
-    client: &Client,
-    team: &TeamId,
-    role: &RoleId,
-    managing_role: &RoleId,
-    perm: RoleManagementPermission,
+#[deprecated(note = "role_owners is deprecated")]
+pub unsafe fn role_owners(
```
gknopf-aranya, line 1315: With the new rank-based approach another way of looking at role owners would be any roles that have a higher rank than the specified role. But since the policy doesn't support iteration it's simpler to keep this as a no-op. We could potentially implement that query with `map` but it doesn't seem worth it.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1620,14 +1537,20 @@ pub fn revoke_role(
 
 /// Create a channel label.
 ///
+/// The `managing_role_id` parameter is accepted for backward compatibility
+/// but is ignored in the rank-based authorization model. The label is
+/// created with a default rank of the command author's rank minus one.
```
gknopf-aranya: Since this API does not allow the user to specify a rank.

chip-so (commented):
Mostly just skimmed the client implementations so far, but I thought these were worth noting.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1661,81 +1611,107 @@ pub fn delete_label(client: &Client, team: &TeamId, label_id: &LabelId) -> Resul
...
-            .inner
-            .team(team.into())
-            .device(device.into())
-            .assign_label(label_id.into(), op.into()),
-    )?;
+    client
+        .rt
+        .block_on(client.inner.team(team.into()).change_rank(
+            object_id.into(),
+            aranya_daemon_api::Rank::new(old_rank),
```
gknopf-aranya: If we're using `i64` in the C API and the policy, it's worth considering whether we also want to use `i64` instead of Rank in the Rust and IPC APIs. I think it's better to have a `Rank` type for type safety since it's easy to implement in Rust. An `.into()` conversion would make this code less verbose.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Yeah, I'm trying not to break backward compatibility here. But it seems like there's a point where if the functionality changes enough we're better off removing the function entirely. Allowing compilation to continue while silently breaking old functionality at runtime isn't ideal.

crates/aranya-client/src/client/team.rs
gknopf-aranya: I have some uncommitted changes on sp1 that were intended to resolve some linting/CI issues. Since sp1 is down I'm not able to access fixes that may have been made already for this.

crates/aranya-client/src/client/team.rs
```diff
@@ -109,15 +108,59 @@ impl Team<'_> {
 
 impl Team<'_> {
     /// Adds a device to the team with an optional initial role.
+    ///
+    /// The device is assigned a default rank of the command author's
```
gknopf-aranya: This doc isn't accurate since it can be `rank - 1` or `author - 1` depending on whether the device has a role assigned to it. Please check accuracy of the docs for adding devices and creating labels in both the aranya-client Rust API docs as well as the C API docs.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1204,13 +1190,18 @@ pub fn hello_subscription_config_builder_set_periodic_interval(
...
-/// @param[in] owning_role the ID of the owning role
-/// @param[in] roles_out returns a list of roles that own `role`
+/// @param[in] owning_role (ignored) formerly the owning role
+/// @param[in] roles_out returns a list of default roles
 /// @param[in,out] roles_len the number of roles written to the buffer.
 ///
 /// @relates AranyaClient.
+#[deprecated(note = "use `setup_default_roles_no_owning_role` instead")]
+#[allow(deprecated)]
 pub unsafe fn setup_default_roles(
```
gknopf-aranya, line 1276: Could call `setup_default_roles_no_owning_role()` here while ignoring `owning_role` to simplify this implementation. There is no need to re-implement this logic here when the method is deprecated.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: It's unfortunate that `role_owners()` isn't behind preview. That would have allowed us to remove it. It probably should have been behind that feature flag.

gknopf-aranya: Consensus is to remove `role_owners()` entirely. It's better to cause an application that uses it to fail to compile than to cause unexpected behavior.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: We're going to remove `role_owners()`:
https://github.com/aranya-project/aranya/pull/715#discussion_r2795253813

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: We're going to remove `role_owners()`:
https://github.com/aranya-project/aranya/pull/715#discussion_r2795253813

crates/aranya-client/tests/common/mod.rs
```diff
@@ -78,43 +78,49 @@ impl DevicesCtx {
         // Add the admin as a new device, and assign its role.
         info!("adding admin to team");
         owner_team
-            .add_device(self.admin.pk.clone(), Some(roles.admin().id))
+            .add_device_with_rank(
+                self.admin.pk.clone(),
+                Some(roles.admin().id),
+                Rank::new(799),
```
gknopf-aranya: `.into()` or `Rank::from(799)`

crates/aranya-daemon/src/api.rs
```diff
@@ -1321,134 +1286,54 @@ impl DaemonApi for Api {
...
-            .context("unable to add role owner")?;
-        self.effect_handler.handle_effects(graph, &effects).await?;
-
-        Ok(())
-    }
-
-    #[cfg(feature = "preview")]
     #[instrument(skip(self), err)]
-    async fn remove_role_owner(
+    async fn change_rank(
```
gknopf-aranya, line 1318: We weren't able to check that the `role_rank > device_rank` invariant is upheld for all cases when the rank of a role/device is changed due to a lack of iteration in the policy. So, we decided not to implement that control in the policy. Is it worth querying role/device ranks here to uphold the desired invariant at the daemon level? I have mixed feelings about this. One one hand, I prefer to keep the security controls in the `policy.md` as much as possible. On the other hand, if we're not considering this a security control and more of a best-practice, preventing users from using the API in unintended ways that could compromise the security of the system seems like an improvement. cc: @mooreoak, @djohnson-spideroak

mooreoak: We could implement it in the action as a safe guard. If we want to implement it I would do it there, but it is not a control just a lint.

crates/aranya-client/tests/tests.rs
```diff
@@ -407,24 +412,38 @@ async fn test_add_devices() -> Result<()> {
...
-    operator
+    // Owner assigns operator role (owner has AssignRole).
+    owner
         .sync_now(team.admin.aranya_local_addr().await?, None)
         .await
-        .context("operator unable to sync with admin")?;
+        .context("owner unable to sync with admin")?;
+    owner
+        .device(team.operator.id)
+        .assign_role(roles.operator().id)
```
gknopf-aranya, line 488: Why is there a new role assignment that wasn't here before? I thought this test is being updated to use the new APIs without changing the behavior of the test.

crates/aranya-client/tests/tests.rs
```diff
@@ -1939,211 +1754,758 @@ async fn test_admin_cannot_remove_last_owner() -> Result<()> {
...
+            devices.admin.pk.clone(),
+            Some(roles.admin().id),
+            Rank::new(699),
+        )
+        .await?;
+    owner_team
+        .add_device_with_rank(
+            devices.operator.pk.clone(),
+            Some(roles.operator().id),
+            Rank::new(699),
```
gknopf-aranya: Ensure these ranks are always set to the same value with a `const`

crates/aranya-client/tests/tests.rs
```diff
@@ -1939,211 +1754,758 @@ async fn test_admin_cannot_remove_last_owner() -> Result<()> {
...
+        .add_device_with_rank(devices.operator.pk.clone(), None, Rank::new(500))
+        .await?;
+
+    let owner_addr = devices.owner.aranya_local_addr().await?;
+    admin_team.sync_now(owner_addr, None).await?;
+
+    // Admin (rank 699) tries to assign operator role (rank 700) to operator device (rank 500)
+    // Admin outranks the device (699 > 500) but NOT the role (699 < 700)
+    // Should fail because author must outrank both targets
+    match admin_team
```
gknopf-aranya: We should also show that assign role will fail if the author outranks the role but not the device, or the device but not the role. For completeness, this test should also show that the assign role succeeds if the author outranks both.

crates/aranya-client/tests/tests.rs
```diff
@@ -1939,211 +1754,758 @@ async fn test_admin_cannot_remove_last_owner() -> Result<()> {
...
+        )
+        .await?;
+
+    let owner_addr = devices.owner.aranya_local_addr().await?;
+    admin_team.sync_now(owner_addr, None).await?;
 
-    // Add admin as owner of member role
+    // Admin (rank 799) tries to create label with rank 900 -- should fail
+    match admin_team
+        .create_label_with_rank(text!("too_high_label"), Rank::new(900))
```
gknopf-aranya: Or use arithmetic like `rank + 1` to guarantee the rank is higher. That's easier to reason about than comparing 799 to 900.

crates/aranya-client/tests/tests.rs
gknopf-aranya, line 1: High-level note on these tests to define rank `const` values and use arithmetic to guarantee desired rank relationships rather than setting hard-coded ranks.

crates/aranya-client/tests/tests.rs
```diff
@@ -771,13 +827,12 @@ async fn test_remove_team() -> Result<()> {
     {
         let admin = devices.admin.client.team(team_id);
 
-        // Role assignment should fail
+        // Label creation should fail after team removal.
```
gknopf-aranya: Not sure what "team removal" means here. What was removed? The team or the label? Terminate team is the proper way to refer to terminating a team. I don't see any new code added to this test that removes anything.

crates/aranya-daemon/src/api.rs
```diff
@@ -796,21 +795,20 @@ impl DaemonApi for Api {
         }
     }
 
-    #[cfg(feature = "preview")]
     #[instrument(skip(self), err)]
     async fn create_role(
```
gknopf-aranya, line 798: Let's mark this as deprecated in the daemon IPC as well

gknopf-aranya: Nevermind, deprecated stuff is removed from this API since it doesn't need to be backward compatible.

crates/aranya-daemon/src/api.rs
```diff
@@ -796,21 +795,20 @@ impl DaemonApi for Api {
...
-        owning_role: api::RoleId,
+        rank: api::Rank,
     ) -> api::Result<api::Role> {
         let graph = self.check_team_valid(team).await?;
 
         let effects = self
             .client
             .actions(graph)
-            .create_role(role_name, RoleId::transmute(owning_role))
+            .create_role(role_name, rank.value())
```
gknopf-aranya: May want to convert to `i64` inside the action itself

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: We're going to remove role_owners():
https://github.com/aranya-project/aranya/pull/715#discussion_r2795253813

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Or we can call the old method `setup_default_roles_deprecated()` so the new method can take the preferred name of `setup_default_roles()`. Old code will no longer compile because the parameters are different and application code will naturally migrate to the new method over time.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: `bug!` will return an error for release builds and panic for debug builds. Better than only catching the bug in debug builds.

crates/aranya-daemon/src/api.rs
gknopf-aranya: In this context is lint a runtime error outside of the policy in actions.rs? Normally a lint would be displayed at compile time, but there isn't a way to catch this at compile time.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: If we were to rename the old function and break old code, we might as well just get rid of it, right?

If we really want the preferred name on the new function without breaking, we _could_ do some macro trickery to overload the name based on number of arguments.

gknopf-aranya: Yeah, I think we should either remove the old one or wait a release cycle before grabbing the name we want when the deprecated method is removed.

crates/aranya-client/src/client/team.rs
gknopf-aranya: Changing this to a `saturating_sub()`

gknopf-aranya: There is a `arithmetic_side_effects` lint that could be enabled, but this could flag a bunch of arithmetic operations in other places.

crates/aranya-client/tests/tests.rs
gknopf-aranya: This is because the admin has `AddDevice` but not `AssignRole` permission.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Done. Testing all these cases.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Since we don't have any deprecated permissions yet, there's nothing to change in the code for this. I'll leave some documentation on the strategy for the future when a permission is deprecated.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: cc: @djohnson-spideroak, @mooreoak, @chip-so

crates/aranya-daemon/src/actions.rs
```diff
@@ -148,6 +149,103 @@ where
...
+    CE: aranya_crypto::Engine + Send + Sync + 'static,
+{
+    /// Checks that demoting a role's rank won't violate the role_rank > device_rank invariant.
+    ///
+    /// If `object_id` is a role, verifies that no device assigned to that role
+    /// has a rank >= `new_rank`. Returns an error if the invariant would be violated.
+    async fn check_role_demotion_invariant(&self, object_id: BaseId, new_rank: Rank) -> Result<()> {
+        // Check if object_id is a role
+        let is_role = self
+            .call_session_action(policy::query_team_roles())
```
gknopf-aranya: Can we call the actions directly instead of via session commands? Since the action is already implemented, this seems cleaner.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: We're going to remove the deprecated method so we can use the desired `setup_default_roles()` right away. Application code can update by removing the owning role param when they get a compiler error.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -262,33 +297,40 @@ pub enum ChanOp {
...
+/// # Deprecation Strategy
+///
+/// When a permission needs to be deprecated:
+///
+/// 1. Mark the variant with `#[deprecated(since = "X.Y.Z", note = "reason")]`
+/// 2. Keep the variant in the enum so existing code continues to compile
+/// 3. Keep the corresponding permission in `policy.md` so enum conversions
+///    between APIs and the policy continue to work
+/// 4. Log a `warn!` message when the deprecated permission is used
+/// 5. Document the migration path in the deprecation note
```
djohnson-spideroak: I think for the docs here, we probably dont need the full strategy. Just something along the lines of "when we deprecate an enum value, here is what you will see." I would like to have a separate doc of our deprecation strategy in aranya-docs or README, that would be useful.

crates/aranya-client/tests/common/mod.rs
```diff
@@ -78,43 +116,37 @@ impl DevicesCtx {
         // Add the admin as a new device, and assign its role.
         info!("adding admin to team");
         owner_team
-            .add_device(self.admin.pk.clone(), Some(roles.admin().id))
+            .add_device_with_rank(self.admin.pk.clone(), Some(roles.admin().id), 799.into())
```
djohnson-spideroak: use the constant defined above

crates/aranya-client/tests/tests.rs
```diff
@@ -466,11 +516,30 @@ async fn test_add_device_with_initial_role_requires_delegation() -> Result<()> {
         .await
         .context("admin unable to sync with owner")?;
 
+    // Admin should succeed at adding membera with member role
+    // because admin outranks the member role and has both AddDevice and AssignRole perms.
+    admin_team
+        .add_device_with_rank(
```
djohnson-spideroak, line 585: Admin starts with AddDevice?

crates/aranya-client/tests/tests.rs
```diff
@@ -466,11 +516,30 @@ async fn test_add_device_with_initial_role_requires_delegation() -> Result<()> {
...
+        .context("admin should be able to add device with member role (admin outranks member)")?;
+
+    // Admin (rank 799) should fail when trying to add a device with the owner role
+    // because admin does not outrank the owner role.
     match admin_team
-        .add_device(devices.membera.pk.clone(), Some(roles.member().id))
+        .add_device_with_rank(
+            devices.memberb.pk.clone(),
+            Some(roles.owner().id),
+            900.into(),
```
djohnson-spideroak: nit: constant for owner rank

crates/aranya-client/tests/tests.rs
```diff
@@ -1555,7 +1647,7 @@ async fn test_remove_perm_from_default_role() -> Result<()> {
 
     // Admin cannot add operator
     admin_team
-        .add_device(devices.operator.pk, None)
+        .add_device_with_rank(devices.operator.pk, None, DEFAULT_LABEL_RANK.into())
```
djohnson-spideroak: DEFAULT_LABEL_RANK?

crates/aranya-client/tests/tests.rs
```diff
@@ -1629,521 +1715,1173 @@ async fn test_owner_cannot_revoke_owner_role() -> Result<()> {
...
 
     Ok(())
 }
 
-/// Test that role management permissions can be assigned and revoked correctly.
+/// Devices cannot assign labels to themselves.
 #[test(tokio::test(flavor = "multi_thread"))]
-async fn test_assign_and_revoke_role_management_permission() -> Result<()> {
-    let mut devices = DevicesCtx::new("test_assign_and_revoke_role_management_permission").await?;
+async fn test_assign_label_to_device_self_rejected() -> Result<()> {
```
djohnson-spideroak, line 1948: Is this testing that assigning a label to self is rejected, or that assigning a label to self without the sufficient rank is rejected?

crates/aranya-client/tests/tests.rs
gknopf-aranya: This constant was defined so that hard-coded label ranks aren't used throughout the tests. It's not really a default in the sense that if no rank is specified it's the default that used. It is more of a default value used for testing purposes.

gknopf-aranya: Oh I see, that should only be used for labels, not the operator device.

crates/aranya-client/tests/tests.rs
gknopf-aranya: It's testing that self-assignment is rejected. The mechanism is rank-based — `author_can_operate_on_target` uses strict `>`, so a device can never outrank itself. There's no separate self-assignment check in the policy; the rank comparison naturally prevents it.

crates/aranya-daemon-api/src/service.rs
gknopf-aranya: Good idea. Trimmed the inline docs to just describe what you'll see (the `#[deprecated]` attribute and migration note on the variant). I'll open an issue to track adding a full deprecation strategy doc to aranya-docs.

gknopf-aranya: Good point. API stability and deprecation approaches are system level requirements.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Good catch — the comment is misleading. Admin already has `AddDevice` from `setup_default_roles` (policy line 1689). The test is granting `AssignRole` because that's the permission admin is missing to add a device *with an initial role*. I'll update the comment to clarify.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Good catch — this was using a label rank for a device. Fixed to `DEFAULT_OPERATOR_DEVICE_RANK`.

crates/aranya-client/tests/tests.rs
gknopf-aranya: The `900` here is intentionally arbitrary — it doesn't correspond to any defined rank. The test is verifying that admin can't add a device with the owner role because admin doesn't outrank the owner role (rank 999,999). The device rank doesn't matter for this check. I'll add a comment to clarify.

crates/aranya-client/tests/common/mod.rs
gknopf-aranya: Fixed — replaced all magic numbers in `add_all_device_roles` with `DEFAULT_ADMIN_DEVICE_RANK`, `DEFAULT_OPERATOR_DEVICE_RANK`, and `DEFAULT_MEMBER_DEVICE_RANK`.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Updated — replaced `900` with `DEFAULT_MEMBER_DEVICE_RANK`. The test fails on the role rank check regardless of device rank, so using a named constant makes that clearer.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Updated the doc comments on this test and the nearby rank-enforcement tests (`test_assign_role_self_rejected`, `test_owner_cannot_revoke_owner_role`, `test_delete_label_requires_permission`, `test_cannot_assign_role_twice`) to explain *why* the operation fails, not just that it's expected to fail.

examples/rust/aranya-example/src/main.rs
```diff
@@ -376,83 +378,103 @@ async fn main() -> Result<()> {
...
+        "custom role has {} permission(s): {:?}",
+        role_perms.len(),
+        role_perms
+    );
+    assert!(
+        role_perms
+            .iter()
+            .any(|p| matches!(p, Permission::CanUseAfc)),
+        "expected custom role to have CanUseAfc permission"
+    );
```
djohnson-spideroak, line 459: Are there any other assertions in the example? I dont know that its the best place to use assert.

crates/aranya-daemon/src/test.rs
```diff
@@ -373,7 +371,7 @@ impl TestCtx {
 
         owner
             .actions()
-            .add_device(DeviceKeyBundle::try_from(&admin.pk)?, None)
+            .add_device_with_rank(DeviceKeyBundle::try_from(&admin.pk)?, None, 0.into())
```
djohnson-spideroak: Is it correct to use 0 as the rank here?

examples/rust/aranya-example/src/main.rs
gknopf-aranya: Agreed — replaced the assert with a log. The `info!` above already prints the permission list, so the assert was redundant. Examples should show usage, not test behavior.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -1885,6 +1911,37 @@ pub unsafe fn add_device_to_team(
     Ok(())
 }
 
+/// Add a device to the team with an explicit rank.
+///
+/// Permission to perform this operation is checked against the Aranya policy.
```
chip-so: Should this describe the restrictions on the rank this can assign?

examples/rust/aranya-example/src/main.rs
gknopf-aranya: Updated to use `anyhow::ensure\!` instead — returns an error rather than panicking, consistent with the rest of the example's error handling.

gknopf-aranya: Actually, reverted back to `assert!` — the example already uses `.expect()` in several places which also panics, so `assert!` is consistent with the existing pattern.

gknopf-aranya: The example doesn't appear to have a consistent error handling paradigm at the moment. `anyhow!` with `?` is used in some places for returning an error while `expect` is used in other places to panic if there is an error. `assert!` is consistent with the panic style of error handling. While panicking isn't ideal for programs intended to be run in a production environment, the bar for error handling in an example is lower. We want the program to exit with an error/panic right away if anything goes wrong because the example is always expected to succeed.

crates/aranya-client/tests/common/mod.rs
```diff
@@ -25,6 +25,44 @@ const SYNC_INTERVAL: Duration = Duration::from_millis(100);
...
+// =============================================================================
+// Rank constants for tests
+// =============================================================================
+//
+// These constants define rank values used in tests. They are derived from the
+// role ranks defined in policy.md to ensure the invariant `role_rank > device_rank`
+// is always satisfied.
+//
+// Role ranks (from policy.md):
+//   - Owner role:    999_999 (MAX_RANK - 1)
```
chip-so: I guess the owner rank is never useful since there's no constant for it? Should it be tested?

crates/aranya-daemon/src/test.rs
gknopf-aranya: It should be `admin_role_rank - 1` or something similar to ensure the device rank is below the default role rank.

crates/aranya-client/tests/common/mod.rs
chip-so: There is no constant for the owner role rank (or the related creator device rank).

crates/aranya-client/tests/afc.rs
```diff
@@ -40,7 +40,7 @@ async fn test_afc_create_assign_revoke_delete_label() -> Result<()> {
     assert_eq!(memberb_team.labels().await?.iter().count(), 0);
 
     let label_id = owner_team
-        .create_label(text!("label1"), default_roles.owner().id)
+        .create_label_with_rank(text!("label1"), 500.into())
```
chip-so: `500` should be `DEFAULT_LABEL_RANK`?

crates/aranya-daemon/src/test.rs
gknopf-aranya: As I'm looking into this more, the reason I defined rank consts in the unit tests is partially because they are defined but not exported by the policy interface. So, they needed to be redefined in the unit tests. This isn't ideal since I'd rather have the default ranks in unit tests match the defaults used for the default roles in the policy. As far as I know, the policy doesn't currently have a way to directly export a const in the policy interface. cc: @jdygert-spok, @chip-so

chip-so: I guess it depends on what you consider the "policy interface". `Machine` and `Module` do allow access to the globals, as they are collected during compilation. But it's not exposed by the runtime AFAIK.

(I replied to this earlier so I thought you saw it, but github queues comments so it showed up later)

(double edit: or maybe it did? It's in the correct position in the thread so I can't tell)

gknopf-aranya: Opened an issue for allowing the policy interface to export const definitions:
https://github.com/aranya-project/aranya-core/issues/577
In the meantime, I've added todo comments where they have been manually redefined.

crates/aranya-client/tests/tests.rs
```diff
@@ -1629,521 +1726,1182 @@ async fn test_owner_cannot_revoke_owner_role() -> Result<()> {
...
+
+    // Give operator ChangeRank permission
+    owner_team
+        .add_perm_to_role(roles.operator().id, Permission::ChangeRank)
+        .await?;
+
+    let owner_addr = devices.owner.aranya_local_addr().await?;
+    operator_team.sync_now(owner_addr, None).await?;
+
+    let label_obj: ObjectId = ObjectId::transmute(high_label);
```
chip-so: I keep seeing this, and I wanted to check my understanding - is `transmute()` just there to avoid having to make ~n<sup>2</sup> `From` impls for all the ID types?

Also it shouldn't be necessary to specify the type before the `=`, right?

examples/rust/aranya-example/src/main.rs
gknopf-aranya: Keeping `assert!` for now — it's consistent with the existing `.expect()` calls in the example that also panic on failure.

The example currently mixes error handling styles (`?` with anyhow vs `.expect()`/`assert!` panics). Standardizing that is a broader cleanup unrelated to the rank/roles changes in this PR, so it's better handled separately to keep this PR focused.

Opened #732 to track standardizing error handling across the example.

crates/aranya-client/tests/tests.rs
```diff
@@ -1629,521 +1726,1182 @@ async fn test_owner_cannot_revoke_owner_role() -> Result<()> {
...
+    // Step 3: Query permissions from old role and copy to new role
+    let old_perms = owner_team.query_role_perms(old_role.id).await?;
+    assert_eq!(old_perms.len(), 2, "old role should have 2 permissions");
+    for perm in old_perms {
+        owner_team.add_perm_to_role(new_role.id, perm).await?;
+    }
+
+    // Verify new role has the same permissions
+    let new_perms = owner_team.query_role_perms(new_role.id).await?;
+    assert_eq!(new_perms.len(), 2, "new role should have 2 permissions");
```
chip-so: This doesn't technically check that the permissions are the same. It just checks that it has the same number of permissions. Though I suppose it's a fair assumption.

crates/aranya-client/tests/tests.rs
```diff
@@ -1629,521 +1726,1182 @@ async fn test_owner_cannot_revoke_owner_role() -> Result<()> {
...
+        .add_device_with_rank(
+            devices.operator.pk.clone(),
+            Some(roles.operator().id),
+            DEFAULT_OPERATOR_DEVICE_RANK.into(),
+        )
+        .await?;
+
+    // Create a high-rank label (same as admin role rank)
+    let high_label = owner_team
+        .create_label_with_rank(text!("high_label"), DEFAULT_ADMIN_ROLE_RANK.into())
```
chip-so: Slightly confusing use of `DEFAULT_ADMIN_ROLE_RANK` for the rank of a label.

crates/aranya-client/tests/tests.rs
```diff
@@ -1629,521 +1726,1182 @@ async fn test_owner_cannot_revoke_owner_role() -> Result<()> {
...
+    // Verify label was created
+    let label = owner_team.label(label_id).await?;
+    assert!(label.is_some(), "label should exist");
+
+    // Verify default rank is author_rank - 1 (owner device rank 1000000 - 1 = 999999)
+    let label_obj: ObjectId = ObjectId::transmute(label_id);
+    let rank = owner_team.query_rank(label_obj).await?;
+    assert_eq!(
+        rank,
+        999999.into(),
```
chip-so: Seems like this would be the place for a creator device rank or owner role rank constant.

chip-so (commented):
Through capi and client. Will look at the rest later.

crates/aranya-daemon/src/test.rs
gknopf-aranya: @chip-so I've been referring to the `policy.rs` file that is generated from the `policy.md` as the policy interface. That contains actions, enums, types, etc. but not any globals/const values.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Yes. The APIs should probably mention rank and permission requirements explicitly rather than vaguely referencing the policy. I'm assuming it was written this way to avoid becoming out of date as the policy changes.

gknopf-aranya: I've added rank restriction docs to `add_device_with_rank`, `create_label_with_rank`, and `create_role` in both the Rust client and CAPI. There are more APIs that need similar changes — I'll open an issue to track the broader doc update.

gknopf-aranya: Created #734 to track this. Updated all API docs on this PR to replace the generic "Permission to perform this operation is checked against the Aranya policy" with specific permission and rank requirements derived from the policy (e.g. `AddDevice` permission, rank must be <= caller's rank, caller must outrank target, etc.).

crates/aranya-client/tests/common/mod.rs
gknopf-aranya: Good catch, adding that in.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Correct about both observations. I'll remove the `: ObjectId` types since that should already be known.

crates/aranya-client/tests/common/mod.rs
gknopf-aranya: Added `MAX_RANK`, `DEFAULT_OWNER_ROLE_RANK`, and `DEFAULT_OWNER_DEVICE_RANK` constants (in both `tests/common/mod.rs` and `daemon/src/test.rs`). Replaced all magic `999_999` values.

crates/aranya-client/tests/afc.rs
gknopf-aranya: Fixed — importing and using `DEFAULT_LABEL_RANK` instead of the hardcoded `500`.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Done — using `DEFAULT_OWNER_ROLE_RANK` here now.

crates/aranya-client/tests/tests.rs
gknopf-aranya: Good point — replaced with `DEFAULT_OPERATOR_ROLE_RANK + 100` and a comment explaining the intent (the label needs to outrank the operator).

crates/aranya-client/tests/tests.rs
gknopf-aranya: Adding `Eq` and `PartialEq` to the enum so perms can be compared directly. I'll assume the perms don't need to be sorted before comparing due to them being set in a particular order.

gknopf-aranya: The order is an implementation detail of the factdb, so sorting before comparing could be a good way to make this test more reliable in case that ever changes.

crates/aranya-daemon/src/test.rs
gknopf-aranya: @jdygert-spok, @chip-so would it be possible to update the `policy-ifgen` tool to include globals in the generated interface so we don't need to plumb them out from the policy via a different compiler interface?

jdygert-spok: Yes, it's possible

gknopf-aranya: Opened aranya-project/aranya-core#582 to track exposing policy globals via `policy-ifgen`.

gknopf-aranya: Using globals from the policy interface is a nice-to-have but doesn't block landing the rank changes because we have the work-around of redefining them in the client/daemon code.

gknopf-aranya: Marking as resolved since TODOs are documented and this is non-blocking.

crates/aranya-metrics/src/main.rs
```diff
@@ -312,59 +312,65 @@ async fn run_demo_body(ctx: DemoContext) -> Result<()> {
...
-    let _admin_addr = ctx.admin.aranya_local_addr().await?;
-    let _operator_addr = ctx.operator.aranya_local_addr().await?;
 
-    // setup sync peers.
+    // Owner adds admin and operator to the team, then delegates
+    // remaining operations to them.
     info!("adding admin to team");
-    owner.add_device(ctx.admin.pk, Some(admin_role.id)).await?;
+    owner
+        .add_device_with_rank(ctx.admin.pk, Some(admin_role.id), 100.into())
```
chip-so: Maybe these should be constants too?

examples/c/example.c
```diff
@@ -424,9 +418,9 @@ AranyaError init_team(Team *t) {
         return err;
     }
 
-    // add admin to team.
-    err = aranya_add_device_to_team(&owner->client, &t->id, admin->pk,
-                                    admin->pk_len, &admin_role_id);
+    // add admin to team with rank 100.
+    err = aranya_add_device_to_team_with_rank(
+        &owner->client, &t->id, admin->pk, admin->pk_len, &admin_role_id, 100);
```
chip-so: Also possibly constants here. Especially for example code.

examples/rust/aranya-example/src/main.rs
```diff
@@ -372,89 +371,113 @@ async fn main() -> Result<()> {
...
+    admin_team.sync_now(owner_addr, None).await?;
+    operator_team.sync_now(owner_addr, None).await?;
+    membera_team.sync_now(owner_addr, None).await?;
+    memberb_team.sync_now(owner_addr, None).await?;
 
     // add membera to team.
     info!("adding membera to team");
     owner_team
-        .add_device(membera.pk.clone(), Some(member_role.id))
+        .add_device_with_rank(membera.pk.clone(), Some(member_role.id), Rank::new(100))
```
chip-so: Probably constants here, too.

examples/rust/aranya-example/src/main.rs
```diff
@@ -372,89 +371,113 @@ async fn main() -> Result<()> {
...
+    let role_perms = owner_team.query_role_perms(custom_role.id).await?;
+    info!(
+        "custom role has {} permission(s): {:?}",
+        role_perms.len(),
+        role_perms
+    );
+    assert!(
+        role_perms
+            .iter()
+            .any(|p| matches!(p, Permission::CanUseAfc)),
```
chip-so: You should be able to use `role_perms.contains(&Permission::CanUseAfc)` since `PartialEq` is implemented on `Permission`.

crates/aranya-metrics/src/main.rs
gknopf-aranya: Yep, will update here and elsewhere. I tried squashing all these a few times but claude didn't apply the feedback generally enough. This seems like a good convention to include in any coding guidelines we provide to claude.

gknopf-aranya:
> I think this looks good. Mostly just some opportunities for more constant-ification.

@chip-so thank you, responded to your feedback with a few commits

crates/aranya-daemon/src/test.rs
gknopf-aranya: Update on this: planning to query the ranks directly instead of redefining the rank const values in the code.

examples/c/example.c
```diff
@@ -427,10 +423,18 @@ AranyaError init_team(Team *t) {
         return err;
     }
 
-    // add admin to team.
+    // Query the admin role rank and add admin to team.
+    AranyaObjectId admin_role_object_id = {.id = admin_role_id.id};
+    int64_t admin_role_rank = 0;
+    err = aranya_query_rank(&owner->client, &t->id, &admin_role_object_id,
```
chip-so, line 429: I think I would omit this for the examples specifically. The constants above don't come from the policy in the first place, so there's really no reason to calculate them dynamically.

gknopf-aranya:
> I understand why you did this, but it really does make the tests a little less clear when you're reading them.

After discussing with @mooreoak today, we realized that exporting the rank globals via the policy interface would prevent future changes to the rank values since any change would be a breaking API change. Redefining the ranks in the unit tests is simpler but brittle since future policy changes could break it. Querying ranks and performing arithmetic on them seemed like the best approach given the constraints we're working with. A helper method that stores all the ranks at the beginning of each test could simplify this a bit.

examples/c/example.c
gknopf-aranya: While the constants don't come from the policy, ensuring that they are below the role ranks defined in the policy is important since the ranks need to be low enough that the objects can be operated on.

chip-so: Hm. That is a good point. This is probably better, then.

PR #717, opened 2026-02-11
Refactor Sync Using Transport Traits

nikki-aranya:
This PR introduces three new transport traits, `SyncConnector` establishes connections to other peers (potentially reusing an existing connection), `SyncListener` accepts incoming connections, and `SyncStream` allows for sending and receiving data.

As a consequence of changing the boundary from the old SyncState, the code that used to be in hello.rs got merged into the QUIC code, which then got merged into `SyncManager` and the new `SyncServer`. A future PR will address making the poll sync vs hello sync vs push sync more clear, as well as aranya-core changes to address some of this.

As this PR stands, `SyncManager` handles all orchestration and scheduling, `SyncServer` handles accepting connections and replying, handing off any scheduling to the `SyncManager`, `SyncClient` handles the actual sync logic and protocol state, and then there are lightweight implementations implementing QUIC Sync, with `SyncHandle` allowing the `SyncServer` as well as any upstream code to talk to the `SyncManager`.

### DIFF
--- a/crates/aranya-daemon/src/sync/transport/quic/connector.rs
+++ b/crates/aranya-daemon/src/sync/transport/quic/connector.rs
@@ -112,7 +112,10 @@ impl SyncConnector for QuicConnector {
             trace!(?peer, "establishing new QUIC connection");
             let mut conn = {
                 // Set the current `GraphId` we're operating on in the PSK store.
-                let _guard = self.psk_store.set_team(TeamId::transmute(peer.graph_id));
+                let _guard = self
+                    .psk_store
+                    .set_team(TeamId::transmute(peer.graph_id))
+                    .await;
                 self.client
                     .connect(Connect::new(addr).with_server_name(addr.ip().to_string()))
                     .await?
--- a/crates/aranya-daemon/src/actions.rs
+++ b/crates/aranya-daemon/src/actions.rs
@@ -25,7 +25,7 @@ use crate::{
 
 /// Container for complex AQC channel creation results.
 #[derive(Debug)]
-pub(crate) struct SessionData {
+pub struct SessionData {
     /// The serialized messages
     #[cfg(feature = "afc")]
     pub ctrl: Vec<Box<[u8]>>,


crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
...
+    type Error: std::error::Error + Send + Sync + 'static;
 
-#[cfg(feature = "preview")]
-use super::GraphId;
-use super::{Result, SyncManager, SyncPeer};
+    fn peer(&self) -> super::SyncPeer;
 
-pub(crate) mod quic;
+    async fn send(&mut self, data: &[u8]) -> Result<(), Self::Error>;
+    async fn receive(&mut self, buffer: &mut Vec<u8>) -> Result<(), Self::Error>;
```
gknopf-aranya: Allocation of `Vec` could be unbounded. Better to have the caller allocate a buffer and pass it in:
```suggestion
    async fn recv(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error>; 
```
The allocated buffer size could be `MAX_SYNC_MESSAGE_SIZE`.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
...
+    async fn finish(&mut self) -> Result<(), Self::Error>;
```
gknopf-aranya: While QUIC has `finish()` since this is intended to be a generic transport trait, `close()` is a more transport-agnostic method name:
```suggestion
    async fn close(&mut self) -> Result<(), Self::Error>;
```

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
...
+}
+
+#[async_trait::async_trait]
+pub(crate) trait SyncTransport: Send + Sync + 'static {
+    type Error: std::error::Error + Send + Sync + 'static;
+    type Stream: SyncStream<Error = Self::Error>;
+
+    async fn connect(&self, peer: super::SyncPeer) -> Result<Self::Stream, Self::Error>;
```
gknopf-aranya, line 58: While QUIC and TcpStream have the concept of streams, UDP does not. If the underlying transport does not support streams, will this fallback to returning a socket or the entire transport as a stream?

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
...
+}
+
+#[async_trait::async_trait]
+pub(crate) trait SyncListener: Send + Sync + 'static {
+    type Error: std::error::Error + Send + Sync + 'static;
+    type Stream: SyncStream<Error = Self::Error>;
+
+    fn local_addr(&self) -> std::net::SocketAddr;
```
gknopf-aranya: This only works for IP-based transports.

gknopf-aranya: Returning a string for the address and letting the caller parse it into a more specific type based on the transport is one option.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
...
-use aranya_runtime::Address;
-use aranya_runtime::{PolicyStore, Sink, StorageProvider};
+#[async_trait::async_trait]
+pub(crate) trait SyncStream: Send + Sync + 'static {
+    type Error: std::error::Error + Send + Sync + 'static;
 
-#[cfg(feature = "preview")]
-use super::GraphId;
-use super::{Result, SyncManager, SyncPeer};
+    fn peer(&self) -> super::SyncPeer;
```
gknopf-aranya, line 25: `SyncPeer` assumes the peer has an IP address.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
...
 
-pub(crate) mod quic;
+    async fn send(&mut self, data: &[u8]) -> Result<(), Self::Error>;
+    async fn receive(&mut self, buffer: &mut Vec<u8>) -> Result<(), Self::Error>;
+    async fn finish(&mut self) -> Result<(), Self::Error>;
+}
+
+#[async_trait::async_trait]
+pub(crate) trait SyncTransport: Send + Sync + 'static {
```
gknopf-aranya: Assuming we're only supporting IP transports, this trait could include `bind()` since the server will need to bind to an address in order to accept connections:
```
async fn bind(&self, addr: Self::Addr) -> Result<Self::Listener, Self::Error>;  
```
Since this is specific to a server and not a client, it may make more sense to put `bind()` under a `SyncServer` or `SyncListener` trait.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
...
-#[cfg(feature = "preview")]
-use std::time::Duration;
-use std::{fmt, future::Future};
+pub(crate) mod quic;
 
-#[cfg(feature = "preview")]
-use aranya_runtime::Address;
-use aranya_runtime::{PolicyStore, Sink, StorageProvider};
+#[async_trait::async_trait]
+pub(crate) trait SyncStream: Send + Sync + 'static {
```
gknopf-aranya, line 20: Let's add a `local_addr()` method to `SyncStream`. We currently only have a way to get the address of `SyncListener`:
`fn local_addr(&self) -> Self::Addr;`

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya, line 1: More docs are needed for each method in the generic transport traits. This is an important new interface that's being added by the PR. In order to implement a new transport, a developer will need a good understanding of what each method is that must be implemented.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
...
+#[async_trait::async_trait]
+pub(crate) trait SyncTransport: Send + Sync + 'static {
+    type Error: std::error::Error + Send + Sync + 'static;
+    type Stream: SyncStream<Error = Self::Error>;
+
+    async fn connect(&self, peer: super::SyncPeer) -> Result<Self::Stream, Self::Error>;
+}
+
+#[async_trait::async_trait]
+pub(crate) trait SyncListener: Send + Sync + 'static {
```
gknopf-aranya, line 67: Could support shutting down the listener:
`async fn shutdown(&mut self) -> Result<(), Self::Error>; `

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya: Wondering if this should be `fn peer_addr(&self) -> Self::Addr;` instead of returning `SyncPeer` to return the peer's network address. A generic transport trait shouldn't need to have anything sync-specific in it.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,32 @@
 //! This module contains all generic syncer transport traits, as well as any transport-specific syncer implementations.
```
gknopf-aranya: We should leave a note here that these transport traits are intended to be implemented via a reliable transport such as QUIC or TCP that guarantees delivery via retransmission and is IP addressable. UDP and non-IP transports are not supported. This causes me to question whether making the transport more generic is something we really need. The only other IP-based protocol with retransmission I can think of that's widely used is TCP, but we'd need to implement some other encryption mechanism on top of that to replace the PSK encryption we have with QUIC.

gknopf-aranya: I think we should think carefully about what transports we may want to support in the future and if this generic transport is getting us closer to that goal.

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: This isn't really close though, close is Drop, finish is if you have side effects (or a no-op)

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: > This only works for IP-based transports.

So does SyncPeer/Addr afaik, so we'd have to have a larger discussion about all that.

gknopf-aranya: The `NetworkIdentifier` type was introduced in the application API so that non-IP addresses could be used in the future. It resolves internally to a `Addr` because we currently only support QUIC. Now that a generic transport is being introduced, I think `NetworkIdentifier` (which is a `String`) is a more general choice for this trait.

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya: Nevermind, we wouldn't be able to sync over UDP unless reliability is added at the application layer in which case a stream can be supported.

Cargo.toml
```diff
@@ -66,6 +66,7 @@ aranya-util = { version = "4.1.0", path = "crates/aranya-util", default-features
 
 # External dependencies
 anyhow = { version = "1.0.86" }
+async-trait = { version = "0.1" }
```
gknopf-aranya: Is the reason for adding the `async-trait` dep to make the transport `dyn`? From what I can tell using this dep makes the trait interface simpler, but has a performance and allocation cost.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,191 @@
...
+        server_keys: Arc<PskStore>,
+    ) -> Result<(Self, SharedConnectionMap), Error> {
+        // Create shared connection map and channel for connection updates
+        let (conns, conn_rx) = SharedConnectionMap::new();
+
+        // Create Server Config
+        let mut server_config = ServerConfig::builder()
+            .with_no_client_auth()
+            .with_cert_resolver(Arc::new(NoCertResolver::default()));
+        server_config.alpn_protocols = vec![ALPN_QUIC_SYNC.to_vec()]; // Set field directly
```
gknopf-aranya, line 67: If this is a generic implementation of a QUIC transport that is sync-agnostic, `ALPN_QUIC_SYNC` seems like something to be passed in externally rather than hard-coded here.

crates/aranya-daemon/src/sync/transport/quic/stream.rs
```diff
@@ -0,0 +1,50 @@
+use bytes::Bytes;
+use futures_util::AsyncReadExt as _;
+use s2n_quic::stream::{BidirectionalStream, ReceiveStream, SendStream};
+
+use super::Error;
+use crate::sync::{transport::SyncStream, SyncPeer};
+
+pub(crate) struct QuicStream {
+    peer: SyncPeer,
```
gknopf-aranya, line 13: Use of `SyncPeer` ties this QUIC transport to syncing when it seems like it should be sync-agnostic.

crates/aranya-daemon/src/sync/transport/quic/stream.rs
```diff
@@ -0,0 +1,50 @@
...
+        self.recv
+            .read_to_end(buffer)
+            .await
+            .map_err(Error::Receive)?;
+        Ok(())
+    }
+
+    async fn finish(&mut self) -> Result<(), Self::Error> {
+        self.send.close().await.map_err(Error::Finish)?;
+        Ok(())
```
gknopf-aranya: Same as:
```suggestion
        self.send.close().await.map_err(Error::Finish)
```

crates/aranya-daemon/src/sync/transport/quic/stream.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    async fn receive(&mut self, buffer: &mut Vec<u8>) -> Result<(), Self::Error> {
+        buffer.clear();
+        self.recv
+            .read_to_end(buffer)
+            .await
+            .map_err(Error::Receive)?;
+        Ok(())
```
gknopf-aranya: Same as:
```suggestion
        self.recv
            .read_to_end(buffer)
            .await
            .map_err(Error::Receive)
```

crates/aranya-daemon/src/sync/transport/quic/stream.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    async fn finish(&mut self) -> Result<(), Self::Error> {
```
gknopf-aranya, line 76: It looks like there's currently no way to signal the receiver to close their side of the stream. QUIC has this but perhaps we don't need it if we weren't using that before for the sync protocol.

gknopf-aranya: On the other hand, if we want to implement a generic transport trait, then having a way to expose the full set of close/shutdown functionality from underlying libraries makes sense. It's worth mentioning as a TODO for future improvements to the transport traits.

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya, line 1: Questioning the organization of sync -> transport -> quic since this indicates that the transport traits are intended to be used by sync. While it's true that in the near term we're considering adding more transports to the syncer, a generic transport should be agnostic to the protocol. E.g. what if we decide to use this for AFC? Or make this a lib that an application can use to send data? Here's another way to organize this:
```
src/sync (sync requester/responder/protocol that assumes use of a generic transport)
src/transport/mod.rs (transport traits)
src/transport/quic/ (specific transport impls)
```

crates/aranya-daemon/src/sync/transport/quic/connections.rs
```diff
@@ -48,7 +48,7 @@ impl SharedConnectionMap {
     ///
     /// If the handle does not match the one in the map,
     /// it has been replaced and does not need to be removed.
-    pub(super) async fn remove(&mut self, peer: SyncPeer, handle: Handle) {
+    pub(super) async fn remove(&self, peer: SyncPeer, handle: Handle) {
```
gknopf-aranya: How do you remove an entry from a connection map without mutating it?

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -0,0 +1,115 @@
+use std::sync::Arc;
+
+use anyhow::Context as _;
+use aranya_daemon_api::TeamId;
```
gknopf-aranya, line 5: The QUIC transport shouldn't need to know about `TeamId`. That is at the protocol level.

gknopf-aranya: I see that `TeamId`/`GraphId` are required for securing the transport with PSKs. So it seems like we need a way to pass in a custom implementation of the PSK operations from the protocol level rather than implementing it here in the transport.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
nikki-aranya: Interior mutability/idk but it compiles fine

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -102,6 +119,58 @@ impl<ST, PS, SP, EF> SyncManager<ST, PS, SP, EF> {
         }
     }
 
+    #[cfg(feature = "preview")]
+    fn add_hello_subscription(
```
gknopf-aranya, line 121: Seems like we'll want more `instrument` tracing for the methods in this module

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -110,50 +179,197 @@ impl<ST, PS, SP, EF> SyncManager<ST, PS, SP, EF> {
...
+        let message = SyncType::Hello(SyncHelloType::Unsubscribe {
+            graph_id: peer.graph_id,
+        });
+
+        self.send_hello_request(peer, message).await
+    }
+
+    #[cfg(feature = "preview")]
+    async fn send_hello_notification(&mut self, peer: SyncPeer, head: Address) -> Result<()> {
+        trace!(?peer, "sending hello notifications to peer");
```
gknopf-aranya: `instrument` can include the function name and any args rather than manually tracing here

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: I want to say that's the implementation's problem not a problem with the trait but I'll think on it

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: This abstraction should still work for UDP, I might write up a new implementation for that

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: I think either that or type Addr;

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya: A peer could DoS attack us by sending a huge response if we don't limit the buffer size.

Cargo.toml
nikki-aranya: I'm not sure, this was the only way I could get it to behave but that was when I was rearranging stuff, I can see if I can get it to work now without

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya: I don't think it will work for regular UDP because there isn't a way to detect dropped packets and resend them. You could implement a custom UDP transport that has retransmission built-in but that's essentially what QUIC is already.

crates/aranya-daemon/src/sync/handle.rs
```diff
@@ -134,20 +164,45 @@ impl SyncHandle {
...
     /// Broadcast hello notifications to all subscribers of a graph.
     #[cfg(feature = "preview")]
     pub(crate) async fn broadcast_hello(&self, graph_id: GraphId, head: Address) -> Response {
         self.send(ManagerMessage::BroadcastHello { graph_id, head })
             .await
     }
 
+    /// Tell the [`SyncManager`] to add this peer to their subscriptions.
+    #[cfg(feature = "preview")]
+    pub(super) async fn hello_subscribe_request(
```
gknopf-aranya, line 168: It would be nice to group all the sync hello code under different impls and modules so fewer preview flags are used for it.

crates/aranya-daemon/src/sync/transport/mod.rs
jdygert-spok: If `buf` is expected to only hold a sync protocol message then the caller should determine the max size.

crates/aranya-daemon/src/sync/transport/quic/stream.rs
nikki-aranya: Can you define what you mean by "sync-agnostic"? It's mostly just keeping track of what peer this specific quic stream is talking to, especially if we happen to make a big pool of streams in the manager in the future, it would be nice to ask "what peer are you talking to?"

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: You're saying that I shouldn't be making the assumption that the transport is driven by the SyncManager/SyncServer?

crates/aranya-daemon/src/sync/transport/quic/connector.rs
nikki-aranya: This is part of why I kept bringing up the question of "hey, how do we handle stuff other than PSKs/mTLS because I know this is a problem", I don't have an answer for that question other than, this is what the code was previously doing so I copied over the set_team stuff

crates/aranya-daemon/src/sync/handle.rs
nikki-aranya: True, this might also be better served once I get started on PR3 since I'm gonna investigate breaking out hello.rs again, along with poll.rs as you asked about in DMs

crates/aranya-daemon/src/sync/transport/quic/stream.rs
gknopf-aranya: What I mean by sync-agnostic, is that the transport implementation shouldn't know anything about the sync protocol, sync peers, team ids, etc. Addressing is needed to talk to peers, but that would be a `Addr` or `NetworkAddress`. It could just be that this `SyncPeer` type is already sync agnostic but has an incorrect name that implies it isn't (I didn't look at how it is defined yet).

Cargo.toml
jdygert-spok: Should be able to do without. You may need to write the trait definition manually with `-> impl Future` to add in `Send` bounds etc.

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya: Right. If it's truly a generic transport implementation, it shouldn't matter who is using it. All that matters is that you want to send/recv from a peer over a network.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
gknopf-aranya: I suppose if it compiles it should be fine. I was wondering if it might not actually be deleting something from the connection map if the compiler indicates that nothing is mutating the connection map.

crates/aranya-daemon/src/sync/transport/mod.rs
jdygert-spok: I don't think we do want to make them completely abstract transports, they are only for syncing. We might even end up wanting to do sync-specific stuff on them.

crates/aranya-daemon/src/sync/handle.rs
gknopf-aranya: Yeah, agree that doesn't need to be on this PR. And who knows, maybe the feature will be moved out of preview by that point. I do think we should be more careful how features behind preview flags are added to the codebase initially to minimize the amount of flags.

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya: Sync-specific things seem like something that can be implemented at the protocol/application layer rather than in the transport layer. Do you have some examples of sync specific things that would be more optimal at the transport layer?

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: bind adds too many complications, it's better to let MyServer::new() do the binding

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: Can't find a good reason to add this/use this, if you need the addr of the stream, just do .peer().addr

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya: And that's valid because `MyServer::new()` is outside the scope of these trait definitions? So it can initialize the server however it needs to?

crates/aranya-daemon/src/sync/transport/mod.rs
gknopf-aranya: That gives you the address of the peer but not your local address. The server should know it's local_addr() but it could be convenient to allow someone to figure out both peer addresses with only a `SyncStream` object if they don't have direct access to the server object.

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: Correct

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: If that's needed in the future we can add it, I don't really want to add an #[allow(dead_code)] because the traits are pub(crate) and internal right now, but right now it's basically just a stub regardless of how it's implemented

crates/aranya-daemon/src/sync/transport/quic/listener.rs
nikki-aranya: see: the previous discussion about bind(), this is kinda just the necessary glue for QUIC sync/the QUIC transport, we could probably make it more generic but I don't see much of a need

crates/aranya-daemon/src/sync/transport/quic/stream.rs
nikki-aranya: SyncPeer's as close to generic as I think we need for the moment, upon further reflection; see: the Addr discussion

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya, line 1: Since the `client.rs` module has been removed, which new abstraction does it fall under? Even though we have a generic transport, thinking of the implementation in terms of a client and server still seems like a useful way to organize the code.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
```diff
@@ -23,135 +24,104 @@ use crate::sync::SyncPeer;
...
 
-/// Thread-safe map for sharing QUIC connections between sync peers.
-///
-/// This map stores persistent QUIC connections indexed by [`SyncPeer`].
-#[derive(Clone, Debug)]
+/// Thread-safe map that stores a [`Handle`] to the connection to allow for reuse across multiple
+/// connection requests to a peer.
+#[derive(Clone, Debug, Default)]
 pub(crate) struct SharedConnectionMap {
-    tx: mpsc::Sender<ConnectionUpdate>,
```
gknopf-aranya, line 31: I believe this was originally a performance optimization to not require obtaining the connection map in order to send an update to it. Since we don't support initiating parallel syncs the case it could help with is a simultaneous outgoing sync, and multiple incoming syncs all establishing a new connection at the same time. @ssweetney-spok, @jdygert-spok what are your thoughts on this change? My perspective is that even if this is a desirable change I'd prefer to keep these broad refactoring changes in a PR that's separate from any behavior/optimization/algorithmic changes we may want to make. It's easier to vouch for changes that appear to improve the organization of the syncer than it is to prove that behavioral changes haven't introduced a bug or slowdown somewhere.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
```diff
@@ -23,135 +24,104 @@ use crate::sync::SyncPeer;
...
-    /// If the handle does not match the one in the map,
-    /// it has been replaced and does not need to be removed.
-    pub(super) async fn remove(&mut self, peer: SyncPeer, handle: Handle) {
+    /// Note that this will skip removing it if the passed handle does not match the stored handle
+    /// for the peer.
+    pub(super) async fn remove(&self, peer: SyncPeer, handle: Handle) {
         match self.handles.lock().await.entry(peer) {
-            Entry::Vacant(_) => {}
+            Entry::Vacant(_) => {
+                trace!(?peer, "remove called but no connection in map");
```
gknopf-aranya: If this is unexpected but not an error, `warn!` could make more sense

crates/aranya-daemon/src/sync/transport/quic/connections.rs
```diff
@@ -23,135 +24,104 @@ use crate::sync::SyncPeer;
...
-
-                (entry.insert(handle).clone(), acceptor)
-            }
-        };
-
-        debug!("created new quic connection");
-        self.tx.send((peer, acceptor)).await.ok();
-
-        handle
+    /// If a handle already exists and is able to be pinged, this will close the passed connection
```
gknopf-aranya: This comment mentions a ping, but I don't see a ping in the implementation of the function

crates/aranya-daemon/src/sync/transport/quic/client.rs
nikki-aranya: Almost all this code now lives in the SyncManager, minus connect() which is now QuicTransport

crates/aranya-daemon/src/sync/transport/quic/connections.rs
gknopf-aranya, line 1: As far as I can tell, the only thing QUIC specific in the connection map implementation is the ping. Seems feasible to refactor it in a way that can be used by multiple transports in the future (if generic connection types can be stored). Not suggesting to tackle that on this PR though.

crates/aranya-daemon/src/sync/transport/quic/connections.rs
nikki-aranya: dangit, forgot to update the comment, thought I did that

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,229 @@
...
+use super::{
+    ConnectionUpdate, Error, PskStore, QuicStream, SharedConnectionMap, SyncListener,
+    ALPN_QUIC_SYNC,
+};
+use crate::sync::{Addr, GraphId, SyncPeer};
+
+type AcceptResult = (SyncPeer, StreamAcceptor, Option<BidirectionalStream>);
+
+#[derive(Debug)]
+pub(crate) struct QuicListener {
```
gknopf-aranya, line 43: Code and docs are using listener and server interchangeably. Let's pick one for consistency.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        // Build up the `ServerConfig` so we can initialize the TLS server.
+        let mut server_config = ServerConfig::builder()
+            .with_no_client_auth()
+            .with_cert_resolver(Arc::new(NoCertResolver::default()));
+        server_config.alpn_protocols = vec![ALPN_QUIC_SYNC.to_vec()]; // Set field directly
+        server_config.preshared_keys =
+            PresharedKeySelection::Required(Arc::clone(&server_keys) as _);
+
+        #[allow(deprecated)]
```
gknopf-aranya, line 71: Are we using deprecated functionality?

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,229 @@
...
+            conn_tx,
+        ))
+    }
+
+    /// Accepts an incoming bidirectional stream from this [`SyncPeer`].
+    async fn accept_pending_stream(peer: SyncPeer, mut acceptor: StreamAcceptor) -> AcceptResult {
+        trace!(?peer, "waiting for bidirectional stream");
+
+        let stream = tokio::time::timeout(
+            Duration::from_secs(30),
```
gknopf-aranya: Magic number, let's define a const for this

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,229 @@
...
+        let identity = get_conn_identity(&mut conn)?;
+        let active_team = self
+            .server_keys
+            .get_team_for_identity(&identity)
+            .context("no active team for accepted connection")?;
+
+        debug!(?remote, ?active_team, "authenticated incoming connection");
+
+        let peer_addr =
+            tokio::time::timeout(Duration::from_secs(5), extract_return_address(&mut conn))
```
gknopf-aranya: Another magic number for this timeout

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Task management vs. networking code needed for initiating a sync request seems like a different abstraction.

crates/aranya-daemon/src/sync/transport/quic/psk.rs
```diff
@@ -172,7 +174,7 @@ impl server::SelectsPresharedKeys for PskStore {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
gknopf-aranya, line 182: `Default` seems okay as long as it doesn't result in any PSKs initialized with all zeros. I don't think it will since it should just be an empty map.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -287,23 +283,18 @@ impl Daemon {
...
-            psk_store,
-            (server.local_addr().into(), client_addr),
-            syncer_recv,
-            conns,
-        )?;
-
-        Ok((client, server, syncer, peers, recv_effects))
+        let (handle, recv) = SyncHandle::channel(128);
+
+        let (listener, conns, conn_tx) = QuicListener::new(server_addr, Arc::clone(&psk_store))
```
gknopf-aranya: I see a listener and a transport object defined here. Typically, there would be a client/server or a transport which handles both send/recv and effectively combines both the client and server into the transport. Not sure I'm following why this was abstracted this way.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,69 @@
...
-use aranya_runtime::Address;
-use aranya_runtime::{PolicyStore, Sink, StorageProvider};
+pub(crate) mod quic;
 
-#[cfg(feature = "preview")]
-use super::GraphId;
-use super::{Result, SyncManager, SyncPeer};
+/// A reliable, ordered, bidirectional byte stream tied to a specific peer.
+///
+/// A single stream can be used for an entire sync conversation/exchange, dictated by the two peers'
```
gknopf-aranya: We usually refer to this as a sync request/response or handshake:
```suggestion
/// A single stream can be used for an entire sync request/response handshake, dictated by the two peers'
```

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,69 @@
...
+/// [`SyncManager`]. The protocol currently only needs one round trip, but this may change as the sync
```
gknopf-aranya: "The protocol currently only needs one round trip"
This is a protocol detail that a generic transport shouldn't need to care about. It may be relevant to the task manager that is driving the sync requests/responses.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,69 @@
...
-/// Types that contain additional data that are part of a [`SyncManager`] object.
-pub(crate) trait SyncState<PS, SP, EF>: Sized + fmt::Debug
-where
-    PS: PolicyStore,
-    SP: StorageProvider,
-{
-    /// Syncs with the peer.
+    /// Sends a message to the peer.
+    async fn send(&mut self, data: &[u8]) -> Result<(), Self::Error>;
+    /// Receives a message from the peer, returning the number of bytes written.
```
gknopf-aranya: The caller must allocate a fixed-size `buffer`. If the buffer is too small, this will return a `BufferToSmall` error.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,69 @@
...
+pub(crate) trait SyncListener: Send + Sync + 'static {
+    /// The specific error type this stream uses.
+    type Error: std::error::Error + Send + Sync + 'static;
+    /// The stream type returned from accepting a connection.
+    type Stream: SyncStream<Error = Self::Error>;
+
+    /// The local address the listener/server is bound to.
+    fn local_addr(&self) -> super::Addr;
+
+    /// Accept a connection from a peer to send and receive data.
```
gknopf-aranya, line 74: Worth documenting what `None`, `Some`, `Err` return values mean.

crates/aranya-daemon/src/sync/server.rs
```diff
@@ -0,0 +1,225 @@
+//! This module contains the [`SyncServer`] that handles incoming sync requests from peers.
+//!
+//! The server listens for incoming connections and processes requests. This includes poll syncing,
+//! in which we respond with a sampling of missing commands, as well as hello syncing, which we
```
gknopf-aranya: I don't think there's a need to mention initiating hello sync in this module since that is the responsibility of the task manager and client. Handling receiving of a hello sync message is the responsibility of this module.

crates/aranya-daemon/src/sync/server.rs
```diff
@@ -0,0 +1,225 @@
...
+use super::SyncManager;
+use super::{
+    transport::{SyncListener, SyncStream},
+    Addr, Error, SyncHandle, SyncPeer,
+};
+use crate::{aranya::Client, sync::SyncResponse};
+
+/// Handles listening for connections from peers and responding to them.
+///
+/// Uses a [`SyncHandle`] to offload hello sync scheduling and operations to the [`SyncManager`].
```
gknopf-aranya: Not following why the server is offloading operations to the task manager. I would expect the task manager to be responsible for all scheduling which offloads sync requests to the client. The server should just be passively responding to sync requests and updating the graph (no scheduling required).

crates/aranya-daemon/src/sync/server.rs
```diff
@@ -0,0 +1,225 @@
...
+                graph_id,
+            } => {
+                peer.check_request(graph_id)?;
+                self.handle
+                    .hello_subscribe_request(peer, graph_change_delay, duration, schedule_delay)
+                    .await?;
+            }
+            SyncHelloType::Unsubscribe { graph_id } => {
+                peer.check_request(graph_id)?;
+                self.handle.hello_unsubscribe_request(peer).await?;
```
gknopf-aranya: If a peer wants to unsubscribe from receiving hello sync messages from us, we can immediately remove them from the subscriptions map. Don't see a need to schedule that since it is not periodic or a network I/O operation.

crates/aranya-daemon/src/sync/server.rs
```diff
@@ -0,0 +1,225 @@
...
+            }
+            SyncHelloType::Hello { head, graph_id } => {
+                peer.check_request(graph_id)?;
+                self.handle.sync_on_hello(peer, head).await?;
```
gknopf-aranya: Now I understand why the server needed a handle to the sync manager in order to send a sync request after receiving a sync hello. It makes sense, but slightly breaks the abstraction of thinking of the sync server as a passive server since it is indirectly initiating client requests in the other direction.

crates/aranya-daemon/src/sync/server.rs
gknopf-aranya: Ah, it was to initiate a sync request in the other directly after receiving a sync hello:
https://github.com/aranya-project/aranya/pull/717/changes#r2865828350

crates/aranya-daemon/src/sync/transport/quic/client.rs
nikki-aranya: I've tried to separate out networking code before and it never really fit, in this prototype or with the original PR that got split up, I'll see if I can address this in PR3 along with poll.rs/hello.rs

crates/aranya-daemon/src/sync/transport/quic/connections.rs
nikki-aranya: Fixed the connection map after our discussion, the problem is it was always non-deterministic which I've now fixed (and explained the logic of)

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: In a PR that's goal is to improve organization and abstractions, combining these into a single abstraction feels like a step backwards (only for this specific change). Most of the rest of the changes seem like an improvement.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
nikki-aranya: the listener listens for connections (but uses the s2n_quic::Server for those, not my naming) and accepts them, the server is a broader object that includes replying on those connections, seems straightforward to me

crates/aranya-daemon/src/sync/transport/quic/listener.rs
nikki-aranya: Yes, this has been deprecated for a while, should be removed once mTLS lands

crates/aranya-daemon/src/daemon.rs
nikki-aranya: The listener is a new abstraction introduced by the transport traits as part of the broader server, it listens for connections and accepts them (doing all the stuff that accepting a connection entails like connection dedupe); this logic was originally part of the monolithic Quic/Server

crates/aranya-daemon/src/sync/transport/quic/psk.rs
nikki-aranya: We were already implicitly relying on it in tests, this just adds the trait https://github.com/aranya-project/aranya/blob/7a5fa131c3354630c7f66f5b2bd64c2bd6c31981/crates/aranya-daemon/src/test.rs#L234

crates/aranya-daemon/src/sync/transport/mod.rs
nikki-aranya: That's fair, I can take it or leave it

crates/aranya-daemon/src/daemon.rs
gknopf-aranya: In order for a generic transport to receive as expected, I believe the listener would be part of the transport abstraction. If you have a transport without a listener, how is it going to accept connections and receive data from peers?

crates/aranya-daemon/src/sync/server.rs
nikki-aranya: We "handle" hello syncs here, but the manager is expected to keep track of all the subscriptions and when to sync, that's its job, same as it is for normal peers, a poll doesn't need to be offloaded, it can be responded to directly, idk

nikki-aranya: I guess a good way to describe this is the manager initiates syncs (hence offloading sync_on_hello and hello subscription scheduling), the server responds to syncs

crates/aranya-daemon/src/sync/server.rs
nikki-aranya: Yeah, see the above reply

crates/aranya-daemon/src/sync/server.rs
gknopf-aranya: Yeah, and I think it's worth mentioning that this is not a passive server since there are certain types of requests it can receive that can cause a request to be sent the other way (since that isn't typical in a normal client/server architecture).

crates/aranya-daemon/src/sync/server.rs
nikki-aranya: The server shouldn't need to care what subscriptions there are, is the main reason for this design, plus it allows the manager to potentially do future cleanup if needed

crates/aranya-daemon/src/daemon.rs
```diff
@@ -287,23 +283,24 @@ impl Daemon {
...
+            .context("unable to initialize QUIC sync listener")?;
+        let server = SyncServer::new(listener, client.clone(), handle.clone());
+
+        let transport = QuicTransport::new(
+            client_addr,
+            server.local_addr(),
             conns,
+            conn_tx,
+            psk_store,
+            server.local_addr(),
```
benz-spideroak: The same parameter is passed twice. Will they ever have different values?

crates/aranya-daemon/src/sync/handle.rs
```diff
@@ -91,10 +114,17 @@ pub(crate) struct SyncHandle {
 
 impl SyncHandle {
     /// Creates a new [`SyncHandle`] for sending messages.
-    pub(crate) fn new(sender: mpsc::Sender<Callback>) -> Self {
+    #[allow(unused)]
+    pub(crate) const fn new(sender: mpsc::Sender<Callback>) -> Self {
         Self { sender }
     }
 
+    /// Crates a new mpsc channel and [`SyncHandle`] for sending messages.
```
benz-spideroak:
```suggestion
    /// Creates a new mpsc channel and [`SyncHandle`] for sending messages.
```

crates/aranya-daemon/src/sync/handle.rs
```diff
@@ -134,28 +164,54 @@ impl SyncHandle {
...
+        self.send(ManagerMessage::HelloSubscribeRequest {
+            peer,
+            graph_change_debounce,
+            duration,
+            schedule_delay,
+        })
+        .await
+    }
+
+    /// Tell the [`SyncManager`] to add this peer to their subscriptions.
```
benz-spideroak: Update doc.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -0,0 +1,196 @@
...
+use tokio::sync::mpsc;
+use tracing::{debug, error, trace, warn};
+
+use super::{
+    listener::{ConnectionUpdate, SharedConnectionMap},
+    Error, PskStore, QuicStream, SyncTransport, ALPN_QUIC_SYNC,
+};
+use crate::sync::{Addr, SyncPeer};
+
+#[derive(Clone, Debug)]
```
benz-spideroak: Does this need to derive Clone?

crates/aranya-daemon/src/sync/mod.rs
```diff
@@ -1,51 +1,45 @@
...
-    #[error("Could not receive sync response: {0}")]
-    ReceiveSyncResponse(Box<Error>),
-
     /// Peer sent an empty response.
     #[cfg(feature = "preview")]
     #[error("peer sent empty response")]
     EmptyResponse,
 
+    #[error(transparent)]
+    AranyaClient(#[from] aranya_runtime::ClientError),
```
jdygert-spok: Error handling could use some cleaning up, sometimes we construct this and sometimes we end up wrapping it into an `Other`.

crates/aranya-daemon/src/sync/transport/mod.rs
```diff
@@ -1,56 +1,75 @@
...
-#[cfg(feature = "preview")]
-use aranya_runtime::Address;
-use aranya_runtime::{PolicyStore, Sink, StorageProvider};
+/// A reliable, ordered, bidirectional byte stream tied to a specific peer.
+///
+/// A single stream can be used for an entire sync request/response handshake, dictated by the two
+/// peers' [`SyncManager`].
+///
+/// Abstracts over any reliable, ordered transport (QUIC, TCP, WebSocket, UDS, etc). Unreliable
+/// protocols like raw UDP are not suitable without an additional reliability layer.
```
jdygert-spok: This doesn't seem to communicate how the stream is message-oriented. (I assume it is at least)

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -0,0 +1,195 @@
...
+                let mut map = self.conns.lock().await;
+                match map.entry(peer) {
+                    Entry::Vacant(e) => {
+                        e.insert(new_handle.clone());
+                        (new_handle, Some(new_acceptor))
+                    }
+                    Entry::Occupied(mut e) => {
+                        let existing_alive = e.get_mut().ping().is_ok();
+                        // We initiated this connection, so it wins if we're the lower-addressed peer.
+                        let outbound_wins = !existing_alive || self.local_addr < peer.addr;
```
gknopf-aranya, line 147: I believe this is designed to ensure both peers pick the same connection. This works with direct connections. However, if there is a NAT, proxy, ISP, etc. between them it may not work.

crates/aranya-daemon/src/daemon.rs
```diff
@@ -287,23 +283,19 @@ impl Daemon {
...
-
-        Ok((client, server, syncer, peers, recv_effects))
+        let (handle, recv) = SyncHandle::channel(128);
+
+        let (listener, conns, conn_tx) = QuicListener::new(server_addr, Arc::clone(&psk_store))
+            .await
+            .context("unable to initialize QUIC sync listener")?;
+        let server = SyncServer::new(listener, client.clone(), handle.clone());
+
+        let connector =
```
gknopf-aranya, line 295: Would have expected the connection map to be created first and passed into client and server. Since connections are shared by the client and server.

crates/aranya-daemon/src/sync/transport/quic/stream.rs
```diff
@@ -0,0 +1,84 @@
...
+        self.stream.read_exact(buf).await.map_err(|error| {
+            warn!(peer = ?self.peer, %error, "receive failed");
+            Error::Receive(error)
+        })?;
+        Ok(len)
+    }
+
+    async fn finish(&mut self) -> Result<(), Self::Error> {
+        debug!(peer = ?self.peer, "closing stream");
+        self.stream.close().await.map_err(|error| {
```
gknopf-aranya:
```suggestion
        self.stream.finish().await.map_err(|error| {
```
The stream trait documents this as a half-close, but it is currently a full close:
```
    /// Receives a message from the peer, returning the number of bytes written. The caller
    /// allocates a fixed-size buffer, and if the received message is too large, the implementation
    /// is expected to throw an error.
    async fn receive(&mut self, buffer: &mut [u8]) -> Result<usize, Self::Error>;
    /// Signals that no more data will be sent on this stream.
    ///
    /// For transports that support half-close (e.g. QUIC, TCP), this notifies the remote peer that
    /// sending is complete. Transports without half-close can implement this as a no-op.
    async fn finish(&mut self) -> Result<(), Self::Error>;
```

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -1,310 +1,464 @@
...
+            }
+            true
+        });
+
+        // Loop through all subscribers and send them a hello notification.
+        for peer in &subscribers {
+            if let Err(error) = self.client.send_hello_notification(*peer, head).await {
+                warn!(?peer, %error, "failed to send hello notification");
+            } else if let Some(sub) = self.hello_subscriptions.get_mut(peer) {
+                sub.last_notified = Instant::now();
```
gknopf-aranya, line 465: If `last_notified` isn't updated if `send_hello_notification()` fails, I believe another hello broadcast will be sent immediately. This could overwhelm a peer that already isn't responsive. We should consider an exponential backoff or updating `last_notified` anyways so another hello message isn't sent until after a delay.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,276 @@
...
+                    match result {
+                        Ok((peer, acceptor, Some(stream))) => {
+                            debug!(?peer, "accepted stream, re-queueing acceptor");
+                            self.pending_accepts.spawn(Self::accept_pending_stream(peer, acceptor));
+                            return Some(Ok(QuicStream::new(peer, stream)));
+                        }
+                        Ok((peer, _acceptor, None)) => {
+                            debug!(?peer, "stream acceptor completed with no stream");
+                        }
+                        Err(error) => warn!(%error, "stream acceptor task panicked"),
```
gknopf-aranya: Is it a bug if the stream acceptor task panics?
```suggestion
                        Err(error) => error!(%error, "stream acceptor task panicked"),
```

crates/aranya-daemon/src/sync/client.rs
```diff
@@ -0,0 +1,292 @@
...
+        debug!(?peer, "starting sync");
+
+        // Connect to the peer.
+        let mut stream = self.connector.connect(peer).await.map_err(|error| {
+            warn!(?peer, %error, "failed to connect to peer");
+            Error::transport(error)
+        })?;
+
+        let mut requester = SyncRequester::new(peer.graph_id, Rng);
+        let mut buf = vec![0u8; MAX_SYNC_MESSAGE_SIZE].into_boxed_slice();
```
gknopf-aranya: This is allocated per sync. Since we don't support parallel sync yet, this could be allocated once and passed in by the caller. When we're ready to support parallel sync the caller could manage a buffer pool.

nikki-aranya: I wasn't sure whether to do that or not, probably should since I already put in the effort of reducing allocations compared to mainline

crates/aranya-daemon/src/sync/manager.rs
```diff
@@ -1,310 +1,464 @@
...
+
+            // Tell the client that we encountered an invalid graph.
+            self.client.invalid_graphs().insert(peer.graph_id);
+        }
+    }
+
+    /// Returns the peer cache map for tests that need it.
+    #[cfg(test)]
+    pub(crate) fn get_peer_caches(&self) -> crate::aranya::PeerCacheMap {
+        self.client.client.caches_for_test()
```
gknopf-aranya, line 199: Or return the client with a getter/accessor method:
```suggestion
        self.client().caches_for_test()
```

crates/aranya-daemon/src/sync/client.rs
gknopf-aranya: I wouldn't say this change needs to be on this PR. I'm pointing out a couple improvements which may have already been issues prior to the refactor that are easier to spot while reviewing this diff. Worth adding a TODO comment at least.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
gknopf-aranya: Would be more reliable to break this tie based on QUIC connection IDs or device IDs.

gknopf-aranya: I think QUIC connection IDs is the way to go since that is randomly generated and guaranteed to be available to the QUIC transport. Device IDs ties the transport implementation to something Aranya-specific which we're generally wanting to avoid.

crates/aranya-daemon/src/sync/manager.rs
nikki-aranya: meh, they're internal, they don't need to be pretty, and client() would be client.client anyways so eh

crates/aranya-daemon/src/sync/transport/quic/connector.rs
gknopf-aranya: @nikki-aranya, @jdygert-spok, @djohnson-spideroak I think we're going to want to open an issue to resolve this. Seems like a bug to me.

nikki-aranya: Added

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -0,0 +1,198 @@
...
+                    Entry::Vacant(e) => {
+                        e.insert(new_handle.clone());
+                        (new_handle, Some(new_acceptor))
+                    }
+                    Entry::Occupied(mut e) => {
+                        let existing_alive = e.get_mut().ping().is_ok();
+                        // We initiated this connection, so it wins if we're the lower-addressed peer.
+                        // TODO(nikki): This semi-fixes an existing bug, but we need a better way
+                        // than raw addresses, this will break with NAT.
+                        // https://github.com/aranya-project/aranya/issues/754
```
gknopf-aranya, line 146: nit: not the same format we use to reference issues elsewhere in the codebase

nikki-aranya: I don't know that there's a good way unless we standardize on a VSCode plugin that flags all TODOs in the codebase or something, all I remember is it's like `#754` after the : or something

crates/aranya-daemon/src/sync/transport/mod.rs
jdygert-spok: This still says "byte stream"

crates/aranya-daemon/src/daemon.rs
```diff
@@ -286,24 +282,28 @@ impl Daemon {
...
-            psk_store,
-            (server.local_addr().into(), client_addr),
-            syncer_recv,
-            conns,
-        )?;
-
-        Ok((client, server, syncer, peers, recv_effects))
+        // Create a `SharedConnectionMap` to allow for reusing QUIC connections.
+        let conns: SharedConnectionMap = Arc::default();
+        let (conn_tx, conn_rx) = mpsc::channel(32);
```
jdygert-spok: Seems funky to me at first glance that these are separate.

crates/aranya-daemon/src/test.rs
```diff
@@ -118,9 +121,14 @@ impl TestDevice {
         device: &TestDevice,
         must_receive: Option<usize>,
     ) -> Result<Vec<Effect>> {
+        let mut buffer = vec![0u8; MAX_SYNC_MESSAGE_SIZE].into_boxed_slice();
```
jdygert-spok: nit: store buffer in `TestDevice`

crates/aranya-daemon/src/sync/types.rs
```diff
@@ -24,3 +40,19 @@ pub(crate) enum SyncResponse {
     /// Failure.
     Err(String),
 }
+
+/// Storage for a subscription to hello messages.
+#[derive(Debug, Clone)]
+#[cfg(feature = "preview")]
```
jdygert-spok: nit: cfg above derive

(I'm curious what mechanism makes this work as-is...)

crates/aranya-daemon/src/sync/mod.rs
```diff
@@ -70,15 +77,20 @@ impl From<std::convert::Infallible> for Error {
...
-                .downcast_ref::<ClientError>()
-                .is_some_and(|err| matches!(err, ClientError::ParallelFinalize)),
-            _ => false,
-        }
+        debug_assert!(
+            !matches!(self, Self::Other(err) if err
+            .downcast_ref::<ClientError>()
+            .is_some_and(|e| matches!(e, ClientError::ParallelFinalize))),
+            "ParallelFinalize should not be wrapped in Other"
+        );
```
jdygert-spok: It looks like we can get rid of the `Other` variant now pretty easily, so then we could remove this.

I do wonder if the finalize error could reasonably sneak into the `Transport` variant. I think not really, but I am just a little wary.

crates/aranya-daemon/src/sync/handle.rs
```diff
@@ -91,10 +114,17 @@ pub(crate) struct SyncHandle {
 
 impl SyncHandle {
     /// Creates a new [`SyncHandle`] for sending messages.
-    pub(crate) fn new(sender: mpsc::Sender<Callback>) -> Self {
+    #[allow(unused)]
+    pub(crate) const fn new(sender: mpsc::Sender<Callback>) -> Self {
         Self { sender }
     }
```
jdygert-spok: Why keep unused?

crates/aranya-daemon/src/sync/server.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Uses a [`SyncHandle`] to offload hello sync scheduling and operations to the [`SyncManager`].
+#[derive_where(Debug; SL)]
+pub(crate) struct SyncServer<SL, PS, SP> {
+    /// The Aranya client and peer cache, alongside invalid graph tracking.
+    client: Client<PS, SP>,
+    /// The listener that yields incoming streams.
+    listener: SL,
+    /// Handle to allow sending messages to the [`SyncManager`].
+    #[allow(dead_code, reason = "only used in preview right now")]
+    handle: SyncHandle,
```
jdygert-spok: Could be more precise with

```rust
#[cfg_attr(not(feature = "preview"), expect(dead_code))]
```

crates/aranya-daemon/src/sync/server.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub(crate) async fn serve(mut self, ready: ready::Notifier) {
+        info!("sync server listening for incoming connections");
+        ready.notify();
+
+        while let Some(stream_result) = self.listener.accept().await {
+            match stream_result {
+                Ok(stream) => {
+                    let peer = stream.peer();
+                    trace!(?peer, "accepted stream");
+                    if let Err(e) = self.handle_stream(stream).await {
```
jdygert-spok: Notable change that we only handle one stream at a time. I believe previously we had concurrency everywhere except the part that needed to interact with the `ClientState`.

crates/aranya-daemon/src/sync/client.rs
```diff
@@ -0,0 +1,308 @@
...
+        let cmd_count = self
+            .process_sync_data(peer, &data, &mut requester, &mut sink)
+            .await?;
+
+        // Send all processed effects to the Daemon API.
+        let effects = sink
+            .collect()
+            .map_err(|e| Error::EffectsSink(Box::new(e)))?;
+        let effects_count = effects.len();
+        if let Err(error) = self.send_effects.send((peer.graph_id, effects)).await {
```
jdygert-spok, line 302: I wonder if we should just send `Policy::Effect` and do the conversion in the effect handler. Although when we get back effects from actions we need to convert them there anyways :/

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,264 @@
...
+    fn local_addr(&self) -> Addr {
+        self.local_addr
+    }
+
+    #[allow(
+        clippy::disallowed_macros,
+        reason = "tokio::select! uses core::unreachable"
+    )]
+    async fn accept(&mut self) -> Option<Result<Self::Stream, Self::Error>> {
+        loop {
```
jdygert-spok: Looks like we never return an error from this, and the caller would only log it anyway, so we can remove the error type here.

Cargo.toml
```diff
@@ -67,6 +67,7 @@ aranya-util = { version = "5.0.0", path = "crates/aranya-util", default-features
...
```
jdygert-spok: Why'd you add this back?

crates/aranya-daemon/src/sync/server.rs
```diff
@@ -71,16 +68,35 @@ where
         info!("sync server listening for incoming connections");
         ready.notify();
 
-        while let Some(stream_result) = self.listener.accept().await {
-            match stream_result {
-                Ok(stream) => {
+        let mut tasks = JoinSet::new();
+        loop {
```
jdygert-spok: Seems like a good spot for `aranya_util::task::scope`:

```rust
aranya_util::task::scope(async |s| {
    while let Some(stream) = self.listener.accept().await {
        let peer = stream.peer();
        trace!(?peer, "accepted stream");
        let client = self.client.clone();
        let handle = self.handle.clone();
        s.spawn(async move {
            if let Err(e) = Self::handle_stream(client, handle, stream).await {
                warn!(?peer, error = %e.report(), "error handling sync request");
            }
        });
    }
})
.await;
```

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,257 @@
...
+            rustls::{server::PresharedKeySelection, ServerConfig},
+        },
+    },
+    stream::BidirectionalStream,
+};
+use tokio::{task::JoinSet, time::Duration};
+use tracing::{debug, error, trace};
+
+#[cfg(doc)]
+use super::QuicConnector;
```
jdygert-spok: nit: just write `super::QuicConnector` in the doc comment.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -0,0 +1,257 @@
...
+    async fn accept_pending_stream(peer: SyncPeer, mut acceptor: StreamAcceptor) -> AcceptResult {
+        trace!(?peer, "waiting for bidirectional stream");
+
+        let stream = tokio::time::timeout(
+            PENDING_ACCEPT_TIMEOUT,
+            acceptor.accept_bidirectional_stream(),
+        )
+        .await
+        .ok()
+        .and_then(|r| r.ok().flatten());
```
jdygert-spok: We should log the accept error. Also I think we don't want to throw away the acceptor when it returns an error, only when it returns None.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+impl SyncConnector for QuicConnector {
+    type Error = Error;
+    type Stream = QuicStream;
+
+    async fn connect(&self, peer: SyncPeer) -> Result<Self::Stream, Self::Error> {
+        // Set the current `GraphId` we're operating on in the PSK store.
+        self.psk_store.set_team(TeamId::transmute(peer.graph_id));
```
jdygert-spok: Since this takes `&self`, is it possible to initiate two connections at once, using the wrong team on the psk store for one of them?

nikki-aranya: technically, yes, if you called it directly, that's why we need to hurry up with mTLS. In practice, the current implementation is single task due to SyncManager::next()

jdygert-spok: Added guard to make sure it can't happen.

PR #719, opened 2026-02-12
metrics: add linux

djohnson-spideroak:
Add linux support to the metrics crate. Uses /proc and sysinfo for gathering the metrics.

### DIFF
--- a/crates/aranya-metrics/src/harness.rs
+++ b/crates/aranya-metrics/src/harness.rs
@@ -6,8 +6,10 @@
 //! separate process, it's impossible to fully remove the effect of measuring a process. See the
 //! [observer problem] for more details.
 //!
-//! Specifically, this uses `proc_pidinfo` on MacOS and `/proc/{PID}/stat` on Linux to collect CPU
-//! and memory usage, falling back to the `sysinfo` crate for disk usage stats.
+//! Specifically, this uses `proc_pidinfo` on macOS and `/proc/{PID}/stat` +
+//! `/proc/{PID}/smaps_rollup` on Linux to collect CPU and memory usage (with
+//! PSS-based detailed memory breakdown when available on kernel >= 4.14),
+//! falling back to the `sysinfo` crate for disk usage stats.
 //!
 //! [observer problem]: https://w.wiki/Ekxn
 use std::{collections::HashMap, fmt, time::Instant};
@@ -75,6 +77,20 @@ struct ProcessMetrics {
     disk_read_bytes: u64,
     /// The amount of data written from disk.
     disk_write_bytes: u64,
+    /// Proportional share size in bytes (from smaps_rollup; 0 if unavailable).
+    pss_bytes: u64,
+    /// PSS of anonymous (heap/stack) pages in bytes.
+    pss_anon_bytes: u64,
+    /// PSS of file-backed pages in bytes.
+    pss_file_bytes: u64,
+    /// PSS of shared memory pages in bytes.
+    pss_shmem_bytes: u64,
+    /// Bytes currently swapped out.
+    swap_bytes: u64,
+    /// Private dirty (non-reclaimable) pages in bytes.
+    private_dirty_bytes: u64,
+    /// Shared clean (freely reclaimable) pages in bytes.
+    shared_clean_bytes: u64,
     /// The number of processes this collection represents.
     process_count: usize,
     // TODO(nikki): these need to be collected inside aranya-daemon
@@ -92,6 +108,13 @@ impl Default for ProcessMetrics {
             virtual_memory_bytes: 0,
             disk_read_bytes: 0,
             disk_write_bytes: 0,
+            pss_bytes: 0,
+            pss_anon_bytes: 0,
+            pss_file_bytes: 0,
+            pss_shmem_bytes: 0,
+            swap_bytes: 0,
+            private_dirty_bytes: 0,
+            shared_clean_bytes: 0,
             process_count: 1,
         }
     }
@@ -102,10 +125,13 @@ impl fmt::Display for ProcessMetrics {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "User Time: {}ms, System Time: {}ms, Physical Memory: {}, Disk Reads: {}, Disk Writes: {}",
+            "User Time: {}ms, System Time: {}ms, Physical Memory: {}, PSS: {}, Swap: {}, \
+             Disk Reads: {}, Disk Writes: {}",
             self.cpu_user_time_us as f64 / 1000.0,
             self.cpu_system_time_us as f64 / 1000.0,
             scale_bytes(self.physical_memory_bytes),
+            scale_bytes(self.pss_bytes),
+            scale_bytes(self.swap_bytes),
             scale_bytes(self.disk_read_bytes),
             scale_bytes(self.disk_write_bytes)
         )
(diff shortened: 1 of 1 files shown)


crates/aranya-metrics/src/harness.rs
```diff
@@ -292,17 +292,130 @@ impl ProcessMetricsCollector {
...
+
+        let utime_ticks: u64 = fields[11].parse()?;
+        let stime_ticks: u64 = fields[12].parse()?;
+        let vsize: u64 = fields[20].parse()?;
+        let rss_pages: u64 = fields[21].parse()?;
+
+        // SAFETY: sysconf with _SC_CLK_TCK is always a valid call.
+        let clock_ticks_per_sec = unsafe { libc::sysconf(libc::_SC_CLK_TCK) };
+        // SAFETY: sysconf with _SC_PAGESIZE is always a valid call.
+        let page_size = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
```
djohnson-spideroak: I am not sure if both of these are always available on linux, or just POSIX compliant linux.

crates/aranya-metrics/src/harness.rs
```diff
@@ -292,17 +292,130 @@ impl ProcessMetricsCollector {
         Ok(())
     }
 
+    /// Collects metrics for a specific process using the Linux `/proc` filesystem.
+    #[cfg(target_os = "linux")]
+    fn collect_process_metrics(&mut self, pid: Pid, metrics: &mut ProcessMetrics) -> Result<()> {
```
gknopf-aranya, line 536: Seems like we can move the linux-specific code into a `linux.rs` module and gate it all with `#[cfg(target_os = "linux")]` once.

crates/aranya-metrics/src/harness.rs
```diff
@@ -292,17 +292,130 @@ impl ProcessMetricsCollector {
...
+
+        #[allow(clippy::cast_sign_loss)]
+        {
+            let ticks = clock_ticks_per_sec as u64;
+            // Convert clock ticks to microseconds: value * 1_000_000 / ticks_per_sec
+            process_metrics.cpu_user_time_us = utime_ticks * 1_000_000 / ticks;
+            process_metrics.cpu_system_time_us = stime_ticks * 1_000_000 / ticks;
+            process_metrics.virtual_memory_bytes = vsize;
+            process_metrics.physical_memory_bytes = rss_pages * page_size as u64;
+        }
```
djohnson-spideroak: fix to remove lint ignore

crates/aranya-metrics/src/harness.rs
```diff
@@ -292,17 +292,130 @@ impl ProcessMetricsCollector {
...
+    fn collect_process_metrics(&mut self, pid: Pid, metrics: &mut ProcessMetrics) -> Result<()> {
+        use std::collections::hash_map::Entry;
+
+        use crate::backend::DebugLogType;
+
+        // First, let's collect metrics for the individual process.
+        let mut process_metrics = ProcessMetrics::default();
+
+        // Collect what we can using /proc, and fall back to sysinfo for disk stats.
+        self.collect_native_linux_metrics(pid, &mut process_metrics)?;
```
gknopf-aranya, line 545: This appears to be running all the metrics for both `/proc` and `sysinfo` while falling back to `sysinfo` if they aren't available in `/proc`. It seems like it would be more efficient to only run the specific `sysinfo` ones that aren't already provided by `/proc`. Is there no way to tell ahead of time which ones are available?

crates/aranya-metrics/src/harness.rs
djohnson-spideroak: I agree, I think that would be good as follow up. It would mean doing the same to put the macos stuff in its own file as well, and that is a lot of diff noise.

crates/aranya-metrics/src/harness.rs
```diff
@@ -292,17 +292,130 @@ impl ProcessMetricsCollector {
...
+        self.collect_sysinfo_disk_metrics(pid, &mut process_metrics)?;
+
+        // Aggregate this process's metrics towards the total.
+        metrics.cpu_user_time_us += process_metrics.cpu_user_time_us;
```
gknopf-aranya: Thinking about whether there's a way to loop over the fields instead of manually performing arithmetic on each one. If the fields were stored in map it would be easier to iterate over them for `metrics` and `process_metrics`.

gknopf-aranya: E.g. something like this:
```
for metric in metrics {
    metrics[metric.name] += process_metrics[metric.name]
}
```

crates/aranya-metrics/src/harness.rs
gknopf-aranya, line 536: Could add tracing to indicate `linux` and `/proc` in case that helps with troubleshooting where the metrics came from.

crates/aranya-metrics/src/harness.rs
```diff
@@ -292,17 +292,130 @@ impl ProcessMetricsCollector {
...
+    ) -> Result<()> {
+        use std::fs;
+
+        let stat_path = format!("/proc/{}/stat", pid.pid);
+        let stat_contents =
+            fs::read_to_string(&stat_path).map_err(|e| anyhow!("Failed to read {stat_path}: {e}"))?;
+
+        // The comm field (field 2) is wrapped in parentheses and may contain spaces or
+        // other parentheses. Find the last ')' to reliably skip past it.
+        let rest = stat_contents
```
gknopf-aranya: Not following what this parsing logic is doing fully. Seems like it could be refactored into a `.parse()` with more docs on what the format is.

crates/aranya-metrics/src/harness.rs
```diff
@@ -292,17 +292,130 @@ impl ProcessMetricsCollector {
...
+        //   21 = rss      (field 24) - resident set size in pages
+        if fields.len() < 22 {
+            return Err(anyhow!(
+                "Not enough fields in /proc/{}/stat (got {})",
+                pid.pid,
+                fields.len()
+            ));
+        }
+
+        let utime_ticks: u64 = fields[11].parse()?;
```
gknopf-aranya: Is there a reason for manually parsing `/proc` data rather than using a lib like:
https://docs.rs/metrics-process/latest/metrics_process/
https://docs.rs/procfs/latest/procfs/

crates/aranya-metrics/src/harness.rs
```diff
@@ -426,3 +539,79 @@ impl ProcessMetricsCollector {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
gknopf-aranya, line 863: Should we have a test that proves `sysinfo` is used as a backup if `/proc` isn't available?

crates/aranya-metrics/src/harness.rs
djohnson-spideroak: I dont think that is really in scope for this PR. If we want to refactor this we should do it in a different PR.

crates/aranya-metrics/src/harness.rs
djohnson-spideroak: Sysinfo doesnt give us everything we need. CPU measurements from sysinfo are percentages. We could swap to sysinfo for the memory but if we are already reading /proc, we may as well grab it from there. Sysinfo is mostly used for the per-pid disk stats.

crates/aranya-metrics/src/harness.rs
djohnson-spideroak: Unless we need features from those crates, I dont see a reason to add a dependency if we can help it. That first one you linked looks interesting if we want to offload the cross-platform work though. In this case, I think its simple enough for us to just parse it.

crates/aranya-metrics/src/harness.rs
djohnson-spideroak: Sysinfo isnt being used as a backup, it is being used where proc doesnt have the data we need.

crates/aranya-metrics/src/harness.rs
djohnson-spideroak: It seems difficult to confuse if you are running this on macos or linux, and this is not really meant as an external tool we would expect customers to run.

crates/aranya-metrics/src/harness.rs
gknopf-aranya: Reviewing usage of a `/proc` parsing lib seems easier than verifying the parsing here. I'm not very familiar with the `/proc` format.

crates/aranya-metrics/src/harness.rs
```diff
@@ -314,17 +374,138 @@ impl ProcessMetricsCollector {
         Ok(())
     }
 
+    /// Collects metrics for a specific process using the Linux `/proc` filesystem.
+    #[cfg(target_os = "linux")]
+    fn collect_process_metrics(&mut self, pid: Pid, metrics: &mut ProcessMetrics) -> Result<()> {
+        use std::collections::hash_map::Entry;
+
+        use crate::backend::DebugLogType;
```
nikki-aranya, line 539: I'm never sure how I feel about inner use statements, given it's only for this specific target it seems fine

djohnson-spideroak: Yeah, I considered moving this one outside of this scope and came to the same conclusion.

crates/aranya-metrics/src/harness.rs
```diff
@@ -135,6 +135,66 @@ fn scale_bytes(bytes: u64) -> String {
...
+#[derive(Debug)]
+struct ProcStat {
+    /// Clock ticks spent in user mode (field 14).
+    utime_ticks: u64,
+    /// Clock ticks spent in kernel mode (field 15).
+    stime_ticks: u64,
+    /// Virtual memory size in bytes (field 23).
+    vsize: u64,
+    /// Resident set size in pages (field 24).
+    rss_pages: u64,
```
ssweetney-spok, line 175: > Resident Set Size: number of pages the process has
                     in real memory.  This is just the pages which count
                     toward text, data, or stack space.  This does not
                     include pages which have not been demand-loaded in,
                     or which are swapped out.  This value is **inaccurate**;
                     see /proc/pid/statm below.

- From https://www.man7.org/linux/man-pages/man5/proc_pid_status.5.html

Based off of the description, should we use something else like `/proc/{pid}/smap` instead? I'm not sure how precise the metrics need to be so this may be OK.

Edit: I just saw that `smap` requires enabling a kernel option

PR #720, opened 2026-02-12
chore: add `clippy::arithmetic_side_effects` workspace lint

djohnson-spideroak:
Enable the lint at warn level to catch unchecked integer arithmetic, pushing toward checked/saturating operations and preventing potential overflow panics. Production code uses checked_add/checked_sub where results feed into fallible paths, saturating_add for metrics counters, and point-allows with justification for provably-bounded operations. Test modules and examples receive broader allows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,24 +37,24 @@ unwrap_used = "warn"
 wildcard_imports = "warn"
 
 [workspace.dependencies]
-aranya-fast-channels = { version = "0.17.1", features = ["alloc"] }
-aranya-afc-util = { version = "0.20.0", features = ["alloc"] }
-aranya-capi-core = { version = "0.7.0" }
-aranya-capi-codegen = { version = "0.6.0" }
-aranya-crypto = { version = "0.12.1", features = ["alloc", "fs-keystore", "clone-aead", "std", "tls"] }
-aranya-crypto-ffi = { version = "0.18.0" }
-aranya-device-ffi = { version = "0.18.0" }
-aranya-envelope-ffi = { version = "0.18.0" }
+aranya-fast-channels = { version = "0.18.0", features = ["alloc"] }
+aranya-afc-util = { version = "0.21.0", features = ["alloc"] }
+aranya-capi-core = { version = "0.8.0" }
+aranya-capi-codegen = { version = "0.7.0" }
+aranya-crypto = { version = "0.13.0", features = ["alloc", "fs-keystore", "clone-aead", "std", "tls"] }
+aranya-crypto-ffi = { version = "0.19.0" }
+aranya-device-ffi = { version = "0.19.0" }
+aranya-envelope-ffi = { version = "0.19.0" }
 aranya-id = { version = "0.2.1" }
-aranya-idam-ffi = { version = "0.18.0" }
-aranya-perspective-ffi = { version = "0.18.0" }
-aranya-policy-compiler = { version = "0.19.0" }
-aranya-policy-ifgen = { version = "0.19.0", features = ["serde"] }
-aranya-policy-ifgen-build = { version = "0.12.0" }
-aranya-policy-lang = { version = "0.12.0" }
+aranya-idam-ffi = { version = "0.19.0" }
+aranya-perspective-ffi = { version = "0.19.0" }
+aranya-policy-compiler = { version = "0.20.0" }
+aranya-policy-ifgen = { version = "0.20.0", features = ["serde"] }
+aranya-policy-ifgen-build = { version = "0.13.0" }
+aranya-policy-lang = { version = "0.13.0" }
 aranya-policy-text = { version = "0.1.4" }
-aranya-policy-vm = { version = "0.18.0", features = ["std"] }
-aranya-runtime = { version = "0.19.0", features = ["std", "libc"] }
+aranya-policy-vm = { version = "0.19.0", features = ["std"] }
+aranya-runtime = { version = "0.20.0", features = ["std", "libc"] }
 
 buggy = { version = "0.1.0" }
 spideroak-base58 = { version = "0.2.0" }
--- a/crates/aranya-daemon-api/src/crypto/txp.rs
+++ b/crates/aranya-daemon-api/src/crypto/txp.rs
@@ -65,7 +65,7 @@ impl<CS: CipherSuite> Ctx<CS> {
 
     /// Creates the HPKE encryption context for the client.
     fn client<R: Csprng>(
-        rng: &mut R,
+        rng: R,
         pk: &PublicApiKey<CS>,
         info: &[u8],
     ) -> Result<(Self, Encap<CS>), HpkeError> {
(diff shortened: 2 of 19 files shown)


crates/aranya-metrics/src/main.rs
```diff
@@ -221,6 +221,7 @@ async fn setup_demo(team_name: &str) -> Result<(Vec<Pid>, DemoContext)> {
 
     const CLIENT_NAMES: [&str; 5] = ["owner", "admin", "operator", "member_a", "member_b"];
     let mut contexts: [Option<ClientCtx>; CLIENT_NAMES.len()] = Default::default();
+    #[allow(clippy::arithmetic_side_effects)] // 5 + 1 cannot overflow
     let mut daemon_pids: Vec<Pid> = Vec::with_capacity(CLIENT_NAMES.len() + 1);
```
gknopf-aranya: Number of daemons could change. `saturating_add()` should work here.

gknopf-aranya: Or probably `check_add()` since that would be way too many daemons if it overflows.

examples/rust/aranya-example/src/main.rs
```diff
@@ -187,6 +187,7 @@ async fn main() -> Result<()> {
     };
 
     let sync_interval = Duration::from_millis(100);
+    #[allow(clippy::arithmetic_side_effects)] // 100ms * 6 cannot overflow
```
gknopf-aranya: Someone could set the sync interval to something higher

examples/rust/aranya-example-multi-node/src/membera/main.rs
```diff
@@ -116,7 +116,12 @@ async fn main() -> Result<()> {
                 break;
             }
         }
-        sleep(3 * SLEEP_INTERVAL).await;
+        sleep(
+            SLEEP_INTERVAL
+                .checked_mul(3)
+                .expect("sleep interval should not overflow"),
+        )
+        .await;
```
djohnson-spideroak: We could probably ignore these, but I dont see too much of a downside of having checked arithmetic here. It does make it a bit more cluttered though.

crates/aranya-util/src/addr.rs
```diff
@@ -567,7 +587,13 @@ const fn base16(x: u8) -> u8 {
     if x < 10 {
         base10(x)
     } else {
-        x - 10 + b'a'
+        match x.checked_sub(10) {
+            Some(offset) => match offset.checked_add(b'a') {
+                Some(v) => v,
+                None => panic!("base16 overflow"),
+            },
+            None => panic!("base16 underflow"),
```
djohnson-spideroak: Thoughts on these panics? Seems non-optimal, but a panic is probably a better outcome than getting a bad result. Either way, we hav a debug assert above that x <= 15.

gknopf-aranya: If there is always a base-16 representation of a u8 then these panics should never be reached.

gknopf-aranya: Panicking is generally okay for infallible operations since it indicates that something went really wrong. E.g. undefined program execution or possibly corrupted memory/hardware.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -682,6 +682,7 @@ pub unsafe fn get_key_bundle(
 }
 
 /// The size in bytes of an ID converted to a human-readable base58 string.
+#[allow(clippy::arithmetic_side_effects)] // constant expression
 pub const ARANYA_ID_STR_LEN: usize = (ARANYA_ID_LEN * 1375) / 1000 + 1;
```
jdygert-spok: I don't think the lint fires on const expressions

crates/aranya-daemon-api/src/crypto/txp.rs
```diff
@@ -319,7 +319,7 @@ where
         // rekey when we're halfway there.
         let max = Seq::max::<<CS::Aead as Aead>::NonceSize>();
         let seq = ctx.seal.seq().to_u64();
-        seq >= max / 2
+        seq >= max.checked_div(2).expect("division by 2 should not fail")
```
jdygert-spok: lint doesn't fire on division with nonzero rhs

crates/aranya-daemon-api/src/crypto/txp.rs
```diff
@@ -329,7 +329,7 @@ where
         self.ctx = Some(ctx);
         // Rekeying takes so long (relatively speaking, anyway)
         // that this should never overflow.
-        self.rekeys += 1;
+        self.rekeys = self.rekeys.saturating_add(1);
```
jdygert-spok: Might as well use `checked_add(1).assume(...)?`

crates/aranya-metrics/src/main.rs
```diff
@@ -221,7 +221,12 @@ async fn setup_demo(team_name: &str) -> Result<(Vec<Pid>, DemoContext)> {
 
     const CLIENT_NAMES: [&str; 5] = ["owner", "admin", "operator", "member_a", "member_b"];
     let mut contexts: [Option<ClientCtx>; CLIENT_NAMES.len()] = Default::default();
-    let mut daemon_pids: Vec<Pid> = Vec::with_capacity(CLIENT_NAMES.len() + 1);
+    let mut daemon_pids: Vec<Pid> = Vec::with_capacity(
+        CLIENT_NAMES
+            .len()
+            .checked_add(1)
+            .expect("capacity should not overflow"),
+    );
```
jdygert-spok: Can make it a const expr to ensure no overflow.

```suggestion
    let mut daemon_pids: Vec<Pid> = Vec::with_capacity(const { CLIENT_NAMES.len() + 1 });
```

examples/rust/aranya-example-multi-node/src/membera/main.rs
jdygert-spok:
```suggestion
        sleep(const { SLEEP_INTERVAL.checked_mul(3).unwrap() }).await;
```

Cargo.toml
```diff
@@ -33,6 +33,7 @@ panic = "warn"
 unsafe_derive_deserialize = "warn"
 undocumented_unsafe_blocks = "warn"
 unwrap_used = "warn"
+arithmetic_side_effects = "warn"
```
jdygert-spok: nit: sort these

crates/aranya-util/src/addr.rs
jdygert-spok: IIRC addr.rs was written in a somewhat convoluted way to ensure that no path could panic, so I'm hesitant to add in panics.

djohnson-spideroak: I can ignore this file if we think the bounds checks + general construction of the Addr is sufficient.

jdygert-spok (approved):

jdygert-spok:
Ugh I think github somehow dropped my comment on missing some lints that were cfg'd mac-only, and I forgot about it.

PR #723, opened 2026-02-13
ci: automate sprint issue transfer

### LINKED ISSUE #708: pm: automate transferring issues to next sprint
We used to have a Golang script that moved the issues over for us.
Once we closed that repo and migrated over to Rust, we no longer had a script that could automatically move issues to the next sprint.

The downsides of not having this automated are:
- Increased developer overhead each sprint required to manually move dozens of issues to the next sprint before sprint planning
- Frequent reminders to developers who may have forgotten to move their issues before sprint planning starts
- Losing track of issues that get lost in previous sprints (sometimes 2-3 or more sprints behind that are never brought into the current sprint again). This is a project management risk.
- A more disorganized backlog that may include issues that should be closed in favor of other issues.

@jdygert-spok suggested that rewriting the script in TypeScript could be a possibility since that is the preferred way to interact with the GitHub APIs.

Another option is to bring the old Go script back from the old repo so it can run on its own.

cc: @djohnson-spideroak, @jdygert-spok, @chip-so, @ycarmy

gknopf-aranya:
## Summary

- Adds a scheduled GitHub Actions workflow that automatically moves incomplete issues from the previous sprint iteration to the current one using [blombard/move-to-next-iteration](https://github.com/blombard/move-to-next-iteration) (pinned to v0.6.1 by SHA)
- Runs every Monday at 05:00 UTC, and can also be triggered manually via `workflow_dispatch`
- Issues with status "Done" or "Won't Fix" are excluded from transfer

## Setup required

After merging, the following repository-level configuration must be set:

| Setting | Type | Description |
|---------|------|-------------|
| `PROJECT_NUMBER` | Variable | The GitHub Project number from the project URL |
| `ITERATION_FIELD` | Variable | The name of your iteration field (e.g. "Sprint") |
| `PROJECT_PAT` | Secret | A PAT with `project` write scope |

## Test plan

- [ ] Verify the workflow appears in the Actions tab after merge
- [ ] Configure the required variables and secret
- [ ] Trigger the workflow manually via `workflow_dispatch` and confirm issues move correctly
- [ ] Confirm the scheduled Monday run works on the next sprint boundary

Closes #708

### DIFF
--- a/.github/workflows/sprint-transfer.yml
+++ b/.github/workflows/sprint-transfer.yml
@@ -35,9 +35,4 @@ jobs:
           iteration-field: Sprint
           iteration: last
           new-iteration: current
-          excluded-statuses: >-
-            ✅ Done,
-            ❌ Won't Do,
-            🖨️ Duplicate,
-            ❓ Cannot Reproduce,
-            ⛔️ Invalid
+          excluded-statuses: "✅ Done, ❌ Won't Do, 🖨️ Duplicate, ❓ Cannot Reproduce, ⛔️ Invalid"


.github/workflows/sprint-transfer.yml
```diff
@@ -0,0 +1,27 @@
...
+    name: Move issues to next sprint iteration
+    runs-on: ubuntu-latest
+    steps:
+      - name: Move incomplete issues to next iteration
+        # Pinned to v0.6.1 (https://github.com/blombard/move-to-next-iteration)
+        uses: blombard/move-to-next-iteration@e0d7571f7fa517a8da883e25b797b9d588e16538
+        with:
+          owner: aranya-project
+          number: 1
+          token: ${{ secrets.PROJECT_PAT }}
```
gknopf-aranya, line 34: @djohnson-spideroak we would need devops to create this project pat before the workflow would be able to work.

.github/workflows/sprint-transfer.yml
```diff
@@ -0,0 +1,27 @@
...
+          iteration-field: Sprint
+          iteration: last
+          new-iteration: current
+          excluded-statuses: "Done,Won't Fix"
```
gknopf-aranya: Double-check whether these are the only statuses to exclude

.github/workflows/sprint-transfer.yml
```diff
@@ -0,0 +1,27 @@
...
+  contents: read
+
+jobs:
+  move-issues:
+    name: Move issues to next sprint iteration
+    runs-on: ubuntu-latest
+    steps:
+      - name: Move incomplete issues to next iteration
+        # Pinned to v0.6.1 (https://github.com/blombard/move-to-next-iteration)
+        uses: blombard/move-to-next-iteration@e0d7571f7fa517a8da883e25b797b9d588e16538
```
gknopf-aranya, line 24: We may need to create our own fork of this if we're not allowed to use external actions.

.github/workflows/sprint-transfer.yml
```diff
@@ -0,0 +1,27 @@
+name: "Transfer issues to next sprint"
+
+on:
+  # Run at midnight PST (08:00 UTC) every Wednesday
```
gknopf-aranya: Is PST the correct timezone? Sometimes I forget what timezone everyone is in.

.github/workflows/sprint-transfer.yml
gknopf-aranya: Also "Duplicate", "Cannot Reproduce", "Invalid"

.github/workflows/sprint-transfer.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Move incomplete issues to next iteration
+        # Pinned to v0.6.1 (https://github.com/blombard/move-to-next-iteration)
+        uses: blombard/move-to-next-iteration@e0d7571f7fa517a8da883e25b797b9d588e16538
+        with:
+          owner: aranya-project
+          number: 1
+          token: ${{ secrets.PROJECT_PAT }}
+          iteration-field: Sprint
+          iteration: last
```
gknopf-aranya, line 36: I believe the previous iteration is referenced with @previous not @last.

.github/workflows/sprint-transfer.yml
gknopf-aranya: Fixed. Updated to midnight CST (06:00 UTC) since HQ is in Kansas City.

.github/workflows/sprint-transfer.yml
gknopf-aranya: Added Duplicate, Cannot Reproduce, and Invalid to the excluded statuses.

.github/workflows/sprint-transfer.yml
gknopf-aranya: Verified against the action docs — `last` is the correct value. The valid options are `last` or `current` for the `iteration` field.

.github/workflows/sprint-transfer.yml
gknopf-aranya: Updated the TODO with minimum PAT permissions (project read/write). Also added a `sprint-management` GitHub Environment and branch filter so the secret is only accessible from main. DevOps still needs to create the environment and PAT.

.github/workflows/sprint-transfer.yml
gknopf-aranya: Added a todo for this.

.github/workflows/sprint-transfer.yml
```diff
@@ -0,0 +1,38 @@
+name: "Transfer issues to next sprint"
+
+on:
+  # Run at midnight CST (06:00 UTC) every Wednesday
+  schedule:
+    - cron: "0 6 * * 3"
```
chip-so: Looks like github supports the cron syntax where you can use the three-letter abbreviation for the day of week. Might be a little clearer to have `wed` instead of `3`.

Also shouldn't this run every other week? I don't think there's a way to do that with cron syntax, so you'll probably have to use some other gate on the `step` below to have it only run on sprint planning weeks.

gknopf-aranya: It's okay to run this job every week. The current sprint isn't updated by GitHub until Wednesday morning every other week. Running this again the next week between sprints would attempt to move issues to the next sprint again (but there shouldn't be anything to move if everything was already moved).

gknopf-aranya: `wed` would be more clear, I'll use that

.github/workflows/sprint-transfer.yml
```diff
@@ -0,0 +1,38 @@
...
+          #   1. Create a GitHub Environment named "sprint-management" with a branch
+          #      protection rule limiting deployment to the main branch
+          #   2. Create a PAT with the following minimum permissions:
+          #      - project (read/write) - required to read project items and update iteration fields
+          #   3. Store the PAT as an environment secret named PROJECT_PAT
+          token: ${{ secrets.PROJECT_PAT }}
+          iteration-field: Sprint
+          iteration: last
+          new-iteration: current
+          excluded-statuses: "Done,Won't Fix,Duplicate,Cannot Reproduce,Invalid"
```
jdygert-spok: These statuses don't match ours.

.github/workflows/sprint-transfer.yml
chip-so: Ah, I see. So the action is fundamentally gated by the sprint period we set up. Sounds good, then.

gknopf-aranya: Yep. Limiting this to only run once every 2 weeks would be an optimization that doesn't affect the behavior. The only behavior change I can think of is if someone adds/moves an issue into the old sprint, then it could get moved to the new sprint later. This is probably a good thing, since putting an issue in an old sprint was most likely unintentional.

.github/workflows/sprint-transfer.yml
gknopf-aranya: Added the emojis

gknopf-aranya:
Then again, we don't want arbitrary PRs to have the ability to write to our project management board. So isolating this workflow could be beneficial. I'd rather not move the PAT to an environment where manual approval is required. As long as the PAT is only accessible from the protected `main` branch, that seems secure enough.

.github/workflows/sprint-transfer.yml
```diff
@@ -35,4 +35,9 @@ jobs:
           iteration-field: Sprint
           iteration: last
           new-iteration: current
-          excluded-statuses: "Done,Won't Fix,Duplicate,Cannot Reproduce,Invalid"
+          excluded-statuses: >-
+            ✅ Done,
+            ❌ Won't Do,
+            🖨️ Duplicate,
+            ❓ Cannot Reproduce,
+            ⛔️ Invalid
```
chip-so: I'm not sure this will do what you're expecting. YAML `>-` preserves internal newlines and the string is [parsed naively by splitting on ","](https://github.com/blombard/move-to-next-iteration/blob/master/index.js#L14). So I think this will only match the first item in the list.

chip-so: I was thinking maybe it was worth forking the action to allow this to accept an actual YAML list, but it turns out this is a [limitation of GitHub's `@action/core`](https://github.com/actions/toolkit/tree/main/packages/core#inputsoutputs). It can only accept strings and bools.

gknopf-aranya: If it's too much of a pain to get this to work with emojis I suppose we could remove them from the project management board.

chip-so: I don't think the emojis matter one way or another, this is just a YAML line folding whitespace issue. The emojis should work as long as it's all on one line.

PR #726, opened 2026-02-18

jdygert-spok:


gknopf-aranya (dismissed):
lgtm, needs a few more approvals cc: @djohnson-spideroak, @chip-so, @benz-spideroak, @ssweetney-spok, @nikki-aranya

PR #731, opened 2026-02-19
Add Correlation ID for tracing

jnewell-spok:
This is laying some groundwork for the observability work.

The whole purpose is to be able to trace an operation from client -> daemon -> client using a given ID.

The addition of this tracing_id tied to RPC calls, or generated within the application, will make this much easier in debugging

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -1522,7 +1522,7 @@ pub unsafe fn query_role_perms(
 ) -> Result<(), imp::Error> {
     let perms = client
         .rt
-        .block_on(client.inner.team(team.into()).query_role_perms(role.into()))?;
+        .block_on(client.inner.team(team.into()).role_perm(role.into()))?;
 
     if *perms_len < perms.len() {
         *perms_len = perms.len();
@@ -1707,7 +1707,7 @@ pub fn query_rank(
         client
             .inner
             .team(team.into())
-            .query_rank(aranya_client::ObjectId::from(object_id)),
+            .rank(aranya_client::ObjectId::from(object_id)),
     )?;
     Ok(Rank(rank.value()))
 }
--- a/crates/aranya-client/src/client/team.rs
+++ b/crates/aranya-client/src/client/team.rs
@@ -326,7 +326,7 @@ impl Team<'_> {
 
     /// Queries all permissions assigned to a role.
     #[instrument(skip(self))]
-    pub async fn query_role_perms(&self, role_id: RoleId) -> Result<Vec<Permission>> {
+    pub async fn role_perm(&self, role_id: RoleId) -> Result<Vec<Permission>> {
         let perms = self
             .client
             .daemon
@@ -349,6 +349,7 @@ impl Team<'_> {
     /// the role. To effectively change a role's rank, create a new role with
     /// matching permissions at the desired rank, assign the new role to the
     /// devices that had the old role, then delete the old role.
+    #[allow(private_bounds)]
     #[instrument(skip(self))]
     pub async fn change_rank(
         &self,
@@ -371,8 +372,9 @@ impl Team<'_> {
     }
 
     /// Queries the rank of an object.
+    #[allow(private_bounds)]
     #[instrument(skip(self))]
-    pub async fn query_rank(&self, object_id: impl ToObjectId) -> Result<Rank> {
+    pub async fn rank(&self, object_id: impl ToObjectId) -> Result<Rank> {
         self.client
             .daemon
             .query_rank(rpc_context(), self.id, object_id.to_object_id().into_api())
--- a/crates/aranya-client/src/tests/afc.rs
+++ b/crates/aranya-client/src/tests/afc.rs
@@ -33,7 +33,7 @@ async fn test_afc_create_assign_revoke_delete_label() -> Result<()> {
     memberb_team.sync_now(owner_addr, None).await?;
     assert_eq!(memberb_team.labels().await?.iter().count(), 0);
 
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -153,7 +153,7 @@ async fn test_afc_uni_chan_create() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -220,7 +220,7 @@ async fn test_afc_uni_send_chan_seal_open() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -297,7 +297,7 @@ async fn test_afc_uni_chan_delete() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -382,7 +382,7 @@ async fn test_afc_uni_chan_revoke_label() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -486,7 +486,7 @@ async fn test_afc_uni_chan_delete_label() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -596,7 +596,7 @@ async fn test_afc_uni_chan_remove_devices() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -715,7 +715,7 @@ async fn test_afc_uni_chan_revoke_role() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -832,7 +832,7 @@ async fn test_afc_uni_chan_change_role_without_perm() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -951,7 +951,7 @@ async fn test_afc_uni_multi_send_chans() -> Result<()> {
         .await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(default_roles.member().id).await?;
+    let member_role_rank = owner_team.rank(default_roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
 
--- a/crates/aranya-client/src/tests/client.rs
+++ b/crates/aranya-client/src/tests/client.rs
@@ -86,7 +86,7 @@ async fn test_sync_now() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.admin().id)
+        .rank(roles.admin().id)
         .await?;
     owner
         .add_device(
@@ -102,7 +102,7 @@ async fn test_sync_now() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.operator().id)
+        .rank(roles.operator().id)
         .await?;
     owner
         .add_device(
@@ -136,7 +136,7 @@ async fn test_sync_now() -> Result<()> {
         .context("admin unable to sync with owner")?;
 
     // Now we should be able to successfully create a label (admin has CreateLabel perm).
-    let member_role_rank = admin.query_rank(roles.member().id).await?;
+    let member_role_rank = admin.rank(roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     admin
@@ -170,7 +170,7 @@ async fn test_add_remove_sync_peers() -> Result<()> {
 
     // Add a command to graph by having the owner create a label.
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(roles.member().id).await?;
+    let member_role_rank = owner_team.rank(roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     owner_team.create_label(text!("label1"), label_rank).await?;
@@ -261,7 +261,7 @@ async fn test_role_create_assign_revoke() -> Result<()> {
         .find(|r| r.name == "admin")
         .ok_or_else(|| anyhow::anyhow!("no admin role"))?;
 
-    let member_role_rank = owner_team.query_rank(roles.member().id).await?;
+    let member_role_rank = owner_team.rank(roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
 
@@ -277,7 +277,7 @@ async fn test_role_create_assign_revoke() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.admin().id)
+        .rank(roles.admin().id)
         .await?;
     owner
         .add_device(
@@ -437,7 +437,7 @@ async fn test_add_devices() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.admin().id)
+        .rank(roles.admin().id)
         .await?;
     owner
         .add_device(
@@ -458,7 +458,7 @@ async fn test_add_devices() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.operator().id)
+        .rank(roles.operator().id)
         .await?;
     admin
         .add_device(
@@ -489,7 +489,7 @@ async fn test_add_devices() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.member().id)
+        .rank(roles.member().id)
         .await?;
     for (name, kb, device_id) in [
         ("membera", team.membera.pk.clone(), team.membera.id),
@@ -548,13 +548,13 @@ async fn test_add_device_with_initial_role_requires_outranking() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.admin().id)
+        .rank(roles.admin().id)
         .await?;
     let member_role_rank = devices
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.member().id)
+        .rank(roles.member().id)
         .await?;
     owner_team
         .add_device(
@@ -668,7 +668,7 @@ async fn test_query_functions() -> Result<()> {
     devices.add_all_device_roles(team_id, &roles).await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(roles.member().id).await?;
+    let member_role_rank = owner_team.rank(roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let label_id = owner_team.create_label(text!("label1"), label_rank).await?;
@@ -779,7 +779,7 @@ async fn test_add_team() -> Result<()> {
     info!(?team_id);
 
     let roles = devices.setup_default_roles(team_id).await?;
-    let member_role_rank = owner.query_rank(roles.member().id).await?;
+    let member_role_rank = owner.rank(roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
 
@@ -789,7 +789,7 @@ async fn test_add_team() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.admin().id)
+        .rank(roles.admin().id)
         .await?;
     owner
         .add_device(
@@ -805,7 +805,7 @@ async fn test_add_team() -> Result<()> {
         .owner
         .client
         .team(team_id)
-        .query_rank(roles.operator().id)
+        .rank(roles.operator().id)
         .await?;
     owner
         .add_device(
@@ -886,7 +886,7 @@ async fn test_remove_team() -> Result<()> {
 
     let owner = devices.owner.client.team(team_id);
     let roles = devices.setup_default_roles(team_id).await?;
-    let member_role_rank = owner.query_rank(roles.member().id).await?;
+    let member_role_rank = owner.rank(roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
 
@@ -899,7 +899,7 @@ async fn test_remove_team() -> Result<()> {
             .owner
             .client
             .team(team_id)
-            .query_rank(roles.operator().id)
+            .rank(roles.operator().id)
             .await?;
         owner
             .add_device(
@@ -914,7 +914,7 @@ async fn test_remove_team() -> Result<()> {
             .owner
             .client
             .team(team_id)
-            .query_rank(roles.admin().id)
+            .rank(roles.admin().id)
             .await?;
         owner
             .add_device(
@@ -1004,13 +1004,13 @@ async fn test_multi_team_sync() -> Result<()> {
         .owner
         .client
         .team(team_id1)
-        .query_rank(roles1.admin().id)
+        .rank(roles1.admin().id)
         .await?;
     let operator_role_rank_t1 = devices
         .owner
         .client
         .team(team_id1)
-        .query_rank(roles1.operator().id)
+        .rank(roles1.operator().id)
         .await?;
     team1
         .add_device(
@@ -1042,13 +1042,13 @@ async fn test_multi_team_sync() -> Result<()> {
         .owner
         .client
         .team(team_id2)
-        .query_rank(roles2.admin().id)
+        .rank(roles2.admin().id)
         .await?;
     let operator_role_rank_t2 = devices
         .owner
         .client
         .team(team_id2)
-        .query_rank(roles2.operator().id)
+        .rank(roles2.operator().id)
         .await?;
     team2
         .add_device(
@@ -1115,7 +1115,7 @@ async fn test_multi_team_sync() -> Result<()> {
     admin1.sync_now(owner_addr, None).await?;
 
     // Now we should be able to successfully create a label (admin has CreateLabel perm).
-    let member_role_rank_t1 = admin1.query_rank(roles1.member().id).await?;
+    let member_role_rank_t1 = admin1.rank(roles1.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank_t1 = Rank::new(member_role_rank_t1.value().saturating_sub(1));
     admin1
@@ -1164,7 +1164,7 @@ async fn test_multi_team_sync() -> Result<()> {
     admin2.sync_now(owner_addr, None).await?;
 
     // Now we should be able to successfully create a label (admin has CreateLabel perm).
-    let member_role_rank_t2 = admin2.query_rank(roles2.member().id).await?;
+    let member_role_rank_t2 = admin2.rank(roles2.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank_t2 = Rank::new(member_role_rank_t2.value().saturating_sub(1));
     admin2
@@ -1229,7 +1229,7 @@ async fn test_hello_subscription() -> Result<()> {
     // Admin performs an action that will update their graph - create a label
     // (admin has permission to create labels)
     info!("admin creating a test label");
-    let member_role_rank = admin_team.query_rank(roles.member().id).await?;
+    let member_role_rank = admin_team.rank(roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let test_label = admin_team
@@ -1384,7 +1384,7 @@ async fn test_hello_subscription_schedule_delay() -> Result<()> {
 
     // Admin creates first label
     info!("admin creating first test label");
-    let member_role_rank = admin_team.query_rank(roles.member().id).await?;
+    let member_role_rank = admin_team.rank(roles.member().id).await?;
     // Label rank must be lower than the member role rank so all team members can operate on it.
     let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let test_label_1 = admin_team
@@ -1570,7 +1570,7 @@ async fn test_create_role() -> Result<()> {
     let roles = devices.setup_default_roles(team_id).await?;
 
     let owner_team = devices.owner.client.team(team_id);
-    let member_role_rank = owner_team.query_rank(roles.member().id).await?;
+    let member_role_rank = owner_team.rank(roles.member().id).await?;
 
     let expected_rank = Rank::new(member_role_rank.value().saturating_sub(1));
     let test_role = owner_team
@@ -1579,7 +1579,7 @@ async fn test_create_role() -> Result<()> {
         .expect("expected to create role");
 
     // Verify the rank can be queried back.
-    let rank = owner_team.query_rank(test_role.id).await?;
+    let rank = owner_team.rank(test_role.id).await?;
     assert_eq!(rank, expected_rank);
 
     owner_team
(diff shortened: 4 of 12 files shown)


Cargo.toml
```diff
@@ -80,6 +80,7 @@ metrics = { version = "0.24" }
 metrics-util = { version = "0.20" }
 postcard = { version = "1", default-features = false, features = ["use-std", "heapless", "experimental-derive"] }
 pretty_assertions = { version = "1.4" }
+rand = { version = "0.8" }
```
gknopf-aranya, line 83: Note that this cannot be used to generate random numbers for anything cryptographic

crates/aranya-client/src/trace.rs
gknopf-aranya, line 1: It appears a `trace.rs` module has been added to both the client and daemon crates. If there is significant overlap between these implementations, this seems like something to include in the `aranya-utils` crate which is shared by the other crates.

crates/aranya-client/src/util.rs
```diff
@@ -188,3 +199,42 @@ macro_rules! impl_vec_into_iter_wrapper {
...
+/// - Returns context for use in RPC calls
+pub(crate) fn rpc_context() -> context::Context {
+    let mut ctx = context::current();
+    ctx.deadline = Instant::now()
+        .checked_add(IPC_TIMEOUT)
+        .expect("IPC_TIMEOUT should not overflow");
+    let mut rng = rand::thread_rng();
+    ctx.trace_context = TraceContext {
+        trace_id: TarpcTraceId::random(&mut rng),
+        span_id: SpanId::random(&mut rng),
```
gknopf-aranya: If the trace and span id are both randomly generated at the same time, why are both needed? I'd understand the need for both if they applied to different spans/contexts but it appears they have the same lifetime.

crates/aranya-client/src/trace.rs
```diff
@@ -0,0 +1,114 @@
...
+///
+/// ```rust,ignore
+/// use aranya_client::trace::generate_trace_id;
+/// use tracing::info;
+///
+/// let trace_id = generate_trace_id();
+/// info!(%trace_id, "starting operation");
+/// // Use trace_id for RPC calls...
+/// ```
+pub fn generate_trace_id() -> TraceId {
```
gknopf-aranya: This appears to be dead code that is only used in tests. The IPC context is generating the trace ID differently.

crates/aranya-daemon/src/api.rs
```diff
@@ -416,6 +418,16 @@ impl Api {
         }
         Ok(GraphId::transmute(team))
     }
+
+    /// Extracts the trace ID from a tarpc context and logs a receive event.
+    ///
+    /// This should be called at the start of each API handler to ensure trace ID
+    /// is available throughout the request's execution.
+    fn setup_trace_context(&self, ctx: &context::Context, method: &'static str) -> TraceId {
```
gknopf-aranya: I think what we'll want to do is go from aranya-client span -> IPC trace ID -> aranya-daemon span. The trace ID is just used to associate a client span with a daemon span. Everything else carries the trace ID automatically inside a span per API call rather than manually passing the trace ID in the daemon.

crates/aranya-daemon/src/trace.rs
```diff
@@ -0,0 +1,127 @@
...
+
+impl From<String> for TraceId {
+    fn from(s: String) -> Self {
+        Self::new(s)
+    }
+}
+
+tokio::task_local! {
+    #[doc(hidden)]
+    pub static TRACE_CONTEXT: Option<TraceId>;
```
gknopf-aranya: Let's track the trace ID per span created in each API call rather than per task.

crates/aranya-daemon/src/trace.rs
```diff
@@ -0,0 +1,127 @@
...
+///
+/// ```rust,ignore
+/// use tarpc::context;
+/// use aranya_daemon::trace::extract_trace_id;
+///
+/// let ctx = context::current();
+/// let trace_id = extract_trace_id(&ctx).to_string();
+/// ```
+pub fn extract_trace_id(ctx: &tarpc::context::Context) -> TraceId {
+    // Try to extract trace ID from context metadata if available
```
gknopf-aranya: It doesn't look like the trace ID is being extracted from the context here

crates/aranya-daemon/src/trace.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_trace_id_creation() {
+        let trace_id = TraceId::new("test-123");
```
gknopf-aranya: Is the trace ID randomly generated or set manually as a string?

gknopf-aranya: A randomly generated trace ID guarantees uniqueness which is a nice property to have if we want to trace each IPC call. But, knowing exactly which API method the request originated in is also good information to have. Both the client and daemon already know the name of the API method so that doesn't need to be sent via the IPC context. So I think the only thing that would need to be sent via the IPC context is a randomly generated trace ID. If it's randomly generated, it could be an int or AranyaId type rather than a string.

crates/aranya-client/src/util.rs
```diff
@@ -188,3 +199,42 @@ macro_rules! impl_vec_into_iter_wrapper {
...
+        sampling_decision: SamplingDecision::Sampled,
```
gknopf-aranya: What is sampling decision for?

crates/aranya-client/tests/trace_tests.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    use crate::common::DeviceCtx;
+
+    #[test_log::test(tokio::test)]
+    async fn test_trace_id_in_logs() {
+        let trace_id = generate_trace_id();
+        info!(%trace_id, "starting test operation");
+
+        debug!(%trace_id, "performing first RPC call");
+        tokio::time::sleep(tokio::time::Duration::from_millis(5)).await;
```
gknopf-aranya: The sleeps in this test make it take longer without any benefit.

crates/aranya-client/tests/trace_tests.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        debug!(%trace_id, "performing second RPC call");
+        tokio::time::sleep(tokio::time::Duration::from_millis(5)).await;
+
+        info!(%trace_id, "test operation completed");
+    }
+
+    #[tokio::test]
+    async fn test_create_team_rpc_correlation() -> anyhow::Result<()> {
```
gknopf-aranya: What is this supposed to be testing? It appears that a trace ID is generated and used for tracing alongside some Aranya team operations. But, there isn't any demonstration of integrated tracing in the aranya client or daemon here. The tracing part of the test could be simplified to:
```
        let create_team_trace_id = generate_trace_id();
        info!(%create_team_trace_id, "starting create_team operation");
```
But all that proves is that a trace ID can be generated and logged. It would be more interesting to prove that the aranya client and daemon both use the same trace ID for an IPC call.

crates/aranya-client/src/util.rs
jnewell-spok: not entirely sure. I just included it as it is part of the API.

crates/aranya-client/src/util.rs
jnewell-spok: it's part of the api. wasn't planning on using it for anything.

crates/aranya-client/src/trace.rs
jnewell-spok: good point

crates/aranya-client/tests/trace_tests.rs
jnewell-spok: removed

crates/aranya-client/src/util.rs
gknopf-aranya: If it's not used, then generating a random number each time wastes cycles. Do you need to set it or can you let it set itself to a default value?

crates/aranya-client/tests/trace_tests.rs
jnewell-spok: I'm not sure this test even needs to exist. only way that we can check that it actually did what it was supposed to is by looking at the logs.

gknopf-aranya: Is there a way to parse the logs in a test without manually looking at them?

crates/aranya-daemon/src/trace.rs
gknopf-aranya: Actually, it looks like tarpc already has definitions for the trace and span ids so we could use those without defining a new type:
https://docs.rs/tarpc-trace/latest/tarpc_trace/struct.Context.html#structfield.span_id

crates/aranya-client/src/util.rs
gknopf-aranya: Looks like you can use `TraceContext::default()` first then manually set the desired fields or:
```
TraceContext {
    <set fields>,
    ...
}
```
Where `...` sets the defaults on fields not defined.

crates/aranya-client/tests/trace_tests.rs
gknopf-aranya: Looks like there's a lib to help with testing actual trace output:
https://docs.rs/tracing-test/latest/tracing_test/
If you decide to use it, please include it has a dev/test dep only, not a production dep.

gknopf-aranya:
It looks like this library is designed to support tracing with tarpc:
https://docs.rs/tarpc-trace/latest/tarpc_trace/index.html

crates/aranya-daemon/src/trace.rs
jnewell-spok: fixed.

crates/aranya-client/src/util.rs
jnewell-spok: modified to just use defaults for the other values that we aren't using

crates/aranya-client/src/util.rs
jnewell-spok: fixed to use default value

crates/aranya-client/tests/trace_tests.rs
jnewell-spok: this is very helpful. It is allowing me to pull the trace_ids from the logs and compare them with an assertion. put a note in that it is only for testing purposes.

gknopf-aranya: In some `Cargo.toml` files we're able to include deps in the `[dev-dependencies]` section so they aren't imported into production code.

crates/aranya-daemon/src/trace.rs
jnewell-spok: cleaned up to use the tarpc traceid definition.

crates/aranya-daemon/src/trace.rs
jnewell-spok: modified to be per api call

crates/aranya-daemon/src/api.rs
jnewell-spok: this makes sense, removed the manual passing of the trace id.

crates/aranya-client/src/trace.rs
jnewell-spok: you are right, this function is not needed, I was initially thinking there would be an applcation created trace_id, but working through it I found it to not be needed.

crates/aranya-client/src/trace.rs
jnewell-spok: I found that the trace module wasn't needed for the client, since the id generation is done in IPC context. The daemon still has it for the id extraction

crates/aranya-client/src/util.rs
```diff
@@ -188,3 +199,25 @@ macro_rules! impl_vec_into_iter_wrapper {
...
+    let mut ctx = context::current();
+    ctx.deadline = Instant::now()
+        .checked_add(IPC_TIMEOUT)
+        .expect("IPC_TIMEOUT should not overflow");
+    let mut rng = rand::thread_rng();
+    ctx.trace_context = TraceContext {
+        trace_id: TarpcTraceId::random(&mut rng),
+        ..TraceContext::default()
+    };
+    tracing::trace!(rpc.trace_id = %ctx.trace_context.trace_id, "RPC: SendRequest");
```
gknopf-aranya: Is this the only place the client logs the trace ID before invoking the RPC? If the goal is to associate client and daemon logs via the trace ID, I think we're going to want to ensure they both at least have one log with the trace ID with `info!` instead of `trace!`. From there, the verbosity can be configured by setting different trace levels. This depends on which tracing levels we want to be able to correlate the tracing ID for. My assumption was that we'd want to correlate the trace ID for all log levels, but this implementation appears to only be able to correlate it for `TRACE`.

crates/aranya-daemon/src/api.rs
```diff
@@ -416,85 +417,107 @@ impl Api {
...
+        let trace_id = trace::extract_trace_id(ctx);
+        trace::record_current_span_trace_id(trace_id);
+        trace!(rpc.method = method, rpc.trace_id = %trace_id, "RPC: ReceiveRequest");
```
gknopf-aranya: Probably should be `info!` so we can correlate the trace ID for more log levels

crates/aranya-daemon/src/api.rs
```diff
@@ -416,85 +417,107 @@ impl Api {
...
 impl DaemonApi for Api {
     //
     // Misc
     //
 
-    #[instrument(skip(self), err)]
-    async fn version(self, context: context::Context) -> api::Result<api::Version> {
+    #[instrument(skip(self), err, fields(trace_id = tracing::field::Empty))]
+    async fn version(self, ctx: context::Context) -> api::Result<api::Version> {
+        self.setup_trace_context(&ctx, "DaemonApi.version");
```
gknopf-aranya: `instrument` should already include the method name in any tracing within it

crates/aranya-daemon/src/api.rs
```diff
@@ -416,85 +417,107 @@ impl Api {
...
 }
 
 impl DaemonApi for Api {
     //
     // Misc
     //
 
-    #[instrument(skip(self), err)]
-    async fn version(self, context: context::Context) -> api::Result<api::Version> {
+    #[instrument(skip(self), err, fields(trace_id = tracing::field::Empty))]
```
gknopf-aranya: Why ignore the trace_id in `instrument`? That seems like a valid way of passing the trace ID into the tracing logs.

crates/aranya-daemon/src/api.rs
```diff
@@ -663,16 +692,18 @@ impl DaemonApi for Api {
             .await
             .context("unable to add device to team")?;
         self.effect_handler.handle_effects(graph, &effects).await?;
+        trace!(?graph, "added device to team");
```
gknopf-aranya, line 704: Why add this trace at the end of `add_device_to_team()` but not the other methods in the API?

crates/aranya-daemon/src/api.rs
```diff
@@ -416,85 +417,107 @@ impl Api {
         }
         Ok(GraphId::transmute(team))
     }
+
+    /// Extracts the trace ID from a tarpc context and logs a receive event.
+    ///
+    /// This should be called at the start of each API handler to ensure trace ID
+    /// is available throughout the request's execution.
+    fn setup_trace_context(&self, ctx: &context::Context, method: &'static str) -> TraceId {
```
gknopf-aranya: The returned `TraceId` doesn't appear to be used anywhere:
```
self.setup_trace_context(&ctx, "DaemonApi.version");
```

gknopf-aranya: `&self` doesn't appear to be used anywhere in this method which implies this doesn't need to be a method on `&self` in `api.rs` it could be a method implemented in `trace.rs`.

crates/aranya-daemon/src/trace.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub use tarpc::trace::TraceId;
+use tracing::Span;
+
+/// Extracts the trace ID from a tarpc context.
+///
+/// The trace ID is generated by the client and included in the request.
+/// If no trace ID is found in the context, a new one is generated.
+pub fn extract_trace_id(ctx: &tarpc::context::Context) -> TraceId {
+    ctx.trace_context.trace_id
```
gknopf-aranya: Implementing a method to extract the value of a field is overkill

crates/aranya-daemon/src/trace.rs
```diff
@@ -0,0 +1,43 @@
...
+pub fn record_current_span_trace_id(trace_id: TraceId) {
+    record_trace_id(&Span::current(), trace_id);
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_trace_id_creation() {
```
gknopf-aranya: These tests are covering an edge-case where the trace ID is empty. The normal case is that the trace ID would be set by the client and logged in a span.

crates/aranya-daemon/src/trace.rs
```diff
@@ -0,0 +1,43 @@
...
+/// Extracts the trace ID from a tarpc context.
+///
+/// The trace ID is generated by the client and included in the request.
+/// If no trace ID is found in the context, a new one is generated.
+pub fn extract_trace_id(ctx: &tarpc::context::Context) -> TraceId {
+    ctx.trace_context.trace_id
+}
+
+/// Records the trace ID on the provided span.
+pub fn record_trace_id(span: &Span, trace_id: TraceId) {
```
gknopf-aranya: `record_trace_id` is only called by `record_current_span_trace_id`. Doesn't seem like it needs to be its own function:
```
/// Records the trace ID on the current span.
pub fn record_current_span_trace_id(trace_id: TraceId) {
     &Span::current().record("trace_id", tracing::field::display(trace_id));
}
```

gknopf-aranya: A better place to test trace ID correlation would be in the client. Then you can compare trace IDs after a round-trip RPC call initiated by the client -> tracing in the daemon -> then returning to the client.

crates/aranya-daemon/src/api.rs
jnewell-spok: yes, I agree. changed

crates/aranya-daemon/src/trace.rs
jnewell-spok: definitely. it also is only used in one place. Modified to just put the extraction inline and remove the function

crates/aranya-client/tests/tests.rs
```diff
@@ -61,8 +132,77 @@ async fn test_client_rand() -> Result<()> {
...
+        let mut buf = [0u8; SEED_IKM_SIZE];
+        owner.client.rand(&mut buf).await;
+        buf
+    };
+    let qs_cfg = CreateTeamQuicSyncConfig::builder()
+        .seed_ikm(seed_ikm)
+        .build()?;
+    let owner_cfg = CreateTeamConfig::builder().quic_sync(qs_cfg).build()?;
+
+    let team = owner.client.create_team(owner_cfg).await?;
```
gknopf-aranya: Setting up a team in order to test trace ID correlation may be more complexity than what's needed. A simpler API to test with could be `version()`.

crates/aranya-client/tests/tests.rs
```diff
@@ -61,8 +132,77 @@ async fn test_client_rand() -> Result<()> {
...
+        }
+    }
+    match (daemon_trace_id, client_trace_id) {
+        (Some(daemon_trace_id), Some(client_trace_id)) => {
+            eprintln!(
+                "trace id capture summary: daemon_trace_id={}, client_trace_id={}",
+                daemon_trace_id, client_trace_id
+            );
+            assert_eq!(
+                client_trace_id, daemon_trace_id,
```
gknopf-aranya: These trace IDs could also be equal if neither one is set since they are both initialized as `None`. I think this test should fail if either the client or daemon trace ID isn't set.

crates/aranya-daemon/src/trace.rs
jnewell-spok: it is being tested within the client: test_basic_rpc_operations

jnewell-spok: cleaned up record_trace_id

crates/aranya-client/tests/tests.rs
```diff
@@ -61,8 +132,77 @@ async fn test_client_rand() -> Result<()> {
...
+                "client and daemon trace ids should match for create_team"
+            );
+        }
+        (daemon_trace_id, client_trace_id) => {
+            eprintln!(
```
gknopf-aranya: Don't see a reason to print this information out unless the test fails.

crates/aranya-client/tests/tests.rs
```diff
@@ -12,22 +12,93 @@
...
-use aranya_daemon_api::text;
-use test_log::test;
+use aranya_daemon_api::{text, SEED_IKM_SIZE};
+use serial_test::serial;
 use tracing::{debug, info};
+use tracing_subscriber::fmt::MakeWriter;
 
 use crate::common::{sleep, DeviceCtx, DevicesCtx, SLEEP_INTERVAL};
 
+fn parse_trace_id(line: &str) -> Option<String> {
```
gknopf-aranya: The parsing of trace IDs seems too complex. Would it help to configure JSON format then parse the JSON instead?

crates/aranya-client/tests/tests.rs
jnewell-spok: good idea. didn't think about doing that. switched it up

gknopf-aranya: That's perfect for this isolated tracing test. Had another idea to verify the trace IDs at the end of existing unit tests like create_team so there's less boilerplate code we need to add. But, I don't think we should attempt that until the trace ID parsing is simplified enough to do something like:
```
let client_trace_id = parse_trace_id(client_log);
let daemon_trace_id = parse_trace_id(daemon_log);
assert_eq!(client_trace_id, daemon_trace_id);
```

crates/aranya-client/tests/tests.rs
```diff
@@ -21,13 +26,79 @@ use aranya_client::{
...
+    let raw = rest[..end].trim_matches(|c| c == '"' || c == ',' || c == '}' || c == ':');
+    if raw.is_empty() {
+        None
+    } else {
+        Some(raw.to_string())
+    }
+}
+
+#[derive(Clone)]
+struct SharedWriter {
```
gknopf-aranya: A lot of this boilerplate trace test code probably belongs in `mod.rs`

crates/aranya-client/tests/tests.rs
jnewell-spok: great idea. I was intending on using json logs for observability anyway.

jnewell-spok: modified to use serde_json

crates/aranya-client/tests/tests.rs
jnewell-spok: moved some of this code and also pulled out some parsing into a function to use from mod.rs

crates/aranya-client/tests/tests.rs
jnewell-spok: created a case for the match when both ids are None.

crates/aranya-client/tests/tests.rs
jnewell-spok: after the matching Some/Some case and the None/None case, this is a failure case. I'm modifying it to bail with message instead of printing and returning Ok

crates/aranya-client/tests/tests.rs
gknopf-aranya: Looks much simpler with the json change. Nice work.

crates/aranya-daemon/src/api.rs
jnewell-spok: moved function to trace.rs

jnewell-spok: removed the return

crates/aranya-daemon/src/api.rs
jnewell-spok: added trace to all other methods in api

crates/aranya-daemon/src/trace.rs
jnewell-spok: fixed test

crates/aranya-daemon/src/api.rs
jnewell-spok: not sure what you mean. Isn't this just initially setting trace_id as empty but setting it within the function later?

crates/aranya-client/src/util.rs
jnewell-spok: changed to info!

crates/aranya-daemon/src/api.rs
jnewell-spok: fixed up setup_trace_context to not have the method as an input anymore

crates/aranya-daemon/src/api.rs
gknopf-aranya: Yeah, why unset/set it when it is already set by default?

jnewell-spok: ah, ok. I see what you mean. I'll remove this

crates/aranya-client/src/util.rs
```diff
@@ -188,3 +199,25 @@ macro_rules! impl_vec_into_iter_wrapper {
...
+/// Creates a new tarpc context with trace metadata.
+///
+/// This function generates a fresh context for each RPC call, including
+/// a unique trace ID that will be propagated through the daemon and
+/// correlated in all logs.
+///
+/// The trace ID is generated client-side and logged, enabling end-to-end
+/// request tracking across client and daemon components.
+pub(crate) fn rpc_context() -> context::Context {
+    let mut ctx = context::current();
```
gknopf-aranya, line 212: I believe `context::current()` and `TraceContext::default()` already generate a random trace ID automatically. So this change does not seem to be required. The previous code was already able to configure the IPC timeout. We can remove the `rand` dep that was added.

crates/aranya-client/Cargo.toml
```diff
@@ -57,9 +58,11 @@ aranya-daemon = { path = "../aranya-daemon", features = ["experimental", "previe
 spideroak-base58 = { workspace = true }
 
 backon = { workspace = true }
+serde_json = "1"
```
gknopf-aranya: There is a workspace dep for this:
```suggestion
serde_json = { workspace = true }
```

crates/aranya-client/tests/tests.rs
```diff
@@ -61,8 +64,49 @@ async fn test_client_rand() -> Result<()> {
     Ok(())
 }
 
+// Trace-correlation test for client<->daemon RPC metadata propagation.
+#[tokio::test]
+#[serial]
+async fn test_basic_rpc_operations() -> Result<()> {
```
gknopf-aranya: This test is essentially checking that a random trace ID is generated, passed into the daemon, then logged in the tracing output. It is a very thorough end-to-end test. However, it may be more complex than what we need. It is validating functionality that is already guaranteed by the tarpc and tracing libraries.

I think we can simplify to this:
- In a client unit test, call `context::current()` twice and demonstrate that the generated trace ID is not the same (this proves it is randomly generated by the tarpc library as we expect)
- Introduce a test-only method into the daemon API called `test_trace_id()` that returns the trace ID from the client and logs it to tracing output. Verify the trace ID generated by the client matches the trace ID returned by the daemon. You can manually run this test with output enabled to confirm the trace IDs match in the tracing output.

We can infer from this that:
- The trace ID is always randomly generated
- The trace ID is passed from the client to the daemon via the API as we expect
- Any trace ID field/span sent to the tracing lib will be included in the tracing output

This allows us to remove the following complexity:
- Tracing log captures
- JSON parsing
- SharedWriter
- Introduction of a new `rand` dep

crates/aranya-client/tests/tests.rs
```diff
@@ -61,8 +64,49 @@ async fn test_client_rand() -> Result<()> {
...
+            );
+        }
+    }
+
+    Ok(())
+}
+
 /// Tests sync_now() by showing that an admin cannot perform operations until it syncs with the owner.
-#[test(tokio::test(flavor = "multi_thread"))]
+#[tokio::test(flavor = "multi_thread")]
```
gknopf-aranya: This is dropping `test_log` from the unit tests which is useful for debugging them. Please revert this for all the unit tests.

crates/aranya-daemon/src/trace.rs
```diff
@@ -0,0 +1,29 @@
...
+//! This module provides helpers for propagating RPC trace IDs
+//! throughout the daemon's execution.
+
+pub use tarpc::trace::TraceId;
+use tracing::Span;
+
+/// Extracts and records RPC trace context on the current span.
+///
+/// Also emits a receive log with method and trace ID for client/daemon correlation.
+pub fn setup_trace_context(ctx: &tarpc::context::Context, method: &'static str) {
```
gknopf-aranya: Passing in each function name manually as a string is difficult to maintain. `#[instrument]` includes the function name in the span be default. So `Span::current()` has the function name. Example:
```
let span = Span::current();                                                                                                 
let method = span.metadata().map(|m| m.name()).unwrap_or("unknown");                                                        
span.record("trace_id", tracing::field::display(ctx.trace_context.trace_id));                                               
tracing::info!(rpc.method = method, rpc.trace_id = %ctx.trace_context.trace_id, "RPC: ReceiveRequest");  
```

examples/rust/aranya-example/run.bash
```diff
@@ -31,4 +31,5 @@ daemon="${current_dir}/target/release/aranya-daemon"
 example="${current_dir}/target/release/aranya-example"
 
 echo "Running aranya-example with daemon: ${daemon}"
-"${example}" "${daemon}"
+# Enable logging to see trace_id and other useful output from both daemon and example
+ARANYA_EXAMPLE=info ARANYA_DAEMON=info "${example}" "${daemon}"
```
gknopf-aranya: Missing newline at end of file

crates/aranya-client/tests/afc.rs
```diff
@@ -16,14 +19,24 @@ use {
...
+    let captured = String::from_utf8_lossy(&logs.lock().expect("poisoned")).into_owned();
+    let _ = find_rpc_trace_ids_for_api_name(&captured, "DaemonApi.version");
+
     // create team.
     let team_id = devices.create_and_add_team().await?;
 
     // create default roles
     let default_roles = devices.setup_default_roles(team_id).await?;
 
+    devices
```
gknopf-aranya: Is this a flake fix? Doesn't seem related to the tracing changes.

jnewell-spok: yeah, it was added because it was complaining about the function being unused.

examples/rust/aranya-example/run.bash
jnewell-spok: fixed

crates/aranya-client/Cargo.toml
jnewell-spok: fixed

crates/aranya-daemon/src/trace.rs
jnewell-spok: this is fixed. no longer passing the function name

crates/aranya-client/tests/tests.rs
jnewell-spok: reverted.

crates/aranya-client/src/util.rs
jnewell-spok: I tried removing this rpc_id generation, and it ends up with an empty rpc_id.

gknopf-aranya: Good point. Looks like the default does not generate a random trace ID. I think the idiomatic way to do this would be `TraceId::random()` which still depends on `rand` so we can keep that as a dep.
```
use tarpc::trace::{Context as TraceContext, TraceId};                                                                           
                                               
pub(crate) fn rpc_context() -> context::Context {                                                                               
    let mut ctx = context::current();                                                                                           
    ctx.deadline = Instant::now()                                                                                               
        .checked_add(IPC_TIMEOUT)                                                                                               
        .expect("IPC_TIMEOUT should not overflow");                                                                             
    ctx.trace_context.trace_id = TraceId::random(&mut rand::thread_rng());                                                      
    tracing::info!(rpc.trace_id = %ctx.trace_context.trace_id, "RPC: SendRequest");                                             
    ctx                                                                                                                         
}   
```

jdygert-spok: It looks like tarpc expects to pull information from an active opentelemetry span. We should probably evaluate what opentelemetry provides and what we can do with it for observability.

gknopf-aranya: Oh, that's probably what generates the random trace ID when it's enabled.

crates/aranya-client/tests/tests.rs
jnewell-spok: I added a round trip function that just handles the ids on either end, and compares them without dealing with the logs. Take a look and let me know if you think this will suffice for an end to end test.

crates/aranya-client/src/client.rs
```diff
@@ -233,18 +222,32 @@ impl Client {
...
+            .get_device_id(rpc_context())
             .await
             .map_err(IpcError::new)?
             .map_err(aranya_error)
             .map(DeviceId::from_api)
     }
 
+    /// Returns a pair of trace IDs `(client_sent, daemon_received)` for
+    /// validating RPC trace propagation.
+    pub async fn test_trace_id(&self) -> Result<(String, String)> {
```
gknopf-aranya, line 235: This can go behind a test feature flag

crates/aranya-client/tests/afc.rs
gknopf-aranya: Adding a function to a test to resolve an unused warning seems to be working around the problem rather than solving it. Valid options are:
- Remove the unused function
- Ignore the unused warning if the function is truly used but the compiler can't detect it for some reason

Is there a specific change on this PR that caused the function to no longer be detected as used by the compiler? Opening a new PR just for this change isn't ideal, but it doesn't seem like it belongs on this PR.

crates/aranya-client/tests/tests.rs
```diff
@@ -61,6 +61,21 @@ async fn test_client_rand() -> Result<()> {
...
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_trace_id_round_trip() -> Result<()> {
+    let work_dir = tempfile::tempdir()?;
+    let owner = DeviceCtx::new("trace-roundtrip", "owner", work_dir.path().join("owner")).await?;
+
+    let (client_trace_id, daemon_trace_id) = owner.client.test_trace_id().await?;
+    println!("client trace_id: {client_trace_id}");
+    println!("daemon trace_id: {daemon_trace_id}");
```
```suggestion
    info!("client trace_id: {client_trace_id}");
    info!("daemon trace_id: {daemon_trace_id}");
```

crates/aranya-daemon-api/src/service.rs
```diff
@@ -392,6 +392,10 @@ pub trait DaemonApi {
     async fn get_public_key_bundle() -> Result<PublicKeyBundle>;
     /// Gets the public device id.
     async fn get_device_id() -> Result<DeviceId>;
+    /// Returns the trace ID received in the current RPC context.
+    ///
+    /// Intended for test/debug validation of client<->daemon trace propagation.
+    async fn test_trace_id() -> Result<String>;
```
gknopf-aranya, line 399: This can be put behind a test feature flag

crates/aranya-client/tests/tests.rs
jnewell-spok: fixed

crates/aranya-daemon-api/src/service.rs
jnewell-spok: done

crates/aranya-client/src/client.rs
jnewell-spok: done

crates/aranya-daemon/src/api.rs
```diff
@@ -421,31 +422,48 @@ impl DaemonApi for Api {
...
     }
 
     #[instrument(skip(self), err)]
-    async fn get_device_id(self, _: context::Context) -> api::Result<api::DeviceId> {
+    async fn get_device_id(self, ctx: context::Context) -> api::Result<api::DeviceId> {
+        trace::setup_trace_context(&ctx);
         self.device_id().map(api::DeviceId::transmute)
     }
 
+    #[cfg(feature = "preview")]
```
gknopf-aranya:
```suggestion
    #[cfg(test)]
```

crates/aranya-daemon-api/src/service.rs
```diff
@@ -392,6 +392,11 @@ pub trait DaemonApi {
     async fn get_public_key_bundle() -> Result<PublicKeyBundle>;
     /// Gets the public device id.
     async fn get_device_id() -> Result<DeviceId>;
+    /// Returns the trace ID received in the current RPC context.
+    ///
+    /// Intended for test/debug validation of client<->daemon trace propagation.
+    #[cfg(feature = "preview")]
```
```suggestion
    #[cfg(test)]
```

crates/aranya-client/tests/afc.rs
jnewell-spok: there isn't any specific change I know of that is causing it. I'll remove this addition, to see if I can track down what it is.

crates/aranya-daemon/src/api.rs
jnewell-spok: oops, fixed

crates/aranya-daemon-api/src/service.rs
jnewell-spok: fixed

crates/aranya-daemon/src/api.rs
jdygert-spok: `cfg(test)` only applies to a crate when you're running unit tests on that crate specifically.

gknopf-aranya (approved):

crates/aranya-daemon/Cargo.toml
```diff
@@ -21,6 +21,7 @@ default = [
 
 # Preview features
 preview = ["aranya-daemon-api/preview"]
+
```
djohnson-spideroak: nit: diff noise

crates/aranya-daemon/src/api.rs
```diff
@@ -457,77 +475,86 @@ impl DaemonApi for Api {
     #[instrument(skip(self), err)]
     async fn add_sync_peer(
         self,
-        _: context::Context,
+        ctx: context::Context,
         peer: Addr,
         team: api::TeamId,
         cfg: api::SyncPeerConfig,
     ) -> api::Result<()> {
         let graph = self.check_team_valid(team).await?;
```
djohnson-spideroak, line 484: Why do some functions have `setup_trace_context` and others dont?

jnewell-spok: I fixed this so they all have it.

crates/aranya-daemon/Cargo.toml
jnewell-spok: fixed

chip-so (approved):

crates/aranya-daemon/src/trace.rs
```diff
@@ -0,0 +1,29 @@
...
+use tracing::Span;
+
+/// Extracts and records RPC trace context on the current span.
+///
+/// Also emits a receive log with trace ID for client/daemon correlation.
+pub fn setup_trace_context(ctx: &tarpc::context::Context) {
+    let trace_id = ctx.trace_context.trace_id;
+    Span::current().record("trace_id", tracing::field::display(trace_id));
+    tracing::info!(rpc.trace_id = %trace_id, "RPC: ReceiveRequest");
+}
```
jdygert-spok, line 16: tarpc already does this for each request:
https://github.com/google/tarpc/blob/4d3b75bdca8ed1a2d28d688b8427835317bbbc81/tarpc/src/server.rs#L208-L224

PR #735, opened 2026-02-22
Fix memory exhaustion DoS attack

Sajjon:
> [!NOTE] 
> Co-authored with ChatGTP

# Summary
This PR fixes a memory-exhaustion DoS in QUIC `sync`/`hello` receive paths by enforcing strict maximum inbound message sizes.

# Attack Vector (fixed)
A malicious peer that can establish QUIC sync connectivity (for example, a compromised team member with valid PSK material) could send an oversized stream payload during sync or hello flows.

Before this change, several code paths used unbounded `read_to_end` into a `Vec<u8>`. That allowed attacker-controlled growth of heap memory until severe memory pressure or OOM, causing daemon instability or crash.

## Impacted paths were:
server.rs (incoming sync request)
client.rs (incoming sync response)
hello.rs (hello responses, including async notification flow)

## Root Cause
Unbounded stream reads (read_to_end) were used on attacker-influenced network input without any explicit upper bound.

# Fix
## Added bounded wire-size limits:
`MAX_SYNC_WIRE_MESSAGE_SIZE` in:
- client.rs
- server.rs

`MAX_HELLO_WIRE_MESSAGE_SIZE` in:
- hello.rs

## Replaced unbounded reads with:
`recv.take((MAX as u64) + 1).read_to_end(...)`

## Added explicit oversize handling:
- sync request/response: return error and fail processing
- hello notification task path: warn and drop oversized response
- The +1024 headroom preserves compatibility for small envelope overhead beyond runtime payload limits while still preventing unbounded allocation.

# Security Impact
This closes a high-impact DoS class where a peer could force unbounded memory allocation via oversized QUIC payloads.

# Validation
`cargo test` passes

### DIFF
--- a/crates/aranya-daemon/src/sync/hello.rs
+++ b/crates/aranya-daemon/src/sync/hello.rs
@@ -171,10 +171,20 @@ where
         };
         // Read the response to avoid race condition with server
         let mut response_buf = Vec::new();
-        recv.take((MAX_HELLO_WIRE_MESSAGE_SIZE as u64) + 1)
+        let cap_plus_one = MAX_HELLO_WIRE_MESSAGE_SIZE + 1;
+
+        recv.take(cap_plus_one as u64)
             .read_to_end(&mut response_buf)
             .await
             .with_context(|| format!("failed to read hello {} response", operation_name))?;
+
+        debug_assert!(
+            response_buf.len() <= cap_plus_one,
+            "bounded read invariant violated: len={} cap_plus_one={}",
+            response_buf.len(),
+            cap_plus_one
+        );
+
         if response_buf.len() > MAX_HELLO_WIRE_MESSAGE_SIZE {
             return Err(anyhow!(
                 "hello {operation_name} response too large: {} > {} bytes",
@@ -314,8 +324,10 @@ where
 
             // Read the response to avoid race condition with server
             let mut response_buf = Vec::new();
+            let cap_plus_one = MAX_HELLO_WIRE_MESSAGE_SIZE + 1;
+
             if let Err(e) = recv
-                .take((MAX_HELLO_WIRE_MESSAGE_SIZE as u64) + 1)
+                .take(cap_plus_one as u64)
                 .read_to_end(&mut response_buf)
                 .await
             {
@@ -326,6 +338,14 @@ where
                 );
                 return;
             }
+
+            debug_assert!(
+                response_buf.len() <= cap_plus_one,
+                "bounded read invariant violated: len={} cap_plus_one={}",
+                response_buf.len(),
+                cap_plus_one
+            );
+
             if response_buf.len() > MAX_HELLO_WIRE_MESSAGE_SIZE {
                 warn!(
                     ?peer,
--- a/crates/aranya-daemon/src/sync/transport/quic/client.rs
+++ b/crates/aranya-daemon/src/sync/transport/quic/client.rs
@@ -28,8 +28,7 @@ use super::{PskStore, SharedConnectionMap, SyncState, ALPN_QUIC_SYNC};
 use crate::{
     aranya::Client,
     sync::{
-        transport::quic, Addr, Callback, Error, GraphId, Result, SyncManager, SyncPeer,
-        SyncResponse,
+        Addr, Callback, Error, GraphId, Result, SyncManager, SyncPeer, SyncResponse, quic::server::MAX_SYNC_WIRE_MESSAGE_SIZE, transport::quic
     },
 };
 
@@ -44,12 +43,6 @@ pub(crate) struct QuicState {
     /// This store is modified by [`crate::api::DaemonApiServer`].
     store: Arc<PskStore>,
 }
-
-/// Upper bound for wire-encoded sync request/response envelopes.
-///
-/// `MAX_SYNC_MESSAGE_SIZE` bounds runtime sync payloads; the wire envelope adds a small
-/// serialization overhead.
-const MAX_SYNC_WIRE_MESSAGE_SIZE: usize = MAX_SYNC_MESSAGE_SIZE.saturating_add(1024);
 
 impl QuicState {
     /// Get a reference to the PSK store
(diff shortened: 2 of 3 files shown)


crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -45,6 +45,12 @@ pub(crate) struct QuicState {
     store: Arc<PskStore>,
 }
 
+/// Upper bound for wire-encoded sync request/response envelopes.
+///
+/// `MAX_SYNC_MESSAGE_SIZE` bounds runtime sync payloads; the wire envelope adds a small
+/// serialization overhead.
+const MAX_SYNC_WIRE_MESSAGE_SIZE: usize = MAX_SYNC_MESSAGE_SIZE.saturating_add(1024);
```
gknopf-aranya: There is already a `MAX_SYNC_MESSAGE_SIZE` defined.

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -338,9 +344,18 @@ where
         trace!("client receiving sync response from QUIC sync server");
 
         let mut recv_buf = Vec::new();
-        recv.read_to_end(&mut recv_buf)
+        recv.take((MAX_SYNC_WIRE_MESSAGE_SIZE as u64) + 1)
```
gknopf-aranya: Why is 1 added to `MAX_SYNC_WIRE_MESSAGE_SIZE`?

crates/aranya-daemon/src/sync/transport/quic/client.rs
```diff
@@ -338,9 +344,18 @@ where
...
+            .read_to_end(&mut recv_buf)
             .await
             .context("failed to read sync response")?;
+        if recv_buf.len() > MAX_SYNC_WIRE_MESSAGE_SIZE {
```
gknopf-aranya, line 354: If it's possible for `recv_buf` to have more than `MAX_SYNC_WIRE_MESSAGE_SIZE` allocated to it, then this is likely not resolving the memory exhaustion attack. This check seems like it should be a panic/assertion verifying the buffer is never larger than expected based on the limit specified by `take()` rather than an error. In theory, the panic should never occur. If it does, that indicates there is a bug in the implementation.

crates/aranya-daemon/src/sync/transport/quic/server.rs
```diff
@@ -36,6 +36,12 @@ use crate::{
     },
 };
 
+/// Upper bound for wire-encoded sync request/response envelopes.
+///
+/// `MAX_SYNC_MESSAGE_SIZE` bounds runtime sync payloads; the wire envelope adds a small
+/// serialization overhead.
+const MAX_SYNC_WIRE_MESSAGE_SIZE: usize = MAX_SYNC_MESSAGE_SIZE.saturating_add(1024);
```
gknopf-aranya: Same const defined for both the client and server?

crates/aranya-daemon/src/sync/transport/quic/client.rs
Sajjon: If we used `take(MAX)` and got `buf.len() == MAX`, we couldn’t tell:
* valid message of exactly `MAX` bytes, or
* oversized message truncated at `MAX`

With `take(MAX + 1)`, an oversized message can include that extra byte, so `buf.len()` becomes `MAX + 1`, and we can reliably reject it.

So the extra byte is like a "discriminator" allowing us to tell the difference.

crates/aranya-daemon/src/sync/transport/quic/client.rs
Sajjon: Right, but that is limit of the runtime layer, right? this is the daemon layer, and the `SyncResponse::Ok(data)` will add some overhead "envelop" bytes, right? So total wire bytes can be: `runtime_payload + envelope_overhead`

Which can be slightly larger than `MAX_SYNC_MESSAGE_SIZE`.

crates/aranya-daemon/src/sync/transport/quic/client.rs
Sajjon: I added `debug_assert!`s

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: That's a good point. We could try some arithmetic such as `MAX_SYNC_MESSAGE_SIZE + WIRE_OVERHEAD` so this continues to work if the runtime message size increases.

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Given the discriminant discussion here:
https://github.com/aranya-project/aranya/pull/735#discussion_r2849942124
I think we'll want to return an error at runtime if a potential DoS attack is detected.

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: Thanks for explaining this, some inline documentation on this would be useful.

crates/aranya-daemon/src/sync/transport/quic/client.rs
gknopf-aranya: And we'll want to ensure the bytes received are dropped/ignored since that error condition would likely occur due to mismatched syncer versions with different `MAX_SYNC_MESSAGE_SIZE` or a DoS attack meaning the bytes are potentially malicious/invalid.

crates/aranya-daemon/src/sync/hello.rs
```diff
@@ -164,9 +171,28 @@ where
         };
         // Read the response to avoid race condition with server
         let mut response_buf = Vec::new();
-        recv.read_to_end(&mut response_buf)
+        let cap_plus_one = MAX_HELLO_WIRE_MESSAGE_SIZE + 1;
+
+        recv.take(cap_plus_one as u64)
```
gknopf-aranya, line 176: It looks like this PR is solving the same `read_to_end()` issue in 3 places. There is an opportunity to refactor into a helper method. The helper method would have a few benefits:
- Better code reuse
- More centralized security controls that are less likely to break in the future as we refactor the syncer
- A single place where docs can be added regarding DoS, the discriminant byte check, why read_to_end() shouldn't be used directly, etc.

PR #738, opened 2026-02-25
Fix typos and ensure no more are introduced

Sajjon:
This PR consist of two commits:
- First commit fixes types in a few files
- Second commit adds prevention of introduction of new typos, by a local pre-commit job and a CD job. The pre-commit step is mentioned in README - but perhaps should be added to https://github.com/aranya-project/.github/blob/main/CONTRIBUTING.md

If you like this, I can do the same for your other repos too, both `core` and `embedded` has quite a few typos too.

### DIFF
--- a/crates/aranya-client-capi/ctest/execute_test.sh
+++ b/crates/aranya-client-capi/ctest/execute_test.sh
@@ -17,10 +17,11 @@ daemon_pids=()
 tmpdir=""
 
 # Track all child processes for cleanup
+# shellcheck disable=SC2329
 cleanup() {
     local exit_code=$?
     echo "Cleaning up processes..."
-    
+
     # First attempt: SIGTERM for clean shutdown
     # Kill tracked daemon PIDs
     for pid in "${daemon_pids[@]}"; do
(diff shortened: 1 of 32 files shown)


.github/workflows/tests.yml
```diff
@@ -23,6 +23,14 @@ concurrency:
...
 
 jobs:
+  spelling:
+    name: Spell Check with Typos
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Actions Repository
+        uses: actions/checkout@v4
+      - name: Spell Check Repo
+        uses: crate-ci/typos@v1.43.5
```
gknopf-aranya, line 33: For supply-chain security, we don't allow running actions directly from 3rd party repos. We can fork the repo at a specific version into the aranya-project then use that. This ensures the code cannot change without us manually updating the fork and pinning to the new version.

_typos.toml
gknopf-aranya, line 1: Would prefer not to clutter the repo root by adding the file here.

.github/workflows/tests.yml
Sajjon: 💯 
I did the same when I was at Radix (a blockchain company)

https://github.com/radixdlt/sargon/blob/main/.github/workflows/test.yml

I'll revert this change and move the config file

gknopf-aranya: In order to fork the repo, we may need a project admin to help. cc: @djohnson-spideroak, @bduffy-spideroak

gknopf-aranya: Nevermind, looks like there is a `typos` CLI tool that could be used instead:
https://crates.io/crates/typos

Sajjon: > Nevermind, looks like there is a `typos` CLI tool that could be used instead: https://crates.io/crates/typos

@gknopf-aranya ? what? yes, this PR is all about https://crates.io/crates/typos it is the CLI tool I'm adding usage of in both pre-commit and here in .github action. So your message did not make any sense, it IS using that CLI tool :)

gknopf-aranya: I thought `crate-ci/typos@v1.43.5` is a 3rd party action. I'm suggesting installing the tool directly in CI with `cargo install` like we do with other Rust CLI tools. For example, here's how we install `cargo make`:
https://github.com/aranya-project/aranya/blob/main/.github/actions/setup/action.yml#L55
This works around the security requirement of not using 3rd party actions.

gknopf-aranya: This approach seems simpler to me than getting a fork of the action setup.

Issue #696, opened 2026-01-22
ci: run shellcheck


jdygert-spok:
Run shellcheck in CI.

Track both aranya and aranya-core or open new issue for aranya-core.

PR #739, opened 2026-02-25
chore: fix mac-only and shellcheck lints

jdygert-spok:


gknopf-aranya (approved):

crates/aranya-client-capi/ctest/execute_test.sh
gknopf-aranya, line 1: Why aren't these lints caught by shellcheck in CI?

jdygert-spok: https://github.com/aranya-project/aranya/issues/696

gknopf-aranya: Fixing lints before adding shellcheck to CI doesn't make sense to me since this could easily brake on a subsequent PR without anyone noticing.

jdygert-spok: I'm tired of seeing these lints firing locally but trying to limit my yak shaving.

Issue #741, opened 2026-02-26
client: move #[doc(hidden)] behind capi feature


gknopf-aranya:
Rather than hiding methods behind `#[doc(hidden)]` that are only used by the C API, that code can be placed behind a `capi` feature. This ensures the Rust application cannot use any methods it isn't supposed to, since they won't be compiled as long as the `capi` feature isn't enabled. The C API can always include the `capi` feature when importing the `aranya-client` so it has the added functionality it depends on.

cc: @djohnson-spideroak, @jdygert-spok

jdygert-spok:
After mtls and #742, the only remaining cases will be the `data` methods. IMO we should just add methods to the rust api that will let us pass in the buffer to be copied into, rather than collecting them out to those types.

gknopf-aranya:
Closing since this does not appear to be needed.

PR #742, opened 2026-02-26
aranya-client: make AFC `ChannelId` an aranya-id

jdygert-spok:
Missed this one in #682.

Also cleans up some unused `doc(hidden)`.

### DIFF
--- a/crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
+++ b/crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
@@ -79,7 +79,6 @@ impl CreateTeamQuicSyncConfigBuilder {
     /// Sets the PSK seed mode.
     ///
     /// This method will be removed soon since certificates will be used instead of PSKs in the future.
-    #[doc(hidden)]
     pub fn mode(&mut self, mode: CreateSeedMode) {
         self.mode = mode;
     }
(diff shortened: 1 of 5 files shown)


crates/aranya-client/src/client/label.rs
```diff
@@ -53,11 +53,6 @@ impl Labels {
     pub fn __data(&self) -> &[Label] {
         &self.labels
     }
-
-    #[doc(hidden)]
-    pub fn __into_data(self) -> Box<[Label]> {
```
gknopf-aranya, line 58: That was used by the C API at some point. Is that no longer the case?

jdygert-spok: Yep, just `__data` is used

crates/aranya-client-capi/src/imp/config/team/quic_sync.rs
```diff
@@ -79,7 +79,6 @@ impl CreateTeamQuicSyncConfigBuilder {
     /// Sets the PSK seed mode.
     ///
     /// This method will be removed soon since certificates will be used instead of PSKs in the future.
-    #[doc(hidden)]
```
gknopf-aranya, line 82: Is this used by both the Rust application and the C API? If so, strange that this was still behind `#[doc(hidden)]`.

jdygert-spok: This is just a method defined in `aranya-client-capi` for the implementation of the C api, so none of this is really public. The corresponding method in the true Rust api is correctly marked hidden.

gknopf-aranya: Another pro for a `capi` feature that gives compilation warnings if functions that aren't supposed to be used, are used:
https://github.com/aranya-project/aranya/issues/741

jdygert-spok: This is in `aranya-client-capi` not `aranya-client`

PR #746, opened 2026-03-05
fix: add generation counter tests and fix RevokeLabelFromDevice

### LINKED ISSUE #549: Write tests for generation counters
We keep a generation counter for devices to allow reassigning a label to a device if it has been removed and re-added. We have a test that checks that duplicate label assignments fail, but we don't have one that tests that it succeeds for removed and re-added devices. AFAIK this is the only use of generation counters currently.

cc: @djohnson-spideroak @jdygert-spok @gknopf-aranya

gknopf-aranya:
## Summary

Closes #549.

- **Fix `RevokeLabelFromDevice` generation check** -- The command previously used `check exists` without validating the device generation counter, allowing revocation of stale label assignments after a device was removed and re-added. Now uses `check_unwrap query` + `label_assignment_matches_gen` to reject stale revocations.
- **Embed device generation in `AssignLabelToDevice` command fields** -- The action now captures the device's generation counter at authoring time and the policy block verifies it matches the current generation. This prevents braid reordering (RemoveDevice priority 400 > AssignLabelToDevice priority 100) from silently assigning a label to a different membership epoch when a concurrent device removal changes the generation.
- **Add `query_device_generation` test-only API** -- New ephemeral action/command/effect wired through all layers (policy, daemon, client) behind a `test-utils` feature flag. Enables tests to assert generation counter values at each step.
- **Add five integration tests** covering generation counter behavior:
  - `test_duplicate_label_assignment_rejected` -- assigning the same label twice fails
  - `test_device_generation_counter_label_reassignment` -- stale labels become invisible after remove/re-add; reassignment succeeds
  - `test_add_device_to_team_twice_rejected` -- adding an existing device fails
  - `test_revoke_stale_label_assignment_rejected` -- revoking a stale label fails (validates the bug fix)
  - `test_concurrent_label_assign_and_device_removal` -- concurrent label assignment and device removal on different devices; verifies braid reordering invalidates the assignment via generation counter mismatch

## Test plan

- [x] All new tests pass: `TMPDIR=/dev/shm cargo test -p aranya-client`
- [x] Full test suite passes (53/53): `TMPDIR=/dev/shm cargo test -p aranya-client`
- [x] Clippy clean on changed crates: `cargo clippy -p aranya-client -p aranya-daemon -p aranya-daemon-api --all-features`
- [ ] CI passes

### DIFF
--- a/crates/aranya-client/Cargo.toml
+++ b/crates/aranya-client/Cargo.toml
@@ -53,7 +53,7 @@ tracing = { workspace = true }
 
 [dev-dependencies]
 aranya-client = { path = ".", features = ["preview"] }
-aranya-daemon = { path = "../aranya-daemon", features = ["experimental", "preview"] }
+aranya-daemon = { path = "../aranya-daemon", features = ["experimental", "preview", "test-utils"] }
 spideroak-base58 = { workspace = true }
 
 backon = { workspace = true }
--- a/crates/aranya-client/src/client/device.rs
+++ b/crates/aranya-client/src/client/device.rs
@@ -225,4 +225,17 @@ impl Device<'_> {
             .map_err(IpcError::new)?
             .map_err(aranya_error)
     }
+
+    /// Returns the generation counter for the device, if any.
+    #[cfg(test)]
+    pub async fn generation(&self) -> Result<Option<i64>> {
+        let gen = self
+            .client
+            .daemon
+            .query_device_generation(create_ctx(), self.team_id, self.id)
+            .await
+            .map_err(IpcError::new)?
+            .map_err(aranya_error)?;
+        Ok(gen)
+    }
 }
--- a/crates/aranya-daemon/src/actions.rs
+++ b/crates/aranya-daemon/src/actions.rs
@@ -254,6 +254,18 @@ where
         .in_current_span()
     }
 
+    /// Invokes `query_device_generation`.
+    #[cfg(feature = "test-utils")]
+    #[instrument(skip(self))]
+    fn query_device_generation(
+        &self,
+        device_id: DeviceId,
+    ) -> impl Future<Output = Result<Vec<Effect>>> + Send {
+        self.call_session_action(policy::query_device_generation(device_id.as_base()))
+            .map_ok(|SessionData { effects, .. }| effects)
+            .in_current_span()
+    }
+
     /// Invokes `query_rank`.
     #[allow(clippy::type_complexity)]
     #[instrument(skip(self))]
--- a/crates/aranya-daemon/src/policy.md
+++ b/crates/aranya-daemon/src/policy.md
@@ -1088,6 +1088,60 @@ ephemeral command QueryRank {
 }
 ```
 
+### Device Generation Queries
+
+##### `query_device_generation`
+
+Returns the generation counter for a device. This query is intended
+for testing only and should not be exposed via any public APIs.
+
+<!-- TODO: Add conditional compilation to the policy language so
+     test-only actions/commands can be gated at the policy level. -->
+
+```policy
+// Emits `QueryDeviceGenerationResult` with the generation counter
+// for the device. If the device does not have a generation counter,
+// then no effect is emitted.
+ephemeral action query_device_generation(device_id id) {
+    publish QueryDeviceGeneration {
+        device_id: device_id,
+    }
+}
+
+effect QueryDeviceGenerationResult {
+    // The device's unique ID.
+    device_id id,
+    // The device's generation counter.
+    generation int,
+}
+
+ephemeral command QueryDeviceGeneration {
+    fields {
+        device_id id,
+    }
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        check team_exists()
+
+        let maybe_gen = query DeviceGeneration[device_id: this.device_id]
+        if maybe_gen is None {
+            finish {}
+        } else {
+            let gen = unwrap maybe_gen
+            finish {
+                emit QueryDeviceGenerationResult {
+                    device_id: this.device_id,
+                    generation: gen.generation,
+                }
+            }
+        }
+    }
+}
+```
+
 ## Roles and Permissions
 <!-- Section contains: Role facts, permissions, assignment/revocation, default roles -->
 
@@ -3126,6 +3180,8 @@ fact LabelAssignedToDevice[label_id id, device_id id]=>{op enum ChanOp, device_g
 // - It is an error if the device has already been granted
 //   permission to use this label for its current generation.
 // - It is an error if the device is not permitted to use AFC.
+// - It is an error if the device's generation has changed since
+//   the action was authored (stale membership epoch).
 //
 // # Required Permissions
 //
@@ -3133,10 +3189,13 @@ fact LabelAssignedToDevice[label_id id, device_id id]=>{op enum ChanOp, device_g
 //
 // Additionally, the target device must have `CanUseAfc` permissions
 action assign_label_to_device(device_id id, label_id id, op enum ChanOp) {
+    // Bind the command to this membership epoch.
+    let gen = get_device_gen(device_id)
     publish AssignLabelToDevice {
         device_id: device_id,
         label_id: label_id,
         op: op,
+        device_gen: gen,
     }
 }
 
@@ -3164,6 +3223,8 @@ command AssignLabelToDevice {
         // The channel operations the device is allowed to use
         // the label for.
         op enum ChanOp,
+        // Device generation at authoring time; rejects stale epochs.
+        device_gen int,
     }
 
     seal { return seal_command(serialize(this)) }
@@ -3185,12 +3246,14 @@ command AssignLabelToDevice {
         // The target device must be able to use AFC.
         check device_has_perm(this.device_id, Perm::CanUseAfc)
 
+        // Reject if the device's membership epoch changed (e.g. reordered remove/re-add).
+        let current_gen = get_device_gen(this.device_id)
+        check this.device_gen == current_gen
+
         let existing_assignment = query LabelAssignedToDevice[
             label_id: this.label_id,
             device_id: this.device_id,
         ]
-
-        let current_gen = get_device_gen(this.device_id)
 
         if existing_assignment is Some {
             // Only allow reuse when the stored assignment is
@@ -3205,6 +3268,7 @@ command AssignLabelToDevice {
             // - `author` outranks `this.device_id`
             // - `this.device_id` refers to a device that exists
             // - `this.label_id` refers to a label that exists
+            // - the command's generation matches the device's current generation
             // - the existing assignment is stale because the device has
             //   been re-provisioned
             finish {
@@ -3233,13 +3297,14 @@ command AssignLabelToDevice {
             // - `author` outranks `this.device_id`
             // - `this.device_id` refers to a device that exists
             // - `this.label_id` refers to a label that exists
+            // - the command's generation matches the device's current generation
             finish {
                 create LabelAssignedToDevice[
                     label_id: this.label_id,
                     device_id: this.device_id,
                 ]=>{
                     op: this.op,
-                    device_gen: current_gen,
+                    device_gen: this.device_gen,
                 }
 
                 emit AssignedLabelToDevice {
@@ -3317,10 +3382,18 @@ command RevokeLabelFromDevice {
         // We need to get label info before deleting
         let label = check_unwrap query Label[label_id: this.label_id]
 
-        check exists LabelAssignedToDevice[
+        let assignment = check_unwrap query LabelAssignedToDevice[
             label_id: this.label_id,
             device_id: this.device_id,
         ]
+        // Reject revocations of stale label assignments. When a
+        // device is removed and re-added its generation counter is
+        // incremented, which implicitly invalidates all prior label
+        // assignments. Revoking one of those stale assignments
+        // would delete the fact row and prevent a future (valid)
+        // assign from reusing it via the `device_gen < current_gen`
+        // update path.
+        check label_assignment_matches_gen(this.device_id, assignment.device_gen)
 
         // At this point we believe the following to be true:
         //
@@ -3329,6 +3402,7 @@ command RevokeLabelFromDevice {
         // - `author` outranks `this.device_id`
         // - `author` outranks `this.label_id`
         // - `this.label_id` refers to a label that exists
+        // - the label assignment matches the device's current generation
         finish {
             delete LabelAssignedToDevice[
                 label_id: this.label_id,
--- a/crates/aranya-client/src/tests/client.rs
+++ b/crates/aranya-client/src/tests/client.rs
@@ -3013,3 +3013,428 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
 
     Ok(())
 }
+
+/// Assigning the same label to a device twice should fail because the
+/// label assignment already exists with the same generation.
+#[test_log::test(tokio::test(flavor = "multi_thread"))]
+async fn test_duplicate_label_assignment_rejected() -> Result<()> {
+    let mut devices = DevicesCtx::new("test_duplicate_label_assignment_rejected").await?;
+    let team_id = devices.create_and_add_team().await?;
+    let roles = devices
+        .setup_default_roles(team_id)
+        .await
+        .context("unable to setup default roles")?;
+    devices.add_all_device_roles(team_id, &roles).await?;
+
+    let owner_team = devices.owner.client.team(team_id);
+    let member_role_rank = owner_team.query_rank(roles.member().id).await?;
+    let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
+    let label = owner_team.create_label(text!("label"), label_rank).await?;
+
+    // First assignment should succeed.
+    owner_team
+        .device(devices.membera.id)
+        .assign_label(label, ChanOp::SendRecv)
+        .await
+        .expect("first label assignment should succeed");
+
+    // Second assignment of the same label should fail.
+    let err = owner_team
+        .device(devices.membera.id)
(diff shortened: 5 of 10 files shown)


crates/aranya-client/src/client/device.rs
```diff
@@ -225,4 +225,19 @@ impl Device<'_> {
             .map_err(IpcError::new)?
             .map_err(aranya_error)
     }
+
+    /// Queries the generation counter for the device.
+    ///
+    /// This is a test-only API and is not part of the public client interface.
+    #[cfg(feature = "test-utils")]
+    pub async fn query_device_generation(&self) -> Result<Option<i64>> {
```
gknopf-aranya: If this is only used by unit tests, this can be implemented as a test helper method rather than adding it to `device.rs`.

crates/aranya-daemon/src/policy.md
```diff
@@ -3317,10 +3370,11 @@ command RevokeLabelFromDevice {
         // We need to get label info before deleting
         let label = check_unwrap query Label[label_id: this.label_id]
 
-        check exists LabelAssignedToDevice[
+        let assignment = check_unwrap query LabelAssignedToDevice[
             label_id: this.label_id,
             device_id: this.device_id,
         ]
+        check label_assignment_matches_gen(this.device_id, assignment.device_gen)
```
gknopf-aranya, line 3396: This fixes a bug where it was possible to revoke a label from a device that was removed/re-added. After removing the device, all label assignments should be invalidated due to the incremented generation counter. So, attempting to revoke a label that was assigned during a previous generation should fail.

crates/aranya-daemon/src/policy.md
```diff
@@ -1088,6 +1088,59 @@ ephemeral command QueryRank {
 }
 ```
 
+### Device Generation Queries
+
+##### `query_device_generation`
+
+Returns the generation counter for a device.
```
gknopf-aranya: Mention this query should only be used for testing and not exposed via any public APIs.

crates/aranya-daemon/src/policy.md
```diff
@@ -1088,6 +1088,59 @@ ephemeral command QueryRank {
...
+
+<!-- TODO: Add conditional compilation to the policy language so
```
gknopf-aranya, line 1098: Should we open an issue for this?

crates/aranya-daemon/src/policy.md
```diff
@@ -1088,6 +1088,59 @@ ephemeral command QueryRank {
...
+
+    seal { return seal_command(serialize(this)) }
+    open { return deserialize(open_envelope(envelope)) }
+
+    policy {
+        check team_exists()
+
+        let maybe_gen = query DeviceGeneration[device_id: this.device_id]
+        if maybe_gen is None {
+            finish {}
```
gknopf-aranya, line 1131: Should we test the case where the generation counter is queried before the device is added to the team?

crates/aranya-daemon/src/actions.rs
```diff
@@ -254,6 +254,19 @@ where
         .in_current_span()
     }
 
+    /// Invokes `query_device_generation`.
+    #[cfg(feature = "test-utils")]
+    #[allow(clippy::type_complexity)]
```
gknopf-aranya: What is this for?

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+        .await
+        .context("unable to setup default roles")?;
+    devices.add_all_device_roles(team_id, &roles).await?;
+
+    let owner_team = devices.owner.client.team(team_id);
+    let member_role_rank = owner_team.query_rank(roles.member().id).await?;
+    let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
+    let label_id = owner_team
+        .create_label(text!("dup_label"), label_rank)
+        .await?;
```
```suggestion
    let label = owner_team
        .create_label(text!("label"), label_rank)
        .await?;
```

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+
+    // First assignment should succeed.
+    owner_team
+        .device(devices.membera.id)
+        .assign_label(label_id, ChanOp::SendRecv)
+        .await?;
```
gknopf-aranya: `.expect()` for a better error if this fails

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+    let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
+    let label_id = owner_team
+        .create_label(text!("gen_label"), label_rank)
+        .await?;
+
+    // Generation counter should start at 0.
+    let gen = owner_team
+        .device(devices.membera.id)
+        .query_device_generation()
+        .await?;
```
gknopf-aranya: Unwrap with `expect()`

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+        )
+        .await
+        .expect_err("adding same device twice should fail");
+    assert!(matches!(err, aranya_client::Error::Aranya(_)), "{err:?}");
+
+    Ok(())
+}
+
+/// Revoking a stale label assignment (from before device removal and
+/// re-add) should fail because the generation counter doesn't match.
```
gknopf-aranya: Revoking a label from a device that was removed from the team should also fail.

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+        .device(devices.membera.id)
+        .query_device_generation()
+        .await?;
+    assert_eq!(gen, Some(0), "initial generation should be 0");
+
+    // Assign label to membera.
+    owner_team
+        .device(devices.membera.id)
+        .assign_label(label_id, ChanOp::SendRecv)
+        .await?;
```
gknopf-aranya: expect

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+
+    // Remove membera from team.
+    owner_team
+        .device(devices.membera.id)
+        .remove_from_team()
+        .await?;
```
gknopf-aranya: expect

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+    let label_rank = Rank::new(member_role_rank.value().saturating_sub(1));
+    let label_id = owner_team
+        .create_label(text!("stale_label"), label_rank)
+        .await?;
+
+    // Generation should start at 0.
+    let gen = owner_team
+        .device(devices.membera.id)
+        .query_device_generation()
+        .await?;
```
gknopf-aranya: expect

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+            Some(roles.member().id),
+            device_rank,
+        )
+        .await?;
+
+    // Generation counter should not change on re-add.
+    let gen = owner_team
+        .device(devices.membera.id)
+        .query_device_generation()
+        .await?;
```
gknopf-aranya: expect

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+        // Generation increments on removal.
+        let gen = owner_team
+            .device(devices.membera.id)
+            .query_device_generation()
+            .await?;
+        assert_eq!(
+            gen,
+            Some(i + 1),
+            "generation should be {} after removal at iteration {i}",
+            i + 1
```
gknopf-aranya: Don't compute `i+1` twice

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+            .await?;
+
+        // Generation should not change on re-add.
+        let gen = owner_team
+            .device(devices.membera.id)
+            .query_device_generation()
+            .await?;
+        assert_eq!(
+            gen,
+            Some(i + 1),
```
gknopf-aranya: Use previously computed `i+1`

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,331 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+            "generation should still be {} after re-add at iteration {i}",
+            i + 1
+        );
+    }
+
+    // Final check: generation should be 10.
+    let gen = owner_team
```
gknopf-aranya: This check is redundant, already performed in loop

crates/aranya-daemon/src/policy.md
```diff
@@ -1088,6 +1088,60 @@ ephemeral command QueryRank {
...
 ```
 
+### Device Generation Queries
+
+##### `query_device_generation`
+
+Returns the generation counter for a device. This query is intended
+for testing only and should not be exposed via any public APIs.
+
+<!-- TODO(#747): Add conditional compilation to the policy language so
```
gknopf-aranya: This issue would be in the aranya-core repo, but I'm not sure if this is a feature we plan to add yet so please close this.

crates/aranya-daemon/src/policy.md
gknopf-aranya: Test added

crates/aranya-daemon/src/policy.md
gknopf-aranya: Adding a comment about why we shouldn't allow revocation of a stale label assignment.

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,358 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+        )
+        .await
+        .expect_err("adding same device twice should fail");
+    assert!(matches!(err, aranya_client::Error::Aranya(_)), "{err:?}");
+
+    Ok(())
+}
+
+/// Revoking a label from a device that was removed from the team should
+/// fail, and revoking a stale label assignment (from before device removal
```
gknopf-aranya: It's not clear from this comment when the stale label is being assigned/revoked. Is it assigned before the device is removed and revoked after the device is re-added?

crates/aranya-daemon/src/policy.md
chip-so: An issue for developing a spec, sure. I can think of several ways to do this.

gknopf-aranya: Opened an issue here to close out this thread:
https://github.com/aranya-project/aranya-docs/issues/178

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,360 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+    // Remove membera from team.
+    owner_team
+        .device(devices.membera.id)
+        .remove_from_team()
+        .await?;
+
+    // Generation counter increments on removal.
+    let gen = owner_team
+        .device(devices.membera.id)
+        .query_device_generation()
```
chip-so: Seems weird to access the `membera` device via `owner_team` after it's been removed. But it looks like `Team::device()` doesn't check membership so this works anyway. I don't know if that's a lurking footgun or intended behavior.

crates/aranya-client/tests/tests.rs
```diff
@@ -3027,3 +3027,360 @@ async fn test_outranks_but_missing_permission() -> Result<()> {
...
+        .device(devices.membera.id)
+        .revoke_label(label)
+        .await
+        .expect_err("revoking stale label assignment should fail");
+    assert!(matches!(err, aranya_client::Error::Aranya(_)), "{err:?}");
+
+    Ok(())
+}
+
+/// Removing and re-adding a device 10 times should increment the
```
chip-so: I don't think doing this 10 times proves anything. Correct operation should be sufficiently shown by doing it once. And that's probably sufficiently tested in the tests above.

gknopf-aranya: Removing this test since the behavior is covered sufficiently by other tests.

crates/aranya-client/tests/tests.rs
gknopf-aranya: `Team::device()` is a handle to a device ID that doesn't require team membership. It's how we're able to query the device generation counter for member A's device ID. The policy will return an error if any invalid operations are attempted with a device that has been removed from the team. Queries are permitted since they don't change graph state.

chip-so: Right, I understand what's actually happening. Only being able to access a device query through a team object that the device isn't a member of is still quite strange and surprising.

gknopf-aranya: I think this strange behavior emerges from attempting to map Rust objects onto a policy interface. The Rust objects have different lifetimes than the objects in the policy. Preventing this could require querying the policy to see whether a device is on the team before allowing an object to be initialized. This would prevent querying the object generation counter on the device object though, which means we'd need a redundant team-level query on the team object. We had to choose somewhere to put each query, and placing this particular query on the device instead of the team made the most sense since it is an attribute of each device.

gknopf-aranya: At any rate, this seems like a complaint about the design/architecture of the client APIs in general since this applies to several other APIs. Not something in scope for resolving on this PR.

chip-so: I'd call it an object modeling mismatch, but I think we see the same problem. The rust code models a Device as something that must belong to a team (it has a non-optional `team_id` field), and the policy does not.

I agree that this is out of scope for the PR, but I thought it was worth bringing up.

jdygert-spok: > the policy does not.

I don't really think this is right given that anything which happens in policy is on a graph.

Even though we've removed a device from the team, there is still data associated with them for that team. So it doesn't seem right to me to say that calling `.team(foo).device(bar)` should require that bar is an active member on foo.

chip-so: Hmmm okay yeah I see what you mean. Then I think we probably wrongly called it "remove_device". It should have been something more like "deactivate" or "disable".

gknopf-aranya:
> Why do we need to explicitly check the counter if the effect of the counter is tested?

The observed effect could occur for a different wrong reason such as a bug. Seems better to verify the generation counter is incrementing as expected to make sure we don't have an unknown bug somewhere.

chip-so (approved):

crates/aranya-client/src/client/device.rs
```diff
@@ -225,4 +225,17 @@ impl Device<'_> {
             .map_err(IpcError::new)?
             .map_err(aranya_error)
     }
+
+    /// Queries the generation counter for the device.
+    #[cfg(feature = "test-utils")]
+    pub async fn query_device_generation(&self) -> Result<Option<i64>> {
```
jdygert-spok: nit: name this `generation` to match how `role` is named.

crates/aranya-client/Cargo.toml
```diff
@@ -19,8 +19,8 @@ default = [
     "aranya-daemon-api/default",
 ]
 
-# Enable debugging.
-debug = []
+# Test utilities (not part of the public API)
+test-utils = ["aranya-daemon-api/test-utils"]
```
jdygert-spok: I hesitate to add a feature which easily lets users accidentally use unstable APIs. But I guess that's what `preview` already is so this doesn't really hurt any more than that.

I also want to avoid features for combination explosion reasons.

I wonder if we should just make it a hidden method instead? :/

gknopf-aranya: Not compiling test code into a production executable is preferred both for saving space and ensuring production code doesn't depend on test code. Hiding it discourages users from using the method but does not prevent them from doing so. They could accidentally forget to stop calling the method when transitioning into production which wouldn't be ideal.

gknopf-aranya: I suppose someone could also forget to disable a test-utils feature. But auditing features seems easier than auditing usage of hidden methods.

jdygert-spok: Another possible approach: We remove the test-utils feature from aranya-client, instead using `cfg(test)` on the client methods and moving these new tests to be unit tests instead of integration tests.

gknopf-aranya: The problem with this is that both unit tests and integration tests need access to `DeviceCtx`. Rather than copying that code into the unit tests, I think it would be better to convert all the integration tests into unit tests so they have access to `cfg(test)` code. I think it would be better to first migrate the integration tests to unit tests on another PR, then add the new unit test for the generation counter here after that lands.

gknopf-aranya: Opened a PR to migrate aranya-client integration tests to unit tests:
https://github.com/aranya-project/aranya/pull/756

jdygert-spok: This can be addressed now

crates/aranya-client/Cargo.toml
```diff
@@ -19,9 +19,6 @@ default = [
     "aranya-daemon-api/default",
 ]
 
-# Enable debugging.
-debug = []
```
gknopf-aranya, line 23: Revert

crates/aranya-client/Cargo.toml
gknopf-aranya: Resolved

PR #748, opened 2026-03-05
fix: delegate member/label operations from owner to admin and operator in example

gknopf-aranya:
Have the admin add members and create labels, and the operator assign labels, instead of having the owner do everything. This matches the delegation pattern already used in aranya-metrics and the multi-node example.

### DIFF
--- a/examples/rust/aranya-example/src/main.rs
+++ b/examples/rust/aranya-example/src/main.rs
@@ -428,9 +428,10 @@ async fn main() -> Result<()> {
         .assign_role(member_role.id)
         .await?;
 
-    // Sync membera and memberb so they see their team membership and roles.
-    membera_team.sync_now(owner_addr, None).await?;
-    memberb_team.sync_now(owner_addr, None).await?;
+    // Sync membera and memberb with the operator so they see their
+    // team membership (added by admin) and role assignments (by operator).
+    membera_team.sync_now(operator_addr, None).await?;
+    memberb_team.sync_now(operator_addr, None).await?;
 
     // Demo custom roles.
     info!("demo custom roles functionality");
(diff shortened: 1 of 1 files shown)


examples/rust/aranya-example/src/main.rs
```diff
@@ -508,19 +507,24 @@ async fn main() -> Result<()> {
...
+    let admin_device_rank = admin_team.query_rank(admin.id).await?;
+    // Label rank must be lower than the admin's device rank so the admin can operate on it.
+    let label_rank = Rank::new(admin_device_rank.value().saturating_sub(1));
+    let label3 = admin_team.create_label(text!("label3"), label_rank).await?;
+
+    // Operator assigns the label to membera and memberb.
     let op = ChanOp::SendRecv;
-    info!("assigning label to membera");
-    owner_team
+    operator_team.sync_now(owner_addr, None).await?;
```
gknopf-aranya: The operator should sync with the admin here to get the label creation command since it's no longer generated by the owner.

examples/rust/aranya-example/src/main.rs
```diff
@@ -392,27 +392,41 @@ async fn main() -> Result<()> {
...
-    owner_team
+    // Admin adds membera and memberb to the team (without roles, since
+    // the admin does not have AssignRole permission).
+    info!("admin adding membera to team");
+    let member_role_rank = admin_team.query_rank(member_role.id).await?;
+    admin_team
         .add_device(
             membera.pk.clone(),
-            Some(member_role.id),
+            None,
```
gknopf-aranya, line 402: The default roles intentionally separate `AddDevice` (admin) and `AssignRole` (operator) to prevent privilege escalation, but `add_device(initial_role_id)` combines both operations under one author. This makes `initial_role_id` effectively owner-only and may encourage users to create custom roles with both permissions, contradicting our security guidelines. Recommendation: remove `initial_role_id` from `add_device` and require a separate `assign_role` call.

ssweetney-spok (approved):

chip-so (approved):

examples/rust/aranya-example/src/main.rs
gknopf-aranya: Opening an issue for more discussion on this topic:
https://github.com/aranya-project/aranya/issues/749
Doesn't block the current PR.

PR #750, opened 2026-03-17

jdygert-spok:


crates/aranya-daemon/src/policy.rs
ssweetney-spok, line 1: Should the tests using `Team::query_rank` be updated to use the new exported constants?

There should still be at least 1 test that uses `Team::query_rank` though.

gknopf-aranya: Seems like a good idea. It was deferred during the custom roles rank work because that feature wasn't available yet.

jdygert-spok: Let's do this on a follow-up

djohnson-spideroak: Agreed on followup, please ensure we have an issue created to close this discussion.

jdygert-spok: #753

PR #757, opened 2026-03-19
Replace `tarpc`

nikki-aranya:
This PR replaces our current RPC layer with a much simpler but equivalent layer that uses postcard and the existing transport code to send and receive commands.

Note that this shouldn't change any behavior, the only difference is that we're no longer relying on tarpc or MessagePack. The encryption, API surface, and UDS setup are all the same. We're also still using the underlying `Framed` code from tokio-util.

Functionally, all this does is add a macro that generates two enums, a Request and a Response enum, the necessary trait for the daemon to implement, and the few functions needed to make the DaemonApiClient work.

Before:
601 packages, `Finished release profile [optimized] target(s) in 1m 28s`
After:
583 packages, `Finished release profile [optimized] target(s) in 1m 16s`

Note that future PRs such as replacing the internal s2n-quic build with quinn in the mTLS migration will also reduce compilation times and dependency complexity, as well as some ongoing work to simplify monomorphization in aranya-core.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -414,7 +414,6 @@ dependencies = [
  "buggy",
  "bytes",
  "ciborium",
- "derive-where",
  "futures-util",
  "postcard",
  "semver",
--- a/crates/aranya-daemon-api/Cargo.toml
+++ b/crates/aranya-daemon-api/Cargo.toml
@@ -39,7 +39,6 @@ buggy = { workspace = true }
 anyhow = { workspace = true }
 bytes = { workspace = true, features = ["serde"] }
 ciborium = { workspace = true }
-derive-where = { workspace = true }
 futures-util = { workspace = true }
 postcard = { workspace = true }
 semver = { workspace = true }


nikki-aranya:
Looks like this change shaves 2 minutes off of unit tests in CI, based on very rough numbers, which is extremely welcome

crates/aranya-client/src/client/device.rs
```diff
@@ -89,9 +89,8 @@ impl Device<'_> {
     pub async fn public_key_bundle(&self) -> Result<PublicKeyBundle> {
         self.client
             .daemon
-            .device_public_key_bundle(create_ctx(), self.team_id, self.id)
+            .device_public_key_bundle(self.team_id, self.id)
```
gknopf-aranya, line 92: @nikki-aranya, @jnewell-spok this context change conflicts with the tracing ID change that's being introduced here https://github.com/aranya-project/aranya/pull/731

crates/aranya-client/src/client.rs
```diff
@@ -53,11 +48,6 @@ use crate::{
     util::ApiConv as _,
 };
 
-/// IPC timeout of 1 year (365 days).
-// A large value helps resolve IPC calls timing out when there are long-running
-// operations happening in the daemon.
-const IPC_TIMEOUT: Duration = Duration::from_secs(365 * 24 * 60 * 60);
```
gknopf-aranya, line 59: Will the new API never time out?

AGENTS.md
```diff
@@ -59,16 +59,15 @@ Key sections in the policy: Devices & Identity, Roles & Permissions, Teams, AFC/
...
 6. If all commands succeed, the action is committed atomically. If any command fails, the entire action is rolled back.
 
-On **every other device** (runs on sync):
-7. The command envelope arrives via sync. The `open` block deserializes it.
-8. The `policy` block runs again -- same checks, same finish logic. Effects are emitted on the receiving device too, so every device can react to the outcome. The receiving device must be able to fully evaluate the command with only the command fields and its local FactDB. This is why command fields cannot contain plaintext secrets.
+On **every other device** (runs on sync): 7. The command envelope arrives via sync. The `open` block deserializes it. 8. The `policy` block runs again -- same checks, same finish logic. Effects are emitted on the receiving device too, so every device can react to the outcome. The receiving device must be able to fully evaluate the command with only the command fields and its local FactDB. This is why command fields cannot contain plaintext secrets.
```
gknopf-aranya, line 70: Not sure why this part changed on this PR

crates/aranya-daemon/src/api.rs
```diff
@@ -1,9 +1,9 @@
-//! Implementation of daemon's `tarpc` API.
+//! Implementation of daemon's RPC API.
```
gknopf-aranya, line 1: With this change it may be better to refer to this as a UDS or IPC API rather than an RPC. RPC implies remote procedure calls from another machine but this is a local API only.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -365,202 +366,253 @@ pub enum Perm {
...
+            )*
+        }
+    };
+}
+
+rpc! {
+    /* Miscellaneous */
     /// Returns the daemon's version.
-    async fn version() -> Result<Version>;
+    fn version() -> Version;
```
gknopf-aranya, line 704: Why is `async` removed? Why is a `Result` no longer returned? Seems like we would still want to `.await` on API calls and handle errors.

crates/aranya-daemon-api/src/crypto/txp.rs
```diff
@@ -196,139 +175,168 @@ enum Side {
     Client,
 }
 
+/// A message (request) sent by the client to the server.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[non_exhaustive]
+enum ClientMsg {
```
gknopf-aranya: The diff makes it look like this is new code, but it was only moved.

gknopf-aranya:
If we want to replace tarpc with our own API, I think we're going to want to test more edge-cases. This currently verifies that the API works with the aranya client under nominal usage.  I'm guessing tarpc had unit test coverage that could give us insight into the types of things to test. Examples:
- What happens if a method is called and cancelled before it completes? Will future requests still work?
- What happens if many requests are made in parallel? Is the API able to buffer everything, or does it drop requests?
- What happens if a method loops forever and never returns?

crates/aranya-daemon-api/src/service.rs
```diff
@@ -365,202 +366,253 @@ pub enum Perm {
...
+            .ok_or_else(|| {
+                ClientError::Transport(io::Error::new(
+                    io::ErrorKind::UnexpectedEof,
+                    "daemon closed connection",
+                ))
+            })
+    }
+}
+
+macro_rules! rpc {
```
gknopf-aranya, line 635: `service.rs` is intended to define just the interface of the API that `aranya-client` interacts with. The boilerplate code for handling requests/responses seems like it belongs in a different module.

crates/aranya-daemon/src/api.rs
jdygert-spok: I don't think it necessarily needs to be limited to the same machine.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -365,202 +366,253 @@ pub enum Perm {
...
+                match req {
+                    $(
+                        $(#[$meta])*
+                        DaemonApiRequest::$name { $($arg),* } => {
+                            DaemonApiResponse::$name(self.$name($($arg),*).await)
+                        }
+                    )*
+                    // cfg'd-out variants need a catch-all
+                    #[allow(unreachable_patterns)]
+                    _ => DaemonApiResponse::UnknownRequest,
```
jdygert-spok: Wouldn't we error earlier when trying to deserialize to `DaemonApiRequest` since the variant is cfg'd out?

crates/aranya-daemon-api/src/service.rs
```diff
@@ -365,202 +366,253 @@ pub enum Perm {
...
+        conn.recv()
+            .await
+            .map_err(ClientError::Transport)?
+            .ok_or_else(|| {
+                ClientError::Transport(io::Error::new(
+                    io::ErrorKind::UnexpectedEof,
+                    "daemon closed connection",
+                ))
+            })
+    }
```
jdygert-spok, line 632: This is not actually concurrently usable like the tarpc client was, since it locks the connection.

crates/aranya-daemon-api/src/service.rs
jdygert-spok: They are still there in the expansion.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -365,202 +366,253 @@ pub enum Perm {
...
+                $name(Result<$ret>),
+            )*
+        }
+
+        /// The handler trait, needs to be implemented server-side.
+        #[allow(unused_variables, async_fn_in_trait)]
+        pub trait DaemonApi {
+            $(
+                $(#[$meta])*
+                async fn $name(&self, $($arg: $ty),*) -> Result<$ret>;
```
jdygert-spok: Now that we have control over it, I think it would be useful to add something like `type Error: Into<ApiError>`, so we can implement the methods returning an error type which is nicer to work with, while the `ApiError` is the one that actually gets serialized.

crates/aranya-daemon/src/api.rs
```diff
@@ -183,27 +179,34 @@ impl DaemonApiServer {
...
-                    let requests = ch
-                        .requests()
-                        .inspect_err(|err| warn!(error = %err.report(), "channel failure"))
-                        .take_while(|r| future::ready(r.is_ok()))
-                        .filter_map(|r| async { r.ok() });
-                    let mut requests = pin!(requests);
-                    while let Some(req) = requests.next().await {
-                        reqs.spawn(req.execute(api.clone().serve()));
+                s.spawn(async move {
+                    loop {
```
jdygert-spok: This seems to no longer be able to process requests from one connection concurrently.

jdygert-spok: I think tarpc would also cancel an operation if the requester dropped their future. Unsure whether we need that.

crates/aranya-client/src/client.rs
nikki-aranya: I wasn't sure if this actually did anything since we don't use the context for anything, but I looked it up and I guess it does, I can add it pretty easily

crates/aranya-daemon-api/src/service.rs
nikki-aranya: Yeah, async and Result get handled in the macro so they don't need to be added to the function definition

crates/aranya-client/src/client.rs
jdygert-spok: I believe we decided we wanted none of the calls to time out. (Although the best we could do for tarpc was set it very high.)

gknopf-aranya: We were setting the timeout to a really large value, so I don't think it matters if the API never never times out. That might actually be better since this was a workaround for tarpc requiring a timeout.

crates/aranya-daemon-api/src/service.rs
nikki-aranya: We never actually used the concurrency afaik, since we .await the calls every time

crates/aranya-daemon-api/src/service.rs
gknopf-aranya: Makes sense. That's better since it's more concise and less error-prone.

crates/aranya-daemon/src/api.rs
nikki-aranya: Probably should keep the cancellation timeout, it's good practice

crates/aranya-daemon-api/src/service.rs
jdygert-spok: IDK if we used them concurrently ourselves in practice, but we definitely exposed it to the user since the client methods take `&self`.

jdygert-spok: e.g. this would actually send requests concurrently:

```rust
join!(
    client.get_public_key_bundle(...),
    client.team(...).add_sync_peer(...),
    client.team(...).labels(),
);
```

crates/aranya-daemon/src/api.rs
jdygert-spok: To be clear, not a timeout but a signal triggered by dropping the future that made the request.

crates/aranya-daemon-api/src/service.rs
gknopf-aranya: That means not exposing concurrency to the user would be a breaking API change. On the other hand, it doesn't seem like a good idea to expose concurrency without testing that it works.

crates/aranya-daemon-api/src/service.rs
nikki-aranya: oops, yeah, I was focused on the actual codegen

crates/aranya-daemon-api/src/service.rs
nikki-aranya: I was thinking about that, probably should

crates/aranya-daemon-api/src/crypto/txp.rs
nikki-aranya: yeah, I had to reorganize it since I removed a lot of the code that was no longer necessary

crates/aranya-daemon/src/api.rs
gknopf-aranya: I don't think we would use this API for remote connections. Aren't there different authentication/encryption requirements for that case? Seems more likely that a HTTPS REST endpoint, `tarpc`, or QUIC would be used for a remote connection.

jdygert-spok: I think we're conflating two things under "api" and "rpc".

1. the code for setting up the transport for the API to be served specifically over a unix socket
2. the implementation of `DaemonApi` describing the response to each request

I think these should be independent so this module probably ought to be split up.

gknopf-aranya: I see. In the future, we may want an authenticated endpoint to forward requests into the same `DaemonApi` to control it (without the UDS specific code that is here).

gknopf-aranya:
@nikki-aranya are you planning to add support for bidirectional async API calls on this PR? The desire for an async IPC API is something the team has discussed before. An example would be `sync_now()` returning right away then later sending the application an async notification indicating that `sync_now()` has completed. Seems fine to add support for that in a follow-up PR but I do want to confirm that the API you're adding would be compatible with adding that feature in the future.

PR #768, opened 2026-03-31
client: fix new query method naming consistency

jdygert-spok:
Other client query methods don't have a `query_` prefix.

### DIFF
--- a/examples/rust/aranya-example-multi-node/src/bin/main.rs
+++ b/examples/rust/aranya-example-multi-node/src/bin/main.rs
@@ -40,9 +40,6 @@ async fn main() -> Result<()> {
     let example = workspace.join("example.env");
     env.generate(&example).await?;
 
-    // Set environment variables before spawning child processes.
-    env.set();
-
     // Start a daemon for each device.
     let tmp = tempdir()?;
     let mut daemons = Vec::with_capacity(env.devices().count());
@@ -55,7 +52,7 @@ async fn main() -> Result<()> {
 
         // Start daemon.
         info!("starting {} daemon", device.name);
-        let child = daemon(&release, &cfg).context("expected to spawn daemon")?;
+        let child = daemon(&release, &cfg, &env).context("expected to spawn daemon")?;
         daemons.push(child);
     }
     // Wait for daemons to start.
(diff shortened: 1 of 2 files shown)


crates/aranya-client/src/client/team.rs
```diff
@@ -372,7 +372,7 @@ impl Team<'_> {
 
     /// Queries the rank of an object.
     #[instrument(skip(self))]
-    pub async fn query_rank(&self, object_id: impl ToObjectId) -> Result<Rank> {
+    pub async fn rank(&self, object_id: impl ToObjectId) -> Result<Rank> {
```
jdygert-spok, line 377: This should probably return `Result<Option<Rank>>` like `label` does.

There's ~100 call sites to update in our tests :/

There's also the question of what to do about the C api method `query_rank`. We didn't already resolve this since we don't seem to have any other C function like this. I guess we skipped `query_label` there.

gknopf-aranya: Rank is a property of all objects in the policy. It is not optional.

jdygert-spok: How should we handle the case where we don't know that object ID? (ID was made up or was not yet synced or has since been deleted). `query_label` returns `None` if there's no label under the LabelID. I think that's similar.

gknopf-aranya: I'd expect querying for a rank of an object that doesn't exist to return an error. `None` would make sense if the object existed but didn't have a rank set, but all objects that exist should have a rank set (enforced by the policy). Label isn't quite the same because a device could either have a label assigned to it or not.

jdygert-spok: `query_label` looks up the existence of a label, not whether a device has labels.

gknopf-aranya: Ah, then `Result<Option<Rank>>` seems okay. Consistency is important. And getting a `None` return is easier to diagnose than handling different error variants if an object doesn't exist.

gknopf-aranya: The other option is to not have `query_label` return `Option`. But this hides the root-cause of non-existence in an error that the caller would need to parse. That seems less ideal.

gknopf-aranya: @djohnson-spideroak, @jdygert-spok if we want to make this change, I'd prefer to get it in before the current release since it is a breaking change.

jdygert-spok: #770

PR #771, opened 2026-03-31
client: clean up label query API consistency

gknopf-aranya:
## Summary

- Change `label()` to return `Result<Label>` instead of `Result<Option<Label>>` across the daemon API, daemon implementation, and client library
- `device_role()` remains `Result<Option<Role>>` since a device may legitimately have no role assigned
- Add `DoesNotExist` error variant to `api::Error`, `aranya_client::Error`, and the C API (`ARANYA_ERROR_DOES_NOT_EXIST`) so callers can distinguish "not found" from other errors
- Use `DoesNotExist` for label, device public key bundle, and rank queries when the resource is missing
- Add `AranyaLabel` opaque type to the C API with `imp::Label` wrapper (using `CString` for null-terminated name), with accessor methods (`label_get_id`, `label_get_name`, `label_get_author`)
- Add `team_label` C API function to query a label by ID
- `team_label_exists` matches on the `DoesNotExist` error variant to return `false` instead of propagating the error
- Add C example coverage for the new `team_label` query in the AFC example

Follows from the discussion in #768.
Depends on #778.

## Test plan

- [ ] `cargo clippy -p aranya-client -p aranya-daemon -p aranya-client-capi` passes with `-D warnings`
- [ ] `cargo test -p aranya-client` passes
- [ ] Rust example runs (`cargo make run-rust-example`)
- [ ] C example runs (`cargo make run-capi-example`)

### DIFF
--- a/crates/aranya-client-capi/src/api/defs.rs
+++ b/crates/aranya-client-capi/src/api/defs.rs
@@ -107,7 +107,7 @@ impl From<&imp::Error> for Error {
             imp::Error::Client(err) => match err {
                 aranya_client::Error::Ipc(_) => Self::Ipc,
                 aranya_client::Error::Aranya(_) => Self::Aranya,
-                aranya_client::Error::DoesNotExist(_) => Self::DoesNotExist,
+                aranya_client::Error::DoesNotExist => Self::DoesNotExist,
                 aranya_client::Error::Bug(_) => Self::Bug,
                 aranya_client::Error::Config(_) => Self::Config,
                 aranya_client::Error::Other(_) => Self::Other,
@@ -438,27 +438,38 @@ impl From<ChanOp> for aranya_client::ChanOp {
 /// A label.
 #[aranya_capi_core::derive(Cleanup)]
 #[aranya_capi_core::opaque(size = 112, align = 8)]
-pub type Label = Safe<aranya_client::Label>;
+pub type Label = Safe<imp::Label>;
 
 /// Get ID of label.
 ///
 /// @param[in] label the label
 ///
 /// @relates AranyaLabel
+#[aranya_capi_core::no_ext_error]
 pub fn label_get_id(label: &Label) -> LabelId {
-    label.deref().id.into()
+    label.id.into()
 }
 
-// TODO(#777): add label_get_name once string accessors return
-// null-terminated strings or provide a length.
+/// Get name of label.
+///
+/// The resulting string is null-terminated and must not be freed.
+///
+/// @param[in] label the label
+///
+/// @relates AranyaLabel
+#[aranya_capi_core::no_ext_error]
+pub fn label_get_name(label: &Label) -> *const c_char {
+    label.name.as_ptr()
+}
 
 /// Get the author of a label.
 ///
 /// @param[in] label the label
 ///
 /// @relates AranyaLabel
+#[aranya_capi_core::no_ext_error]
 pub fn label_get_author(label: &Label) -> DeviceId {
-    label.deref().author_id.into()
+    label.author_id.into()
 }
 
 /// Label ID.
@@ -2333,7 +2344,7 @@ pub fn team_label(
     let label = client
         .rt
         .block_on(client.inner.team(team.into()).label(label_id.into()))?;
-    Label::init(label_out, label);
+    Label::init(label_out, imp::Label::from(label));
     Ok(())
 }
 
@@ -2355,7 +2366,7 @@ pub unsafe fn team_label_exists(
         .block_on(client.inner.team(team.into()).label(label.into()))
     {
         Ok(_) => Ok(true),
-        Err(aranya_client::Error::DoesNotExist(_)) => Ok(false),
+        Err(aranya_client::Error::DoesNotExist) => Ok(false),
         Err(e) => Err(e.into()),
     }
 }
--- a/crates/aranya-client/src/error.rs
+++ b/crates/aranya-client/src/error.rs
@@ -24,8 +24,8 @@ pub enum Error {
     Aranya(#[from] AranyaError),
 
     /// The requested resource does not exist.
-    #[error("does not exist: {0}")]
-    DoesNotExist(String),
+    #[error("does not exist")]
+    DoesNotExist,
 
     /// A configuration error happened.
     #[error("configuration error")]
@@ -77,7 +77,7 @@ pub struct AranyaError {
 
 pub(crate) fn aranya_error(err: api::Error) -> Error {
     match err {
-        api::Error::DoesNotExist(msg) => Error::DoesNotExist(msg),
+        api::Error::DoesNotExist(_) => Error::DoesNotExist,
         err => Error::Aranya(err.into()),
     }
 }
--- a/crates/aranya-daemon-api/src/service.rs
+++ b/crates/aranya-daemon-api/src/service.rs
@@ -31,6 +31,7 @@ pub type CE = DefaultEngine;
 pub type CS = <DefaultEngine as Engine>::CS;
 
 /// An error returned by the API.
+// TODO: add more error variants as needed for control flow.
 #[derive(Serialize, Deserialize, Debug)]
 pub enum Error {
     /// The requested resource does not exist.
--- a/examples/c/example.c
+++ b/examples/c/example.c
@@ -859,6 +859,11 @@ AranyaError run_afc_example(Team *t) {
     err = aranya_label_get_id(&label, &queried_label_id);
     EXPECT("error getting label ID", err);
 
+    const char *queried_label_name = NULL;
+    err = aranya_label_get_name(&label, &queried_label_name);
+    EXPECT("error getting label name", err);
+    printf("created label '%s'\n", queried_label_name);
+
     aranya_label_cleanup(&label);
 
     // Tell them both to sync with the owner to see their new label.
--- a/crates/aranya-client-capi/output/aranya-client.h
+++ b/crates/aranya-client-capi/output/aranya-client.h
@@ -929,15 +929,16 @@ AranyaError aranya_label_get_id(const struct AranyaLabel *label,
                                 struct AranyaLabelId *__output);
 
 /**
- * Get ID of label.
+ * Get name of label.
+ *
+ * The resulting string is null-terminated and must not be freed.
  *
  * @param[in] label the label
  *
  * @relates AranyaLabel
  */
-AranyaError aranya_label_get_id_ext(const struct AranyaLabel *label,
-                                    struct AranyaLabelId *__output,
-                                    struct AranyaExtError *__ext_err);
+AranyaError aranya_label_get_name(const struct AranyaLabel *label,
+                                  const char **__output);
 
 /**
  * Get the author of a label.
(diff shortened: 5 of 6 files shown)


examples/c/example.c
```diff
@@ -850,6 +850,20 @@ AranyaError run_afc_example(Team *t) {
         goto exit;
     }
 
+    // Query the label to verify it was created correctly.
+    AranyaLabel label;
+    err = aranya_team_label(&owner->client, &t->id, &label_id, &label);
+    EXPECT("error querying label", err);
+
+    AranyaLabelId queried_label_id;
+    err = aranya_label_get_id(&label, &queried_label_id);
```
gknopf-aranya, line 859: Calling this before initializing the label could cause undefined behavior. Opened an issue for making the API safer:
https://github.com/aranya-project/aranya/issues/772

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2282,11 +2317,34 @@ pub unsafe fn team_labels(
...
+/// Returns the label metadata for the given label ID.
+/// Returns an error if the label does not exist.
+///
+/// @param[in] client the Aranya Client
+/// @param[in] team the team's ID
+/// @param[in] label_id the label ID to query
+/// @param[out] label_out returns the label
+///
+/// @relates AranyaClient.
+pub fn team_label(
```
djohnson-spideroak, line 2338: is the name here consistent with other query functions?

crates/aranya-daemon/src/api.rs
```diff
@@ -1138,13 +1138,13 @@ impl DaemonApi for Api {
...
-            Ok(Some(api::Label {
+            Ok(api::Label {
                 id: api::LabelId::from_base(e.label_id),
                 name: e.label_name.clone(),
                 author_id: api::DeviceId::from_base(e.label_author_id),
-            }))
+            })
         } else {
-            Ok(None)
+            Err(anyhow!("label not found").into())
```
djohnson-spideroak: We would probably want a "not found" error

examples/c/example.c
```diff
@@ -850,6 +850,22 @@ AranyaError run_afc_example(Team *t) {
...
+    EXPECT("error getting label ID", err);
+
+    const char *queried_label_name = NULL;
+    err = aranya_label_get_name(&label, &queried_label_name);
+    EXPECT("error getting label name", err);
+    printf("created label '%s'\n", queried_label_name);
+
+    aranya_label_cleanup(&label);
```
djohnson-spideroak, line 867: Is the label name free'd here?

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -2296,11 +2354,10 @@ pub unsafe fn team_label_exists(
...
     label: &LabelId,
 ) -> Result<bool, imp::Error> {
-    let label_result = client
+    Ok(client
         .rt
-        .block_on(client.inner.team(team.into()).label(label.into()))?;
-    let exists = label_result.is_some();
-    Ok(exists)
+        .block_on(client.inner.team(team.into()).label(label.into()))
+        .is_ok())
```
jdygert-spok: This does not tell you whether the label exists, there are other errors that can occur.

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -430,6 +430,41 @@ impl From<ChanOp> for aranya_client::ChanOp {
...
+///
+/// The resulting string must not be freed.
+///
+/// @param[in] label the label
+///
+/// @relates AranyaLabel
+#[aranya_capi_core::no_ext_error]
+pub fn label_get_name(label: &Label) -> *const c_char {
+    label.deref().name.as_ptr().cast()
+}
```
jdygert-spok, line 463: `Text` is not null terminated. (Maybe it should be so we can do this?)

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Then we can return the error directly and let the caller check the error type if needed. If there is no error, the caller can assume the label exists.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Yes, `team_roles`, `team_devices`

examples/c/example.c
gknopf-aranya: Yes

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: We need to add the error cases that distinguish this.

gknopf-aranya: Yep, I'm adding the logic as we speak

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: Just saw `role_get_name` does this too :/

gknopf-aranya: Yep, this is following the conventions already established for strings

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya, line 2359: The API has `team_label_exists()` but not `team_role_exists()`

gknopf-aranya: I don't think we can add a `team_device_exists()` because the policy doesn't know about a device until it is added to the team. `team_devices()` returns a list of devices added to the team which we can use to determine whether a device was added to the team.

gknopf-aranya: We could add a `team_device_is_on_team()` method to check if a specific device is on the team without needing to query for all devices on the team. I suggest that we add these APIs later since they can be added in a backward compatible way.

crates/aranya-client-capi/src/api/defs.rs
jdygert-spok: #775

examples/c/example.c
gknopf-aranya: Marking as resolved since this was an existing issue prior to this PR. The safety issue only occurs if the API is misused. Resolving this issue can be done in a backward compatible way after the release.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: https://github.com/aranya-project/aranya/issues/776

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Can this be done post-release?

crates/aranya-daemon/src/api.rs
gknopf-aranya: Added a new error variant for this. Needed for device/role as well.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: `get_role_name()` is an existing method that already had this UB for a long time (we just didn't know). We'll resolve that in a follow-up PR. I'll add a TODO to introduce `label_get_name()` and `label_get_author()` after the source of UB is resolved so we don't introduce any new methods with known UB in this release.

jdygert-spok: Why remove `label_get_author`?

gknopf-aranya: Reverted. Thought that returned a string as well for a sec.

crates/aranya-daemon/src/api.rs
```diff
@@ -1213,14 +1215,13 @@ impl DaemonApi for Api {
...
             trace!(?graph, "queried label");
-            Ok(Some(api::Label {
+            Ok(api::Label {
                 id: api::LabelId::from_base(e.label_id),
                 name: e.label_name.clone(),
                 author_id: api::DeviceId::from_base(e.label_author_id),
-            }))
+            })
         } else {
-            trace!(?graph, "queried label (none)");
```
gknopf-aranya, line 1222: Revert

crates/aranya-daemon-api/src/service.rs
```diff
@@ -31,17 +31,21 @@ pub type CE = DefaultEngine;
 pub type CS = <DefaultEngine as Engine>::CS;
 
 /// An error returned by the API.
-// TODO: enum?
 #[derive(Serialize, Deserialize, Debug)]
-pub struct Error(String);
+pub enum Error {
+    /// The requested resource does not exist.
+    DoesNotExist(String),
```
gknopf-aranya, line 38: Not sure how I feel about special-casing just the DoesNotExist error

gknopf-aranya: I would prefer to either not add a new DoesNotExist error variant here (requires the client to parse the error string to figure out what variant it is) or create new error variants for all the possible errors (larger refactor).

jdygert-spok: We can do this incrementally, and we don't need a variant for _all_ possible variants, just ones that need to be distinguished for control flow.

crates/aranya-daemon-api/src/service.rs
```diff
@@ -31,17 +31,21 @@ pub type CE = DefaultEngine;
 pub type CS = <DefaultEngine as Engine>::CS;
 
 /// An error returned by the API.
-// TODO: enum?
```
gknopf-aranya, line 34: This becomes a todo to add more error variants

crates/aranya-client-capi/src/api/defs.rs
```diff
@@ -430,6 +435,41 @@ impl From<ChanOp> for aranya_client::ChanOp {
...
+
+/// Get name of label.
+///
+/// The resulting string is null-terminated and must not be freed.
+///
+/// @param[in] label the label
+///
+/// @relates AranyaLabel
+#[aranya_capi_core::no_ext_error]
+pub fn label_get_name(label: &Label) -> *const c_char {
```
jdygert-spok, line 461: Could put `no_ext_error` on the other getters too.

crates/aranya-client/src/error.rs
```diff
@@ -23,6 +23,10 @@ pub enum Error {
     #[error("daemon error")]
     Aranya(#[from] AranyaError),
 
+    /// The requested resource does not exist.
+    #[error("does not exist: {0}")]
+    DoesNotExist(String),
```
jdygert-spok: Worth thinking a bit about if we want to shape this any different since it is part of the public api.

gknopf-aranya: Callers probably don't need access to the raw string. It is useful to be able to tell if it's a device/role that doesn't exist (but this information should already be known in the context of the original query).

gknopf-aranya: If we provide info about the type of object that does not exist, then that could become less maintainable as more objects are introduced. A generic `DoesNotExist` applies to any object/resource.

crates/aranya-client-capi/src/api/defs.rs
gknopf-aranya: Added this to the other getters

gknopf-aranya:
Rerunning CI jobs. Failed due to a known/reported flake.

crates/aranya-client/src/error.rs
gknopf-aranya: Removed the string from the public facing API error.

PR #773, opened 2026-04-01
fix: use device ID for QUIC connection tie-breaking

### LINKED ISSUE #754: bug: quic connection tie-break on network address is brittle
Tie-breaking QUIC connections based on peer addresses will not work for situations where there is not a direct connection between peers: NAT, ISP, proxy, etc.

Use the QUIC connection ID to tie-break the QUIC connections instead.
This ID is randomly generated so it should be unique for each QUIC connection.
This implies it will always have a deterministic ordering comparison that will break the tie.

Tie-breaking based on the device ID is another option, but it seemed like the QUIC connection ID may be more readily available to the QUIC transport without jumping through extra hoops.

Considering a scenario where a peer maliciously sets two connections to the same ID.
But, the impact is only possibly closing the wrong connection with a malicious peer. There doesn't seem to be a downside to closing connections with a malicious peer anyways.

https://github.com/aranya-project/aranya/pull/717/changes#r2948041410

cc: @djohnson-spideroak, @jdygert-spok, @nikki-aranya, @ssweetney-spok

gknopf-aranya:
## Summary

Replace network address comparison with device ID comparison for QUIC connection tie-breaking when both peers connect simultaneously. Addresses are brittle behind NAT, proxies, and ISPs.

### Changes

- Add `ConnectionInfo` struct (port + `DeviceId`) exchanged after the QUIC handshake via postcard serialization. This type can be extended with additional fields in the future.
- Connector sends `ConnectionInfo`, listener sends back its `DeviceId` (also via postcard)
- Both sides use device ID ordering to deterministically pick a winner: the peer with the lower device ID keeps its outbound connection
- Use typed `DeviceId` throughout instead of raw `[u8; 32]` — bytes conversion only at the wire boundary
- Remove unused `local_addr` field from `QuicConnector`
- Pass local `DeviceId` through `ConnectionPool` to both connector and listener

### Why device ID?

- Cryptographically unique and stable per device
- Transport-agnostic — works with both s2n-quic and quinn
- Both peers agree on ordering without additional coordination
- QUIC wire-protocol connection IDs (RFC 9000) are not accessible via public APIs in either s2n-quic or quinn
- s2n-quic's internal connection ID is a sequential counter, not suitable for tie-breaking

**Note:** This is a sync wire protocol change. Peers running previous versions will not be able to connect to peers running this version.

Closes #754

## Test plan

- [ ] `cargo clippy -p aranya-daemon` passes with `-D warnings`
- [ ] `cargo test -p aranya-client` passes
- [ ] Multi-node sync tests pass


crates/aranya-daemon/src/sync/transport/quic/connections.rs
```diff
@@ -11,15 +11,17 @@ type ConnectionUpdate = (SyncPeer, StreamAcceptor);
 /// Shared state for coordinating QUIC connections between connector and listener.
 pub(crate) struct ConnectionPool {
     conns: SharedConnectionMap,
+    local_device_id: [u8; 32],
```
gknopf-aranya: Would rather store `DeviceId` in a struct and serialize/convert to bytes right before putting them on the wire.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -24,10 +25,8 @@ pub(crate) struct QuicConnector {
     pool: ConnectorPool,
     /// Allows authenticating the identity of a given `GraphId`.
     psk_store: Arc<PskStore>,
-    /// The return port we want the peer to connect to us on.
-    return_port: Bytes,
-    /// The local address of the server, since it should be infallible.
-    local_addr: Addr,
+    /// Serialized connection info to send to peers.
+    conn_info_bytes: Bytes,
```
gknopf-aranya, line 29: Why store Bytes here when it can be serialized on-demand before putting the data on the wire?

gknopf-aranya: Oh, it's re-used later so it's fine to cache this in advance.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -121,14 +125,27 @@ impl SyncConnector for QuicConnector {
...
-            debug!(?peer, "QUIC handshake complete, sent return port");
+            // Receive the listener's device ID for tie-breaking.
+            let remote_id_bytes = conn
+                .accept_receive_stream()
+                .await?
+                .context("unable to accept device ID stream")?
+                .receive()
+                .await?
+                .context("peer didn't send device ID")?;
+            let remote_device_id: DeviceId =
```
gknopf-aranya: We still need to be able to tie-break even if the peer sends an invalid device ID. I think the right approach here is to close the peer's connection if they send an invalid device ID.

PR #786, opened 2026-04-07
mTLS 2

### LINKED ISSUE #647: quinn: implement mTLS certs

### LINKED ISSUE #648: quinn: implement quinn transport for syncer

### LINKED ISSUE #663: ci: remove RUSTSEC-2025-0134 exception during quinn migration

jdygert-spok:
Migrate from our `s2n-quic` fork to `quinn`, and from using preshared keys to using mutual TLS.

Closes #647.
Closes #648.
Closes #663.


crates/aranya-daemon/src/sync/transport/quic/mod.rs
```diff
@@ -69,3 +71,87 @@ impl From<std::convert::Infallible> for Error {
...
+
+/// Creates a QUIC transport configuration with no idle timeout.
+///
+/// Connections are kept open indefinitely without keep-alive pings.
+/// This is appropriate for daemon-to-daemon connections where resources
+/// are not constrained and connection reuse is desired.
+fn transport_config() -> Arc<TransportConfig> {
+    let mut transport_config = TransportConfig::default();
+    // Disable idle timeout - connections stay open forever.
+    transport_config.max_idle_timeout(None);
```
jdygert-spok, line 153: > WARNING: If a peer or its network path malfunctions or acts maliciously, an infinite idle timeout can result in permanently hung futures!

crates/aranya-client-capi/CMakeLists.txt
```diff
@@ -31,6 +31,17 @@ corrosion_import_crate(
...
+
+corrosion_import_crate(
+    MANIFEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../aranya-certgen/Cargo.toml
+    CRATES aranya-certgen
+    CRATE_TYPE bin
+    FEATURES cli
+)
+
+add_executable(Aranya::Certgen ALIAS aranya-certgen)
+
```
jdygert-spok, line 44: Move to tests

jdygert-spok: If we are stabilizing and supplying aranya-certgen to the user then it can stay here. Is that the plan?

crates/aranya-client/tests/scale/init.rs
```diff
@@ -182,22 +197,56 @@ impl TestCtx {
...
             _work_dir: work_dir,
         })
     }
 }
+
+/// Shared certificate authority for generating device certificates.
+struct TestCertAuthority {
+    ca: CaCert,
+    root_certs_dir: PathBuf,
+}
```
jdygert-spok, line 216: Copied this from `src/tests/common.rs`. Could move the scale tests into `src/tests` to have access to it.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -1,194 +1,111 @@
...
-                .await
-                .context("timed out waiting for return address")?
-                .context("unable to extract return address")?;
-        let peer = SyncPeer::new(peer_addr, GraphId::transmute(active_team));
+        let graph_id = tokio::time::timeout(RESOLVE_GRAPH_ID_TIMEOUT, extract_graph_id(&mut conn))
+            .await
+            .context("timed out waiting for graph ID")?
+            .context("unable to extract graph ID")?;
+        // TODO(mtls): Key just on addr?
+        let peer = SyncPeer::new(Addr::from(remote), graph_id);
```
jdygert-spok: I'm currently sending the graph ID over the connection so we can associate each connection with one `SyncPeer`. This is currently unnecessary but removing it would require more changes in the sync traits elsewhere which would then need to be reverted if/when we do make the connections per-graph. I guess we could try to make the graph ID optional in the sync traits.

crates/aranya-daemon/src/util.rs
```diff
@@ -114,77 +19,32 @@ impl TryClone for fs_keystore::Store {
...
-        out.sort_unstable();
-
-        expected.sort_unstable();
-
-        assert_eq!(out, expected);
+// TODO(mtls): Persist?
+#[derive(Debug, Default)]
+pub struct TeamConfigStore {
+    set: RwLock<HashSet<TeamId>>,
+}
```
jdygert-spok: This needs to be persisted, currently we must call `client.add_team` every time the daemon restarts.

jdygert-spok: It is now initialized with the list of graphs from the storage provider.

jdygert-spok:
Opened tracking issue for future work: #787.

crates/aranya-daemon/src/sync/transport/quic/mod.rs
jdygert-spok: #789

crates/aranya-daemon/src/sync/transport/quic/stream.rs
```diff
@@ -33,52 +30,36 @@ impl SyncStream for QuicStream {
...
-            .await
-            .map_err(Error::Send)?;
-        self.stream
-            .send(Bytes::copy_from_slice(data))
-            .await
-            .map_err(|error| {
-                warn!(peer = ?self.peer, %error, "send failed");
-                Error::Send(error)
-            })
+        self.send.write_all(&len.to_be_bytes()).await?;
```
djohnson-spideroak, line 33: do we use big-endian normally?

jdygert-spok: It's split somewhat evenly in aranya-core. Note this was kept from #717.

crates/aranya-client-capi/CMakeLists.txt
djohnson-spideroak: We do want to stabilize it eventually, initially I think we can keep it "preview"

crates/aranya-client/tests/scale/init.rs
djohnson-spideroak: Yeah, I think in a followup we would want to consolidate the tests and such.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -1,162 +1,138 @@
...
+                        e.insert(new_conn.clone());
+                        (new_conn.clone(), Some(new_conn))
                     }
                     Entry::Occupied(mut e) => {
-                        let existing_alive = e.get_mut().ping().is_ok();
+                        let existing_alive = e.get().close_reason().is_none();
                         // We initiated this connection, so it wins if we're the lower-addressed peer.
                         // TODO(nikki): This semi-fixes an existing bug, but we need a better way
                         // than raw addresses, this will break with NAT.
                         // https://github.com/aranya-project/aranya/issues/754
```
djohnson-spideroak, line 109: Is this still relevant?

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -1,194 +1,111 @@
...
 use crate::sync::{Addr, GraphId, SyncPeer};
 
-/// The amount of time we wait trying to accept a bidirectional stream from the peer connecting to
-/// us before we time out (they may be really really slow, or busy doing other things).
-const PENDING_ACCEPT_TIMEOUT: Duration = Duration::from_secs(30);
-
 /// The amount of time we wait trying to resolve the connecting peer's address (and waiting for them
 /// to send the port they want) before we time out.
-const RESOLVE_PEER_ADDR_TIMEOUT: Duration = Duration::from_secs(5);
+const RESOLVE_GRAPH_ID_TIMEOUT: Duration = Duration::from_secs(5);
```
djohnson-spideroak: comment needs to be updated

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -1,194 +1,111 @@
...
+    endpoint: quinn::Endpoint,
     /// Receiver to register new connections from the [`QuicConnector`](super::QuicConnector).
     pool: ListenerPool,
     /// Queue to allow awaiting a number of potential streams concurrently until one resolves.
-    pending_accepts: JoinSet<AcceptResult>,
+    #[derive_where(skip(Debug))]
+    accepting: SelectAll<Accepting>,
+    /// Set of incoming connections which have not yet completed.
+    #[derive_where(skip(Debug))]
+    connecting: SelectAll<Connecting>,
```
djohnson-spideroak: Does this queue by time of connection for accepting? It seems like its random based on the comment. Eventually we will want this to probably be a queue based on connection time.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
djohnson-spideroak: > This is currently unnecessary but removing it would require more changes in the sync traits elsewhere which would then need to be reverted if/when we do make the connections per-graph.

To clarify, even though we are associating it with a graph, we are not sharing connections only per-graph?

jdygert-spok: Currently, connections are keyed by `(Addr, GraphId)`, IOW connections are per-graph.

With the current mTLS implementation, connections don't _need_ to be per-graph, since we don't yet support per-graph certs. So we _could_ change it to reuse connections between graphs.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
jdygert-spok: I believe they will be yielded in the order they finish connecting.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
jdygert-spok: Yes, we are keeping the same tie breaking behavior for now.

crates/aranya-client-capi/CMakeLists.txt
jdygert-spok: Keeping this as-is and stabilizing certgen soon.

crates/aranya-daemon/src/sync/transport/quic/listener.rs
```diff
@@ -210,39 +129,26 @@ impl SyncListener for QuicListener {
...
-                    trace!(?remote, "raw connection accepted from server");
-
-                    if let Err(error) = self.register_connection(conn).await {
-                        error!(?remote, ?error, "failed to accept connection");
-                        handle.close(application::Error::UNKNOWN);
+                Some(incoming) = self.endpoint.accept() => {
+                    self.accept_incoming(incoming);
+                }
+                Some(conn) = self.connecting.next() => {
+                    if let Err(error) = self.register_connection(conn.clone()).await {
```
djohnson-spideroak: Do we want to block on register_connection here? We would block this loop until the graph ID is received from the peer for the duration of the timeout.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -1,162 +1,138 @@
...
+            // TODO(mtls): timeout?
+            let new_conn = self
+                .endpoint
+                .connect_with(config, addr, peer.addr.host())?
+                .await?;
+            new_conn
+                .open_uni()
                 .await?
-                .send(self.return_port.clone())
+                .write(peer.graph_id.as_bytes())
```
djohnson-spideroak: should use write_all

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -1,162 +1,138 @@
...
+        let addr = find_matching_ip_version(addrs, self.local_addr)
+            .context("no resolved address matches local endpoint IP version")?;
+        trace!(%addr, "resolved peer address");
 
         // Check for an existing live connection, cleaning up dead ones.
         // Drops the lock before doing async connection work.
         let reuse = {
             // Hold the lock across this entire operation
             let mut map = self.pool.conns.lock().await;
             match map.entry(peer) {
```
djohnson-spideroak: I think this is wrong, we are not doing the lookup based on the resolved address here. The peer argument is potentially a DNS name, and listener.rs keys on IP because that is what it gets out of quinn.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -1,162 +1,138 @@
...
                     }
                     Entry::Occupied(mut e) => {
-                        let existing_alive = e.get_mut().ping().is_ok();
+                        let existing_alive = e.get().close_reason().is_none();
                         // We initiated this connection, so it wins if we're the lower-addressed peer.
                         // TODO(nikki): This semi-fixes an existing bug, but we need a better way
                         // than raw addresses, this will break with NAT.
                         // https://github.com/aranya-project/aranya/issues/754
-                        let outbound_wins = !existing_alive || self.local_addr < peer.addr;
+                        let outbound_wins = !existing_alive || self.local_addr < addr;
```
djohnson-spideroak, line 110: What if local address is 0.0.0.0? Related to the NAT break, if we didnt bind to the same address on the wire, this isnt going to have the behavior we expect.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
jdygert-spok: Agreed, I brought this up recently too. I can try to fix here but was planning to leave for follow up.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
jdygert-spok: Ugh yeah, this definitely isn't good. Bumping up the priority to fix #754 but I still want to do it separately to not drag out this PR.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
jdygert-spok: #795.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,418 @@
...
+/// # Arguments
+/// * `cert_path` - Path to the device certificate PEM file (may contain certificate chain)
+/// * `key_path` - Path to the private key PEM file (PKCS#8 or SEC1/EC format)
+///
+/// # Returns
+/// A tuple of (certificate chain, private key)
+///
+/// # Errors
+/// Returns an error if the files cannot be read or parsed.
+// TODO: Support loading multiple device certificates for devices that join multiple teams.
```
djohnson-spideroak: nit: we should have an issue number here

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,418 @@
...
+
+    // Load private key.
+    // With the `alloc` feature enabled on rustls-pki-types, PrivateKeyDer uses
+    // Zeroizing<Vec<u8>> internally, ensuring key bytes are zeroized on drop.
+    let key_file = File::open(key_path).map_err(|e| CertError::OpenFile {
+        path: key_path.to_path_buf(),
+        source: e,
+    })?;
+
+    let key = PrivateKeyDer::from_pem_reader(key_file).map_err(|e| CertError::ParseKey {
```
djohnson-spideroak, line 207: does key impl `Zeroize` type here?

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,418 @@
...
+    certs: Vec<CertificateDer<'static>>,
+    key: PrivateKeyDer<'static>,
+) -> Result<rustls::ClientConfig> {
+    let config = rustls::ClientConfig::builder()
+        .with_root_certificates(roots)
+        .with_client_auth_cert(certs, key)
+        .map_err(CertError::BuildClientConfig)?;
+
+    debug!("built rustls client config for mTLS");
+    Ok(config)
```
djohnson-spideroak, line 241: Do we need ALPN here?

djohnson-spideroak (commented):
still reviewing

crates/aranya-daemon/src/api.rs
```diff
@@ -548,27 +530,27 @@ impl DaemonApi for Api {
...
+
+        Ok(())
     }
 
     #[instrument(skip(self), err)]
     async fn remove_team(self, _: context::Context, team: api::TeamId) -> api::Result<()> {
-        if let Some(data) = &self.quic {
-            self.remove_team_quic_sync(team, data)?;
+        if !self.teams.remove(team) {
+            return Err(anyhow!("team {team} was not present").into());
```
djohnson-spideroak, line 547: Do we want to return here? If the syncer has a different list, we could end up with stale team entries.

jdygert-spok: I don't think that is able to happen.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
jdygert-spok: `Zeroize` yes, `ZeroizeOnDrop`, no. But it is defined by rustls and we just pass it into rustls code so its out of our hands.

crates/aranya-daemon/src/sync/transport/quic/certs.rs
jdygert-spok: Those are set here: https://github.com/aranya-project/aranya/blob/3a9557138710e5ba3a006be540917fab05fd0285/crates/aranya-daemon/src/sync/transport/quic/mod.rs#L92-L120

crates/aranya-client-capi/ctest/execute_test.sh
```diff
@@ -97,12 +103,28 @@ if [ -n "$daemon_names" ]; then
     tmpdir=$(mktemp -d)
     echo "Using temp directory: $tmpdir"
 
+    # Create certificate directory
+    root_certs_dir="$tmpdir/root_certs"
+    mkdir -p "$root_certs_dir"
+
+    # Generate CA certificate
+    echo "Generating CA certificate..."
+    "$certgen_path" ca --cn "Test CA" --output "$root_certs_dir/ca"
```
NWPlayer123, line 112: Does this generate a new cert every time? Seems like that could get unbounded fast and should prob be cleaned up (cargo make task?), I don't know whether to do so automatically cuz sometimes it's nice to go back when you're debugging something

Cargo.lock
```diff
@@ -3196,6 +3088,61 @@ dependencies = [
...
+checksum = "b9e20a958963c291dc322d98411f541009df2ced7b5a4f2bd52337638cfccf20"
+dependencies = [
+ "bytes",
+ "cfg_aliases",
+ "pin-project-lite",
+ "quinn-proto",
+ "quinn-udp",
+ "rustc-hash 2.1.1",
+ "rustls",
+ "socket2 0.6.0",
```
NWPlayer123, line 3104: probably worth updating tokio so everything's on socket2 0.6, prob out of scope for this PR though, thiserror 1.X can be removed if/when tarpc gets migrated off, rustc-hash is weird because age needs to be updated but the published crates don't have it updated even though it's updated in their Cargo.toml, and crypto would need to get updated to getrandom 0.3, and I think we're forever stuck with multiple versions of windows-sys

crates/aranya-daemon/src/sync/transport/quic/certs.rs
```diff
@@ -0,0 +1,418 @@
...
+    PrivateKeyDer<'static>,
+)> {
+    let root_store = load_root_certs(&config.root_certs_dir)?;
+    let (device_certs, device_key) = load_device_cert(&config.device_cert, &config.device_key)?;
+    Ok((Arc::new(root_store), device_certs, device_key))
+}
+
+#[cfg(test)]
+#[allow(clippy::expect_used, clippy::unwrap_used)]
+mod tests {
```
NWPlayer123, line 296: do we want these here when we're already creating a separate tests.rs

crates/aranya-daemon/src/sync/transport/quic/connector.rs
```diff
@@ -1,171 +1,140 @@
...
-            let (handle, acceptor) = {
-                // Hold the lock across this entire operation
-                let mut map = self.pool.conns.lock().await;
+            let (handle, acceptor) = self.pool.conns.with_map(|map| {
                 match map.entry(peer) {
                     Entry::Vacant(e) => {
-                        e.insert(new_handle.clone());
-                        (new_handle, Some(new_acceptor))
+                        e.insert(new_conn.clone());
+                        (new_conn.clone(), Some(new_conn))
```
NWPlayer123, line 102: I don't *love* the cloning but it seems like quinn::Endpoint was designed for that so eh

crates/aranya-daemon/src/sync/transport/quic/mod.rs
```diff
@@ -69,3 +70,88 @@ impl From<std::convert::Infallible> for Error {
...
+        .map_err(|e| Error::Endpoint(format!("failed to create server endpoint: {e}")))?;
+
+    let local_addr = endpoint
+        .local_addr()
+        .map_err(|e| Error::Endpoint(format!("unable to get local address: {e}")))?;
+
+    debug!("created unified QUIC endpoint with mTLS at {}", local_addr);
+
+    let (connector_pool, listener_pool) = connections::pool(32);
+
```
NWPlayer123, line 138: should probably be a const generic (or I guess just a variable passed in)

crates/aranya-daemon/src/sync/transport/quic/certs.rs
jdygert-spok: This seems a fine place for them, they are very narrow tests.

crates/aranya-daemon/src/sync/transport/quic/connector.rs
jdygert-spok: Yeah, it's like how we were cloning the handle before, but `quinn` just doesn't split it up into handle/acceptor.

crates/aranya-client-capi/ctest/execute_test.sh
jdygert-spok: The tempdir gets removed by the cleanup function called at exit.

crates/aranya-daemon/src/sync/transport/quic/mod.rs
jdygert-spok: Meh, it just means if you somehow create new connections too quickly without letting the server register them then you'll get throttled. I don't think it needs to be configurable.

chip-so (approved):

PR #796, opened 2026-04-16
chore: rustfmt 2024

jdygert-spok:


chip-so (approved):
So is it just sorting capitalized imports first now?

jdygert-spok:
> So is it just sorting capitalized imports first now?

That's the main one we see. There's a list of changes here:

https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt.html

djohnson-spideroak (approved):

